PR #1377, opened 2014-12-21
Hide motions from non staff users in early workflow state.

normanjaeckel:


openslides/motion/models.py
```diff
@@ -783,6 +789,12 @@ class State(models.Model):
     icon = models.CharField(max_length=255)
     """A string representing the url to the icon-image."""
 
+    required_permission_to_see = models.CharField(max_length=255, blank=True)
+    """
+    A permission string. If not empty, the user has to have this permission to
+    see a motion in this state.
```
emanuelschuetze: Please add an example in this comment how can use this. In which file you can set this field.
(I tested it successfully after modify motion/signals.py and added this line to state_2_1:
`required_permission_to_see='motion.can_manage_motion',`)

And please add here a note in manual that there is a new possibility to control permissions:
http://docs.openslides.org/de/Motion.html#arbeitsablauf-und-status

emanuelschuetze:
Only one comment. Rest works fine.

normanjaeckel:
Ready to be rebased and merged.

ostcar:
ok.

PR #1380, opened 2014-12-25
Feature: amendments

ostcar:
@emanuelschuetze the backend is done. I added the relevant links into the template. Could you make it pretty?

### DIFF
--- a/openslides/motion/templates/motion/motion_detail.html
+++ b/openslides/motion/templates/motion/motion_detail.html
@@ -29,6 +29,9 @@
                 </span>
             {% endif %}
         {% endif %}
+        {% if motion.is_amendment %}
+            (Amendment of <a href="{{ motion.parent|absolute_url }}">{{ motion.parent.identifier|default:motion.parent }}</a>)
+        {% endif %}
     </small>
     <small class="pull-right">
         <a href="{% url 'motion_list' %}" class="btn btn-mini">
@@ -279,6 +282,23 @@
             {% endif %}
         </h5>
         {{ version.creation_time }}
+
+        {% if 'motion_amendments_enabled'|get_config %}
+            <h5>{% trans 'Amendments' %}:</h5>
+            {% with amendments=motion.amendments.all %}
+                {% if amendments %}
+                    <ul>
+                        {% for amendment in amendments %}
+                            <li><a href="{{ amendment|absolute_url }}">{{ amendment }}</a></li>
+                        {% endfor %}
+                    </ul>
+                {% endif %}
+            {% endwith %}
+            <a class="btn btn-mini btn-primary" href="{% url 'motion_create_amendment' motion.pk %}">
+                <i class="icon-plus icon-white"></i>
+                {% trans 'New amendment' %}
+            </a>
+        {% endif %}
 
         <!-- Support/Unsupport button -->
         {% if perms.motion.can_support_motion and 'motion_min_supporters'|get_config > 0 %}
--- a/openslides/motion/templates/motion/slide.html
+++ b/openslides/motion/templates/motion/slide.html
@@ -77,6 +77,9 @@
     {{ motion.active_version.title }}
     <small>
         {% trans "Motion" %} {{ motion.identifier|default:'' }}
+        {% if motion.is_amendment %}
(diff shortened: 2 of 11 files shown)


openslides/motion/templates/motion/slide.html
```diff
@@ -77,6 +77,9 @@ <h4 class="first">{% trans "Status" %}:</h4>
     {{ motion.active_version.title }}
     <small>
         {% trans "Motion" %} {{ motion.identifier|default:'' }}
+        {% if motion.is_amendment %}
+            (Amendment of <strong>{{ motion.parent }}</strong>)
```
emanuelschuetze: Please remove the strong-tag in slides subtitle (we never use bold style in subtitle).

And please reduce it to motion.parent.identifier (as i proposed). Otherwise you get multiple line breaks in subtitle if you have a long main motion. The identifier is enough. The title is (normally) the same like in the amendment.

openslides/motion/templates/motion/motion_detail.html
```diff
@@ -280,6 +283,20 @@
...
+            {% with amendments=motion.amendments.all %}
+                {% if amendments %}
+                    <ul>
+                        {% for amendment in amendments %}
+                            <li><a href="{{ amendment|absolute_url }}">{{ amendment }}</a></li>
+                        {% endfor %}
+                    </ul>
+                {% endif %}
+            {% endwith %}
+            <a class="btn btn-mini" href="{% url 'motion_create_amendment' motion.pk %}">{% trans 'New amendment' %}</a>
```
emanuelschuetze: Please use blue button with `class="btn btn-mini btn-primary"` and add the "plus"-icon as the "new motion" button.

emanuelschuetze:
2 notes (see above). Rest looks good. Please fix and rebase. I want to make a final test before merge.

openslides/motion/templates/motion/slide.html
ostcar: I prefer that we reference a motion the same way at any place. @normanjaeckel what do you think?

openslides/motion/templates/motion/motion_detail.html
ostcar: done

ostcar:
I removed the strong-tag from the slide and changed the button style.
We need a decision for the question, if the parent motion of a amendment should be illustrated only by his identifier or with the \__unicode\__ methode, that we use in all other places.

emanuelschuetze:
We show motions in  two different ways:
a) long unicode form (e.g. "1 | title")
b) short form in subtitle (e.g. "Motion 1"), see https://github.com/ostcar/OpenSlides/blob/amendments/openslides/motion/templates/motion/motion_detail.html#L20

In this case we discus about an extended subtitle. So please use the same motion style (b) = short form to display the parent identifier.

normanjaeckel:
Normally I would prefer uniformity. But may be we can use a short form in this special case (only subtitle) so that the line does not become too long. If I see this correctly this only touches the motion detail view and the motion detail slide, right?

ostcar:
done.

There is one rarely case that I don't like. When the parent motion doesn't have a identifier, and the identifier for the amendment is created, it becomes the identifier "A 1" (without the identifier of the parent). If the parent then becomes a Identifier (e. g. "B 1") then the identifier from the second amendment will be "B 1 A 2".

If there is a categorie with the prefix "A" then amendments from parents with no identifiers and motions from the categorie A will have the same namespace.

I don't know how to solve this and would prefer to do nothing. One solution would be to set the identifier of a parent when the identifier of a amendment will be set.

normanjaeckel:
I think we should ignore this case. If someone sets or changes the identifier of a parent after creating amendments, he has to reset the identifier of the amendments manually.
We should also ignore the a category prefix might be `A` or `Ä` or something like this. The user has to care about this manually. Maybe someone wants to add a note in the manual.

ostcar:
@normanjaeckel so I can rebase and merge this?

emanuelschuetze:
I make last tests: One thing is still missing. If you create a amendment the category should be the same like the parent motion. Currently it's empty.
After fixing that you can merge. Rest is ok.

ostcar:
done

normanjaeckel:
Rebase and merge.

PR #1381, opened 2014-12-28
Tags for motions

ostcar:
The backend is done. You can alter the tags in the general config tab.

The tags are created in the core app, so we can use the same tabs for assignments or other apps later.

@emanuelschuetze I put the tags as list inside the MotionDetailView as example. Can you make it pretty?

PR #1385, opened 2015-01-02
Merge stable branch

normanjaeckel:
This is to show the merge commit. If it is ok, I will push it directly into master branch.


ostcar:
@normanjaeckel you can push this into the master branch right now. There will be a lot of conflicts in #1381 and #1380 that have to be manually resolved. So it will be best, when I merged them into master.

PR #1386, opened 2015-01-03
use isort

ostcar:
* sorted all imports. Run python make isort to do it again
* check that all imports are sorted with travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,4 @@ script:
   - "DJANGO_SETTINGS_MODULE='tests.settings' coverage run ./manage.py django test"
   - "coverage report -m --fail-under=80"
   - "flake8 --max-line-length=150 --statistics openslides tests"
+  - "isort --line-width 120 -rc -m 4  -c ."
--- a/openslides/users/models.py
+++ b/openslides/users/models.py
@@ -1,10 +1,7 @@
 # TODO: Check every app, that they do not import Group or User from here.
 
-from django.contrib.auth.models import (PermissionsMixin, AbstractBaseUser,
-                                        BaseUserManager)
-
-# TODO: Do not import the Group in here, but in core.models (if necessary)
 from django.contrib.auth.models import Group  # noqa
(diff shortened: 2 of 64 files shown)


openslides/users/models.py
```diff
@@ -1,10 +1,7 @@
 # TODO: Check every app, that they do not import Group or User from here.
 
-from django.contrib.auth.models import (PermissionsMixin, AbstractBaseUser,
-                                        BaseUserManager)
-
-# TODO: Do not import the Group in here, but in core.models (if necessary)
```
normanjaeckel: Please do not remove this TODO or add a short instruction to a specific GitHub issue.

.travis.yml
```diff
@@ -7,3 +7,4 @@ script:
   - "DJANGO_SETTINGS_MODULE='tests.settings' coverage run ./manage.py django test"
   - "coverage report -m --fail-under=80"
   - "flake8 --max-line-length=150 --statistics openslides tests"
+  - "isort --line-width 120 -rc -m 4  -c ."
```
normanjaeckel: I would prefer the following line:
`"isort --line-width 115 --multi_line 3 --check-only --recursive openslides tests"`

Alternative: `"isort --line-width 115 --multi_line 1 --check-only --recursive openslides tests"`

* The long option flags are a little bit more understandable so that we do not have to read the isort docs if we look later at this line.
* I propose to move the recursive flag to the end just before the file definitions because it belongs more to it than the other flags.
* Lines of 120 are a little bit too long for GitHub diff view.
* The file definition `.` is not possible because it also catches files in a virtualenv folder and such stuff.
* Last but not least I like a vertical style more than  the grid style because it is easier to see the changes in a diff later on.

openslides/users/models.py
ostcar: I added https://github.com/OpenSlides/OpenSlides/issues/1387

I don't think that we need a commend in the code, because the code is not our primary issue tracker.

normanjaeckel: Ok.

PR #1389, opened 2015-01-05
Used Bower and gulp to manage third party JS and CSS libraries

normanjaeckel:
Cleaned up settings. Updated development section in README.rst. Removed jQuery, jquery.cookie and bootstrap from repository.

I can move more packages into bower.json if this way is accepted.


emanuelschuetze:
I tested the install instruction from README with bower and gulp. Works without problems.

Only two notes:
- bootstrap.min.js is still included and not replaced by openslides.js
- note that we use only custom components for jQuery UI (see README). I'm not sure if this defineable in bower.

gulpfile.js
```diff
@@ -0,0 +1,58 @@
...
+    uglify = require('gulp-uglify');
+
+// Directory where the results go
+var output_directory = path.join('openslides', 'static');
+
+// Catches all JavaScript files from all bower components and concats them to
+// one file js/openslides-libs.js. In production mode the file is uglified.
+gulp.task('js', function() {
+    return gulp.src(mainBowerFiles({
+            filter: /js$/
```
ostcar:
```javascript
filter: /\.js$/
```

openslides/utils/settings.py.tpl
```diff
@@ -1,49 +1,66 @@
...
 MEDIA_ROOT = os.path.join(OPENSLIDES_USER_DATA_PATH, 'media', '')
 
-HAYSTACK_CONNECTIONS['default']['PATH'] = os.path.join(OPENSLIDES_USER_DATA_PATH, 'whoosh_index', '')
-
 TEMPLATE_DIRS = (
     os.path.join(OPENSLIDES_USER_DATA_PATH, 'templates'),
-    os.path.join(SITE_ROOT, 'templates'))
+)
+
+STATICFILES_DIRS.insert(0, os.path.join(OPENSLIDES_USER_DATA_PATH, 'static'))
```
ostcar:
```python
STATICFILES_DIRS = [os.path.join(OPENSLIDES_USER_DATA_PATH, 'static')] + STATICFILES_DIRS
```

PR #1392, opened 2015-01-05
Fixed bug in ResetPasswordView. Fixed #1391.

### LINKED ISSUE #1391: ResetPasswordView does not work.

normanjaeckel:


tests/users/test_views.py
```diff
@@ -57,6 +57,12 @@ def test_activate(self):
         response = self.client.get('/user/1/status/activate/')
         self.assertEqual(response.status_code, 302)
 
+    def test_reset_password(self):
+        self.admin.default_password = new_password = gen_password()
```
ostcar: Please use a static password and not a generated password here. You don't test gen_password, but the reset_password-view. If the function gen_password is broken, this test could fail.

If you want to test the gen_password function, then please write its own test.

ostcar:
Please fix travis, and then you can merge.

I can live with it if you don't change the test.

tests/users/test_views.py
normanjaeckel: Done.

ostcar:
ok, merge after travis is fixed.

ostcar:
Optional: rebase

PR #1393, opened 2015-01-07
Reduce the usage of bsmselect jquery pluin.

emanuelschuetze:
Use the plugin only for specific <select multiple> form fields.
E.g. not for the group permissions field.

Partly reverted changes from pull request #1381.

### DIFF
--- a/openslides/core/templates/base.html
+++ b/openslides/core/templates/base.html
@@ -136,7 +136,7 @@
     <script type="text/javascript" src="{% static 'js/jquery/jquery.bsmselect.js' %}"></script>
     <script type="text/javascript">
         // use jquery-bsmselect for all <select multiple> form elements
-        $("select[multiple]").bsmSelect({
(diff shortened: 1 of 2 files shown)


ostcar:
I would like to use a cleverer method to choose if bsmselect is used. For example we could use a CSS-class for it.

When we change this line:
https://github.com/OpenSlides/OpenSlides/pull/1393/files#diff-a031633c7495f6f841356dafbfd405d3L139

to: 
```
$("select[multiple].use_bsmselect")
```
then only select elements with the clase 'use_bsmselect' would use bsmselect, but we could load the files in core for all apps and plugins.

emanuelschuetze:
Ok, I'll update this pull request with oskars suggestion.

emanuelschuetze:
Updated with extra css-class. Please review and merge.

ostcar:
ok

PR #1397, opened 2015-01-12
Fixed haystack indexing problem for m2m objects.

emanuelschuetze:
Updated django-haystack to <2.4.

@ostcar 
Thanks for your hint. I merged your test commit into my commit.
One index problem is still open:
If an assignment is created and you added a candidate. The search index is NOT updated with the new candidate. The assignment is listed in index but without candidates. Workaround: edit assignment and save the form again to rebuild index for this object. Than the new candidate is searchable and related to assignment object.
Any idea to fix this?


normanjaeckel:
Ok.

ostcar:
ok

PR #1398, opened 2015-01-13
Improved portable install script

emanuelschuetze:
and fixed AttributeError

### DIFF
--- a/openslides/core/chatbox.py
+++ b/openslides/core/chatbox.py
@@ -3,8 +3,9 @@
 from datetime import datetime
 
 from django.conf import settings
-from django.contrib.sessions.models import Session
 from django.utils.html import urlize
+from django.utils.importlib import import_module
+
 from sockjs.tornado import SockJSConnection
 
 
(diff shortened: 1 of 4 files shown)


ostcar:
ok

emanuelschuetze:
I just updated my pull request because I reverted the openslides.exe (which I committed accidently). So no more changes excepting openslides.exe.

So please merge.

openslides/core/chatbox.py
```diff
@@ -20,14 +21,13 @@ def on_open(self, info):
...
-        session_key = info.get_cookie(settings.SESSION_COOKIE_NAME).value
-        session = Session.objects.get(session_key=session_key)
+        # get the session (compatible with other auth-backends)
+        engine = import_module(settings.SESSION_ENGINE)
         try:
+            session_key = info.get_cookie(settings.SESSION_COOKIE_NAME).value
+            session = engine.SessionStore(session_key)
             self.user = User.objects.get(pk=session.get_decoded().get('_auth_user_id'))
-        except User.DoesNotExist:
+        except (User.DoesNotExist, AttributeError):
```
normanjaeckel: Can you please split up the try-except-blocks so that every error tests only the line where the error might appear (savely).

emanuelschuetze: It is already reduced to the relevant lines of code:
An AttributeError could be arise in line 27 (.value) and 29 (.get_decoded()).
An User.DoesNotExist error could be arise in line 29 (User.objects.get..)
So what is to optimize here?

PR #1408, opened 2015-01-17
Refactored REST api in agenda, core and users app.

normanjaeckel:


openslides/users/models.py
```diff
@@ -72,7 +73,8 @@ class User(SlideMixin, AbsoluteUrlMixin, PermissionsMixin, AbstractBaseUser):
 
     class Meta:
         permissions = (
-            ('can_see', ugettext_noop('Can see users')),
+            ('can_see_name', ugettext_noop('Can see names of users')),
+            ('can_see_extra_data', ugettext_noop('Can see extra data of users')),
```
ostcar: Rename to 'can_see_full_user'

PR #1416, opened 2015-01-21
add isort

ostcar:
See also #1386

When all is ready, I will call the command "python make isort" to sort all imports. This are the only manual changes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,4 @@
+[isort]
+multi_line_output=3
+include_trailing_comma=True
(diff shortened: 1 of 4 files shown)


ostcar:
The Travis Fail was expected. It will work when you give your OK and I actually changed the import-order.

ostcar:
See also: #1386

ostcar:
See also: https://groups.google.com/forum/#!topic/django-developers/EQhbI4cDuX4

ostcar:
I changed the pull request that only isort is installed and a isort-config file is appended. With this, we can use isort to fix the imports in one file but do not have to change any file right now.

We can add the travis check when the release is near.

-> Ready for review.

ostcar:
@normanjaeckel please review (I don't know If you realised my change here)

ostcar:
@normanjaeckel push

setup.cfg
```diff
@@ -0,0 +1,4 @@
+[isort]
+multi_line_output=3
```
normanjaeckel: line-width is missing. And, if you like, you can order the options alphabetically.

ostcar: line-width=79 is the default and does not have to be set in this file

normanjaeckel:
Rebase and merge.

PR #1417, opened 2015-01-21
Show special characters in PDF, like "&". (Fixed #1415)

### LINKED ISSUE #1415: Ampersand is not shown on Agenda PDF.
Use `&` in an agenda title. It disapears on PDF. I think this symbol might be used often.

emanuelschuetze:
Added escape function for agenda, motion, assignment and participants
pdf views.


normanjaeckel:
Ok.

PR #1418, opened 2015-01-21
Updated translations (DE, FR, CS, PT) for 1.7.

emanuelschuetze:


openslides/locale/fr/LC_MESSAGES/django.po
```diff
@@ -1894,245 +1925,262 @@ msgstr "Arrêtez de soumettre de nouvelles motions par les utilisateurs non-empl
...
 
-#: motion/signals.py:76
+#: motion/signals.py:75
+msgid "Activate amendments"
+msgstr "Activer amendements"
+
+#: motion/signals.py:80
+msgctxt "Prefix for the identifier for amendments"
+msgid "A"
+msgstr "un"
```
normanjaeckel: I think this translation is not correct. The Fench prefix should not be `un` but `A`, should it?

openslides/locale/pt/LC_MESSAGES/django.po
```diff
@@ -1890,245 +1921,262 @@ msgstr "Parar a submissão de novas moções por utilizadores não-staff"
...
 
-#: motion/signals.py:76
+#: motion/signals.py:75
+msgid "Activate amendments"
+msgstr "Activar emendas"
+
+#: motion/signals.py:80
+msgctxt "Prefix for the identifier for amendments"
+msgid "A"
+msgstr "A"
```
normanjaeckel: Here the same. It should be `E`.

openslides/locale/fr/LC_MESSAGES/django.po
emanuelschuetze: right, updated!

openslides/locale/pt/LC_MESSAGES/django.po
emanuelschuetze: should be done, right?

emanuelschuetze: ah, that's PT. right "E"! I'll fix it.

emanuelschuetze:
all done. please merge.

ostcar:
ok

PR #1421, opened 2015-01-21
Fixed bug in chatbox.

normanjaeckel:
Please test carefully because I am not sure that this is the solution.


emanuelschuetze:
ok. works!

PR #1423, opened 2015-01-22
New Test order.

ostcar:
* seperate unittests and integration tests
* moved old tests in seperat folder 'old'
* created a testrunner that does not create a testdatabase, if only unittests are run
* wrote some unit- and integration tests as examples
* fixed user.get_short_name() to use the sort order from config
* fixed wrong url_pattern in the user app

Run:
```python make test tests.unit```
or:
```DJANGO_SETTINGS_MODULE=tests.integration.settings python manage.py django test tests.unit```

to run only the unit tests (very fast)

The second command will be a lost easier when #1412 is merged. Then it should be:
```python manage.py test tests.unit```

### CODE REVIEW

**Comment by ostcar** [comment]
I don't know why travis does not find the settings tests/integration/settings.py. I don't want to debug this in the old main script. I fix this when #1412 is merged.

File: openslides/users/models.py [hunk 1]
```diff
@@ -125,9 +127,19 @@ def get_short_name(self):
         E. g.: * Max Mustermann
                * Enders, Christoph
         """
-        # TODO: Order of name. See config.
-        name = ('%s %s' % (self.first_name, self.last_name)).strip()
-        return name or self.username
+        if self.first_name and self.last_name:
+            # The user has a last_name and a first_name
+            user_names = [self.last_name, self.first_name]
```
normanjaeckel: What if the last name or the first name contains whitespaces at the beginning or at the end. They should be stripped away before combining the names.
Proposal also for more readability and less lines of code:
```python
if self.first_name and self.last_name:
    # The user has a last_name and a first_name
    if config['users_sort_users_by_first_name']:
        name = ' '.join(self.first_name.strip(), self.last_name.strip())
    else:
        name = ', '.join(self.last_name.strip(), self.first_name.strip())
elif ...
```

ostcar: done.

I used another implementation, but I think this should be ok for you

PR #1424, opened 2015-01-22
Use a real anonymous user object, instadt of a authentification backend

ostcar:
The advantage of this implementation over the old one is, that we have a real AnonymousUserObject in request.user that we can alter in our code.

The disadvantage is, that I had to copy a lot of code from the django core into our code basis to realize this.


emanuelschuetze:
I just tested it: enable and login with anonymous user. OK.

openslides/global_settings.py
```diff
@@ -71,7 +67,7 @@
     'django.middleware.locale.LocaleMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
-    'django.contrib.auth.middleware.AuthenticationMiddleware',
+    'openslides.users.anonymous.AuthenticationMiddleware',
```
normanjaeckel: Can you please rename the authentication submodule to something like `openslides.users.auth`? This file is not only about anonymous users but authentication of all users.

openslides/users/anonymous.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    Class for anonymous user instances, which have the permissions from the
+    Group 'Anonymous' (pk=1).
+    """
+
+    def get_all_permissions(self, obj=None):
+        """
+        Return the permissions a user is granted by his group membership(s).
+
+        - try to return the permissions for the 'Anonymous' group (pk=1).
```
normanjaeckel: Typo: `Try ...`

openslides/users/anonymous.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        perms = Permission.objects.filter(group__pk=1)
+        if perms is None:
+            return set()
+        # TODO: test without order_by()
+        perms = perms.values_list('content_type__app_label', 'codename').order_by()
+        return set([u'%s.%s' % (ct, name) for ct, name in perms])
```
normanjaeckel: The unicode `u` is not required, is it?

openslides/users/anonymous.py
```diff
@@ -0,0 +1,101 @@
...
+    def process_request(self, request):
+        """
+        Like django.contrib.auth.middleware.AuthenticationMiddleware, but uses
+        our own get_user function.
+        """
+        assert hasattr(request, 'session'), (
+            "The authentication middleware requires session middleware "
+            "to be installed. Edit your MIDDLEWARE_CLASSES setting to insert "
+            "'django.contrib.sessions.middleware.SessionMiddleware' before "
+            "'openslides.users.anonymous.AuthenticationMiddleware'."
```
normanjaeckel: Don't forget to change the module name also here.

normanjaeckel:
Some remarks but rest looks good.

openslides/global_settings.py
ostcar: I like the name "anonymous" more, because this is an AuthenticationMiddleware that (in difference to is original) allows anonymous "authentication".

openslides/users/anonymous.py
ostcar: done

openslides/users/anonymous.py
ostcar: done

openslides/users/anonymous.py
ostcar: see above

openslides/global_settings.py
normanjaeckel: Hm. I think we need a generic term for what is done in this file resp. in the middleware, that means for both: anonymous "authentication" and authetication with credentials. such a generic term can be "auth" or "authentication" or may be something else but not "anonymous".

ostcar: ok. But I want some points!

ostcar:
all done. Can I rebase an merge?

normanjaeckel:
Rebase and merge.

PR #1427, opened 2015-01-24
Django REST framework – config and mediafile

normanjaeckel:


openslides/config/views.py
```diff
@@ -100,3 +101,22 @@ def form_valid(self, form):
             config[key] = form.cleaned_data[key]
         messages.success(self.request, _('%s settings successfully saved.') % _(self.config_collection.title))
         return super(ConfigView, self).form_valid(form)
+
+
+class ConfigViewSet(viewsets.ViewSet):
```
ostcar: This could be a problem. What permission does a user need to see all config values?

Could there be config values which are secret? Maybe in any plugin or for a short period of time? For example who voted for who

normanjaeckel: At the moment we don't have a policy which config values are public and which not. We can open an issue but for now I have to decide whether to grant read access to all or not. If not nobody could even see the welcome text.

ostcar: I think this is a more general problem. It could be, that the config-app can not be accessable via rest, but though another api which is more suable for config values. 

If we add the config now, it will be harder to change the client later. So I would suggest, you remove the config from the rest api for now.

normanjaeckel: Why not adding a flag to every config variable `required_permission_to_see`. Default is None which means that there is no special permission required to see this variable. At the moment there is no variable which has secret content but we (or plugin developers) can add a special permission to variables later on.

ostcar: This could work. But I want to think about this.

ostcar:
Do you want to rebase this? (I would prefer to stay on our one commit policy)
In other case you can merge this. There will probably be some changes before the 2.0 in the rest api, but we should have a full features rest api for now to work with it

PR #1431, opened 2015-01-24
fix #1430 - anonymous user and motion detail view

### LINKED ISSUE #1430: Anonymous not working properly
* Create a motion with admin user.
* Activate anynousmus guest users.
* Try to see the motion with anonymous user:
  * Response HTTP 500, expected HTTP 200

The problem is that the method `motion.is_submitter(person)` tries to check the anonymous user. This fails because this user object is not like the others.

ostcar:


normanjaeckel:
Ok.

normanjaeckel:
Please add a CHANGELOG entry. The merge.

ostcar:
do not merge this before we have a solution for #1432

I don't think that we need a changelog entry for this. The fix was only for a change in the 2.0 release. And I also don't think, that we need a changelog entry for the rework of the anonymous user. That is only for intern use.

normanjaeckel:
Ok. But we forgot CHANGELOG entry in #1424 so we can add it here.

ostcar:
You added it in #1433, so can I rebase and merge here?

normanjaeckel:
Rebase and merge.

PR #1433, opened 2015-01-25
Updated CHANGELOG.

normanjaeckel:
We merged some pull requests without adding entries to the CHANGELOG. I think we should add these entries now and maybe remove some of them later (just before the final release). This is less work than searching for all stuff during release process.


ostcar:
ok. You can merge it now or wait for travis.

Issue #1432, opened 2015-01-24
Anonymous user can not access REST api


normanjaeckel:
* Enable anonymous guest user.
* Try to get agenda item viw REST api:
  * Response HTTP 403, expected HTTP 200

If you use an normal registered user (non manager) it works. Default permissions aren't changed.

PR #1434, opened 2015-01-25
use our anonymous user object for the rest framework

ostcar:
fix: #1432


normanjaeckel:
Looks easy.  Do you like to add a short test? :-)

PR #1435, opened 2015-01-25
Fix AssignmentPollUpdateView

ostcar:
Lets merge this and never speak of it again...

I did not wrote a test, because I don't like our old test environment anymore.


ostcar:
See also: http://de.demo.openslides.org/assignment/3/

PR #1436, opened 2015-01-25
Speed up travis with cache for pip and node

ostcar:
Changed travis to cache the installed python and node packages.

"sudo: false" tells travis to run the commands in a container-based-infrastructure (in which sudo does not work). See http://docs.travis-ci.com/user/workers/container-based-infrastructure/

Added the 'upgrade' argument to pip install, so it does installs new version of python packages if available. This will update the cache.

If we want to 'uninstall' dependencies (which should not be neccessary) the cache can be deleted on the travis webpage.

See also: docs.travis-ci.com/user/caching/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,15 @@
 language: python
+sudo: false
+cache:
+  directories:
+    - node_modules
+    - $HOME/virtualenv/python$TRAVIS_PYTHON_VERSION/lib/python$TRAVIS_PYTHON_VERSION/site-packages/
+    - $HOME/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/
 python:
   - "3.3"
   - "3.4"
 install:
-  - "pip install -r requirements.txt"
+  - "pip install --upgrade --requirement requirements.txt"
   - "npm install"
   - "node_modules/.bin/bower install"
   - "node_modules/.bin/gulp --production"


ostcar:
@normanjaeckel can you merge this or give some comments?

.travis.yml
```diff
@@ -1,9 +1,15 @@
 language: python
+sudo: false
+cache:
+  directories:
+    - node_modules
```
normanjaeckel: What about caching bower_components?

.travis.yml
```diff
@@ -1,9 +1,15 @@
...
+    - $HOME/virtualenv/python$TRAVIS_PYTHON_VERSION/lib/python$TRAVIS_PYTHON_VERSION/site-packages/
```
normanjaeckel: Why not caching the full virtualenv dir:
`- $HOME/virtualenv/`

ostcar: I tried this. The problem is, that the the .virtualenv directory is very big (100MB) which have to be loaded for any run and have to be compressed and send back after each run. This took so long, that there was no speed gain. See the documentation of travis cache for this.

.travis.yml
ostcar: the bower install is very fast. It doesn't needs 3 seconds. For that reason the cache would slow down travis.

See http://docs.travis-ci.com/user/caching/#Things-not-to-cache

PR #1438, opened 2015-01-30
Angular

ostcar:
Only for demonstration, not ready yet.

Rebuild openslides-lib.js, start the server and call: http://localhost:8000/static/angular/index.html#/agenda/new/

Works:
- [X] DetailView
- [X] ListView
- [X] CreateView
- [X] UpdateView
- [x] Audoupdate
- [ ] DeleteView
- [ ] Autoupdate when an element is deleted
- [x] Files should be in there own app folder
- [x] Translation
- [x] Config in root scope
- [x] make tests work again


emanuelschuetze:
Really nice! I tested the agenda CreateView, ListeView, DetailView and Autoupdate. Good work!

ostcar:
@normanjaeckel please merge. There are a lot of things that have to be done better, but so @emanuelschuetze can start with the user app.

I only deactivated all the user urls. All other urls can be accessed as usual.

normanjaeckel:
I would like to split up the various parts of this pull request. The changes in all serializers can be done separately, see #1440. With this one can start developing the user app and I can refactor the other app serializers one by one.

package.json
```diff
@@ -1,14 +1,15 @@
...
+  "devDependencies": {
+    "bower": "~1.3.12",
+    "gulp": "~3.8.10",
+    "gulp-concat": "~2.4.3",
+    "gulp-if": "~1.2.5",
+    "gulp-minify-css": "~0.3.11",
+    "gulp-uglify": "~1.0.2",
+    "main-bower-files": "~2.4.1",
+    "yargs": "~1.3.3",
+    "po2json": "~0.3.2"
```
normanjaeckel: Please explain where we use the converted json files. They are not checked into the repo, are they? Maybe gulp should do the job.

openslides/config/serializers.py
```diff
@@ -0,0 +1,9 @@
+from openslides.utils.rest_api import serializers, VariableField
+
+
+class ConfigSerializer(serializers.Serializer):
```
normanjaeckel: I don't see why we need an extra serializer here. Config list, (retrieve) and update can be easily done in the ConfigViewSet and the cached data are still in a nice format.

openslides/core/views.py
```diff
@@ -24,6 +25,14 @@
 from .serializers import CustomSlideSerializer, TagSerializer
 
 
+class AngularView(utils_views.View):
```
normanjaeckel: Why not use a simple TemplateView? It means less code and plugins may easily override the default index template.

openslides/urls.py
```diff
@@ -1,68 +1,13 @@
-from django.conf import settings
 from django.conf.urls import include, patterns, url
 
-from openslides.core.views import ErrorView
-from openslides.utils.plugins import get_urlpatterns
+from openslides.core.views import AngularView
 from openslides.utils.rest_api import router
 
-handler403 = ErrorView.as_view(status_code=403)
```
normanjaeckel: Can we please use one urls.py which contains the new url to the angular index at the right place?

package.json
ostcar: The converted json files lie in openslides/static and so there are not part of the repo. You can create a gulp job for it if you like. It could look for files "openslides/locale/[^(en)]/javascript.po" and generates files "openslides/static/translation/$1.json".

At the moment po2json is not compatible with the current jed version, see https://github.com/SlexAxton/Jed/issues/36

I would propose we wait until we have some translated po files and look if there is a newer version of po2json, or use jed ~1.0.0

openslides/config/serializers.py
ostcar: I would like to use angular-data to handle the config values. For that, the config have to be in the format like any other model. So I need a list of dicts and not a dict it self. Also, I need the possibility to get inject single config values that were updated and pushed via autoupdate. For that, tornado need to call a get request on a single config value.

I don't see the problem, that we use the the same functions like in any other app here.

openslides/core/views.py
ostcar: The angular template language and the django template language are to similar at that we can send the angular base template though the django template language.

Also, I like the possibility to configure the webserver (nginx, apache) to serve the angular base template. For that reason it lies in static so it can be copied with "collect static" to a place where nginx can handle it.

Also, all other angular templates are in the static folder. As template developer you only have to define with a higher priority to static to override all templates.

I think we can live with 5 lines of extra code instead of 2 lines for a normal template view.

ostcar:
I would like to let the js code for the agenda and assignments in the pull requests. There are good examples and do no harm. But for that to work (as example) I need the changed serializes. They do also no harm and can be easily changed back later when you do the rework of this apps.

We should first merge https://github.com/OpenSlides/OpenSlides/pull/1440 then I change the js code to work with the new tornado-push-dict.

openslides/config/serializers.py
normanjaeckel: Config REST API changes split up in #1441. The format is now a list of dicts as probably required for angular-data.

I dont' use the serializer because manual serializing the config data was pretty easy. For later update method we will have to implement a custom function in any case.

package.json
normanjaeckel: So we should strip this and the jed requirement here and implement ist later?

openslides/core/views.py
normanjaeckel: At the moment the file is fetched from a fixed location in the core. Plugins can't override it if not using an extra webserver like Apache2 or Nginx. Hasn't Django a smart way to deliver static files using the static file finders first? Maybe something like https://docs.djangoproject.com/en/1.7/ref/contrib/staticfiles/#finders-module? I can look for it if you don't have enough time.

package.json
ostcar: No, I want to have the methods in the template yet. In other case we have to rework any template to add ugettext

openslides/config/serializers.py
ostcar: The Rest api is your part, so your implementation is ok for me (as long as it works :) )

I would recommend you to use one system. If you use the serializer system from django rest framework, you should use it here as well.

openslides/core/views.py
ostcar: That would be nice.

normanjaeckel:
#1440 and #1441 ready to be merged before adding AngularJS.

bower.json
```diff
@@ -1,10 +1,22 @@
...
+    "datatables-bootstrap3-plugin": "~0.2.0",
+    "jquery": "~1.11.2",
+    "jquery.cookie": "~1.4.1",
+    "angular": "~1.3.11",
+    "angular-route": "~1.3.11",
+    "angular-data": "~1.5.3",
+    "sockjs": "~0.3.4",
+    "lodash": "~3.0.1",
+    "angular-cookies": "~1.3.11",
+    "jed": "~1.1.0"
```
ostcar: sort and add comments

openslides/locale/de/javascript.po
```diff
@@ -0,0 +1,223 @@
+# SOME DESCRIPTIVE TITLE.
```
normanjaeckel: Please add a nice title like in our old po-files.

openslides/core/views.py

openslides/config/serializers.py
ostcar: done in https://github.com/OpenSlides/OpenSlides/pull/1441

bower.json
ostcar: sorted done

ostcar:
@normanjaeckel ready for next review

openslides/core/views.py
```diff
@@ -24,6 +26,18 @@
...
+    def get(self, *args, **kwargs):
+        """
+        Returns the angular base template.
+
+        As default the template lies in openslides/core/static/templates/index.html
```
normanjaeckel: Proposal:
```python
class IndexView(utils_views.View):
    """
    The primary view for OpenSlides using AngularJS.

    The default base template is 'openslides/core/static/templates/index.html'.
    You can override it by simply adding a custom 'templates/index.html' file
    to the custom staticfiles directory. See STATICFILES_DIRS in settings.py.
    """
    with open(finders.find('templates/index.html')) as index_file:
         content = index_file.read()
    return HttpResponse(content)
```

openslides/users/apps.py
```diff
@@ -29,5 +29,5 @@ def ready(self):
         register_slide_model(User, 'participant/user_slide.html')
 
         # Register viewsets.
-        router.register('users/user', UserViewSet)
-        router.register('users/group', GroupViewSet)
+        router.register('user/user', UserViewSet)
```
normanjaeckel: Hm. The app name is 'users' not 'user'. We also wanted to rename all other apps to plural names (motions, assignments, ...) as discussed in October 2014.

openslides/utils/autoupdate.py
```diff
@@ -101,9 +101,14 @@ def send_object(cls, object_url):
         # Send out internal HTTP request to get data from the REST api.
         for waiter in cls.waiters:
             # Read waiter's former cookies and parse session cookie to new header object.
-            session_cookie = waiter.connection_info.cookies[settings.SESSION_COOKIE_NAME]
             headers = HTTPHeaders()
-            headers.add('Cookie', '%s=%s' % (settings.SESSION_COOKIE_NAME, session_cookie.value))
+            try:
+                session_cookie = waiter.connection_info.cookies[settings.SESSION_COOKIE_NAME]
+            except:
```
normanjaeckel: Please add correct Exception here. I think it is KeyError.

openslides/urls.py
```diff
@@ -1,23 +1,27 @@
...
+    # TODO: add "special" urls, for example pdf views etc.
+
+    url(r'^user.*', AngularView.as_view()),
+)
+
+
+# Depricated. Remove each app from the patterns below, when they should be served
+# via angular.
+
 handler403 = ErrorView.as_view(status_code=403)
```
normanjaeckel: These error views are not deprecated I think. At least the HTTP 500 view might still appear. Maybe we have to change the template.

openslides/core/views.py
```diff
@@ -24,6 +26,18 @@
...
+        """
+        with open(finders.find('templates/index.html')) as f:
+            content = f.read()
+        return HttpResponse(content)
```
normanjaeckel: I think we need a authentication check so that if the user is anonymous and guest users are disabled there should be a redirect to the login page.

openslides/urls.py
```diff
@@ -1,23 +1,27 @@
...
+urlpatterns = patterns(
+    '',
+    url(r'^rest/', include(router.urls)),
+    # TODO: add "special" urls, for example pdf views etc.
+
+    url(r'^user.*', AngularView.as_view()),
+)
+
+
+# Depricated. Remove each app from the patterns below, when they should be served
```
normanjaeckel: Typo: Depr~~i~~ecated

openslides/urls.py
```diff
@@ -1,23 +1,27 @@
...
+from openslides.core.views import AngularView, ErrorView
 from openslides.utils.rest_api import router
 
+
+urlpatterns = patterns(
+    '',
+    url(r'^rest/', include(router.urls)),
+    # TODO: add "special" urls, for example pdf views etc.
+
+    url(r'^user.*', AngularView.as_view()),
```
normanjaeckel: @emanuelschuetze wanted to use the URL 'participants'.

openslides/utils/rest_api.py
```diff
@@ -45,7 +45,7 @@ def get_collection_and_id_from_url(url):
     Raises OpenSlidesError if the URL is invalid.
     """
     path = urlparse(url).path
-    match = re.match(r'^/api/(?P<name>[-\w]+/[-\w]+)/(?P<id>[-\w]+)/$', path)
+    match = re.match(r'^/rest/(?P<name>[-\w]+/[-\w]+)/(?P<id>[-\w]+)/$', path)
```
normanjaeckel: Please rename 'name' to 'collection' here and in line 51. I forgot this in my last pull request.

openslides/urls.py
emanuelschuetze: No, I want to keep "participants" in the GUI! The URL name is not so important for me. Currently, we use as url also "assignment" instead of "election" (like in GUI).

openslides/users/apps.py
ostcar: Are you sure? I can't remember. In this case we would name the app "agenda" to "agendas" and what would be the name for "core"?

openslides/core/views.py
ostcar: The login page has to be handeld in angular as well.

openslides/core/views.py
ostcar: done (but the code is still in the get method)

openslides/utils/autoupdate.py
ostcar: ups. done

openslides/urls.py
ostcar: done

openslides/users/apps.py
normanjaeckel: I am sure. We thought whether to rename the participant app to 'user' or 'users'. @DebVortex advised the plural form so you chose it. Of course we do not rename 'core' to 'cores' for we only have one system core. And 'agenda' is already a (latin) plural form.

openslides/urls.py
ostcar: I use "Deprecated" now

openslides/utils/rest_api.py

openslides/users/apps.py
emanuelschuetze: I like it too - with plural names if it make sense (like we use it in the GUI). That means:
agenda, motions, assignments, users, files and of course one core!

openslides/core/views.py
```diff
@@ -24,6 +26,26 @@
...
+    The primary view for OpenSlides using AngularJS.
+
+    The default base template is 'openslides/core/static/templates/index.html'.
+    You can override it by simply adding a custom 'templates/index.html' file
+    to the custom staticfiles directory. See STATICFILES_DIRS in settings.py.
+    """
+
+    def get(self, *args, **kwargs):
+        """
+        Returns the angular base template.
```
normanjaeckel: This docstring can be removed because the main issue is described in the class docstring.
Or use `Returns an HttpResponse object containing the 'templates/index.html'.`

openslides/urls.py
```diff
@@ -1,23 +1,24 @@
-from django.conf import settings
 from django.conf.urls import include, patterns, url
 
-from openslides.core.views import ErrorView
-from openslides.utils.plugins import get_urlpatterns
+from openslides.core.views import AngularView, ErrorView
```
normanjaeckel: This should be IndexView now.

ostcar: done

openslides/core/views.py

openslides/users/apps.py
ostcar: ok, I changed it back

ostcar:
@normanjaeckel all done

tests/old/users/test_umlaut_user.py
```diff
@@ -1,40 +0,0 @@
...
-    """
-    Tests persons with umlauts in there name.
-    """
-
-    def setUp(self):
-        self.user = User.objects.create_user('äöü', 'äöü')
-        self.client = Client()
-        self.client.login(username='äöü', password='äöü')
-
-    def test_login(self):
```
normanjaeckel: This test and the following look still valid because at the moment we don't use AngularJS for authentication, do we?

ostcar: The first and the second test have no value. They only test the django login view.

The third test has to be removed.

We have to rewrite the login page to angular for the 2.0 in any case. So there is no harm to remove the hole file now.

normanjaeckel:
Pull request merged as first draft for refactoring OpenSlides as single page app. Details have to be discussed later.

ostcar:
@emanuelschuetze can you start with the users app now? If you have any question to angular, feel free to ask.

PR #1440, opened 2015-02-04
Refactoring REST and SockJS

normanjaeckel:


openslides/utils/autoupdate.py
```diff
@@ -71,41 +69,47 @@ def on_close(self):
...
+        This method is called after succesful response of AsyncHTTPClient().
+        See send_object().
         """
-        # TODO: update cookies
+        # TODO: Update cookies of the client.
+        url_parts = response.request.url.split('/')
+        data = {
+            'url': response.request.url,
+            'name': '%s/%s' % (url_parts[-4], url_parts[-3]),
+            'data': json.loads(response.body.decode()),}
```
ostcar: could you write a comment that explains what parts of the url is inside url_parts[-3] and url_parts[-4]?

I can live with this solution but I am not happy. I have some concerns that this will work in any place. For example, what is with the projector? Could it be possible, that we will push a list of models in the future?

openslides/utils/autoupdate.py
```diff
@@ -71,41 +69,47 @@ def on_close(self):
...
+
             # TODO: read to python Morselcookies and why "set-Cookie" does not work
+            headers = HTTPHeaders()
             request_cookies = waiter.request_info.cookies.values()
             cookie_value = ';'.join("%s=%s" % (cookie.key, cookie.value)
                                     for cookie in request_cookies)
             headers.parse_line("Cookie: %s" % cookie_value)
 
+            # TODO: Use host and port as given in the start script
+            host_and_port = settings.OPENSLIDES_HOST_AND_PORT or 'localhost:8000'
```
ostcar: tornado is also a part of openslides. If the tornado runs on one server and django on another, this config value should only specify the domain that tornado should use to connect to django. Could you find a better name for it?

openslides/utils/autoupdate.py
```diff
@@ -71,41 +69,47 @@ def on_close(self):
...
+
             request = HTTPRequest(
-                url=''.join((REST_URL, object_url)),
+                url=''.join(('http://', host_and_port, object_url)),
```
ostcar: "http://" should be part of the config variable, to another protocol (for example https) can be used.

openslides/utils/autoupdate.py
normanjaeckel: Refactored

openslides/utils/autoupdate.py
normanjaeckel: Done.

openslides/utils/autoupdate.py
normanjaeckel: Done.

normanjaeckel:
See #1442.

PR #1442, opened 2015-02-05
Rework of tornado autoupdate functionality

normanjaeckel:


openslides/utils/autoupdate.py
```diff
@@ -58,56 +58,62 @@ def validate_absolute_path(self, root, absolute_path):
...
+            # Read waiter's former cookies and parse session cookie to new header object.
+            session_cookie = waiter.connection_info.cookies[settings.SESSION_COOKIE_NAME]
             headers = HTTPHeaders()
-            # TODO: read to python Morselcookies and why "set-Cookie" does not work
-            request_cookies = waiter.request_info.cookies.values()
-            cookie_value = ';'.join("%s=%s" % (cookie.key, cookie.value)
-                                    for cookie in request_cookies)
-            headers.parse_line("Cookie: %s" % cookie_value)
-
+            headers.add('Cookie', '%s=%s' % (settings.SESSION_COOKIE_NAME, session_cookie.value))
```
ostcar: Are you sure, that we only need to forward the session cookie?

normanjaeckel: I think we have only two cookies at the moment: The session cookie and the csrf cookie. Of these two we need only the session cookie.

PR #1444, opened 2015-02-06
rest cache

ostcar:
open for discussion. I make this pull request ready when this way of caching is reviewed.

When a retrieve-view from a ModelViewSet is called for the first time, the cache is created and returned on later request of the same view.

When a model is changed, the cache is also recreated. For that, all Serializers of the root_rest_element are used.

To know the list of Serializers for a root_rest_element, I tag the serializers with a decorator. An alternative would be to set the list of serializers manually in the model.

### DIFF
--- a/openslides/utils/rest_api.py
+++ b/openslides/utils/rest_api.py
@@ -1,19 +1,69 @@
 import re
+from collections import OrderedDict
 
 from urllib.parse import urlparse
 
+from django.core.cache import cache
 from django.core.urlresolvers import reverse
-from rest_framework import response, routers, serializers, viewsets  # noqa
+from rest_framework import permissions, response, routers, serializers, viewsets  # noqa
+from rest_framework.response import Response
 
 from .exceptions import OpenSlidesError
+from .autoupdate import inform_changed_data
 
 router = routers.DefaultRouter()
+
+
+class ModelViewSet(viewsets.ModelViewSet):
+    """
+    Viewset that unses the cache.
+    """
+    # TODO: cache get_serializer_class() because it is now called twice and there
+    #       could be db-queries in it.
+
+    def retrieve(self, request, *args, **kwargs):
+        # Try to get the data from the cache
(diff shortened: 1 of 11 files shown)


openslides/utils/rest_api.py
```diff
@@ -1,20 +1,70 @@
...
+    def retrieve(self, request, *args, **kwargs):
+        # Try to get the data from the cache
+        cache_key = self.model.get_rest_cache_key(
+            self.kwargs['pk'],
+            self.get_serializer_class())
+        data = cache.get(cache_key)
+
+        if data is None:
+            # If the data is not in the cache then get it from the db and save it into the cache
+            instance = self.get_object()
```
normanjaeckel: I suppose this will not work with our ItemViewSet in the following case but I have not tested it:

* Assume the user has only permission to see normal agenda items but no permission to see orga items.
* He sends a retrieve request on a normal agenda item and gets it. The item will be cached now.
* Someone else changes the item and swich the type to orga item. The cache is updated and is is up to tornado to notify the user. So tornado tries to retrieve the object, gets the cached version and sends it out.
* The user gets the fully updated orga item. But he should get a 403 insead.

The problem might be that the ItemViewSet method check_object_permission() is not called because the REST framework would have called it inside get_object() but this method is also not called because we have valid data in the cache.

May be we should not try to cache serializers by overriding ViewSet code but serializer code directly.

ostcar: In this case there would always be the database query which I like to get rid of.

I created a new commit in which the original object is created from the cached data and the permission check is made on it.

The best way would be if we do not use check_object_permissions from django_rest_framework but use our own method that works with the serialized data. That should be easy and the fastest way.

normanjaeckel:
Possibly restricting the use of viewset and serializer features seam to be the only way to achive our caching goals. Therefor we need a detailed documentation. Maybe we also can do it less complex (not connecting cache keys to models and the like).

ostcar:
This has to be redone anyway...

Issue #1429, opened 2015-01-24
Rename related name to assignment in AssignmentPoll class


normanjaeckel:
Rename it from `poll_set` to `polls` like in motion. Don't forget to update serializer classes.

ostcar:
fixed

Issue #1345, opened 2014-10-11
Cleanup assignments app after the refactoring of the users app


ostcar:
* rename 'person' to 'user'
* get_participants (with queryset)
* see, that the users are sorted
* change id to pk in some templates

ostcar:
all done

PR #1446, opened 2015-02-07
rework of assignment

ostcar:
fixes: #1429, #1345


openslides/assignment/forms.py
```diff
@@ -8,12 +8,12 @@
 
 
 class AssignmentForm(CssClassMixin, forms.ModelForm):
-    posts = forms.IntegerField(
+    member_count = forms.IntegerField(
```
normanjaeckel: What does this mean `member_count`? Do you want to change the label too?

openslides/assignment/models.py
```diff
@@ -15,24 +15,36 @@
...
+    STATUS_ELECTED = 2
+    STATUS_BLOCKED = 3
+    STATUSES = (
+        (STATUS_CANDIDATE, ugettext_lazy('candidate')),
+        (STATUS_ELECTED, ugettext_lazy('elected')),
+        (STATUS_BLOCKED, ugettext_lazy('blocked')),
+    )
+
+    assignment = models.ForeignKey(
+        "Assignment",
```
normanjaeckel: Here we should use single quotes.

openslides/assignment/models.py
```diff
@@ -15,24 +15,36 @@
...
+        choices=STATUSES,
+        default=STATUS_CANDIDATE)
 
     class Meta:
-        unique_together = ("assignment", "person")
+        unique_together = ('assignment', 'user')
 
     def __str__(self):
-        return str(self.person)
+        return "%s <-> %s" % (self.assignment, self.user)
```
normanjaeckel: Here we should use single quotes.

openslides/assignment/models.py
```diff
@@ -44,35 +56,87 @@ def get_root_rest_element(self):
...
+    """
+    Agenda items for this assignment.
+    """
 
     class Meta:
         permissions = (
-            ('can_see_assignment', ugettext_noop('Can see elections')),  # TODO: Add plural s to the codestring
-            ('can_nominate_other', ugettext_noop('Can nominate another person')),
+            ('can_see_assignments', ugettext_noop('Can see elections')),
+            ('can_nominate_other', ugettext_noop('Can nominate another user')),
```
normanjaeckel: @emanuelschuetze would like to see `participant` here.

openslides/assignment/models.py
```diff
@@ -80,125 +144,115 @@ def get_absolute_url(self, link='detail'):
...
+        """
+        Returns True if the user is blockt for candidature.
+
+        Costs one database query.
+        """
+        return self.blocked.filter(pk=user.pk).exists()
 
-    def get_participants(self, only_elected=False, only_candidate=False):
-        candidates = self.assignment_candidates.exclude(blocked=True)
+    def set_candidate(self, user, force=False):
```
normanjaeckel: Where to use `force`?

openslides/assignment/models.py
```diff
@@ -80,125 +144,115 @@ def get_absolute_url(self, link='detail'):
...
+        Delete the connection from the assignment to the user.
+        """
+        self.assignment_related_users.filter(user=user).delete()
 
-        # TODO: rewrite this with a queryset
-        participants = []
-        for candidate in candidates.all():
-            participants.append(candidate.person)
-        return participants
+    def set_phase(self, phase):
```
normanjaeckel: Here we need a short note that the new phase is not saved. Or do we need this method at all?

openslides/assignment/models.py
emanuelschuetze: Yes, please use "participant" here.

openslides/assignment/serializers.py
```diff
@@ -103,8 +102,8 @@ class AssignmentFullSerializer(serializers.ModelSerializer):
     """
     Serializer for assignment.models.Assignment objects. With all polls.
     """
-    assignmentcandidate_set = AssignmentCandidateSerializer(many=True, read_only=True)
-    poll_set = AssignmentAllPollSerializer(many=True, read_only=True)
+    related_user = AssignmentRelatedUserSerializer(many=True, read_only=True)
```
normanjaeckel: This should be `related_users` because this is the name of the field.

openslides/assignment/serializers.py
```diff
@@ -136,5 +135,5 @@ class Meta:
             'status',
             'assignmentcandidate_set',
```
normanjaeckel: `related_users`

openslides/assignment/views.py
```diff
@@ -28,150 +27,183 @@
...
+                    self.request,
+                    _("You were nominated successfully."))
+        else:
+            messages.error(
+                self.request,
+                _("You can not candidate to this assignment"))
+
+
+class AssignmentDeleteCandidateshipView(SingleObjectMixin, RedirectView):
+    required_permission = None  # Any user can withdrawn his candidature
```
normanjaeckel: Typo withdraw~~n~~

normanjaeckel:
Nice work. I made same remarks.

openslides/assignment/models.py
ostcar: done

openslides/assignment/forms.py
ostcar: This fields means exactly what the description and the label says. When the assignment has 8 members from which 4 has to be elected, then member_count is 8.

Maybe we need another field "elected_count"?

openslides/assignment/models.py
ostcar: No. I am starting to use double quotes on strings that are shown to users.

See: https://stackoverflow.com/questions/56011/single-quotes-vs-double-quotes-in-python

openslides/assignment/models.py
ostcar: done

openslides/assignment/models.py
ostcar: I forgot some code. Fixed now

openslides/assignment/models.py
ostcar: No. We decided that models should not save itself but that the code that calls the method has to call model.save() explicit .

In other case
```python
model.set_foo()
model.set_bar()
```
would save model two times.

openslides/assignment/serializers.py
ostcar: Done.

Is it ok that all users with the permission to see assignments can see which user is blocked?

normanjaeckel: I don't know. But why not. So I think they shoould be able to see them.

openslides/assignment/models.py
normanjaeckel: What does "No" mean? That we don't need a short note or that we don't need this method?

ostcar: That we do not need a short note, because there is no need to tell the standard. There should be a node if self.save() would be called.

openslides/assignment/models.py
normanjaeckel: Hm. I don't agree with this as a styling rule. There may be many reasons to do it this or that way (E. g. as a righty I like typing single quotes only with the right hand on standard German keyboard.) but I think most of the code written by @ostcar and me is with mainly single quotes everywhere so we can keep this way ...

openslides/assignment/models.py
ostcar: ups :) We need this method. It makes one test before the value is changed...

openslides/assignment/forms.py
normanjaeckel: At the moment this field decides whether to use a yes-no-abstain poll or not if the config is set to the respective value. I this (default) use case if an assignment has 8 members of which 4 have to be elected you have to set the field to 4 not to 8 to get a normal poll if there are 5 candidates. So the name of the field is not correct with the current implementation.

openslides/assignment/models.py
ostcar: I disagree. There is no harm in using double quotes. There is also no harm when the quotes are different on any string. This coding standard brings on improvment.

Also, of cause all my code is in single quotes, when I starting with this change yet. With your logic we could never change anything.

Recently you changed the style that we do not end a tuple with a comma. I don't like this neither but there is also no harm that you do this (I will not).
https://github.com/OpenSlides/OpenSlides/blob/master/openslides/users/serializers.py#L21

openslides/assignment/forms.py
ostcar: But in this case the app in only for election, maybe we should rename it :)

ostcar: done

openslides/assignment/models.py
normanjaeckel: Ok. Then everybody is free to use single or double quotes everywhere.

openslides/assignment/models.py
```diff
@@ -44,35 +56,87 @@ def get_root_rest_element(self):
...
+
+    description = models.TextField(
+        blank=True,
+        verbose_name=ugettext_lazy("Description"))
+    """
+    Text to describe the assignment.
+    """
+
+    open_posts = models.PositiveSmallIntegerField(
+        verbose_name=ugettext_lazy("Number of members that should be elected"))
```
normanjaeckel: Maybe use: `"Number of members to be elected"` or `"Number of members that are to be elected"` so that it sounds less optional.
The same in the following docstring.

openslides/assignment/views.py
```diff
@@ -356,31 +379,31 @@ def append_to_pdf(self, story):
...
         story.append(Spacer(0, 0.5 * cm))
 
         # Filling table rows...
         data = []
-        polls = assignment.poll_set.filter(published=True)
+        polls = assignment.polls.filter(published=True)
         # 1. posts
         data.append([
             Paragraph("%s:" %
                       _("Number of available posts"), stylesheet['Bold']),
```
normanjaeckel: Here the string has to be changed. Use the same string as in verbose_name in the model field.

normanjaeckel:
Two remarks. You can immendiately rebase (and add a nice and pretty ;-) commit message that also shows that a database reset is required.)

openslides/assignment/models.py
ostcar: done

openslides/assignment/views.py
ostcar: done

PR #1447, opened 2015-02-08
use ui router for more generic url routing

ostcar:
Use ui router.

See https://stackoverflow.com/questions/21023763/difference-between-angular-route-and-angular-ui-router

Highlights:
* Use name patterns in <a>-Elements so like django url names
* use template name per default like in django


normanjaeckel:
As discussed, please rebase (change commit message e. g. to "Used AngularUI Router for mote generic url routing.") and merge into (experimental) master.

Issue #1457, opened 2015-02-13
Refactor users app.


normanjaeckel:
What to do:

- [x] Move user related stuff from core to users app.
- [x] Clean up e. g. move code out of 'api.py' to more specific places.
- [x] Add new serializer for user creation.
- [x] Hide groups 'Anonymous' and 'Registered' from creation serializer.
- [x] Hide groups 'Anonymous' and 'Registered' for update requests. May be use another serializer.
- [x] Add possibility to delete users.
- [x] Add possibility to create, update and delete groups.
- [x] Hide default_password from normal users.
- [x] Clean up a little bit more

See also #1443 #1452

ostcar:
Only the manager should see the default password.

normanjaeckel:
All done, see #1642

PR #1458, opened 2015-02-13
Refactoring users app.

normanjaeckel:
See #1457.


normanjaeckel:
I will fix the old tests if the pull request is accepted. The main problem is that creating a test user does not automaticly add it to the 'Registered' group, but I don't like this auto adding any more. So I would add the group in the tests manually wherever required.

openslides/users/models.py
```diff
@@ -150,3 +160,83 @@ def reset_password(self, password=None):
         if password is None:
             password = self.default_password
         self.set_password(password)
+
+
+def gen_username(first_name, last_name):
```
ostcar: this is not a model. Please make this either a method of the User model or the user manager, or find another place. Django uses often the file 'utls' for this.

normanjaeckel: I thought about bringing model related helper functions together. Like https://github.com/django/django/blob/master/django/contrib/auth/models.py. But I can move these methods to `helpers.py` if you like.

openslides/core/management/commands/migrate.py
```diff
@@ -2,28 +2,23 @@
...
             except (FileExistsError, AttributeError):
-                # If the folder already exist or the settings OPENSLIDES_USER_DATA_PATH
-                # is unknown, then do nothing
+                # If the folder already exists or the settings
+                # OPENSLIDES_USER_DATA_PATH is unknown, just do nothing.
                 pass
-
         super().handle(*args, **options)
-        create_builtin_groups_and_admin()
+        post_migrate_command.send(self)
```
ostcar: Please rename and/or write a command that this should only be used for permissions

openslides/users/csv_import.py
```diff
@@ -62,6 +62,9 @@ def import_users(csvfile):
                             continue
                     user.reset_password()
                     user.save()
+                    warnings.warn(
+                        "Created users are not in the group 'Registered'.",
+                        RuntimeWarning)
```
ostcar: do not put the user into the group, but check the permissions in has_perms()

openslides/users/models.py
normanjaeckel: We can do in in a Manager.

openslides/users/serializers.py
```diff
@@ -45,6 +48,58 @@ class Meta:
...
+            validated_data['first_name'],
+            validated_data['last_name'])
+        # Prepare setup password
+        if not validated_data['default_password']:
+            validated_data['default_password'] = gen_password()
+        # Add group 'Registered'
+        validated_data['groups'].append(get_registered_group())
+        # Perform creation in the database
+        user = super().create(validated_data)
+        user.password = make_password(validated_data['default_password'], '', 'md5')
```
ostcar: put the hashed password into validated_data

openslides/core/management/commands/migrate.py
normanjaeckel: Done.

openslides/users/models.py
normanjaeckel: Done.

openslides/users/csv_import.py
normanjaeckel: Done with a custom auth backend.

openslides/users/serializers.py
normanjaeckel: Done.

normanjaeckel:
Ready for next review.

ostcar:
Some remaks. Please tell me when I should review again.

normanjaeckel:
Ready for next review (if we don't go on anonymous user stuff).

openslides/users/models.py
```diff
@@ -14,20 +17,96 @@
...
+        """
+        try:
+            admin = self.get(username='admin')
+        except self.model.DoesNotExist:
+            admin = self.model()
+            admin.username = 'admin'
+            admin.last_name = 'Administrator'
+            created = True
+        else:
+            created = False
```
ostcar: proposal:
```python
admin, created = self.get_or_create(username='admin', defaults={'last_name: 'Administrator'})
```
The disadvantage is, that a second save is called, when the admin is created. But the code is much more clearer and when this command is called, speed is not so essential.

openslides/users/models.py
```diff
@@ -14,20 +17,96 @@
...
+
+        try:
+            staff = Group.objects.get(pk=4)
+        except Group.DoesNotExist:
+            if created:
+                raise UserError("Admin user can not be created because the "
+                                "group 'Staff' is not available.")
+            else:
+                raise UserError("Admin user can not be reset because the group "
+                                "'Staff' is not available.")
```
ostcar: You should create the Group 'Staff'. When the user calls this script to reset his admin user and gets this error message, he probably does not know what to do.

openslides/users/models.py
```diff
@@ -14,20 +17,96 @@
...
+        else:
+            admin.save()
```
ostcar: If you create the admin user with get_or_create(), you do not need this line

openslides/users/models.py
normanjaeckel: Done.

openslides/users/serializers.py
```diff
@@ -45,6 +49,75 @@ class Meta:
...
+    Serializes data to create new users or update users.
+
+    Do not use this for list or retrieve requests.
+    """
+    groups = PrimaryKeyRelatedField(
+        many=True,
+        queryset=Group.objects.exclude(pk__in=(1, 2)),
+        help_text=ugettext_lazy('The groups this user belongs to. A user will '
+                                'get all permissions granted to each of '
+                                'his/her group.'))
```
ostcar: I think 'group' has to be in plural: 'groups'

openslides/users/models.py
normanjaeckel: Hm. That would be difficult because we don't have a default create_staff_function. And then we need to check Registered group too, because an admin needs more than just some manage permissions. At the moment I don't have enough mood to implement all this logic. You can open an issue if you like.

openslides/users/serializers.py
normanjaeckel: Done. But this text was copied from django.contrib.auth.models.Permission.groups.

openslides/users/models.py
ostcar: #1462

normanjaeckel:
All done.

ostcar:
ok. Please rebase and merge (when travis is green)

Issue #1454, opened 2015-02-12
Provide permission checks in angular templates


emanuelschuetze:
e.g. with ng-permission="perm1 perm2" or ="!perm1"

PR #1459, opened 2015-02-13
Implemented auth via angular

ostcar:
Also added the derective osPerms to check if the current user has permissions

fixes: #1454


openslides/urls.py
```diff
@@ -11,7 +11,7 @@
 urlpatterns = patterns(
     '',
     url(r'^rest/', include(router.urls)),
-    # TODO: add "special" urls, for example pdf views etc.
+    (r'^api/users/', include('openslides.users.urls')),
```
normanjaeckel: Use another prefix here or find a solution without any prefix as discussed.

openslides/utils/views.py
```diff
@@ -76,14 +76,20 @@ class AjaxMixin(object):
...
+    def get_ajax_context(self, **context):
         """
         Returns a dictonary with the context for the ajax response.
         """
-        return kwargs
+        return context
 
     def ajax_get(self, request, *args, **kwargs):
         """
+        Depricated. Use ajax_response instead.
```
normanjaeckel: Typo: Depr~~i~~ecated

openslides/users/views.py
```diff
@@ -499,3 +503,54 @@ def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
         kwargs['user'] = self.request.user
         return kwargs
+
```
normanjaeckel: Please remove old login and logout views from this file and from urls.py.

openslides/urls.py
ostcar: done

openslides/utils/views.py
ostcar: done

openslides/users/views.py
ostcar: done

ostcar:
all done. I had to remove the code to get rid of the old urls.

PR #1465, opened 2015-02-14
Implemented auth via angular

ostcar:
Also added the derective osPerms to check if the current user has permissions.

Removed old django views and urls for user

fixes: #1454


ostcar:
See also: #1459

openslides/core/views.py
```diff
@@ -35,6 +36,14 @@ class IndexView(utils_views.View):
     to the custom staticfiles directory. See STATICFILES_DIRS in settings.py.
     """
 
+    @classmethod
```
normanjaeckel: I would prefer method decorating or using the decorater in the urls.py. See https://docs.djangoproject.com/en/1.7/topics/class-based-views/intro/#decorating-the-class

openslides/global_settings.py
```diff
@@ -11,7 +11,7 @@
 
 AUTHENTICATION_BACKENDS = ('openslides.users.auth.CustomizedModelBackend',)
 
-LOGIN_URL = '/login/'
+LOGIN_URL = '/users/'
```
normanjaeckel: Should this be `LOGIN_URL = '/users/login/'`?

openslides/urls.py
```diff
@@ -11,14 +11,15 @@
 urlpatterns = patterns(
     '',
     url(r'^rest/', include(router.urls)),
-    # TODO: add "special" urls, for example pdf views etc.
+    (r'^users/', include('openslides.users.urls')),
```
normanjaeckel: Can you please use the `url(...)` style?

openslides/users/main_menu.py
```diff
@@ -10,5 +10,5 @@ class UserMainMenuEntry(MainMenuEntry):
     verbose_name = ugettext_lazy('Users')
     required_permission = 'users.can_see_extra_data'
     default_weight = 50
-    pattern_name = 'user_list'
+    pattern_name = 'core_dashboard'
```
normanjaeckel: Is this a permanent change? I think not. If I am right, please add a TODO and an issue.

openslides/users/models.py
```diff
@@ -3,12 +3,13 @@
 from random import choice
 
 from django.contrib.auth.hashers import make_password
-from django.contrib.auth.models import (
+from django.contrib.auth.models import (  # noqa
```
normanjaeckel: Here I would prefere separating the imports that are here just for other files to import them and the really required imports. But maybe isort doesn't like this ...

openslides/users/models.py
```diff
@@ -158,11 +159,11 @@ def get_absolute_url(self, link='detail'):
...
         """
         if link == 'detail':
-            url = reverse('user_detail', args=[str(self.pk)])
+            url = "/users/%s/" % self.pk
         elif link == 'update':
-            url = reverse('user_update', args=[str(self.pk)])
+            url = "/users/%s/edit/" % self.pk
         elif link == 'delete':
-            url = reverse('user_delete', args=[str(self.pk)])
+            url = "/users/%s/del/" % self.pk
```
normanjaeckel: Proposal: We change URL style to use `delete` instead of `del`for all new AngularJS deletion views.

openslides/users/urls.py
```diff
@@ -4,75 +4,10 @@
...
-
-    url(r'^(?P<pk>\d+)/edit/$',
-        views.UserUpdateView.as_view(),
-        name='user_update'),
-
-    url(r'^(?P<pk>\d+)/del/$',
-        views.UserDeleteView.as_view(),
-        name='user_delete'),
-
-    url(r'^(?P<pk>\d+)/reset_password/$',
```
normanjaeckel: The ResetPasswordView is still required because there is no possibility to reset the password via REST API.

openslides/users/urls.py
```diff
@@ -4,75 +4,10 @@
...
-        {'action': 'deactivate'},
-        name='user_status_deactivate'),
-
-    url(r'^(?P<pk>\d+)/status/toggle/$',
-        views.SetUserStatusView.as_view(),
-        {'action': 'toggle'},
-        name='user_status_toggle'),
-
     url(r'^csv_import/$',
         views.UserCSVImportView.as_view(),
```
normanjaeckel: This view can be removed because import should be available via AngularJS.

openslides/users/views.py
```diff
@@ -213,30 +67,10 @@ class UserCSVImportView(CSVImportView):
     """
```
normanjaeckel: Remove this view. See above.

openslides/users/views.py
```diff
@@ -213,30 +67,10 @@ class UserCSVImportView(CSVImportView):
...
     Import users via CSV.
     """
-    import_function = staticmethod(import_users)
     required_permission = 'users.can_manage'
     success_url_name = 'user_list'
     template_name = 'users/user_form_csv_import.html'
-
-
-class ResetPasswordView(SingleObjectMixin, QuestionView):
```
normanjaeckel: Don't remove this view. See above.

openslides/users/views.py
```diff
@@ -478,3 +162,58 @@ def get_form_kwargs(self):
...
+                'success': True,
+                'user_id': self.user.pk})
+        else:
+            context.update({'success': False})
+
+        return super().get_ajax_context(**context)
+
+    def post(self, request, *args, **kwargs):
+        try:
+            data = json.loads(request.body.decode('utf-8'))
```
normanjaeckel: The argument `'utf-8'` can be omitted. Because we don not set the encoding explicitly we should no assume it this way.

openslides/users/views.py
```diff
@@ -478,3 +162,58 @@ def get_form_kwargs(self):
...
         kwargs['user'] = self.request.user
         return kwargs
+
+
+class UserLoginView(AjaxMixin, View):
+    """
+    Login the user via ajax.
+    """
+
+    def get_ajax_context(self, **context):
```
normanjaeckel: Proposal:
```python
def get_ajax_context(self, **context):
    context['success'] = self.success
    if self.success:
        context['user_id'] = self.user.pk
    return super().get_ajax_context(**context)
```

openslides/users/views.py
```diff
@@ -478,3 +162,58 @@ def get_form_kwargs(self):
...
+
+
+class UserLogoutView(AjaxMixin, View):
+    """
+    Logout the user via ajax.
+    """
+
+    def post(self, request, *args, **kwargs):
+        auth_logout(request)
+        return self.ajax_response()
```
normanjaeckel: There should be a respones whether logut was successful or not.

openslides/users/views.py
```diff
@@ -478,3 +162,58 @@ def get_form_kwargs(self):
...
+            if form.is_valid():
+                self.user = form.get_user()
+                auth_login(request, self.user)
+                self.success = True
+            else:
+                self.success = False
+        return self.ajax_response()
+
+
+class UserLogoutView(AjaxMixin, View):
```
normanjaeckel: Why not using `openslides.utils.views.AjaxView`?

openslides/users/views.py
```diff
@@ -478,3 +162,58 @@ def get_form_kwargs(self):
...
+    """
+    Logout the user via ajax.
+    """
+
+    def post(self, request, *args, **kwargs):
+        auth_logout(request)
+        return self.ajax_response()
+
+
+class WhoAmIView(AjaxMixin, View):
```
normanjaeckel: Why not using `openslides.utils.views.AjaxView`?

openslides/users/views.py
```diff
@@ -478,3 +162,58 @@ def get_form_kwargs(self):
...
+    """
+    Returns the user id in the session.
+    """
+
+    def get(self, request, *args, **kwargs):
+        return self.ajax_response()
+
+    def get_ajax_context(self, **context):
+        return super().get_ajax_context(
+            user_id=self.request.user.pk,
```
normanjaeckel: This looks like it raises an error if an anonymous requests this view.

tests/integration/users/test_views.py
```diff
@@ -0,0 +1,86 @@
...
+class TestWhoAmIView(TestCase):
+    url = '/users/whoami/'
+
+    def test_get_anonymous(self):
+        response = self.client.get(self.url)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.content, b'{"user_id": null}')
+
+    def test_get_logedin(self):
```
normanjaeckel: Typo: login

tests/integration/users/test_views.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class TestWhoAmIView(TestCase):
+    url = '/users/whoami/'
+
+    def test_get_anonymous(self):
+        response = self.client.get(self.url)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.content, b'{"user_id": null}')
```
normanjaeckel: Here you can use `response.content.decode()` or better `response.data`.

tests/integration/users/test_views.py
```diff
@@ -0,0 +1,86 @@
...
+        response = self.client.get(self.url)
+
+        self.assertEqual(response.status_code, 405)
+
+    def test_post_anonymous(self):
+        response = self.client.post(self.url)
+
+        self.assertEqual(response.status_code, 200)
+
+    def test_post_logedin(self):
```
normanjaeckel: Typo. Better use `test_post_authenticated_user`

tests/integration/users/test_views.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def test_get(self):
+        response = self.client.get(self.url)
+
+        self.assertEqual(response.status_code, 405)
+
+    def test_post_anonymous(self):
+        response = self.client.post(self.url)
+
+        self.assertEqual(response.status_code, 200)
```
normanjaeckel: This should not be HTTP 200 but 400 or 401 or 4xx.

tests/integration/users/test_views.py
```diff
@@ -0,0 +1,86 @@
...
+    def test_get(self):
+        response = self.client.get(self.url)
+
+        self.assertEqual(response.status_code, 405)
+
+    def test_post_no_data(self):
+        response = self.client.post(self.url)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.content, b'{"success": false}')
```
normanjaeckel: See above.

tests/integration/users/test_views.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def test_post_correct_data(self):
+        response = self.client.post(
```
normanjaeckel: You can use a restframework APIClient for easier formating the request.

tests/integration/users/test_views.py
```diff
@@ -0,0 +1,86 @@
...
+            self.url,
+            json.dumps({'username': 'admin', 'password': 'admin'}),
+            'application/json')
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            json.loads(response.content.decode('utf-8')),
```
normanjaeckel: See above. `response.data` should do it.

tests/old/config/test_config.py
```diff
@@ -120,10 +120,6 @@ def test_get_config_form_overview(self):
         response = self.client_manager.get('/config/')
         self.assertRedirects(response=response, expected_url='/config/general/',
                              status_code=302, target_status_code=200)
-        bad_client = Client()
-        response = bad_client.get('/config/', follow=True)
-        self.assertRedirects(response=response, expected_url='/login/?next=/config/general/',
```
normanjaeckel: Why removing this test? The expected url is '/users/login...'.

openslides/core/views.py
ostcar: This has to be a feature of this view class and not only of the one view in the urls

openslides/global_settings.py
ostcar: no. /users/login/ is a ajax view. Here is a "template-view" needed. '/users' is not optimal. This probably has to be the dashboard, if it is created via angular. At the moment the users view is the best choice.

openslides/urls.py
ostcar: We do not use the url-style when we use include. See line 32ff

openslides/users/main_menu.py
ostcar: This is not permanent. I created this issue: #1467

openslides/users/models.py
ostcar: I used 'isort' to generate the imports. I would like to keep it the way, so this file is ready for the isort check

openslides/users/models.py
ostcar: There will be no delete url anymore. We do not need in the django part and neither in the angular part. I deleted the lines

openslides/users/urls.py
ostcar: See #1468

openslides/users/urls.py
ostcar: I thought we discussed that @emanuelschuetze first creates all apps via angular before he create the csv stuff.

openslides/users/views.py
ostcar: done

openslides/users/views.py
ostcar: angular uses utf-8. What do you propose?

openslides/users/views.py
ostcar: The AjaxView uses the get request. The logout should only be possible via post.

openslides/users/views.py
ostcar: I don't like the implementation of AjaxView (I created it, I know). I don't want to allow the post method.

I think in ajax views we should always explicit decide which request-methods are allowed.

openslides/users/views.py
ostcar: No, I use the value 'null' in angular to work with it. So I can set the current user to the return value of this method and do not have to check for different status codes.

If the view is requested by an anonymous user self.request.user is the AnonymousUser which pk is none

tests/integration/users/test_views.py
ostcar: What do you mean with response.data? When I try it I get an Error:
```
AttributeError: 'HttpResponse' object has no attribute 'data
```
I do not like to decode the return value, because I want to test the exact response of the view and not a result that was altert.

tests/integration/users/test_views.py
ostcar: Why? The user is logged out after the request is done.

tests/integration/users/test_views.py
ostcar: done

tests/integration/users/test_views.py
ostcar: done

tests/old/config/test_config.py
ostcar: The config test does not have to test, that the django login mixin works. I had to fix this, because the login view was altered and I did not want to fix old unneeded tests.

openslides/core/views.py
normanjaeckel: Sorry I don't understand. The IndexView only appears once in the ulrs.py so you can decorate it there. Or you can decorate the dispatch method of the class so that the decorater has effect on every view of this class.

openslides/global_settings.py
normanjaeckel: Ah I understand. Then we should remove the redirection feature in Django and this settings value. Or redirect to  '/' ans let AngularJS do the rest (checking session and displaying the login form if user is not authenticated).

openslides/urls.py
normanjaeckel: I would say this is just old style and we should use url-style everywhere. See line 13.

openslides/users/main_menu.py
normanjaeckel: This issue is about widgets but not about MainMenuEntries. Please add an issue about main menu entries.

ostcar:
@normanjaeckel I'm sry. I used --amend instead of creating a new review commit.

openslides/users/models.py
normanjaeckel: Hm. Ok.

openslides/core/views.py
ostcar: I used this way: https://docs.djangoproject.com/en/1.7/topics/class-based-views/intro/#mixins-that-wrap-as-view

openslides/users/models.py
```diff
@@ -158,11 +159,9 @@ def get_absolute_url(self, link='detail'):
         Returns the URL to the user.
         """
         if link == 'detail':
-            url = reverse('user_detail', args=[str(self.pk)])
+            url = "/users/%s/" % self.pk
         elif link == 'update':
-            url = reverse('user_update', args=[str(self.pk)])
-        elif link == 'delete':
-            url = reverse('user_delete', args=[str(self.pk)])
+            url = "/users/%s/edit/" % self.pk
```
normanjaeckel: Then we do not need the update url too. This is done via patch or put request on detail url.

openslides/global_settings.py
ostcar: At the moment '/' is not handled by angular. '/users/' is the most basic url there is.

openslides/users/urls.py
normanjaeckel: Ok.

openslides/users/views.py
normanjaeckel: Just `json.loads(request.body.decode())`

openslides/urls.py
ostcar: I don't see a benefit for it. For normal urls we want to use the url_pattern and that has to be done in the url-form. But we do not need it for include urls.

Also, this change is not a part of this pull request. Please open a issue in which we can talk about it.

openslides/users/main_menu.py
ostcar: See #1469

openslides/users/views.py
normanjaeckel: May be use a View from Django REST Framework.There you can choose between several HTTP requests and decide, what to do so we don't have to use our own AjaxView or Mixin.

openslides/users/views.py
ostcar: Logout is successfull when the user is not logged in after the request was called.

openslides/users/views.py
normanjaeckel: You can open an issue and we discuss it there.

openslides/users/views.py
normanjaeckel: Ok. Then please add a short note to the docstring.

openslides/users/models.py
ostcar: No. The update view has a template (a form). You can call this url in angular. But you can not call the delete url.

openslides/users/views.py
ostcar: Which is the same: https://docs.python.org/3/library/stdtypes.html#bytes.decode

but ok: done

openslides/users/views.py
ostcar: See: #1470

openslides/users/views.py
ostcar: done

tests/integration/users/test_views.py
ostcar: Can you give me an example for the restframework APIClient?

openslides/users/models.py
normanjaeckel: I don't understand the difference.
If I click on Edit button AngularJS gives me a template. If I click on Delete Button AngularJS does some other magic (confirmation dialog or whethever). Why do we need an URL for the first button but not for the seconde. The Edit template does not have to be connected to an URL. AngularJS can find this without a specific URL if only the button has a respective flag.

tests/old/config/test_config.py
normanjaeckel: This test tests that an unauthorized client is not abled to see the config page. The second element (that he is redirected to the login page) does not need to be tested (here you are right). You can test an 302 HTTP status code if you like.

tests/integration/users/test_views.py
normanjaeckel: See here: https://github.com/OpenSlides/OpenSlides/blob/master/tests/integration/users/test_viewset.py#L47 and especially line 73.

tests/integration/users/test_views.py
normanjaeckel: Hm. That looks like a design decision. See #1471.

tests/integration/users/test_views.py
normanjaeckel: Ah If you use a Django REST Framework View for this (I would prefer) you have access to response.data.

Nevertheless you should not compare  with a bytestring. The problem is the encoding. The bystring is encoded with the (default) encoding of the test file. This is UTF-8 only because Python > 3.x uses this default now and we do not have a magic encoding line at the beginning of the file. This has nothing to do with the encoding of the data coming in. I my opinion the test should not assume things from the testing environment (here: the encoding of the test file).

My preference is using Django REST Framework View and response.data and let the framework do the rest.

openslides/users/views.py
normanjaeckel: See #1471.

openslides/core/views.py
normanjaeckel: The IndexView is not a Mixin. It is used only once in our code.

openslides/global_settings.py
normanjaeckel: Then we need an issue and TODO to change this to '/' when the root is ready via AngularJS.

openslides/urls.py
normanjaeckel: See #1472. (But I don't agree. :-))

normanjaeckel:
Some more remarks.

tests/old/config/test_config.py
ostcar: It is a Feature of the PermissionMixin (which is used by this view) not to allow users which unprivileged permissions to see the page. This is not a feature of the config app.

If we change the behaviour of the PermissionMixin this test should not fail. For example if we change the statuscode to 403 instead of 302

openslides/core/views.py
ostcar: I do not see any problem with my code. It keeps the urls.py clean, what I like.

tests/old/config/test_config.py
normanjaeckel: Ok.

openslides/core/views.py
normanjaeckel: I think we should choose the Django way for decorating (non mixin) classes. That's all.

openslides/global_settings.py
ostcar: I created #1474 because a TODO in the global_settings can easily be overseen but this has to be done before the 2.0 release.

normanjaeckel: Ok.

openslides/users/views.py
ostcar: I added a special form of the Django REST framework APIView. Please create a new commend on the code for this view if needed.

tests/integration/users/test_views.py
ostcar: I use a subclass of the Django Rest Framework APIView know, but we should not use ```.data``` in integration tests. ```.data``` is not the result of the view. See: http://www.django-rest-framework.org/api-guide/responses/#data

I see no problem to use the default encoding of the test file. If we use request.content.decode() it would be exactly the same but with the disadvantages that not the result is tested, but an altered version of it.

tests/integration/users/test_views.py
ostcar: done

ostcar: I did not use ```.data```. We should not use this in integrations tests. See above.

tests/integration/users/test_views.py
normanjaeckel: Using 'data' is recommended by the Django RESt Framework: http://www.django-rest-framework.org/api-guide/testing/#testing-responses

openslides/users/models.py
ostcar: But we *have* a url for the update view.

get_absolute_url did never do some actions but only returned an url to a page on which the user could perform an action.

The question is, if we still need the method 'get_absolute_url'. There should be no place in django where an url is needed. The only place where this could be needed is, when  an AjaxView, APIView or RedurectView returns an url to the client. But I don't think this should be necessary. We should allays return the id to a model and create the url with angular

normanjaeckel: Yeah thats true. Remove this method completly.

ostcar:
@normanjaeckel all done. Please review

openslides/core/views.py
normanjaeckel: I can do it later in #1476 if you like.

openslides/users/models.py
normanjaeckel: See #1477.

openslides/utils/views.py
```diff
@@ -640,3 +650,71 @@ def form_valid(self, form):
         messages.warning(self.request, warning)
         messages.error(self.request, error)
         return super(CSVImportView, self).form_valid(form)
+
+
+class APIViewMeta(type):
```
normanjaeckel: This class is not required any more.

openslides/utils/views.py
```diff
@@ -640,3 +650,71 @@ def form_valid(self, form):
...
+    """
+
+    http_method_names = []
+    """
+    The allowed actions have to be explicitly defined.
+
+    Django allowes the following:
+    http_method_names = ['get', 'post', 'put', 'patch', 'delete', 'head', 'options', 'trace']
+
+    A method will be created for each allowed method. See APIViewMeta for more
```
normanjaeckel: This docstring is outdated.

openslides/utils/views.py
```diff
@@ -640,3 +650,71 @@ def form_valid(self, form):
...
+        """
+        Returns the context for the response.
+        """
+        return context
+
+    def method_call(self, request, *args, **kwargs):
+        """
+        Http method that returns the response object with the context data.
+
+        self.before_METHOD() will be called if it exists. METHOD is the http
```
normanjaeckel: This docstring is outdated.

normanjaeckel:
Some remarks and a typo in the commit message (~~units~~utils).

openslides/utils/views.py
ostcar: done

openslides/utils/views.py
ostcar: done

openslides/utils/views.py
ostcar: done

openslides/core/views.py
ostcar: Please do not do this in the urls. If you want do to in in another way inside the view, its ok for me

ostcar:
There went something wrong with my rebase of #1466 ontop of this. I probably pushed the wrong branch. I look at this on Wednesday. I'm sry  this can not be merged now.

openslides/core/views.py
normanjaeckel: Ok.

normanjaeckel:
See #1479.

PR #1466, opened 2015-02-14
Use django url patterns in angular

### LINKED ISSUE #1453: Provide django-url-names in angular
Please check if it possible to provide the django-url-names (e.g. "user_print") for using with ui-sref in angular templates.

ostcar:
First, you have to merge: #1465

Fixes #1453

### DIFF
--- a/openslides/core/views.py
+++ b/openslides/core/views.py
@@ -1,27 +1,33 @@
+import re
+
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.staticfiles import finders
 from django.core.exceptions import PermissionDenied
-from django.core.urlresolvers import reverse
+from django.core.urlresolvers import get_resolver, reverse
 from django.db import IntegrityError
+from django.http import HttpResponse
 from django.shortcuts import redirect, render_to_response
 from django.template import RequestContext
 from django.utils.importlib import import_module
 from django.utils.translation import ugettext as _
 from haystack.views import SearchView as _SearchView
-from django.http import HttpResponse
 
 from openslides import __version__ as openslides_version
 from openslides.config.api import config
 from openslides.utils import views as utils_views
-from openslides.utils.plugins import get_plugin_description, get_plugin_verbose_name, get_plugin_version
+from openslides.utils.plugins import (
+    get_plugin_description,
+    get_plugin_verbose_name,
+    get_plugin_version,
+)
 from openslides.utils.rest_api import ModelViewSet
 from openslides.utils.signals import template_manipulation
 from openslides.utils.widgets import Widget
 
+from .exceptions import TagException
 from .forms import SelectWidgetsForm
 from .models import CustomSlide, Tag
-from .exceptions import TagException
 from .serializers import CustomSlideSerializer, TagSerializer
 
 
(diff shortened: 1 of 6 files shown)


ostcar:
@normanjaeckel rebased. Please review and merge

openslides/core/views.py
```diff
@@ -353,3 +359,19 @@ def check_permissions(self, request):
...
+class UrlPatternsView(utils_views.APIView):
+    """
+    Returns a dictonary with all url patterns as json.
+    """
+    URL_KWARGS_REGEX = re.compile(r'%\((\w*)\)s')
+
+    def get_context_data(self, **context):
+        result = {}
+        url_dict = get_resolver(None).reverse_dict
+        for url_pattern in filter(lambda key: isinstance(key, str), url_dict.keys()):
```
normanjaeckel: Please rename `url_pattern` to `pattern_name` because the filter function just retrievse names but no patterns.

openslides/core/views.py
```diff
@@ -353,3 +359,19 @@ def check_permissions(self, request):
...
+            url = url_dict[url_pattern][0][0][0]
```
normanjaeckel: Ah I think this won't work properly. The url_dict is a MultiValueDict that means one key contains possibly more than one value.
See [here](https://github.com/django/django/blob/master/django/core/urlresolvers.py#L446): `possibilities = self.reverse_dict.getlist(lookup_view)`
In your case: `possibilities = url_dict.getlist(url_pattern)`
So we need a more complex result.

openslides/core/views.py
ostcar: no, the variable 'url_pattern' is the url-part, not the name.

normanjaeckel: Are you sure?
```
>>> from django.core.urlresolvers import get_resolver
>>> url_dict = get_resolver(None).reverse_dict
>>> list(filter(lambda key: isinstance(key, str), url_dict.keys()))[10]
'projector_clean_scroll'
```
This shows that the result of iterating the filter function are some strings. In my example this string with underscores is exactly the same as [this URL pattern name](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/projector/urls.py#L51). The URL with slashes would be `/projector/clean/scroll/. The URL pattern would be the result of [this function call](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/projector/urls.py#L48). You can get access to the URL and the URL pattern (and its kwargs and this stuff) by evaluating the `url_dict` with help of the URL pattern name as key (or even the view function because all URL patterns are registered twice by Django).

openslides/core/views.py
ostcar: Django allows multiple urls for one pattern_name. This makes sense for patterns with different arguments. We use this feature for agenda_speaker_delete, see https://github.com/OpenSlides/OpenSlides/blob/master/openslides/agenda/urls.py#L58 (this url will be removed in 2.0)

But this feature is not supported by the angular ui router. So the pattern_name in this function has to be unique.

ostcar:
@normanjaeckel please review once more

openslides/core/views.py
normanjaeckel: Ok. But the function should use this `getlist` method. And can you please split up the url getting to make the code more understandable. ... [0][0][0] ... tells me nothing ...

normanjaeckel: And we need somewhere a comment or assertion that nobody uses pattern_names twice.

ostcar: why should it use getlist? You mean I should split up the url getting in more then variable? I don't like it.

We will mostly use url_pattern in angular. when someone defines pattern twice, then he can not access his urls. So he knowns that it does not work.

openslides/core/views.py
```diff
@@ -338,3 +344,20 @@ def check_permissions(self, request):
...
+    Returns a dictonary with all url patterns as json.
+    """
+    URL_KWARGS_REGEX = re.compile(r'%\((\w*)\)s')
+    http_method_names = ['get']
+
+    def get_context_data(self, **context):
+        result = {}
+        url_dict = get_resolver(None).reverse_dict
+        for pattern_name in filter(lambda key: isinstance(key, str), url_dict.keys()):
+            url = url_dict[pattern_name][0][0][0]
```
normanjaeckel: This line is very complex and would be difficult to debug. Can you please split it up to make more clear was these multiple access on the nested list does?

ostcar: No. I won't do it. If you think it is important, then you can feel free to do it. In an extra pull request, of in a separate commit in this one. Or you can send me a patch.

normanjaeckel:
Please rebase (but dont set #1453 to fixed).

ostcar:
done

PR #1473, opened 2015-02-15
Isort

normanjaeckel:
I propose to add isort to TravisCI and then sort (and remove files from the isort.cfg) continuously.


normanjaeckel:
Closed qith respect to #1416.

PR #1481, opened 2015-02-17
Added coverage and flake8 configuration to setup.cfg.

normanjaeckel:
Cleaned up these tools.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 -r requirements_production.txt
 
 # Requirements for development and tests in alphabetical order
-coverage
+coverage>=4.0a5,<4.1.0
 flake8
 isort
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,4 +1,13 @@
+[coverage:run]
+source = openslides
+
+[coverage:html]
+directory = development/tmp/htmlcov
+
+[flake8]
+max_line_length = 150
+
 [isort]
-multi_line_output=3
-include_trailing_comma=True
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -2,7 +2,7 @@
 -r requirements_production.txt
 
 # Requirements for development and tests in alphabetical order
-coverage
+coverage>=4.0a5,<4.1.0
```
ostcar: I would like not to define versions for development dependencies. In different then the productive dependencies, the versions are not tested, so the tools work or do not work irrespective of the file requirements.txt (for the productive dependencies the command 'openslides' checks the dependencies in setup.py which is the same as the versions in requirements_production.txt).

I think it's ok for a developer to call ```pip install -r requirements.txt -U```  from time to time. And there should be no harm, if a new version of development tool breaks our workprocess. At least we know, that a new version was released.

setup.cfg
```diff
@@ -1,4 +1,13 @@
+[coverage:run]
+source = openslides
+
+[coverage:html]
+directory = development/htmlcov
```
ostcar: I understand the folder 'development' as a folder like ~/.config/openslides where all files are, that are needed to run openslides. So this folder can be moved from one instance to another. Of cause, in this case the htmlcov does no harm, but it is not needed neither. For that reason I would like it to be in a different place. What about ```/tmp```?

requirements.txt
normanjaeckel: Yes you are right. We shouldn't use development version.

Here in this case I would apply for a special exception. Coverage 4.0 will now support the use of setup.cfg and I prefer this. The latest alpha release was on 17th Feb. 2015 and the [development](https://bitbucket.org/ned/coveragepy/commits/all) goes straight ahead. When coverage 4.0 final is release I could change this file immendiately.

setup.cfg
normanjaeckel: Ok, we can split up the directory with settings and database on the hand and some tmp files on the other. Maybe we should rename these dirs.
What about `var` for settings and database ... and `tmp` for htmlcov ...?

ostcar: ok. If you like, you can put both dirs inside a directory ```development```

requirements.txt
ostcar: Now I see that you use a development version :) ok, this is a special case.

PR #1486, opened 2015-02-18
Cleaned up users.signals. Fixed #1483.

### LINKED ISSUE #1483: Clean up users/signals.py

normanjaeckel:


openslides/users/signals.py
```diff
@@ -113,79 +112,88 @@ def create_builtin_groups_and_admin(**kwargs):
...
+        'motion.can_manage_motion',
+        'motion.can_see_motion',
+        'motion.can_support_motion',
+        'users.can_manage',
+        'users.can_see_extra_data',
+        'users.can_see_name', )
+    for permission_string in permission_strings:
+        app_label, codename = permission_string.split('.')
+        permission_dict[permission_string] = Permission.objects.get(
+            content_type__app_label=app_label, codename=codename)
```
ostcar: Proposal: find a way only to use one query instead of one query per permission.

This is not so important, because this function is not called so often.

ostcar:
One proposal. You can merge this if you like.

PR #1487, opened 2015-02-18
Cleaned up urls.py. Fixed #1472.

### LINKED ISSUE #1472: Use url() in urls.py

normanjaeckel:


openslides/urls.py
```diff
@@ -40,9 +35,13 @@
...
 
 # TODO: move this patterns into core or the participant app
+js_info_dict = {'packages': []}
+
 urlpatterns += patterns(
     '',
-    (r'^jsi18n/$', 'django.views.i18n.javascript_catalog', js_info_dict),
+    url(r'^jsi18n/$',
+        'django.views.i18n.javascript_catalog',
+        js_info_dict),
```
ostcar: The js_info_dict can be removed right away. We definitely need another way to handle translation in angular

ostcar:
One comment. Please fix it or merge directly

openslides/urls.py
normanjaeckel: Done.

ostcar:
ok

Issue #619, opened 2013-04-24
Simplify get_version()


normanjaeckel:
Can we simpify the get_version function and the VERSION object? At the moment it is a 5-tuple, but it can easily be a string like '1.4b1' so we do not need to do the composition work also in every plugin again.

ostcar:
@normanjaeckel is this an issue for the 2.0 release?

PR #1488, opened 2015-02-18
Updated setup.py and openslides module init, esp. version string. 

normanjaeckel:
Removed git commit id grabbing during development. See #619.


openslides/core/views.py
```diff
@@ -115,14 +114,9 @@ def get_context_data(self, **kwargs):
...
         context = super().get_context_data(**kwargs)
-        # OpenSlides version. During development the git commit id is added.
-        if RELEASE:
-            description = ''
-        else:
-            description = 'Commit %s' % get_git_commit_id()
         context['modules'] = [{'verbose_name': 'OpenSlides',
-                               'description': description,
-                               'version': get_openslides_version()}]
+                               'description': '',
```
ostcar: I think this can be removed. But we have to rewrite this view with angular in any way. For that reason we need an ajax-view to get the version. If you like, you could create it with this pull request.

ostcar:
One proposal. You can merge this if you like.

openslides/core/views.py
normanjaeckel: This can be done in #1492.

normanjaeckel:
See #1506.

PR #1498, opened 2015-03-01
Altered ui-router definitions

ostcar:
1. Allow trailing slash in urls
2. Use integer ids for detail views as default


normanjaeckel:
Ok.

PR #1506, opened 2015-03-05
Added REST API for projector. Introduced new projector API.

normanjaeckel:
Added custom slide projector element class.
Added user slide projector element class.
Renamed SignalConnectMetaClass classmethod get_all_objects to get_all (private API).
Added migrations to core app.
Fixed and wrote tests.
Updated CHANGELOG.


openslides/core/models.py
```diff
@@ -1,62 +1,95 @@
...
-    name = models.CharField(max_length=255, unique=True,
-                            verbose_name=ugettext_lazy('Tag'))
+    name = models.CharField(
+        verbose_name=ugettext_lazy('Tag'),
+        max_length=255,
+        unique=True)
 
     class Meta:
-        ordering = ['name']
+        ordering = ('name',)
```
ostcar: I am ok with that change. But please note, that django changed all tuples to list in there settings:
https://github.com/django/django/commit/9ec8aa5e5d42ac4529846f7eae6bf4982800abff

The discussion which lead to this is here:
https://groups.google.com/forum/#!msg/django-developers/h4FSYWzMJhs/_2iVc4qgfsoJ

normanjaeckel: Ok. thanks. Then should we also change everything to lists?

ostcar: I am not sure. Django still uses tuples in Meta.ordering in there contrib apps. This was only a change for the config. Maybe they will make a decision for "meta-config" to in the next future. I would say we wait until then.

normanjaeckel:
@ostcar Ready for next review. (Enabeling viewset access if an object is on projector is still missing.)

normanjaeckel:
@ostcar Ping.

normanjaeckel:
Things done (mostly) as discussed. Please review and setup client.

openslides/core/views.py
```diff
@@ -46,6 +56,192 @@ def get(self, *args, **kwargs):
...
+        with list of dictionaries to append to the projector config entry.
+        """
+        # Get config entry form projector model, add new elements and try to
+        # serialize. This raises ValidationErrors if the data is invalid.
+        projector_instance = self.get_object()
+        projector_config = projector_instance.config
+        for projector_element in request.data:
+            projector_config.append(projector_element)
+        serializer = self.get_serializer(projector_instance, data={'config': projector_config}, partial=False)
+        serializer.is_valid(raise_exception=True)
```
ostcar: What happens if a exception is raised? Does the user get a 500er?

openslides/core/views.py
```diff
@@ -46,6 +56,192 @@ def get(self, *args, **kwargs):
...
+        with list of dictionaries to write them to the projector config
+        entry. All old entries are deleted.
+        """
+        # Get config entry form projector model, delete old and add new
+        # elements and try to serialize. This raises ValidationErrors if the
+        # data is invalid.
+        projector_instance = self.get_object()
+        projector_config = []
+        for projector_element in request.data:
+            projector_config.append(projector_element)
```
ostcar: Proposal for the lines 112-114
```python
projector_config = [projector_element for projector_element in request.data]
```

openslides/core/views.py
```diff
@@ -46,6 +56,192 @@ def get(self, *args, **kwargs):
...
+        serializer.save()
+        return Response(serializer.data)
+
+    @detail_route(methods=['post'])
+    def clear_and_activate_projector_elements(self, request, pk):
+        """
+        REST API operation to activate projector elements. It expects a POST
+        request to /rest/core/projector/<pk>/clear_and_activate_projector_elements/
+        with list of dictionaries to write them to the projector config
+        entry. All old entries are deleted.
```
ostcar: We talked about keeping "overlays". Comes this feature in the future?

openslides/core/views.py
```diff
@@ -46,6 +56,192 @@ def get(self, *args, **kwargs):
...
+        REST API operation to deactivate projector elements. It expects a
+        POST request to /rest/core/projector/<pk>/deactivate_projector_elements/
+        with list of dictionaries. These are exactly the projector_elements in the
+        config that should be deleted.
+        """
+        # Check the data. It must be a list of dictionaries. Get config
+        # entry from projector model. Pop out the entries that should be
+        # deleted and try to serialize. This raises ValidationErrors if the
+        # data is invalid.
+        if type(request.data) != list or list(filter(lambda item: type(item) != dict, request.data)):
```
ostcar: proposal: use 'is_instance(request.data, list)'

openslides/core/views.py
```diff
@@ -46,6 +56,192 @@ def get(self, *args, **kwargs):
...
+        projector_instance = self.get_object()
+        projector_config = projector_instance.config
+
+        new_projector_config = []
+        for config_entry in projector_config:
+            for entry_to_be_deleted in request.data:
+                if entry_to_be_deleted == config_entry:
+                    break
+            else:
+                new_projector_config.append(config_entry)
```
ostcar: Proposal for lines 135ff
```python
projector = self.get_object()
for entry_to_be_deleted in request.data:
    projector.config.remove(entry_to_be_deleted)
```

openslides/utils/projector.py
```diff
@@ -0,0 +1,66 @@
...
+        assert self.config_entry.get('name') == self.name, (
+            'To get data of a projector element, the correct config entry has to be given.')
+        return {
+            'scripts': self.get_scripts(),
+            'context': self.get_context()}
+
+    def get_scripts(self):
+        """
+        Returns the name of the JS file for this projector element. This app
+        is to be used by the client.
```
ostcar: I do not understand the second sentence. Please check, that it is correct.

openslides/utils/projector.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        Returns all data to be sent to the client. The projector object and
+        the config entry have to be given.
+        """
+        self.projector_object = projector_object
+        self.config_entry = config_entry
+        assert self.config_entry.get('name') == self.name, (
+            'To get data of a projector element, the correct config entry has to be given.')
+        return {
+            'scripts': self.get_scripts(),
```
ostcar: What should this be? One file or a list of files? The plural 'scripts' let me assume, that you mean a list of files. I think one file is enough. So please change it to 'script' or 'javascript' or 'js'

ostcar:
Please add the feature, that projector elements with a specific attribute are not removed from the projector when the method  clear_and_activate_projector_elements is called

openslides/core/views.py
```diff
@@ -46,6 +56,192 @@ def get(self, *args, **kwargs):
...
+            'clear_and_activate_projector_elements',
+            'deactivate_projector_elements',
+            'deactivate_all_projector_elements')
+        if (not request.user.has_perm('core.can_see_projector') or
+                (self.action in manage_methods and
+                 not request.user.has_perm('core.can_manage_projector'))):
+            self.permission_denied(request)
+
+    @detail_route(methods=['post'])
+    def activate_projector_elements(self, request, pk):
```
ostcar: Please rename to 'activate' or at lease 'activate_elements'. So the uri is not /rest/core/projector/1/activate_projector_elements/

openslides/core/views.py
```diff
@@ -46,6 +56,192 @@ def get(self, *args, **kwargs):
...
+        projector_config = projector_instance.config
+        for projector_element in request.data:
+            projector_config.append(projector_element)
+        serializer = self.get_serializer(projector_instance, data={'config': projector_config}, partial=False)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data)
+
+    @detail_route(methods=['post'])
+    def clear_and_activate_projector_elements(self, request, pk):
```
ostcar: Please rename. see above

openslides/core/views.py
```diff
@@ -46,6 +56,192 @@ def get(self, *args, **kwargs):
...
+        projector_config = []
+        for projector_element in request.data:
+            projector_config.append(projector_element)
+        serializer = self.get_serializer(projector_instance, data={'config': projector_config}, partial=False)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data)
+
+    @detail_route(methods=['post'])
+    def deactivate_projector_elements(self, request, pk):
```
ostcar: Please rename. See above.

openslides/core/views.py
```diff
@@ -46,6 +56,192 @@ def get(self, *args, **kwargs):
...
+                    break
+            else:
+                new_projector_config.append(config_entry)
+        serializer = self.get_serializer(projector_instance, data={'config': new_projector_config}, partial=False)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data)
+
+    @detail_route(methods=['post'])
+    def deactivate_all_projector_elements(self, request, pk):
```
ostcar: Please rename. Proposal: /rest/core/projector/1/clear/

openslides/core/views.py
normanjaeckel: I think this depends on the exception. If something like Http404 is raised, the user gets HTTP 404. I have to tested this but used the code as the framework does in `mixins.py`.

openslides/core/views.py
normanjaeckel: Better in an extra issue/pull request. see #1521

openslides/utils/projector.py
normanjaeckel: I am not sure what this should be. Maybe the authors of the client like to get a list of JS objects, callbacks or something like this. Proposal: Let's wait for the implementation of the client and rename it at the end to whatever is suitable (script/scripts/js-files/js-file/js-path ...).

openslides/utils/projector.py
normanjaeckel: Don't know. I will remove the docstring and add one when the client os ready so we know exactly what this method should return.

normanjaeckel:
See #1521.

openslides/core/views.py
normanjaeckel: Done.

openslides/core/views.py
normanjaeckel: Solved.

openslides/core/views.py
normanjaeckel: Solved.

openslides/core/views.py
normanjaeckel: Solved.

openslides/core/views.py
normanjaeckel: Done.

openslides/core/views.py
normanjaeckel: Done.

openslides/core/views.py
normanjaeckel: Solved.

openslides/core/views.py
ostcar: Could you test it?

openslides/core/views.py
ostcar: Please do this in this pull request. It is another API and I would like to start working final API

openslides/utils/projector.py
ostcar: We can define the method right now. It should only return one file. If an element needs more files, the first file can load more.

openslides/utils/projector.py
ostcar: As author of the client I would like, that the only one script is returned in this field.

openslides/core/views.py
normanjaeckel: See tests.unit.core.test_views.ProjectorAPI.test_activate_elements_bad_element.

openslides/utils/projector.py
normanjaeckel: As discussed: Why not using JS-objects? At the moment we load ass JS-files at one time. Later we want to have all JS in one big (uglified) file. So it might be better to add projector element related JS code into on of the normal app js files and then just reference this code as service for AngularJS.

openslides/utils/projector.py
normanjaeckel: May be we can omit this field completely. The client may detect the service by the name of the projector element.

openslides/core/views.py
normanjaeckel: Done.

ostcar:
Please rebase and merge this. I think it is easier to find and fix issues when I am writing the client code.

PR #1511, opened 2015-03-29
Send seperat tree for agenda items

ostcar:


openslides/agenda/views.py
```diff
@@ -16,6 +17,8 @@
 from django.utils.translation import ugettext as _
 from django.utils.translation import ugettext_lazy
 from reportlab.platypus import Paragraph
+from rest_framework.decorators import list_route
```
normanjaeckel: I would like to import these things from openslides.utils.rest_api.

openslides/agenda/views.py
```diff
@@ -811,3 +814,69 @@ def get_queryset(self):
         if not self.request.user.has_perm('agenda.can_see_orga_items'):
             queryset = queryset.exclude(type__exact=Item.ORGANIZATIONAL_ITEM)
         return queryset
+
+    @list_route(methods=['get', 'put'])
+    def tree(self, request):
+        """
+        Returns or set the agenda tree.
```
normanjaeckel: Typo: sets

openslides/agenda/views.py
```diff
@@ -811,3 +814,69 @@ def get_queryset(self):
...
+                    for child in item_children[item.pk]]
+
+        return Response(dict(id=item.pk, children=get_children(item))
+                        for item in item_list if not item.parent)
+
+    def set_tree(self, request):
+        """
+        Sets the agenda tree.
+        """
+        tree = request.data['tree']
```
normanjaeckel: Here we need a permission check on 'agenda.can_manage_agenda'.

openslides/agenda/views.py
```diff
@@ -811,3 +814,69 @@ def get_queryset(self):
...
+                break
+            touched_items.add(item_pk)
+
+            Item.objects.filter(pk=item_pk).update(
+                parent_id=parent_pk,
+                weight=weight)
+        else:
+            status = "ok"
+
+        return Response({'status': status})
```
normanjaeckel: Here we should respond a proper status code and as content {'detail': <message>}.

openslides/agenda/views.py
normanjaeckel: Please add an example tree.

openslides/agenda/views.py
ostcar: I think this is exaggerated. We import other things from other third party libs, too. 

But I can live with it, so: done

openslides/agenda/views.py
ostcar: done. But in the method set_tree

openslides/agenda/views.py
ostcar: done

openslides/agenda/views.py
ostcar: I don't see why.

Won't fix

normanjaeckel: On every other API request the framework does so. HTTP 200 means OK and HTTP 400 means Error. Additionally the framework uses a 'detail' message. So we should do this as the framework does.

ostcar:
all done

normanjaeckel:
Rebase and merge.

PR #1512, opened 2015-03-29
Assignment REST API changes

normanjaeckel:


emanuelschuetze:
Ok. Can you rebase and merge it so that I can use it in my template code?

normanjaeckel:
@ostcar Ping.

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+            'You can not be nominated by other participants anymore.')
+
+    def get_other_user(self, request):
+        """
+        Helper method to get a specific user from request data (not the
+        request.user) so that the views self.candidature_other or
+        self.mark_elected can play with him.
+        """
+        if not isinstance(request.data, dict):
+            raise ValidationError({'detail': _('Invalid data. Expected dictionary, got %s.') % type(request.data)})
```
ostcar: Could you make this line a bit shorter, so it can be read on github?

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+        if request.method == 'POST':
+            message = self.nominate_other(request, user, assignment)
+        else:
+            # request.method == 'DELETE'
+            message = self.delete_other(request, user, assignment)
+        return Response({'detail': message})
+
+    def nominate_other(self, request, user, assignment):
+        if assignment.phase == assignment.PHASE_FINISHED:
+            raise ValidationError({'detail': _('You can not nominate someone to this election because it is finished.')})
```
ostcar: The same here

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+            raise ValidationError({'detail': _('User %s does not want to be an candidate.') % user})
+        assignment.set_candidate(user)
+        return _('User %s was nominated successfully.') % user
+
+    def delete_other(self, request, user, assignment):
+        # To delete candidature status you have to be a manager.
+        if not request.user.has_perm('assignments.can_manage'):
+            self.permission_denied(request)
+        if assignment.phase == assignment.PHASE_FINISHED:
+            raise ValidationError({'detail': _('You can not delete someones candidature to this election because it is finished.')})
```
ostcar: The same here

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+        assignment = self.get_object()
+        if request.method == 'POST':
+            if not assignment.is_candidate(user):
+                raise ValidationError({'detail': _('User %s is not a candidate of this election.') % user})
+            assignment.set_elected(user)
+            message = _('User %s was successfully elected.') % user
+        else:
+            # request.method == 'DELETE'
+            if not assignment.is_elected(user):
+                raise ValidationError({'detail': _('User %s is not an elected candidate of this election.') % user})
```
ostcar: The same here

ostcar:
Maybe we should change the flake8 test, that maximal 114 characters are allowed

tests/integration/assignments/test_viewset.py
```diff
@@ -0,0 +1,270 @@
...
+
+        self.assertEqual(response.status_code, 403)
+
+
+class MarkElected(TestCase):
+    def setUp(self):
+        self.client = APIClient()
+        self.client.login(username='admin', password='admin')
+        self.assignment = Assignment.objects.create(title='test_assignment_Ierohsh8rahshofiejai', open_posts=1)
+        self.user = get_user_model().objects.create_user(
```
ostcar: Why do you need this user? All the test could also be done with the admin user

tests/integration/assignments/test_viewset.py
```diff
@@ -0,0 +1,270 @@
...
+class CandidatureOther(TestCase):
+    def setUp(self):
+        self.client = APIClient()
+        self.client.login(username='admin', password='admin')
+        self.assignment = Assignment.objects.create(title='test_assignment_leiD6tiojigh1vei1ait', open_posts=1)
+        self.user = get_user_model().objects.create_user(
+            username='test_user_eeheekai4Phue6cahtho',
+            password='test_password_ThahXazeiV8veipeePh6')
+
+    def test_invalid_data(self):
```
ostcar: Could you either change the name of the method to 'test_data_empty_dict' (or something else) or add a comment.

The same on the next methods, which also test invalid data

tests/integration/assignments/test_viewset.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def test_more_invalid_data(self):
+        response = self.client.post(
+            reverse('assignment-candidature-other', args=[self.assignment.pk]), {'user': 'string_instead_of_integer'})
+
+        self.assertEqual(response.status_code, 400)
+
+    def test_still_more_invalid_data(self):
+        # ID of a user that does not exist.
+        big_number = '1000000000000000000'
```
ostcar: Please use a number smaler then sys.maxsize (9223372036854775807) because it also proves your point, but saves memory

tests/integration/assignments/test_viewset.py
normanjaeckel: Name of test class changed, comment added.

openslides/assignments/views.py
normanjaeckel: Done.

openslides/assignments/views.py
normanjaeckel: Done.

openslides/assignments/views.py
normanjaeckel: Done.

openslides/assignments/views.py
normanjaeckel: Done.

tests/integration/assignments/test_viewset.py
normanjaeckel: Done.

tests/integration/assignments/test_viewset.py
normanjaeckel: Removed some '0' ...

normanjaeckel:
All done. Ready for last review.

tests/integration/assignments/test_viewset.py
normanjaeckel: And changed everything again.

tests/integration/assignments/test_viewset.py
```diff
@@ -0,0 +1,270 @@
...
+        admin.groups.remove(group_staff)
+
+        response = self.client.post(
+            reverse('assignment-candidature-other', args=[self.assignment.pk]),
+            {'user': self.user.pk})
+
+        self.assertEqual(response.status_code, 403)
+
+    def test_delete_other(self):
+        self.assignment.set_candidate(get_user_model().objects.get(username='test_user_eeheekai4Phue6cahtho'))
```
ostcar: The user object is 'self.user', why do you not use this?

tests/integration/assignments/test_viewset.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def test_delete_other_twice(self):
+        response = self.client.delete(
+            reverse('assignment-candidature-other', args=[self.assignment.pk]),
+            {'user': self.user.pk})
+
+        self.assertEqual(response.status_code, 400)
+
+    def test_delete_other_when_finished(self):
+        self.assignment.set_candidate(get_user_model().objects.get(username='test_user_eeheekai4Phue6cahtho'))
```
ostcar: here the same

tests/integration/assignments/test_viewset.py
```diff
@@ -0,0 +1,270 @@
...
+
+        response = self.client.delete(
+            reverse('assignment-candidature-other', args=[self.assignment.pk]),
+            {'user': self.user.pk})
+
+        self.assertEqual(response.status_code, 200)
+        self.assertFalse(Assignment.objects.get(pk=self.assignment.pk).candidates.filter(username='test_user_eeheekai4Phue6cahtho').exists())
+
+    def test_delete_other_during_voting_non_admin(self):
+        self.assignment.set_candidate(get_user_model().objects.get(username='test_user_eeheekai4Phue6cahtho'))
```
ostcar: Here too

tests/integration/assignments/test_viewset.py
```diff
@@ -0,0 +1,270 @@
...
+        self.assignment.save()
+
+        response = self.client.delete(
+            reverse('assignment-candidature-other', args=[self.assignment.pk]),
+            {'user': self.user.pk})
+
+        self.assertEqual(response.status_code, 400)
+
+    def test_delete_other_during_voting(self):
+        self.assignment.set_candidate(get_user_model().objects.get(username='test_user_eeheekai4Phue6cahtho'))
```
ostcar: Here too

tests/integration/assignments/test_viewset.py
normanjaeckel: Done.

tests/integration/assignments/test_viewset.py
normanjaeckel: Done.

tests/integration/assignments/test_viewset.py
normanjaeckel: Done.

tests/integration/assignments/test_viewset.py
normanjaeckel: Done.

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+            # To withdraw self during voting you have to be a manager.
+            self.permission_denied(request)
+        if not assignment.is_candidate(request.user):
+            raise ValidationError({'detail': _('You are not a candidate of this election.')})
+        assignment.set_blocked(request.user)
+        return _(
+            'You have withdrawn your candidature successfully. '
+            'You can not be nominated by other participants anymore.')
+
+    def get_other_user(self, request):
```
ostcar: I do not like the name of this method. When reading the code where the method is called, it is not clear who this other user is.

I would prefer to call this method "get_user" because there is no reason to think, that the returned user is the request-user

normanjaeckel: Changed.

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+        client has to send {'user': <id>}.
+        """
+        if not request.user.has_perm('assignments.can_manage'):
+            self.permission_denied(request)
+        user = self.get_other_user(request)
+        assignment = self.get_object()
+        if request.method == 'POST':
+            if not assignment.is_candidate(user):
+                raise ValidationError({'detail': _('User %s is not a candidate of this election.') % user})
+            assignment.set_elected(user)
```
ostcar: The python way is to call a method and if it fails to get the Exception with try-except. So I would prefer that "assignment.set_elected()" would raise an error, if the user is not a candidate.

But the method "assignment.set_elected" uses the django method "update_or_create". So either, we say, that if 'set_elected' is called on a new user, the new user has to become a related user of the assignment which is elected, or "assignment.set_elected" should use the django method "update" instead.

I would prefer that if the user is not a candidate, he/she can nevertheless marked as elected.

normanjaeckel: I don't understand what to be done here in views.py. If there is a bug in models.py please open an issue.

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+        return _('User %s was nominated successfully.') % user
+
+    def delete_other(self, request, user, assignment):
+        # To delete candidature status you have to be a manager.
+        if not request.user.has_perm('assignments.can_manage'):
+            self.permission_denied(request)
+        if assignment.phase == assignment.PHASE_FINISHED:
+            raise ValidationError({'detail': _('You can not delete someones candidature to this election because it is finished.')})
+        if not assignment.is_candidate(user) and not assignment.is_blocked(user):
+            raise ValidationError({'detail': _('User %s has no status in this election.') % user})
```
ostcar: What is the case, that you test here?

normanjaeckel: You can only delete an user from candidature/election if he is a candidate or blocked. If he is elected or absent, this method is not allowed.

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+    def nominate_other(self, request, user, assignment):
+        if assignment.phase == assignment.PHASE_FINISHED:
+            raise ValidationError({'detail': _('You can not nominate someone to this election because it is finished.')})
+        if assignment.phase == assignment.PHASE_VOTING and not request.user.has_perm('assignments.can_manage'):
+            # To nominate other during voting you have to be a manager.
+            self.permission_denied(request)
+        if assignment.is_candidate(user):
+            raise ValidationError({'detail': _('User %s is already a candidate of this election.') % user})
+        if assignment.is_blocked(user) and not request.user.has_perm('assignments.can_manage'):
+            raise ValidationError({'detail': _('User %s does not want to be an candidate.') % user})
```
ostcar: It seems, that an elected user can be unelected by any user by calling this method

ostcar: This is relevant when the assignment has more then one post and at first only one candidate is elected and so the assignment is retuned to the PHASE_SEARCH to find more potential candidates

normanjaeckel: I don't understand what to do here in views.py. Someone who is elected, should be be an candidate of a further poll. This is the actual behavior of OpenSlides 1.7.x and I did not want to change anything with this pull request.

normanjaeckel: Ah, is it about the non managers should not mark elected people?

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+
+    def withdraw_self(self, request, assignment):
+        # Withdraw candidature and set self blocked.
+        if assignment.phase == assignment.PHASE_FINISHED:
+            raise ValidationError({'detail': _('You can not withdraw your candidature to this election because it is finished.')})
+        if assignment.phase == assignment.PHASE_VOTING and not request.user.has_perm('assignments.can_manage'):
+            # To withdraw self during voting you have to be a manager.
+            self.permission_denied(request)
+        if not assignment.is_candidate(request.user):
+            raise ValidationError({'detail': _('You are not a candidate of this election.')})
```
ostcar: Is it possible to block yourself in a different way, if you are not a candidate? If not, I would prefer that this two lines are deleted.

openslides/assignments/views.py
normanjaeckel: Fixed.

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+            # request.method == 'DELETE'
+            message = self.withdraw_self(request, assignment)
+        return Response({'detail': message})
+
+    def nominate_self(self, request, assignment):
+        if assignment.phase == assignment.PHASE_FINISHED:
+            raise ValidationError({'detail': _('You can not candidate to this election because it is finished.')})
+        if assignment.phase == assignment.PHASE_VOTING and not request.user.has_perm('assignments.can_manage'):
+            # To nominate self during voting you have to be a manager.
+            self.permission_denied(request)
```
ostcar: It seems inconsistent that you get a 403 if want to nominate yourself on a assignment with phase == voting, but a 400 on a assignment with phase == finished

openslides/assignments/views.py
normanjaeckel: You can send the withdraw request via the API but the current behavior is that you can not bliock yourself if you are not candidate. Therefor we need thees two lines.

openslides/assignments/views.py
normanjaeckel: Why this? During VOTING a manager can nominate himself. During FINISHED he can't.

openslides/assignments/views.py
```diff
@@ -242,6 +243,134 @@ def get_serializer_class(self):
...
+        if assignment.is_candidate(request.user):
+            raise ValidationError({'detail': _('You are already a candidate of this election.')})
```
ostcar: I would prefer to remove the two lines. If you update a nodes content with new content, and the new content is exactly the same as the old one, you would also expect to get a 200. Or if you safe a file that has not changed it would be strange to get an IO error.

So when you update your candidate-status to the status you already have, you should not get an error

openslides/assignments/views.py
ostcar: I mean a user which is not a manager

normanjaeckel: Yeah this user can ask a manager to be upgraded. This is what 403 says: You don't have permissions. Try to get them and then you can do this action.

openslides/assignments/views.py
ostcar: But this is a nice feature, why shouldn't we support it be deleting this two lines?

normanjaeckel: You can open an issue for the discussion if you like.

openslides/assignments/views.py
ostcar: I'm not sure if you understand me. I have the following case:

* A new assignment is created with 2 posts. To be elected, a candidate needs more then 50% of the votes.
* There are some candidates. The phase of the assignment is set to "voting".
* One candidate is elected, the other candidates do not get the needed votes. The assignment is set to the phase "searching" again to get more candidates.
* Any user with the permission to nominate other elects the already elected user.

I think this user will be marked as candidate and therefore will lose this status as elected user

openslides/assignments/views.py
normanjaeckel: Done.

openslides/assignments/views.py
ostcar: I would prefer that you remove line 362 and 363 so an user that is not a candidate can be marked as elected by a manager

openslides/assignments/views.py
normanjaeckel: Please look at new review commit. I think this is fixed.

openslides/assignments/views.py
ostcar: I don't like this neither. If we have a new case where we need this method in the future, do we have to rename it to "get_user_for_candidature_and_election_and_new_case"? A method should be named what it does and not where it is used.

openslides/assignments/views.py
normanjaeckel: No. The reverse method is setting the user as candidate. So e. g. if we have a toggle button that sends POST and DELETE we got an inconsistent result.

openslides/assignments/views.py
normanjaeckel: Changed.

openslides/assignments/views.py
ostcar: ok

openslides/assignments/views.py
ostcar: Are you against this feature? If not, what is the problem to remove the two lines right now?

openslides/assignments/views.py
ostcar: ok

openslides/assignments/views.py
ostcar: This method is only for managers. A manager should be allowed to delete elected users. The manager can always unelect the user and delete him after this.

openslides/assignments/views.py
ostcar: But if I have a relay short election where I know who is elected before I set him as candidate, then I need an unnecessary click.

openslides/assignments/views.py
ostcar: I like it :)

openslides/assignments/views.py
normanjaeckel: Hm. I don't like that we would accept a HTTP DELETE request on an inexisting resource. So won't fix for now.

openslides/assignments/views.py
normanjaeckel: No you can send two HTTP requests in the correct order. I think @emanuelschuetze will apply the client as it was before. There you even have to create a poll to mark someone elected. I do not want to implement big changes to the current behavior.

openslides/assignments/views.py
normanjaeckel: He can do so. But he can't delete users that are not elected and not candidates or blocked. The concept of this pull request is that we have three different api endpoints where we can send POST or DELETE. The endpoint `candidature_other` is to create new candidates or delete them or for historical reason to delete blocked candidates so that the can be nominated again. The endpoint to mark someone elected or not is something different. If we want to have everything in one endpoint everything else should be not like this. I am not against this way but it would be more difficult to implement.

normanjaeckel:
All done (or commented).

PR #1513, opened 2015-04-08
New angular templates (part 1)

emanuelschuetze:
Template work still in progress. This is a first "beta" state of my work for further testing.
Please review and merge!

Users app:
- Show permissions in group form (Fixed #1500)
- Added experimental csv import for users.

Agenda app:
- Agenda sort view.
- Added angular-xeditable for agenda list columns.

General:
- Remove old template and js files. Remove old views.

PR #1517, opened 2015-04-30
Added motion views.

normanjaeckel:
Please have a first look so I can also implement motion update view in this way.

### DIFF
--- a/openslides/motions/models.py
+++ b/openslides/motions/models.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.db import models
 from django.db.models import Max
@@ -81,6 +82,16 @@ class Motion(RESTModelMixin, SlideMixin, AbsoluteUrlMixin, models.Model):
     tags = models.ManyToManyField(Tag)
     """
     Tags to categorise motions.
+    """
+
+    submitters = models.ManyToManyField(settings.AUTH_USER_MODEL, related_name='motion_submitters')
+    """
+    Users who submit this motion.
+    """
+
+    supporters = models.ManyToManyField(settings.AUTH_USER_MODEL, related_name='motion_supporters')
+    """
+    Users who support this motion.
     """
 
     class Meta:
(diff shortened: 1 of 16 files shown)


openslides/motions/serializers.py
```diff
@@ -138,36 +139,76 @@ class MotionSerializer(ModelSerializer):
...
+        Customized method to create a new motion from some data.
+        """
+        motion = Motion()
+        motion.title = validated_data['title']
+        motion.text = validated_data['text']
+        motion.reason = validated_data.get('reason', '')
+        motion.identifier = validated_data.get('identifier')
+        motion.category = validated_data.get('category')
+        motion.workflow = validated_data.get('workflow', int(config['motion_workflow']))
+        motion.reset_state()
```
ostcar: Is there a reason why you call this? It looks like, that it will also be called in the save method

openslides/motions/serializers.py
```diff
@@ -138,36 +139,76 @@ class MotionSerializer(ModelSerializer):
...
+
+        motion.save()
+
+        # Submitter
+        MotionSubmitter.objects.bulk_create(
```
ostcar: We probably need some permission checks here.  For example if supporters are allowed in the state.

This could be done in the validation process.

ostcar:
looks good

openslides/motions/serializers.py
normanjaeckel: I think I have to call it to set the identifier which is done in `set_identifier()`.

openslides/motions/serializers.py
normanjaeckel: Hm. I think a manager should be able to add supporters to a new motion as he likes. For non managers the client has to do something. A check is in `create()` of the viewset class.

openslides/motions/serializers.py
ostcar: Please have a look in the save method (which you call two lines later). If state is None, then reset_state() is called.

openslides/motions/serializers.py
ostcar: ok

normanjaeckel:
Ready for review.

openslides/motions/models.py
```diff
@@ -378,56 +389,37 @@ def get_last_version(self):
...
-    def support(self, person):
+    def support(self, user):
         """
-        Add 'person' as a supporter of this motion.
+        Add 'user' as a supporter of this motion.
         """
         if self.state.allow_support:
-            if not self.is_supporter(person):
-                MotionSupporter(motion=self, person=person).save()
+            if not self.is_supporter(user):
```
ostcar: I think this line can be removed. If not (please test it) it can be merged with line 408

openslides/motions/models.py
```diff
@@ -378,56 +389,37 @@ def get_last_version(self):
...
 
-    def unsupport(self, person):
+    def unsupport(self, user):
         """
-        Remove 'person' as supporter from this motion.
+        Remove 'user' as supporter from this motion.
         """
         if self.state.allow_support:
-            self.supporter.filter(person=person).delete()
+            self.supporters.remove(user)
```
ostcar: What happens if the user is not a supporter? Maybe there need to be a try-except block

ostcar:
ok. But please first merge #1513 and rebase on top of it

openslides/motions/serializers.py
```diff
@@ -134,40 +123,120 @@ class Meta:
             'reason',)
 
 
+class WorkflowField(IntegerField):
```
ostcar: Please use a function based custom validator:
 http://www.django-rest-framework.org/api-guide/validators/#function-based

openslides/motions/serializers.py
```diff
@@ -134,40 +123,120 @@ class Meta:
...
             'state',
             'workflow',
+            'tags',
             'attachments',
             'polls',
             'log_messages',)
+        read_only_fields = ('parent',)  # Some other fields are also read_only. See definitions above.
 
-    def get_workflow(self, motion):
+    def create(self, validated_data):
```
ostcar: Pleas use transaction.atomic() as function decorator

openslides/motions/serializers.py
```diff
@@ -134,40 +123,120 @@ class Meta:
...
+            if validated_data['submitters']:
+                motion.submitters.add(*validated_data['submitters'])
+            else:
+                motion.submitters.add(validated_data['request_user'])
+            motion.supporters.add(*validated_data['supporters'])
+            motion.attachments.add(*validated_data['attachments'])
+            motion.tags.add(*validated_data['tags'])
+        return motion
+
+    def update(self, motion, validated_data):
```
ostcar: Please use transaction

openslides/motions/serializers.py
normanjaeckel: Done.

openslides/motions/serializers.py
normanjaeckel: Done.

openslides/motions/serializers.py
normanjaeckel: Done.

openslides/motions/models.py
normanjaeckel: Still open.

openslides/motions/models.py
normanjaeckel: Still open.

normanjaeckel:
Two very small things still open. @emanuelschuetze You can already have a look at this and try to implement some client side stuff if you like.

openslides/motions/models.py
normanjaeckel: Now done.

openslides/motions/models.py
normanjaeckel: Now done.

normanjaeckel:
Ready for next review and merge.

ostcar:
Please do not rebase your review commits, so I can see what changed

Issue #1518, opened 2015-05-04
/agenda/print does not work for anonymous user 


emanuelschuetze:
Try to get agenda PDF via /agenda/print:
* works for logged in user
* does NOT work for anonymous user (redirects to `/users/?next=%2Fagenda%2Fprint%2F` and does not open PDF)

ostcar:
This is expected behaviour :)

You have to activate the anonymous user. If the anonymous user is not active, you are redirected to the login page, which is the user-list-view (at this development state).

The question is, why can an anonymous user see the agenda, if config['system_enable_anonymous'] == False. But this is an other (new) issue

PR #1519, opened 2015-05-05
Fix anonymous user for rest requests

ostcar:
See #1518


openslides/agenda/models.py
```diff
@@ -163,7 +163,7 @@ def get_title(self):
         Return the title of this item.
         """
         if not self.content_object:
-            agenda_title = self.title
+            agenda_title = self.title or ""
```
normanjaeckel: Why don't we send just 'None' so that the client knows that there isn't any title but the title of the related item?

openslides/users/rest_auth.py
```diff
@@ -0,0 +1,19 @@
+from rest_framework.authentication import BaseAuthentication
+
+from openslides.config.api import config
+
+
+# TODO: Move this into the file 'auth' after fixing circular imports
```
normanjaeckel: Where are these imports? Do we have an issue here?

tests/unit/users/test_serializers.py
```diff
@@ -7,6 +7,7 @@
 from rest_framework.test import APIRequestFactory
 
 
+@patch('openslides.users.rest_auth.config', {'system_enable_anonymous': False})
```
normanjaeckel: I think this is the default, so why do we have to patch it here?

openslides/agenda/models.py
ostcar: Its not for the rest api. The pdf did not work when the title is None

openslides/users/rest_auth.py
ostcar: I think it is because you import the whole rest_api in utils/rest_api.py. The config is imported when the rest_api is initialized. So the file where the Authentication lays is loaded. If this file imports utils/rest_api.py, then there is the circle.

tests/unit/users/test_serializers.py
ostcar: I don't know, but it does not work without the patch. Probably because it is a unittest and the database is not loaded. If you like you can try to fix the config, so it works without a database (I would like it)

openslides/agenda/models.py
normanjaeckel: If the title is None the PDF should print the title of the related item. This looks like an error in the construction of the elements of this app.

openslides/users/rest_auth.py
normanjaeckel: There should no be an import from on of these apps (utils, config, rest_framework) to our users app but only in other direction. Which import line in utils/rest_api.py causes this loop? I can't imagine this error.

tests/unit/users/test_serializers.py
normanjaeckel: Hm. So we just keep this line ...

openslides/agenda/models.py
ostcar: This is not the case for items which are related item (see line 165) but for items, where the title is None. This should never happen, but if there are an item without a content_object and without a title in the database, then there should be no error anyway.

openslides/users/rest_auth.py
ostcar: 1. Some file: import rest_framework.something. -> Load rest_framework_config
2. rest_framework: load config (load openslides.users.auth)
3. openslides.users.auth: import .models
4. openslides.users.models: import openslides.utils.rest_api
5. openslides.utils.rest_api: import rest_framework.something

Step 5 fails, because the config of the rest framework is not loaded yet.

ostcar: Solve this by importing the Permission class from django.contrib.auth and not from models

tests/unit/users/test_serializers.py
ostcar: 1. users.models.User.username: blank=True
2. users.serializers: remove __init__
3. Generate username in validate and not create method
4. Change Docstring
5. Try to merge FullUser and CreateUpdateUser (Test with a database where user is in group with pk 1 or 2)

openslides/users/rest_auth.py
ostcar: done

tests/unit/users/test_serializers.py
ostcar: done

ostcar:
@normanjaeckel ready for next review

normanjaeckel:
Looks good. Some remarks on the review commit.

ostcar:
@normanjaeckel I would like to rebase and merge this. Is that ok for you?

normanjaeckel:
One remark is still open. Rebase can be done.

normanjaeckel:
Please rebase and merge.

PR #1522, opened 2015-05-19
Added two agenda import ways.

emanuelschuetze:
This pull request implement the agenda import function via angularjs per csv file and per copy/paste from textbox. 

Not all issues are ready. This pull request is opened for first review/discussion.
Open tasks:
- [ ] default encoding utf-8 (How can I encode each string from csv file to utf-8; independent from the source encoding of the csv file?)
- [ ] use generic csv column header strings (Currently there are the German header strings from csv file used (e.g. "Titel") which are directly converted while import into JSON object. Then I can access only via obj[i].Titel. How can we solve this to make it more generic? E.g. by fetch the first element of the object? See TODO in importByCSV function.)
- [ ] Is it useful to create all items in bulk mode?


openslides/agenda/static/js/agenda/agenda.js
```diff
@@ -153,6 +153,50 @@ angular.module('OpenSlidesApp.agenda', [])
...
+            item.weight = 0;  // TODO: the rest_api should do this
+            item.tags = [];   // TODO: the rest_api should do this
+            // TODO: create all items in bulk mode
+            Agenda.create(item).then(
+                function(success) {
+                    $scope.importcounter++;
+                }
+            );
+        });
+    }
```
ostcar: When there are some valid and some invalid entries in the csv file, then some Items are imported and other aren't. Do you really want' this behaviour?

emanuelschuetze: Yes, it's the same behaviour like now in 1.7. For future we could use a "bulk mode" to create all items in one request. But I'm not sure if this possible.

openslides/agenda/static/templates/agenda/item-import.html
```diff
@@ -0,0 +1,86 @@
...
+<div id="submenu">
+  <a ui-sref="agenda.item.list" class="btn btn-sm btn-default">
+    <i class="fa fa-angle-double-left fa-lg"></i>
+    <translate>Back to overview</translate>
+  </a>
+</div>
+
+<h2 translate>Import by copy/paste</h2>
+<p translate>Copy and paste your agenda item titles in this textbox.
+Keep each item in a single line.</p>
```
ostcar: The newline between line 11 and line 12 will not be seen in the browser. Did you wan't a newline?

ostcar:
@emanuelschuetze looks good!

openslides/agenda/static/js/agenda/agenda.js
ostcar: Ok. But this means, that when a file with 1000 Entries are imported where 50 Entries are invalid, there are 950 Entries imported and the user has to find the 50 invalid ones. That could be hard.

openslides/agenda/static/templates/agenda/item-import.html
emanuelschuetze: No. It's ok.

openslides/agenda/static/js/agenda/agenda.js
emanuelschuetze: I agree. What's your idea to make a "pre check" of all items without import?

emanuelschuetze:
@ostcar I need help with task 1 and 2 of my initial pull request comment (that means encoding and generic column strings) to finish this pull request. Do you have any idea?

openslides/agenda/static/js/agenda/agenda.js
ostcar: One possible way would be, that Agenda.create() returned the id of the new created element. Then the ids could be saved in an array. If at least one create-request fails, the user could be ask if he want to delete all the elements, which were created successfully.

ostcar:
I have no experience with Javascript an utf-8. Maybe @DebVortex can help?

DebVortex:
I only had to deal with utf-8 in javascript once and if I remember right, I used the snippets provided here: http://ecmanaut.blogspot.de/2006/07/encoding-decoding-utf8-in-javascript.html

emanuelschuetze:
Outsourced 3 TODOs in new issues:
#1531 
#1532 
#1533 

Now, ready to last review.

ostcar:
@emanuelschuetze ok, please rebase an merge

emanuelschuetze:
Rebased.

PR #1526, opened 2015-05-26
Added view to add and remove users from the list of speakers.

normanjaeckel:


emanuelschuetze:
ok

normanjaeckel:
Ready to review.

emanuelschuetze:
ok

openslides/agenda/views.py
```diff
@@ -208,6 +216,137 @@ def get_queryset(self):
...
+            # Retrieve user_id
+            user_id = request.data.get('user')
+
+            # Check permissions and other conditions. Get user instance.
+            if user_id is None:
+                # Add oneself
+                if not self.request.user.has_perm('agenda.can_be_speaker'):
+                    self.permission_denied(request)
+                if item.speaker_list_closed:
+                    raise ValidationError({'detail': _('The list of speakers is closed.')})
```
ostcar: Is this exception caught?

openslides/agenda/views.py
```diff
@@ -208,6 +216,137 @@ def get_queryset(self):
...
+            # Check permissions and other conditions. Get speaker instance.
+            if speaker_id is None:
+                # Remove oneself
+                queryset = Speaker.objects.filter(
+                    item=item, user=self.request.user).exclude(weight=None)
+                try:
+                    # We assume that there aren't multiple entries because this
+                    # is forbidden by the Manager's add method. We assume that
+                    # there is only one speaker instance or none.
+                    speaker = queryset.get()
```
ostcar: Where is the advantage of this? There is no harm in calling queryset.delete()

tests/integration/agenda/test_viewsets.py
```diff
@@ -0,0 +1,179 @@
...
+
+class ManageSpeaker(TestCase):
+    """
+    Tests managing speakers.
+    """
+    def setUp(self):
+        self.client = APIClient()
+        self.client.login(username='admin', password='admin')
+        self.item = Item.objects.create(title='test_title_aZaedij4gohn5eeQu8fe')
+        self.user = get_user_model().objects.create_user(
```
ostcar: You can use this line, but I understand our integration test, that they test the code with out User app. And that there is at least one user with the pk 1 in it (the admin user). So I will continue to import the User app directly.

ostcar:
You can merge this if you like

openslides/agenda/views.py
normanjaeckel: Yes, see also http://www.django-rest-framework.org/api-guide/exceptions/#validationerror.

openslides/agenda/views.py
normanjaeckel: Line 294 is use for both possible results of the line 273 if-clause.

tests/integration/agenda/test_viewsets.py
normanjaeckel: I do not understand this remark, neither why we should bind our authentication app (called users at the moment) together with the other app. This binding is not necessary.

openslides/agenda/views.py
ostcar: ok

openslides/agenda/views.py
ostcar: I know, but line 286 could also use filter instead of get

normanjaeckel: Then I can't catch the Speaker.DoesNotExist exception to validate whether the user is on the list of speakers or a speaker with the given id exists.

PR #1527, opened 2015-06-11
Start the browser on 127.0.0.1

### LINKED ISSUE #1525: Start script should start server on 0.0.0.0 and open browser on localhost
Currently starting openslides on 0.0.0.0 opens browser also with http://0.0.0.0:8000.
It should open the browser on localhost:8000 because 0.0.0.0 can not opened on every browser (e.g. firefox on windows).

Its a blocker for 2.0alpha for the portable/gui version.

ostcar:
Fixed #1525

### DIFF
--- a/openslides/__main__.py
+++ b/openslides/__main__.py
@@ -135,7 +135,7 @@ def start(args):
     execute_from_command_line(['manage.py', 'migrate'])
 
     if not args.no_browser:
-        start_browser('http://0.0.0.0:8000')
+        start_browser('http://127.0.0.1:8000')
 
     # Start the webserver
     execute_from_command_line(['manage.py', 'runserver', '0.0.0.0:8000'])


openslides/__main__.py
```diff
@@ -135,7 +135,7 @@ def start(args):
     execute_from_command_line(['manage.py', 'migrate'])
 
     if not args.no_browser:
-        start_browser('http://0.0.0.0:8000')
+        start_browser('http://127.0.0.1:8000')
```
normanjaeckel: I am not sure if we should use `localhost` here because this uses the local socket instead of a TCP/IP connection. http://stackoverflow.com/questions/3715925/localhost-vs-127-0-0-1

normanjaeckel:
One proposal, then merge.

openslides/__main__.py
emanuelschuetze: Sounds useful for me to use `localhost` instead of `127.0.0.1`.

ostcar: @normanjaeckel Can you find another source? I do not belief what that comment says, because I understood something else from the term "unix domain socket". And also, I think, that this would only make a difference, if the server listens to localhost, and not, that the client connets to localhost, what is mapped to 127.0.0.1 in /etc/hosts

normanjaeckel:
See #1534.

PR #1529, opened 2015-06-11
Added config update view.

normanjaeckel:


openslides/config/views.py
```diff
@@ -129,12 +129,17 @@ def retrieve(self, request, *args, **kwargs):
...
             self.permission_denied(request)
-        else:
-            # TODO: Implement update method
-            self.permission_denied(request)
+        key = kwargs['pk']
+        # Check if pk is a valid config variable key.
+        if key not in config:
+            raise Http404
+        # Change value.
+        config[key] = request.data
```
ostcar: I think we need validation here. This should be possible with the django form fields, as long as we use them

ostcar:
One comment. You can create an issue for this an merge this pull request.

normanjaeckel:
Please review again.

PR #1536, opened 2015-06-12
Fixed FileFinder importer for plugin detection in portable

emanuelschuetze:
Updated setuptools versions for openslides-gui.


emanuelschuetze:
@andkit As discussed please look at the line ` return tuple(x[0] for x in importer.iter_modules())` which seems to need an update for python3.3 (see https://docs.python.org/3.4/library/pkgutil.html#pkgutil.iter_modules).
Your commit has introduced the iter_modules, see https://github.com/OpenSlides/OpenSlides/commit/3584da89c364b368580713d7b356ef739abd4632

andkit:
@emanuelschuetze can you try the following patch (especially if finding plugins actually works, I only dry-tested it in the console):
-    importer = pkgutil.get_importer(path)
-    return tuple(x[0] for x in importer.iter_modules())
+    return tuple(x[1] for x in pkgutil.iter_modules([path]))

This will handle importers that do not provider iter_modules similar to your patch, but it also uses a specialized implementation for the FileFinder importer (standard importer of directories, and possibly any other that registers itself appropriately)

emanuelschuetze:
@andkit Your patch works. Thanks.

emanuelschuetze:
Ready for review and merge.

normanjaeckel:
Looks good (untested).

ostcar:
You can merge it if you like.

I think this code is for the portalbe only and for that reason should be moved there. But this can be done later.

Issue #1469, opened 2015-02-15
Use main menu in angular


ostcar:
We need a solution to get a _generic_ main menu in angular base template. Only installed apps (and also plugins later) should be included into the main menu.

emanuelschuetze:
#1513 shows only a temporary solution because main menu entries are placed manually. We need a more generic solution.

Also we need a solution to show the selected menu entry as activ! This is important for a first alpha of 2.0 - so I set it to "blocker".

ostcar:
@emanuelschuetze I don't see this issue as a blocker. Please add the flag again with an explanation if you think otherwise.

emanuelschuetze:
Now, it's good enough for 2.0a1.

PR #1539, opened 2015-06-16
Delete old stuff

ostcar:
* Removed django templates (with templatetags and template_manipulation_signal)
* Removed django forms
* Removed server side CSV import
* Removed get_absolute_url from all models (with utils.AbsoluteUrlMixin)
* Removed widgets (with account-app and personal_info_widget)
* Removed views
* Removed main_menu
* Removed Django Error pages
* Removed mptt
* Removed OpenSlides 1.x Slide-System (with old overlays and the projector app)
* Make url.py ready for the 2.0 release
* Sorted imports from all files with isort
* Removed old unused tests


openslides/account/widgets.py
```diff
@@ -1,42 +0,0 @@
-from django.contrib.auth.models import AnonymousUser
-from django.utils.translation import ugettext_lazy
-
-from openslides.utils.personal_info import PersonalInfo
-from openslides.utils.widgets import Widget
-
-
-class PersonalInfoWidget(Widget):
```
normanjaeckel: Could you please add an issue to discuss and decide whether we want to have something like a personal info widget or box in the new AngularJS app? Here we also should decide whether the server should provide a special read-only REST view with specific data (my motions, where I am on the list of speakers ...).

openslides/agenda/main_menu.py
```diff
@@ -1,14 +0,0 @@
-from django.utils.translation import ugettext_lazy
-
-from openslides.utils.main_menu import MainMenuEntry
-
-
-class AgendaMainMenuEntry(MainMenuEntry):
```
normanjaeckel: Do we have an issue for discussing the future main menu implementation?

openslides/agenda/models.py
```diff
@@ -4,26 +4,20 @@
...
 from openslides.utils.rest_api import RESTModelMixin
 from openslides.utils.utils import to_roman
-from openslides.users.models import User
 
 
-# TODO: remove mptt after removing the django views and forms
-class Item(RESTModelMixin, SlideMixin, AbsoluteUrlMixin, MPTTModel):
+class Item(RESTModelMixin, SlideMixin, models.Model):
     """
     An Agenda Item
```
normanjaeckel: Docstring is outdated.

openslides/account/widgets.py
ostcar: #1540

openslides/agenda/models.py
```diff
@@ -330,13 +282,12 @@ def _calc_sibling_no(self):
...
-            if prev_sibling.type == self.AGENDA_ITEM:
-                sibling_no += 1
-            prev_sibling = prev_sibling.get_previous_sibling()
-        return sibling_no + 1
+        siblings = (Item.objects.filter(parent=self.parent, type=self.AGENDA_ITEM)
+                        .order_by('weight'))
+        for index, item in enumerate(siblings):
+            if item == self:
+                return index + 1
+        raise RuntimeError("Item {} has to be a sibling of himself".format(self))
```
normanjaeckel: This error sounds strange: Someone can not be a sibling of himself. Also the docstring of this method seems wrong. What does it do?

openslides/agenda/models.py
```diff
@@ -421,8 +362,10 @@ def begin_speach(self):
         self.save()
         # start countdown
         if config['agenda_couple_countdown_and_speakers']:
-            reset_countdown()
-            start_countdown()
+            # TODO: Fix me with the new countdown api
```
normanjaeckel: Please add an issue for this.

openslides/agenda/main_menu.py
ostcar: See #1469

openslides/agenda/models.py
ostcar: done

openslides/agenda/urls.py
```diff
@@ -4,14 +4,7 @@
 
 urlpatterns = patterns(
     '',
-
-    # PDF
     url(r'^print/$',
         views.AgendaPDF.as_view(),
         name='agenda_pdf'),
-
-    # TODO: remove it after implement projector rest api
```
normanjaeckel: This is a special case for me (because this was an implementation for a customer) so I need an extra issue here.

openslides/agenda/models.py
ostcar: I rewrote the hole method, gave it a new name and changed the docstring. Please have another look at it.

openslides/agenda/models.py
ostcar: See #1541

openslides/agenda/urls.py
ostcar: See #1542

openslides/core/urls.py
```diff
@@ -1,46 +1,9 @@
...
-
-    url(r'^dashboard/$',
-        views.DashboardView.as_view(),
-        name='core_dashboard'),
-
-    url(r'^dashboard/select_widgets/$',
-        views.SelectWidgetsView.as_view(),
-        name='core_select_widgets'),
-
-    url(r'^version/$',
```
normanjaeckel: Do we have an issue which shows which "views" have to be reimplemented later? (Version, Search, Chatbox)

openslides/urls.py
```diff
@@ -1,52 +1,17 @@
 from django.conf.urls import include, patterns, url
+from django.views.generic import RedirectView
 
-from openslides.core.views import IndexView, ErrorView
-from openslides.users.views import UserSettingsView, UserPasswordSettingsView
+from openslides.core.views import IndexView
 from openslides.utils.rest_api import router
 
-handler403 = ErrorView.as_view(status_code=403)
```
normanjaeckel: I think we have to keep the error views for the moment because we still have Django views which still can produce errors like 500 oder 403.

openslides/core/urls.py
ostcar: I created #1543 as meta issue for this. Please add more items to it as you like

openslides/projector/models.py
```diff
@@ -1,48 +1,8 @@
...
-
-
 class SlideMixin(object):
     """
-    A Mixin for a Django-Model, for making the model a slide.
-    """
+    Depricated.
 
-    slide_callback_name = None
+    Weill be reused or removed when more slides are implemented for the
```
normanjaeckel: Typo `Will`

openslides/projector/models.py
```diff
@@ -1,48 +1,8 @@
-from django.core.urlresolvers import reverse
-
-from openslides.utils.utils import int_or_none
-
-
 class SlideMixin(object):
     """
-    A Mixin for a Django-Model, for making the model a slide.
-    """
+    Depricated.
```
normanjaeckel: Typo `Deprecated`

openslides/urls.py
ostcar: I talked with @emanuelschuetze about this. We both think, that the error handlers should be removed.

The only views that exist as django views are "ajax-views" or the pdf-views. So we do only need the base template for the error views. To create only a base template for this cases seems to much work.

A 404 can not happen, because all unknown urls are redirected to angular. A 403er can only happen when the user is logged in and requests a pdf which he is forbidden to see. The user can only see the 500er page on a pdf view (which should never happen). The user will never request a "ajax-view" directly.

Django has build in error pages, they are not so pretty but ok for this rare cases.

openslides/projector/models.py
ostcar: done

openslides/projector/models.py
ostcar: Not again... done

openslides/utils/forms.py
```diff
@@ -1,117 +0,0 @@
...
-
-        for (id, text) in super()._get_choices():
-            # This line is only required to translate permission objects
-            text = text.split(' | ')[-1]
-            yield (id, ugettext_lazy(text))
-
-    choices = property(_localized_get_choices, forms.ChoiceField._set_choices)
-
-
-class CleanHtmlFormMixin(object):
```
normanjaeckel: Ah, we need an implementation of bleach or something else for our REST fields.

ostcar: Yes. We should create a SerializerField which filters html with bleach.

openslides/urls.py
normanjaeckel: Ok. Remove it now and let's see if we need it later. Reimplementation would not be a problem.

normanjaeckel:
Everything looks ok. Of course it is a bit scary but that's it.

Please add a note in the CHANGELOG.

Please rebase everything then I will have a last look.

ostcar:
Done. I had no Idea what to write in the changelog. That old stuff is removed is indirectly mentioned, when we write, that we rewrote the old stuff with angular.

PR #1549, opened 2015-06-17
Refactored config API. Removed form_fields. Added extra fields for HT…

normanjaeckel:
…ML rendering like label and help text. Added validation.

### DIFF
--- a/openslides/agenda/signals.py
+++ b/openslides/agenda/signals.py
@@ -1,12 +1,12 @@
 from datetime import datetime
 
-from django import forms
 from django.contrib.contenttypes.models import ContentType
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ValidationError as DjangoValidationError
+from django.core.validators import MaxLengthValidator, MinValueValidator
 from django.utils.translation import ugettext as _
-from django.utils.translation import ugettext_lazy, ugettext_noop
+from django.utils.translation import ugettext_lazy
 
-from openslides.config.api import ConfigCollection, ConfigVariable
+from openslides.config.api import ConfigVariable
 
 from .models import Item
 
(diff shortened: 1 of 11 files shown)


openslides/agenda/signals.py
```diff
@@ -15,65 +15,50 @@ def validate_start_time(value):
     try:
         datetime.strptime(value, '%d.%m.%Y %H:%M')
     except ValueError:
-        raise ValidationError(_('Invalid input.'))
+        raise DjangoValidationError(_('Invalid input.'))
```
ostcar: Why did you rename this? Is there another ValidationError?

openslides/agenda/signals.py
```diff
@@ -15,65 +15,50 @@ def validate_start_time(value):
...
     agenda_start_event_date_time = ConfigVariable(
         name='agenda_start_event_date_time',
         default_value='',
-        form_field=forms.CharField(
-            validators=[validate_start_time, ],
-            widget=forms.DateTimeInput(format='%d.%m.%Y %H:%M'),
-            required=False,
-            label=ugettext_lazy('Begin of event'),
-            help_text=ugettext_lazy('Input format: DD.MM.YYYY HH:MM')))
+        html_input_type='text',  # TODO: Use 'datetime' here. Use a better validator.
```
ostcar: suggestion: remove the "html_" prefix. There could be a non-html client. For example a GTK-Frontend.

openslides/config/api.py
```diff
@@ -173,19 +173,45 @@ def get_field_names(self):
...
+    argument 'on_change' can get a callback which is called every time, the
+    variable is changed. If the argument 'translatable' is set, OpenSlides
+    is able to translate the value during setup of the database if the
+    admin uses the respective command line option.
     """
-    def __init__(self, name, default_value, form_field=None, on_change=None, translatable=False):
+    def __init__(self, name, default_value, html_input_type='text', label=None,
+                 help_text=None, choices=None, validators=None, on_change=None,
+                 translatable=False):
+        assert html_input_type in CONFIG_HTML_INPUT_TYPE_MAPPING, (
```
ostcar: assert raises a AssertionError. I would prefer a ValueError

openslides/config/api.py
```diff
@@ -173,19 +173,45 @@ def get_field_names(self):
...
+            "Invalid value for config attribute html_input_type.")
+        if html_input_type == 'radio':
+            assert choices is not None, ("Either config attribute 'choices' must not be None or "
+                                         "'html_input_type' must not be 'radio'.")
+        else:
+            assert choices is None, ("Either config attribute 'choices' must be None or "
```
ostcar: Here the same

openslides/config/views.py
```diff
@@ -41,17 +61,25 @@ def update(self, request, *args, **kwargs):
...
-        # Validate value.
-        form_field = config.get_config_variables()[key].form_field
-        value = request.data['value']
-        if form_field:
+        # Validate datatype.
+        expected_type = CONFIG_HTML_INPUT_TYPE_MAPPING[config_variable.html_input_type]
+        if not isinstance(value, expected_type):
+            raise ValidationError({
+                'detail': _('Wrong datatype. Expected %s, got %s.') % (expected_type, type(value))})
+        if config_variable.html_input_type == 'radio' and not dict(config_variable.choices).get(value):
```
ostcar: Suggestion: Move the validation code into the config-api, inside the __setitem__ method, so that it is also called at other places.

openslides/motions/signals.py
```diff
@@ -21,146 +21,132 @@ def setup_motion_config(sender, **kwargs):
...
         name='motions_workflow',
         default_value='1',
-        form_field=forms.ChoiceField(
-            widget=forms.Select(),
-            label=ugettext_lazy('Workflow of new motions'),
-            required=True,
-            choices=[(str(workflow.pk), ugettext_lazy(workflow.name)) for workflow in Workflow.objects.all()]))
+        html_input_type='radio',
+        label=ugettext_lazy('Workflow of new motions'),
+        choices=[(str(workflow.pk), ugettext_lazy(workflow.name)) for workflow in Workflow.objects.all()])
```
ostcar: What happens if a new workflow is created at runtime? (I hope this is possible some time :) )

ostcar:
Is the ConfigGroup and ConfigCollection still needed, or could the signal just return an tuple or iterable?

For example:
```python
def setup_agenda_config(...):
    yield ConfigVariable( ....
    yield ConfigVariable(...
```

If the two classes are not needed anymore, could you remove it by this refactory?

openslides/agenda/signals.py
normanjaeckel: In this file there isn't any other ValidationError but in many other files we use the Django REST Framework ValidationError which is different from Django's ValidationError. This is just to to avoid confusion.

openslides/agenda/signals.py
normanjaeckel: Hm. Ok. But then the question is whether we should use another mapping: string instead of text? boolean instead of checkbox? integer instead of number? What do you think?

openslides/config/api.py
normanjaeckel: Ok, I change this.

openslides/config/api.py
normanjaeckel: Here it is something different because we have a complex structure. The values None and 'radio' are permitted in general but not in special cases.

openslides/config/views.py
normanjaeckel: Ok.

openslides/motions/signals.py
normanjaeckel: Then the autoupdate should send that the config has to be updated. This code is called every time you access the config api (except simple retrieving values from the cache).

normanjaeckel:
The question of grouping configs is not solved yet for @emanuelschuetze had to tell us how he want's to sort and group config variables.

openslides/config/api.py
normanjaeckel: Changed.

openslides/config/api.py
normanjaeckel: Changed to ConfigError.

openslides/agenda/signals.py
emanuelschuetze: I prefer the html_input_type name because we currently use only html templates. And there I would to use this type 1:1 without any mapping. If someone wants to write a GKT frontend we can discuss it again ;)

normanjaeckel: Ok. Let's use HTML input types (and this attribute name) for the moment.

normanjaeckel:
Refactored everything (as discussed) so everything has to be reviewed again.

emanuelschuetze:
ok

openslides/agenda/signals.py
ostcar: django restframework recomments this: https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/exceptions.py#L55

PR #1550, opened 2015-06-18
Removed old static files

ostcar:
@emanuelschuetze I was not sure about the images. Can you look at 'core/static/img/', and 'mediafile/static/img/' and tell me which images must stay in core?

I would prefer that they are removed (except of the openslides-logo) and imported with bower and gulp.


emanuelschuetze:
I looked at 'core/static/img' and 'mediafiles/static/img':
Please keep only favicon.png, logo*.png and voting*.png files. The other files can be deleted.

PR #1552, opened 2015-06-18
Projector client

ostcar:
Only some groundwork to be continued.

But please review and merge (after a rebase) so @emanuelschuetze can start.

### DIFF
--- a/openslides/core/projector.py
+++ b/openslides/core/projector.py
@@ -13,24 +13,16 @@ class CustomSlideSlide(ProjectorElement):
     Slide definitions for custom slide model.
     """
     name = 'core/customslide'
-    scripts = 'core/customslide_slide.js'
 
     def get_context(self):
         pk = self.config_entry.get('id')
         if not CustomSlide.objects.filter(pk=pk).exists():
             raise ProjectorException(_('Custom slide does not exist.'))
-        return [{
-            'collection': 'core/customslide',
-            'id': pk}]
+        return {'id': pk}
 
     def get_requirements(self, config_entry):
-        self.config_entry = config_entry
-        try:
-            pk = self.get_context()[0]['id']
-        except ProjectorException:
-            # Custom slide does not exist so just do nothing.
-            pass
-        else:
+        pk = config_entry.get('id')
+        if pk is not None:
             yield ProjectorRequirement(
                 view_class=CustomSlideViewSet,
                 view_action='retrieve',
@@ -42,7 +34,6 @@ class Clock(ProjectorElement):
     Clock on the projector.
     """
     name = 'core/clock'
-    scripts = 'core/clock.js'
 
     def get_context(self):
         return {'server_time': now().timestamp()}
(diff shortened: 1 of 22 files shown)


openslides/utils/projector.py
```diff
@@ -8,13 +8,12 @@ class ProjectorElement(object, metaclass=SignalConnectMetaClass):
     Base class for an element on the projector.
 
     Every app which wants to add projector elements has to create classes
-    subclassing from this base class with different names. The name and
-    scripts attributes have to be set. The metaclass
-    (SignalConnectMetaClass) does the rest of the magic.
+    subclassing from this base class with different names. The name attribute
+    has to be set. The metaclass     (SignalConnectMetaClass) does the rest of
```
normanjaeckel: Bad formating.

ostcar: done

openslides/core/projector.py
```diff
@@ -13,20 +13,17 @@ class CustomSlideSlide(ProjectorElement):
...
-        return [{
-            'collection': 'core/customslide',
-            'id': pk}]
+        return {'id': pk}
 
     def get_requirements(self, config_entry):
         self.config_entry = config_entry
         try:
-            pk = self.get_context()[0]['id']
+            pk = self.get_context()['id']
```
normanjaeckel: Here we can choose: Do we want that the requirement does nothing if the custom slide does not exist or is it no harm because the client might get a 404 without problems?

openslides/core/urls.py
```diff
@@ -7,4 +7,8 @@
     url(r'^core/url_patterns/$',
         views.UrlPatternsView.as_view(),
         name='core_url_patterns'),
+
+    # View for the projectors are handelt by angular.
+    url(r'^projector.*$',
```
normanjaeckel: Can we move this to the global urls.py so that one can see the both entry points (/projector... and /...) at a glance?

openslides/users/projector.py
```diff
@@ -37,8 +27,14 @@ def get_requirements(self, config_entry):
             # User does not exist so just do nothing.
             pass
         else:
-            for item in context:
+            pk = str(context['id'])
```
normanjaeckel: Here the same as for custom slides.

openslides/core/urls.py
ostcar: The base entry point can not be moved into the core.py because it has to be the last entry.

I would like, that only apps are included in the global urls.py and the projector is not an app, but one view of the core app. So I would like to keep it this way.

openslides/core/projector.py
ostcar: done

openslides/core/urls.py
ostcar: done

openslides/users/projector.py
```diff
@@ -12,33 +12,23 @@ class UserSlide(ProjectorElement):
...
-            # User does not exist so just do nothing.
-            pass
-        else:
-            for item in context:
+        pk = config_entry.get('id')
+        if pk is not None:
+            yield ProjectorRequirement(
+                view_class=UserViewSet,
+                view_action='retrive',
+                pk=pk)
```
normanjaeckel: Please have a last check that this is always a string (or add str).

normanjaeckel:
Then rebase and merge.

PR #1538, opened 2015-06-15
Angular template for config app.

emanuelschuetze:
Ready for first review.

My Iast TODOs:
* updated config (requires #1529)
* reset all configs to default values

@normanjaeckel I have some suggestions for you to update the rest-api:
* use new prefix "general_" for all general config keys
* introduce new "type" value for each config field to show the correct html input type in list.html (e.g. radio, checkbox, text, number etc.; see http://www.w3schools.com/html/html_form_input_types.asp)
* rename config names: "motion" -> "motions"; "assignment" -> "assignments"

PR #1556, opened 2015-06-19
Fixed default_value in option request

ostcar:
Fixed missing default_value in option request


emanuelschuetze:
OK, I need it for #1538

PR #1562, opened 2015-06-21
Fix JavaScript errors in chromium (Fixed #1530)

### LINKED ISSUE #1530: Problem with token = in agenda.js
In static/js/agenda/agenda.js there is a function buildTree which contains an `=` in the declaration line. This causes Chromium to crash.

emanuelschuetze:


normanjaeckel:
Ok.

PR #1564, opened 2015-06-21
Added view to create related assignment and motion items. Updated ser…

### LINKED ISSUE #1559: Add motion.agenda_item to rest api
Each motion should be able to reference to one agenda item.

normanjaeckel:
…ializers. Fixes #1559.

### DIFF
--- a/openslides/assignments/serializers.py
+++ b/openslides/assignments/serializers.py
@@ -1,12 +1,15 @@
+from django.contrib.contenttypes.models import ContentType
 from django.db import transaction
 from django.utils.translation import ugettext as _
 
+from openslides.agenda.models import Item
 from openslides.utils.rest_api import (
     DictField,
     IntegerField,
     ListField,
     ListSerializer,
     ModelSerializer,
+    SerializerMethodField,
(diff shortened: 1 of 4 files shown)


emanuelschuetze:
See my comment in #1559

emanuelschuetze:
@ostcar This pull request was still in discussion (see my last comment also in #1559) and was not ready to merge! For future please look carefully.

ostcar:
@emanuelschuetze ok, lets try the revert button...

We should stop creating pull requests for branches we do not want to be merged. Any Idea how we can discuss core which is not a pull request?

emanuelschuetze:
> We should stop creating pull requests for branches we do not want to be merged.

No, this was a real pull request from Norman. But before merge we have to note all comments! And my comment was not answered by Norman.

> Any Idea how we can discuss core which is not a pull request?

For non-core code discussion we should wrote an explicit initial comment and title that this PR is not for merge.

emanuelschuetze:
@normanjaeckel Again: Let's discuss my comment in #1559

PR #1566, opened 2015-06-24
Added projector methods to user model

ostcar:
@emanuelschuetze 

I added the methods 'isProjected()' and 'project()'. See the users- or customslide model how to use them.

You have to add the useClass: BaseModel attribute and the method getResourceName to each Model that should use isProjected() and project().


emanuelschuetze:
Ok.

PR #1574, opened 2015-06-25
Added list slides for motions and assignments.

normanjaeckel:
@emanuelschuetze: Care that the definition of agenda list slide changed. You can reach it now via "agenda/item" without any pk. See agenda/projector.py.

### DIFF
--- a/openslides/agenda/projector.py
+++ b/openslides/agenda/projector.py
@@ -10,11 +10,39 @@ from .views import ItemViewSet
 
 class ItemListSlide(ProjectorElement):
     """
-    Slide definitions for the agenda.
+    Slide definitions for Item model.
+
+    This is only for list slides.
+
+    Set 'id' to None to get a list slide of all root items. Set 'id' to an
+    integer to get a list slide of the children of the metioned item.
+
+    Additionally set 'tree' to True to get also children of children.
     """
-    name = 'agenda/agenda'
+    name = 'agenda/item-list'
+
+    def get_context(self):
+        pk = self.config_entry.get('id')
+        if pk is None:
+            # Root list slide.
+            context = {'tree':  self.config_entry.get('tree', False)}
+        else:
+            # Children slide.
+            if not Item.objects.filter(pk=pk).exists():
+                raise ProjectorException(_('Item does not exist.'))
+            context = {'id': pk, 'tree':  self.config_entry.get('tree', False)}
+        return context
 
     def get_requirements(self, config_entry):
+        pk = config_entry.get('id', 'tree')
+        if pk is None or self.config_entry.get('tree', False):
+            # Root list slide or slide with tree.
+            yield ProjectorRequirement(
+                view_class=ItemViewSet,
+                view_action='tree')
+
+        # Root list slide and children list slide.
+        # Related objects like users and tags are not unlocked.
         yield ProjectorRequirement(
             view_class=ItemViewSet,
             view_action='list')
@@ -24,13 +52,17 @@ class ItemDetailSlide(ProjectorElement):
     """
     Slide definitions for Item model.
 
-    To activate this slide as list of speakers slide, set 'list_of_speakers'
-    true.
+    This is only for detail slides. You have to set 'id'.
+
+    To activate a detail slide as list of speakers slide, set 'id' and set
+    'list_of_speakers' true.
     """
-    name = 'agenda/item'
+    name = 'agenda/item-detail'
 
     def get_context(self):
         pk = self.config_entry.get('id')
+        if pk is None:
+            raise ProjectorException(_('Id must not be None.'))
         if not Item.objects.filter(pk=pk).exists():
             raise ProjectorException(_('Item does not exist.'))
         return {'id': pk, 'list_of_speakers': self.config_entry.get('list_of_speakers', False)}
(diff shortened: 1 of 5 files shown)


emanuelschuetze:
Ok.

ostcar:
@normanjaeckel I think we need the agenda/agenda slide with possible id. In the 1.7 branch we had this slide, which is usefull. You used it on the "Mitgliederversammlung" from the Refugee Law Clinic: http://de.demo.openslides.org/projector/preview/agenda/?pk=5

emanuelschuetze:
@ostcar As I understand it correctly there is a agenda slide AND a item detail slide under "agenda/item" (same like in 1.x):
`Set 'id' to get a detail slide. Omit it to get a list slide.`

So that's fine for me.

normanjaeckel:
@ostcar's question might me how to deal with the summary slide. I can add it (later) to this class if you like. (Meaning you have to set "id": ... and "type": "summary" to get the summary slide. Ok?

ostcar:
@normanjaeckel I do not understand way you merged this two slides together. They need different templates and different js-code. To use the same slide-name but use different templates and js-code if another argument is in the config dict seems  	irreproducible.

We not use agenda/item with a required id attribute (like user, motion, assignment etc) and another slide agenda/agenda with an optional id attribute.

You added the feature to use more then one slide class per app, lets use it.

normanjaeckel:
Changed agenda slides implementation. @ostcar please review again.

openslides/agenda/projector.py
```diff
@@ -24,20 +54,25 @@ class ItemDetailSlide(ProjectorElement):
...
     def get_context(self):
         pk = self.config_entry.get('id')
-        if not Item.objects.filter(pk=pk).exists():
-            raise ProjectorException(_('Item does not exist.'))
-        return {'id': pk, 'list_of_speakers': self.config_entry.get('list_of_speakers', False)}
+        if pk is not None:
+            # Detail slide.
+            if not Item.objects.filter(pk=pk).exists():
+                raise ProjectorException(_('Item does not exist.'))
+            return {'id': pk, 'list_of_speakers': self.config_entry.get('list_of_speakers', False)}
```
ostcar: if pk is None, then None is returned. Is this valid for a get_context method?

openslides/agenda/projector.py
```diff
@@ -10,11 +10,41 @@
...
 class ItemListSlide(ProjectorElement):
     """
-    Slide definitions for the agenda.
+    Slide definitions for Item model.
+
+    This is only for list slides.
+
+    Omit 'id' to get the whole agenda tree. Set 'id' to None to get a list
+    slide of all root items. Set 'id' to an integer to get a list slide of
+    the children of the metioned item.
```
ostcar: I am against a agenda slide which shows not only the children, but also the children of the children. I think this are to many items.

But if you want such a slide, then please also add this feature for each item. For example show item 5 with his children as tree. For that to work id and tree have to be different attributes in the projector_config.

openslides/agenda/projector.py
```diff
@@ -10,11 +10,41 @@
...
+            context = {'id': pk}
+        return context
 
     def get_requirements(self, config_entry):
+        pk = config_entry.get('id', 'tree')
+        if pk == 'tree':
+            # Agenda tree.
+            yield ProjectorRequirement(
+                view_class=ItemViewSet,
+                view_action='tree')
```
ostcar: The tree-rest-request is also needed when only the items with no parents are shown on a slide (pk=None in your implementation). The tree-view is the only way to get the information which items have no parents. (We excluded this information from the item-detail-view.)

openslides/agenda/projector.py
```diff
@@ -24,20 +54,25 @@ class ItemDetailSlide(ProjectorElement):
     """
     Slide definitions for Item model.
 
-    To activate this slide as list of speakers slide, set 'list_of_speakers'
-    true.
+    This is only for detail slides. You have to set 'id'.
+
+    To activate a detail slide as list of speakers slide, set 'id' and set
+    'list_of_speakers' true.
```
ostcar: Proposal. Add another slide (agenda/item-list-of-speakers) for this.

openslides/agenda/projector.py
normanjaeckel: Yes.

openslides/agenda/projector.py
normanjaeckel: Ok. Changed.

openslides/agenda/projector.py
normanjaeckel: Ok, I can use `tree` as a property for its own. To get a nested tree on the projector is very useful and only overheaded if you have a huge agenda. Small events normally only have two or three levels.

openslides/agenda/projector.py
normanjaeckel: This we can do in another pull request. See #1576.

openslides/agenda/projector.py
ostcar: Could you add a comment or explicitly call return None at the end of the function. Because it looks like an error

ostcar:
Please fix travis, and then merge this

openslides/agenda/projector.py
normanjaeckel: Changed.

PR #1579, opened 2015-06-29
Moved config app into the core app

ostcar:
I would like to merge the config app with the core app because they both all always required. This step has to be done before we create the migration files, because I heard there where problem with django migrations that move a Model from one app into another.

utils.rest_api.RestModelMixin had to move out of utils.rest_api because django_rest_framework does not load is config lazy and therefore there was a circular import starting from this line: https://github.com/OpenSlides/OpenSlides/blob/master/openslides/global_settings.py#L181 To really solve this problem django_rest_framework has to load its config lazy or  openslides.users.auth.AnonymousAuthentication has to be in a file with no imports.


CHANGELOG
```diff
@@ -51,6 +51,7 @@ Other:
 - Used setup.cfg for development tools.
 - Removed code for Windows portable version with GUI. Used new repository for
   this.
+- Moved config app into the core app.
```
ostcar: merge with other line "refactor api"

PR #1580, opened 2015-06-29
Added initial migrations for all apps

ostcar:
This is a blocker for the alpha1 release.

I did not create a migration file for the config app, because I like to remove it. See #1579


CHANGELOG
```diff
@@ -51,6 +51,7 @@ Other:
 - Used setup.cfg for development tools.
 - Removed code for Windows portable version with GUI. Used new repository for
   this.
+- Added migrations for all OpenSlides apps.
```
ostcar: delete line

PR #1582, opened 2015-06-29
Let the config app accept values that can be converted to the right type

ostcar:


openslides/core/static/templates/config-form-field.html
```diff
@@ -5,7 +5,7 @@
     <input ng-model="value" ng-change="save(configOption.key, value, type)" id="{{ id }}" type="{{ type }}"
```
emanuelschuetze: you forgot to remvoe this `type`.

emanuelschuetze:
one remark. rest ok. (untested)

emanuelschuetze:
Successfully tested. OK.

PR #1583, opened 2015-06-29
Use item-detail and item-list slides for agenda app.

emanuelschuetze:


ostcar:
@normanjaeckel the agenda-item-detail-slide has to be called agenda/item. On the client side we have a generic function to activate model-slides. This function uses the model-name as slide-name. So for the agenda-item-detail-slide it is agenda/item

emanuelschuetze:
Ready for review and merge.

PR #1588, opened 2015-06-30
Updated python requirements.

ostcar:
This included django 1.8 (but django 1.7 is still supported)

### DIFF
--- a/openslides/motions/models.py
+++ b/openslides/motions/models.py
@@ -179,7 +179,6 @@ class Motion(RESTModelMixin, models.Model):
             use_version.version_number = version_number + 1
 
         # Necessary line if the version was set before the motion got an id.
-        # This is probably a Django bug.
         use_version.motion = use_version.motion
 
         use_version.save()
@@ -339,7 +338,11 @@ class Motion(RESTModelMixin, models.Model):
         attributes, it is populated with the data from the last version
         object if such object exists.
         """
-        new_version = MotionVersion(motion=self, **kwargs)
+        if self.pk is None:
+            # Do not reference the MotionVersion object to an unsaved motion
+            new_version = MotionVersion(**kwargs)
+        else:
+            new_version = MotionVersion(motion=self, **kwargs)
         if self.versions.exists():
             last_version = self.get_last_version()
         else:
--- a/openslides/users/migrations/0002_auto_20150630_0143.py
+++ b/openslides/users/migrations/0002_auto_20150630_0143.py
@@ -1,0 +1,28 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('users', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='user',
+            name='groups',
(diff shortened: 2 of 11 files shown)


openslides/motions/models.py
```diff
@@ -339,7 +339,11 @@ def get_new_version(self, **kwargs):
         attributes, it is populated with the data from the last version
         object if such object exists.
         """
-        new_version = MotionVersion(motion=self, **kwargs)
+        if self.pk is None:
+            # Do not reference the MotionVersion object to an unsaved motion
```
normanjaeckel: This looks strange: Where do we ever try to retrieve a new version for an unsaved motion? See line 154 ans 184 in this file.

If we ever do so, how can we ensure that the reference is set sometime?

openslides/users/migrations/0002_auto_20150630_0143.py
```diff
@@ -0,0 +1,28 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
normanjaeckel: We have a migration here but no model change. Maybe this depends on the Django version so we can omit this change.

normanjaeckel:
Remarks to be discussed.

openslides/motions/models.py
ostcar:
```python
str(Motion())
```
This happens (for example) if you call get_title() from a Motion without a Version. See get_last_version() and get_active_version()

openslides/users/migrations/0002_auto_20150630_0143.py
ostcar: No we can't. Try to remove this file and run OpenSlides with django 1.8.

If a database was create with the alpha the User and Group database was created with the first migration file. So the database tables have to be changed.

openslides/motions/models.py
normanjaeckel: Ok but then what happens if I try to save such a motion. Do we need a fix in save()?

ostcar: I test this. Please do not merge

openslides/users/migrations/0002_auto_20150630_0143.py
normanjaeckel: This is dangerous. If a user migrates database but does not upgrade to Django 1.8 he might run into trouble. I this case it looks ok but we should think about dropping old Django support in future cases.

openslides/motions/models.py
ostcar: This is save. See the line below the deleted comment in my last commit.

Before django 1.8 it looks like a bug, because it was possible to set a unsaved motion for the new MotionVersion, but it was not used. Now, it is not possible to set the unsaved motions. So it is clear, that the line is needed.

ostcar:
travis will be fixed after I rebase this on top of #1591 (after it is merged)

openslides/motions/models.py
normanjaeckel: I do not understand, but WTF ... Merge.

normanjaeckel:
Rebase and merge.

PR #1589, opened 2015-06-30
Fix osPerms to use groups registered and anonymous

### LINKED ISSUE #1586: 'os-perms' does not check the default groups (anonymous/registered)
The directive `os-perms="agenda.can_see"` does not work for the admin who is in staff and registered group. Because the permission "agenda.can_see" is selected in group Registered only. For "agenda.can_mange" (permission from staff group) it works as expected.

Please add also all permissions from the default groups (registered/anonymous) to allPerms in getPerms function in users.js. Then we can use "os-perms" correctly in template (e.g. show menu item "agenda" only if users have at least "agenda.can_see" permission.

ostcar:
Fixes #1586

### DIFF
--- a/openslides/users/static/js/users/users.js
+++ b/openslides/users/static/js/users/users.js
@@ -42,27 +42,12 @@ angular.module('OpenSlidesApp.users', [])
                 }
                 return name;
             },
-            get_full_name: function() {
-                // should be the same as in the python user model.
-                var firstName = _.trim(this.first_name),
-                    lastName = _.trim(this.last_name),
-                    structure_level = _.trim(this.structure_level),
-                    name;
-
-                if (firstName && lastName) {
-                    // TODO: check config
-                    name = [firstName, lastName].join(' ');
-                } else {
-                    name = firstName || lastName || this.username;
-                }
-                if (structure_level) {
-                    name = name + " (" + structure_level + ")";
-                }
-                return name;
-            },
             getPerms: function() {
                 var allPerms = [];
-                _.forEach(this.groups, function(groupId) {
+                var allGroups = this.groups;
+                // Add registered group
+                allGroups.push(2);
+                _.forEach(allGroups, function(groupId) {
                     // Get group from server
                     Group.find(groupId);
                     // But do not work with the returned promise, because in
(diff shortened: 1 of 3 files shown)


openslides/users/static/js/users/users.js
```diff
@@ -42,27 +42,12 @@ angular.module('OpenSlidesApp.users', [])
                 }
                 return name;
             },
-            get_full_name: function() {
```
emanuelschuetze: Why you removed this function?

ostcar: See line 27 of the file

emanuelschuetze: Ok.

openslides/users/views.py
```diff
@@ -110,11 +111,22 @@ def check_permissions(self, request):
         permission to see users and in case of create, update or destroy
         requests the permission to see extra user data and to manage users.
         """
-        if (not request.user.has_perm('users.can_see_name') or
-                (self.action in ('create', 'update', 'destroy') and not
-                 (request.user.has_perm('users.can_manage') and
-                  request.user.has_perm('users.can_see_extra_data')))):
-            self.permission_denied(request)
+        # Any logged in user can retrive groups.
```
normanjaeckel: Hm. I don't like this style (using multiple returns in a difficult function). We can merge this for now but I would like to refactor all permission checks later. May be we should use explicit permission checks in the view methods instead of the general method `check_permission`.

openslides/users/views.py
```diff
@@ -110,11 +111,22 @@ def check_permissions(self, request):
...
+        # Anonymous user can retrive groups when they are activated.
+        if (self.action in ('retrieve', 'list') and
+                (config['general_system_enable_anonymous'] or
+                 self.request.user.is_authenticated())):
+            return
+
+        # Users with the permissions 'can_manage' and 'can_see_extra_data' can
+        # edit groups.
+        if (self.action in ('create', 'update', 'destroy', 'partial_update') and
+                request.user.has_perm('users.can_manage') and
```
normanjaeckel: Please add `request.user.has_perm('users.can_see_name')` just for consistency. We will have to remove this later when we refactor the permission system using inheritance.

normanjaeckel:
One remark, then merge.

openslides/users/views.py
ostcar: done with protest

PR #1591, opened 2015-07-01
Added a way to load angular apps dynamical

ostcar:
With this pull request all OpenSlides apps are loaded dynamical into angular.

This can not be done with gulp, because there has to be a way to install plugins without node and gulp.


CHANGELOG
```diff
@@ -51,6 +51,7 @@ Other:
 - Used setup.cfg for development tools.
 - Removed code for Windows portable version with GUI. Used new repository for
   this.
+- Added possibillity for django apps to append js files and angular apps.
```
normanjaeckel: This can be 'merged' with line 42.

ostcar: Do we want to track the changes from one alpha to another?

openslides/core/urls.py
```diff
@@ -7,10 +7,15 @@
     url(r'^core/url_patterns/$',
         views.UrlPatternsView.as_view(),
         name='core_url_patterns'),
+
     url(r'^core/version/$',
         views.VersionView.as_view(),
         name='core_version'),
 
+    url(r'^angular_js/(?P<app>site|projector)/$',
```
normanjaeckel: The termin `app` may cause missunderstandings (Django app or AngularJS app) so I propose to refer to `OpenSlidesApp` (which is `openslides_app` in lowercase writing:
```
url(r'^core/js_files/(?P<openslides_app>site|projector)/$'
```

openslides/core/views.py
```diff
@@ -308,3 +309,38 @@ def update(self, request, *args, **kwargs):
 
         # Return response.
         return Response({'key': key, 'value': value})
+
+
+class AppsJsView(utils_views.View):
```
normanjaeckel: Please add a docstring.

openslides/core/views.py
```diff
@@ -308,3 +309,38 @@ def update(self, request, *args, **kwargs):
...
+                    app_name=app_config.label))
+
+            # Add all js files that the module needs
+            try:
+                js_files += [
+                    '{static}{path}'.format(
+                        static=settings.STATIC_URL,
+                        path=path)
+                    for path in app_config.js_files]
+            except AttributeError:
```
normanjaeckel: It is not easy to see where the AttributeError belongs to.
```python
try:
    app_js_files = app_config.js_files
except AttributeError:
    # The app needs no js-files
    pass
else:
    js_files += [
        '{static}{path}'.format(
            static=settings.STATIC_URL,
            path=path)
        for path in app_js_files]
```

openslides/core/urls.py
ostcar: done

openslides/core/views.py

openslides/core/views.py

requirements.txt
```diff
@@ -2,7 +2,7 @@
 -r requirements_production.txt
 
 # Requirements for development and tests in alphabetical order
-coverage>=4.0a5,<4.1.0
+coverage==4.0a5
```
normanjaeckel: Why not using the latest release (4.0a6)?

ostcar: This breaks travis. It seams that the 4.0a6 needs allow-external.

I am not sure if this was a temporal problem, If you like, you can update the coverage version with a seperate pull request.

CHANGELOG
normanjaeckel: I think not.

requirements.txt
normanjaeckel: Ok.

CHANGELOG
ostcar: done

openslides/core/urls.py
normanjaeckel: What about (my proposal) using `core/js_files` instead of `angular_js`?

normanjaeckel:
Rebase and check Travis again.

openslides/core/urls.py
ostcar: ups. I did not realize this. I like angular_js for. Because there could be other clients using javascript, but this file is only for the angular client.

ostcar:
@normanjaeckel all done, please merge

openslides/core/urls.py
normanjaeckel: Ok.

PR #1592, opened 2015-07-01
Refactored permission check for REST API viewsets.

normanjaeckel:
Please have a first look but do not merge. If this way is ok, I can implement it for every viewset we have.

This  pull request contains also some clean up.

### DIFF
--- a/openslides/core/views.py
+++ b/openslides/core/views.py
@@ -35,6 +35,8 @@ from .serializers import (
 )
 
 
+# Special Django views
+
 class IndexView(utils_views.CSRFMixin, utils_views.View):
     """
     The primary view for OpenSlides using AngularJS.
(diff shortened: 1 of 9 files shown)


ostcar:
I line this way. Is it possible in this way, that projector_permissions is only called, when the user has not the regular permissions for the view?

Please rebase on top of master to fix travis

openslides/utils/rest_api.py
```diff
@@ -46,8 +49,8 @@ def initial(self, request, *args, **kwargs):
 
         # Ensure that the incoming request is permitted
         self.perform_authentication(request)
-        if not self.check_projector_requirements():
-            self.check_permissions(request)
+        if not self.check_permissions(request) and not self.check_projector_requirements():
```
normanjaeckel: @ostcar As you see here the projector_requirements and evaluated only if the first check returns False.

ostcar:
ok. Rebase and merge

normanjaeckel:
Sorry, this is not ready. I will ping this when it's done.

normanjaeckel:
@ostcar All done (with massive changes): Ready for review.

openslides/core/views.py
```diff
@@ -265,10 +284,25 @@ def determine_metadata(self, request, view):
...
     """
     metadata_class = ConfigMetadata
 
+    def check_view_permissions(self):
+        """
+        Returns True if the user has required permissions.
+        """
+        if self.action in ('list', 'retrieve'):
+            # Everybody is able to list and retrieve the config.
+            result = True
```
ostcar: When anonymous user is deactivated, then this function should return False for anonymous user.

openslides/core/views.py
```diff
@@ -168,71 +216,42 @@ def clear_elements(self, request, pk):
...
-        result = dict(openslides_version=version, plugins=[])
-        # Versions of plugins.
-        for plugin in settings.INSTALLED_PLUGINS:
-            result['plugins'].append({
-                'verbose_name': get_plugin_verbose_name(plugin),
-                'description': get_plugin_description(plugin),
-                'version': get_plugin_version(plugin)})
+        if self.action in ('list', 'retrieve'):
+            # Everybody is able to list and retrieve tags.
+            result = True
```
ostcar: This is False for anonymous users, when they are deactivated

openslides/utils/rest_api.py
```diff
@@ -20,39 +19,39 @@
...
-        Runs anything that needs to occur prior to calling the method handler.
-        """
-        self.format_kwarg = self.get_format_suffix(**kwargs)
 
-        # Ensure that the incoming request is permitted
-        self.perform_authentication(request)
-        if not self.check_projector_requirements():
-            self.check_permissions(request)
-        self.check_throttles(request)
+    def get_permissions(self):
```
ostcar: Please add a docstring

openslides/utils/views.py
```diff
@@ -88,14 +52,33 @@ def as_view(cls, *args, **kwargs):
...
     top_space = 3
     document_title = None
+    required_permission = None
+
+    def check_permission(self, request, *args, **kwargs):
+        """
+        Checks if the user has the required permission.
+        """
+        if self.required_permission is None:
+            return True
```
ostcar: Could you add an explicit permission check. When required_permission is None, then no user should have permissions. There for we can not forget to set a permission.

openslides/utils/views.py
```diff
@@ -88,14 +52,33 @@ def as_view(cls, *args, **kwargs):
...
+        else:
+            return request.user.has_perm(self.required_permission)
```
ostcar: I like this, but I think, that you don't.

If there is a PDF that only the manager can see (for example the user Password PDF), which permission has to be defined in required_permission? Only 'can_manage' without 'can_see'?

You have either to redefine the can_manage permission to also include 'can_see', or change the required_permission attribute to an list or tuple.

I would like, that we redefine the can_manage permission. You do not have to do this in this pull request. You can leave this method as it is and create an issue to redefine the can_manage permission

openslides/core/views.py
normanjaeckel: Done.

openslides/core/views.py
normanjaeckel: Done.

openslides/utils/rest_api.py
normanjaeckel: Done.

openslides/utils/views.py
normanjaeckel: No this is old style code for Django views. We can do this in an extra pull request.

openslides/utils/views.py
normanjaeckel: This is old style code (only copied from the old PermissionMixin). You can refactor this later if you like.

normanjaeckel:
All done.

ostcar:
ok. Rebase and merge

PR #1593, opened 2015-07-05
Added new class RESTElement to configure our REST api

ostcar:
This pull requests adds a new class RESTElement to define elements that can be accessed via the REST api.

I would like this new way for the search implementation because I want to use the collection name and do not like to get this name from the django rest framework router (through the django reverse function). So I created the RESTElement class where all methods can be defined, which a REST element should have. For now, it is only used for the method get_collection_name().

To do this, I also implemented an alternative to utils.dispatch.SignalConnectMetaClass. The SignalConnectMetaClass implemented some functionality that I did not need (like check_permission) and I don't like to use django signals where a simple set() can do the job. The new metaclass RegisterSubclasses is simple (only 10 lines of code) and the Base class on which it is used does not need any extra attributes or methods.

This pull request is a blocker for the implementation of the search.

### DIFF
--- a/openslides/core/rest_api.py
+++ b/openslides/core/rest_api.py
@@ -1,0 +1,21 @@
+from openslides.utils.rest_api import RESTElement
+
+from . import views
+
+
+class CustomSlideRESTElement(RESTElement):
+    viewset = views.CustomSlideViewSet
+
+
+class ProjectorRestElement(RESTElement):
+    viewset = views.ProjectorViewSet
+
+
+class TagRestElement(RESTElement):
+    viewset = views.TagViewSet
+
+
+class ConfigRestElement(RESTElement):
+    viewset = views.ConfigViewSet
+    router_prefix = 'config'
+    app_label = 'core'
--- a/openslides/users/rest_api.py
+++ b/openslides/users/rest_api.py
@@ -1,0 +1,13 @@
+from openslides.utils.rest_api import RESTElement
+
+from . import views
+
+
+class UserRESTElement(RESTElement):
+    viewset = views.UserViewSet
+
+
+class GroupRESTElement(RESTElement):
+    viewset = views.GroupViewSet
+    app_label = 'users'
+    router_prefix = 'group'
--- a/openslides/utils/class_register.py
+++ b/openslides/utils/class_register.py
@@ -1,0 +1,46 @@
+class RegisterSubclasses(type):
+    """
+    Metaclass to register subclasses of a class.
+
+    The metaclass adds an attribute 'subclasses' to the base class. This
+    attribute is a set containing all derivated classes.
+
+    After a new subclass is created, the classmethod on_subclass_created is
+    called.
+
+    Example:
+    class Base(metaclass=RegisterSubclasses):
+        pass
+
+    class Child(Base):
+        pass
+
+    assert Child in Base.subclasses
+    """
+
+    def __init__(cls, *args, **kwargs):
+        # Try to add the cls to cls.subclasses. If it does not have the
+        # attribute subclasses, then this is the base class, which should not
+        # be registered
+        try:
+            cls.subclasses.add(cls)
+        except AttributeError:
+            cls.subclasses = set()
+        else:
+            try:
+                cls.on_subclass_created()
+            except Exception as exception:
+                # If on_subclass_created() raises any exception, then the class
+                # should not be in cls.subclasses.
+                cls.subclasses.remove(cls)
+                # Raise the exception afterwards
+                raise exception
+        super().__init__(*args, **kwargs)
+
+    def on_subclass_created(cls):
+        """
+        This method is called when a new subclass is created.
+
+        Do nothing as default.
+        """
+        pass
--- a/openslides/utils/models.py
+++ b/openslides/utils/models.py
@@ -39,3 +39,18 @@ class RESTModelMixin:
         root_instance = self.get_root_rest_element()
         rest_url = '%s-detail' % type(root_instance)._meta.object_name.lower()
         return reverse(rest_url, args=[str(root_instance.pk)])
+
+    @classmethod
+    def get_collection_name(cls):
+        """
+        Returns the name of the collection.
+
+        This is usualy the name of the django app and the name of the model
+        combined with a slash.
+
+        The collection_name is used to identify an object inside of
+        openslides. For example in the rest api or in the search.
+        """
+        return "{app_name}/{model_name}".format(
+            app_name=cls._meta.app_label,
+            model_name=cls._meta.model_name)
--- a/openslides/utils/rest_api.py
+++ b/openslides/utils/rest_api.py
@@ -19,12 +19,14 @@ from rest_framework.serializers import (  # noqa
     SerializerMethodField,
     ValidationError,
 )
-from rest_framework.viewsets import GenericViewSet as _GenericViewSet  # noqa
-from rest_framework.viewsets import ModelViewSet as _ModelViewSet  # noqa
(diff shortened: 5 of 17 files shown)


openslides/core/rest_api.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class TagRestElement(RESTElement):
+    viewset = views.TagViewSet
+
+
+class ConfigRestElement(RESTElement):
+    viewset = views.ConfigViewSet
+    router_prefix = 'config'
+    app_label = 'core'
```
normanjaeckel: Is it possible to determine the app_label somehow from Django's app loading API?

openslides/users/rest_api.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class UserRESTElement(RESTElement):
+    viewset = views.UserViewSet
+
+
+class GroupRESTElement(RESTElement):
+    viewset = views.GroupViewSet
+    app_label = 'users'
+    router_prefix = 'group'
```
normanjaeckel: You might get these automaticly.

openslides/utils/class_register.py
```diff
@@ -0,0 +1,39 @@
+class RegisterSubclasses(type):
+    """
+    Metaclass to track subclasses of a class.
+
+    The metaclass adds an attribute 'subclasses' to the metaclass. This
```
normanjaeckel: You mean `... to the base class ...`?

openslides/utils/class_register.py
```diff
@@ -0,0 +1,39 @@
+class RegisterSubclasses(type):
+    """
+    Metaclass to track subclasses of a class.
```
normanjaeckel: ` Metaclass to register subclasses of a class.`?

openslides/utils/class_register.py
```diff
@@ -0,0 +1,39 @@
...
+    called.
+
+    Example:
+    class Base(metaclass=RegisterSubclasses):
+        pass
+
+    class Child(Base):
+        pass
+
+    assert Child in Base.subclasses
```
normanjaeckel: I think this would be True: `Child.subclasses is Base.subclasses`.

If I am right this looks missleading.

openslides/utils/class_register.py
```diff
@@ -0,0 +1,39 @@
...
+        # Try to add the cls to cls.subclasses. If it does not have the
+        # attribute subclasses, then this is the base class, which should not
+        # be registered
+        try:
+            cls.subclasses.add(cls)
+        except AttributeError:
+            cls.subclasses = set()
+        else:
+            cls.on_subclass_created()
+        super().__init__(*args, **kwargs)
```
normanjaeckel: This `super` call is always empty, isn't it?

openslides/utils/models.py
```diff
@@ -39,3 +39,18 @@ def get_root_rest_url(self):
...
+
+        This is usualy the name of the django app and the name of the model
+        combined with a slash.
+
+        The collection_name is used to identify an object inside of
+        openslides. For example in the rest api or in the search.
+        """
+        return "{app_name}/{model_name}".format(
+            app_name=cls._meta.app_label,
+            model_name=cls._meta.model_name)
```
normanjaeckel: In line 40 we use `_meta.object_name.lower()` which should be the same as here I think.

openslides/utils/rest_api.py
```diff
@@ -89,6 +90,86 @@ class ViewSet(PermissionMixin, _ViewSet):
...
 
 
+class RESTElement(metaclass=RegisterSubclasses):
+    """
+    Element of our REST api.
+
+    Each element bundels a viewset, a route and optionally a model together.
+    """
+
+    viewset = None
```
normanjaeckel: Your implementation looks like that viewset never should be None. See line 168. So you have to ensure, that all subclasses define this attribute properly.

openslides/utils/rest_api.py
```diff
@@ -89,6 +90,86 @@ class ViewSet(PermissionMixin, _ViewSet):
...
+    """
+    A django REST framework viewset class used in the router.
+    """
+
+    router_prefix = None
```
normanjaeckel: This is also used as router's base_name which is not obvious when reading the docstring.

normanjaeckel:
Hm. This looks complex: How can we search for content that has not a viewset for its own but is available through a viewset of another model (motions' MotionVersion or agenda's Speaker for example (?)).

openslides/core/rest_api.py
ostcar: I also thought about this but did not find a solution.

openslides/users/rest_api.py
ostcar: how?

openslides/utils/class_register.py
ostcar: done

openslides/utils/class_register.py
ostcar: done.

openslides/utils/class_register.py
ostcar: You are right. But why is this missleading?

openslides/utils/class_register.py
ostcar: What do you mean? If you ask if the super call is necessary, then probably not, because I think that type.__init__ does nothing.

openslides/utils/models.py
ostcar: Line 40 is in a "normal" method. This place is in a class method.

I can change line 40 if you like.

openslides/utils/rest_api.py

openslides/utils/rest_api.py

ostcar:
I like to implement a search view, that takes a query-search-string and returns a list of rest-elements (motions/motion 1, users/group 5, etc). It is the task of the client to find out where in the object the searched term is.

openslides/core/rest_api.py
normanjaeckel: You may initiate the `RESTElement` in apps.py not by importing the class but by calling it and use a kind of different implementation (also using a customized __init__()).

openslides/users/rest_api.py
normanjaeckel: The queryset of the GroupViewSet contains the model. The name of the model `group`. It is just like all the other cases you dealed with (Motions, Categories, Users, Items, ...).

openslides/utils/class_register.py
normanjaeckel:
```python
class Me(metaclass=RegisterSubclasses):
    pass

class ChildOne(Me):
    pass

class ChildTwo(Me):
    pass

class Grandchild(ChildOne):
    pass
```
Now it is missleading if `ChildOne.subclasses` returns a set containing also ChildTwo. Also it is missleading that 
`Grandchild.subclasses` contains the whole family but not the grandfather.

May be you want to restrict the subclasses attribute to be available only for the base class may be by deleting the reference or something like that.

openslides/utils/models.py
normanjaeckel: I don't know what is better. It just looks strange.

normanjaeckel:
Hm. But if your search engine is model based, it would return MotionVersion for example. Then you want to use the `get_root_rest_element()` method? If yes the class registering system is here only for root rest elements and not for rest elements in general. We may rename the class from RESTElement to RootRESTElementRegistration or the like.

ostcar:
My search engine will not be model based, but RESTElement based.

What is the difference between a rest element and a root rest element? We use the root_ prefix in for django models where it makes more sense. But I can change the names of the RESTModelMixin methods to get_rest_element and get_rest_url if you like.

normanjaeckel:
Whatever ...

PR #1598, opened 2015-07-07
Appended an _id suffix to all related field names in the rest api

### LINKED ISSUE #1597: Rename ForeignKey fields to FIELDNAME_id
Please rename the field names of foreign keys, that only return the id to FIELDNAME_id.

For example the motion serializer has the field 'category' which is the id of the category.

On the client, I would like to add the category-object to the motion object, so motion.category.name returns the name of the category. But my problem is, that the attribute-name 'category' is already taken.

So it should be the same then in django. FIELD_id is the id of a ForeignKeyField and FIELD is the foreign object.

This is a blocker for #1596

ostcar:
Fixes #1597

This pull requests breaks the angular client in some parts. I will fix it in #1596

### DIFF
--- a/openslides/utils/rest_api.py
+++ b/openslides/utils/rest_api.py
@@ -1,4 +1,5 @@
 import re
+from collections import OrderedDict
 from urllib.parse import urlparse
 
 from rest_framework.decorators import detail_route, list_route  # noqa
@@ -6,14 +7,16 @@ from rest_framework.metadata import SimpleMetadata  # noqa
 from rest_framework.mixins import DestroyModelMixin, UpdateModelMixin  # noqa
 from rest_framework.response import Response  # noqa
 from rest_framework.routers import DefaultRouter
+from rest_framework.serializers import ModelSerializer as _ModelSerializer
 from rest_framework.serializers import (  # noqa
+    MANY_RELATION_KWARGS,
     CharField,
     DictField,
     Field,
     IntegerField,
     ListField,
     ListSerializer,
-    ModelSerializer,
+    ManyRelatedField,
     PrimaryKeyRelatedField,
     RelatedField,
     SerializerMethodField,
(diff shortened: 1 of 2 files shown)


ostcar:
This is a blocker for #1596

ostcar:
Please wait with your review. I have to test something.

ostcar:
@normanjaeckel please review.

Also see https://github.com/tomchristie/django-rest-framework/issues/3121 and https://github.com/tomchristie/django-rest-framework/pull/3122

openslides/utils/rest_api.py
```diff
@@ -30,6 +32,69 @@
 router = DefaultRouter()
 
 
+class IdManyRelatedField(ManyRelatedField):
+    field_name_suffix = '_id'
```
normanjaeckel: Can you please a a docstring here?

openslides/utils/rest_api.py
```diff
@@ -30,6 +32,69 @@
...
+
+    def bind(self, field_name, parent):
+        """
+        Called when the field is bound to the serializer.
+
+        Changes the source  so that the original field name is used (removes
+        the _id suffix).
+        """
+        if field_name:
+            self.source = field_name[:-len(self.field_name_suffix)]
```
normanjaeckel: Can you add a comment or docstring like this:
`In case of many=True relations this class is used child_relation class, see first line in many_init() classmethod below. In this case the field_name is an empty string which of course cannot be shortened.`

openslides/utils/rest_api.py
```diff
@@ -30,6 +32,69 @@
...
+
+    Only works together the our ModelSerializer.
+    """
+    field_name_suffix = '_id'
+
+    def bind(self, field_name, parent):
+        """
+        Called when the field is bound to the serializer.
+
+        Changes the source  so that the original field name is used (removes
```
normanjaeckel: Unnecessary whitespace after `source`.

openslides/utils/rest_api.py
normanjaeckel: May be use also a docstring at the bind method, see docstrings in lin 53 ff.

openslides/utils/rest_api.py
```diff
@@ -30,6 +32,69 @@
...
+        """
+        if field_name:
+            self.source = field_name[:-len(self.field_name_suffix)]
+        super().bind(field_name, parent)
+
+    @classmethod
+    def many_init(cls, *args, **kwargs):
+        """
+        Method from rest_framework.relations.RelatedField That uses our
+        IdManyRelatedField class.
```
normanjaeckel: Add `... instead of rest_framework.relations.ManyRelatedField class`

openslides/utils/rest_api.py
```diff
@@ -30,6 +32,69 @@
...
+        """
+        # TODO: this method can be removed if this is merged and released:
```
normanjaeckel: The issue and pull request were refused so you can remove this comment and may be add a short explanation, that this code is copied ....

openslides/utils/rest_api.py
```diff
@@ -30,6 +32,69 @@
...
+    """
+    ModelSerializer that changes the field names of related fields to
+    FIELD_NAME_id.
+    """
+    serializer_related_field = IdPrimaryKeyRelatedField
+
+    def get_fields(self):
+        fields = super().get_fields()
+        new_fields = type(fields)()
+        self.manipulated_field_names = set()
```
normanjaeckel: I do not see/find where this attribute is used? Is this a misstake?

openslides/utils/rest_api.py
```diff
@@ -30,6 +32,69 @@
...
+
+class ModelSerializer(_ModelSerializer):
+    """
+    ModelSerializer that changes the field names of related fields to
+    FIELD_NAME_id.
+    """
+    serializer_related_field = IdPrimaryKeyRelatedField
+
+    def get_fields(self):
+        fields = super().get_fields()
```
normanjaeckel: Can you please add a docstring and maybe rename new_fields to manipulated_fields or customized_fields or something like that so we don't think these fields are *newley created/initiated* but only some of them are changed/renamed in a special way?

normanjaeckel:
Looks nice (and complex). Some remarks.

openslides/utils/rest_api.py

openslides/utils/rest_api.py

openslides/utils/rest_api.py

openslides/utils/rest_api.py

openslides/utils/rest_api.py

openslides/utils/rest_api.py

openslides/utils/rest_api.py
ostcar: done (in some other way)

normanjaeckel:
Rebase and merge.

Issue #1332, opened 2014-08-17
Outsource plugin url code in urls.py into a separate function


normanjaeckel:
See discussion in #1331.

normanjaeckel:
Urls.py does not contain plugin code any more.

PR #1600, opened 2015-07-10
Updated plugin utils. Enabled plugins to override urlpatterns.

normanjaeckel:


openslides/urls.py
```diff
@@ -1,8 +1,17 @@
...
+from openslides.utils.plugins import get_plugin_urlpatterns
 from openslides.utils.rest_api import router
 
+urlpatterns = []
+
+for plugin in settings.INSTALLED_PLUGINS:
+    plugin_urlpatterns = get_plugin_urlpatterns(plugin)
+    if plugin_urlpatterns:
+        urlpatterns += plugin_urlpatterns
+
```
ostcar: I don't like this. This should at least be in a separate function. See #1332

ostcar:
One comment.

If you like, you can not change the urls.py and open an issue to give plugins the possibility to add urls.

They already can add rest-elements and therefore also add urls.

ostcar:
Please rebase and merge

PR #1602, opened 2015-07-22
Added views to create, update and delete MotionPoll objects.

normanjaeckel:


ostcar:
I don't like that the MotionPoll is not part of the Motion. There can be no MotionPoll without a Motion. Therefore thy should only be accessible through motion.poll.

(The same for assignments.)

So I am against the new route motions/motionpoll

normanjaeckel:
We can change this. Do you want to open an issue for this? At the moment I just do the same as for assignments.

PR #1603, opened 2015-07-24
Fixed error when starting webbrowser. Added info on port to general s…

normanjaeckel:
…tart script message.

### DIFF
--- a/openslides/utils/main.py
+++ b/openslides/utils/main.py
@@ -281,15 +281,19 @@ def start_browser(browser_url):
     Launches the default web browser at the given url and opens the
     webinterface.
     """
-    browser = webbrowser.get()
+    try:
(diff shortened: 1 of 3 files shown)


openslides/utils/main.py
```diff
@@ -281,15 +281,19 @@ def start_browser(browser_url):
     Launches the default web browser at the given url and opens the
     webinterface.
     """
-    browser = webbrowser.get()
+    try:
+        browser = webbrowser.get()
+    except webbrowser.Error as e:
+        print 'Web browser controller error: %s' % e
```
ostcar: print should be called as function and not as statement. Even in python 2.6 and python 2.7

ostcar:
One suggestion. You can fix it or merge immediately

PR #1604, opened 2015-08-31
Set permissions for OPTIONS requests.

normanjaeckel:


ostcar:
looks good. Please merge after travis is fixed

normanjaeckel:
Please review again and merge.

ostcar:
Could you look by the way to any other model.py if blank=True is correct?

ostcar:
In not. Please merge

PR #1625, opened 2015-09-06
Add upload feature to mediafiles module

boehlke:
TODO
- Update does not work when file is not set
- Fix/improve form error handling

### DIFF
--- a/openslides/mediafiles/migrations/0002_auto_20150906_1246.py
+++ b/openslides/mediafiles/migrations/0002_auto_20150906_1246.py
@@ -1,0 +1,20 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('mediafiles', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='mediafile',
+            name='is_presentable',
+        ),
+        migrations.AlterField(
+            model_name='mediafile',
+            name='title',
+            field=models.CharField(max_length=255, null=True, unique=True, blank=True, verbose_name='Title'),
+        ),
+    ]
--- a/openslides/mediafiles/projector.py
+++ b/openslides/mediafiles/projector.py
@@ -16,11 +16,9 @@ class MediafileSlide(ProjectorElement):
     def get_context(self):
         pk = self.config_entry.get('id')
         try:
-            mediafile = Mediafile.objects.get(pk=pk)
+            Mediafile.objects.get(pk=pk)
         except Mediafile.DoesNotExist:
             raise ProjectorException(_('File does not exist.'))
-        if not (mediafile.is_presentable and mediafile.filetype == 'application/pdf'):
-            raise ProjectorException(_('File is not presentable.'))
         return {'id': pk}
 
     def get_requirements(self, config_entry):
--- a/openslides/mediafiles/serializers.py
+++ b/openslides/mediafiles/serializers.py
@@ -1,6 +1,25 @@
-from openslides.utils.rest_api import ModelSerializer, SerializerMethodField
+import mimetypes
 
+from django.db import models as dbmodels
+
+from ..utils.rest_api import FileField, ModelSerializer, SerializerMethodField
 from .models import Mediafile
+
+
+class AngularCompatibleFileField(FileField):
+
+    def to_internal_value(self, data):
+        if data == '':
+            return None
+        return super(AngularCompatibleFileField, self).to_internal_value(data)
+
+    def to_representation(self, value):
+        if value is None:
+            return None
+        return {
+            'name': value.name,
+            'type': mimetypes.guess_type(value.path)[0]
+        }
 
 
 class MediafileSerializer(ModelSerializer):
@@ -8,6 +27,14 @@ class MediafileSerializer(ModelSerializer):
     Serializer for mediafile.models.Mediafile objects.
     """
     filesize = SerializerMethodField()
+
+    def __init__(self, *args, **kwargs):
+        """
+        This constructor overwrites the FileField field serializer to return the file meta data in a way that the
+        angualarjs upload module likes
+        """
+        super(MediafileSerializer, self).__init__(*args, **kwargs)
+        self.serializer_field_mapping[dbmodels.FileField] = AngularCompatibleFileField
 
     class Meta:
         model = Mediafile
@@ -18,8 +45,7 @@ class MediafileSerializer(ModelSerializer):
             'uploader',
             'filesize',
             'filetype',
-            'timestamp',
-            'is_presentable',)
+            'timestamp',)
 
     def get_filesize(self, mediafile):
         return mediafile.get_filesize()
--- a/openslides/mediafiles/static/templates/mediafiles/mediafile-form.html
+++ b/openslides/mediafiles/static/templates/mediafiles/mediafile-form.html
@@ -9,22 +9,25 @@
 </div>
 
 <form name="mediafileForm">
+
+  <div class="form-group">
+    <div ng-if="mediafile.id">
+      <span translate>Current value: </span>{{ mediafile.filename }}
+    </div>
+    <input type="file" ngf-select ng-model="mediafile.mediafile" required/>
+  </div>
+
   <div class="form-group">
     <label for="inputTitle" translate>Title</label>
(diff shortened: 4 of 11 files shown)


openslides/mediafiles/migrations/0002_auto_20150906_1246.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
normanjaeckel: These two lines can be removed.

openslides/mediafiles/projector.py
```diff
@@ -16,11 +16,9 @@ class MediafileSlide(ProjectorElement):
     def get_context(self):
         pk = self.config_entry.get('id')
         try:
-            mediafile = Mediafile.objects.get(pk=pk)
+            Mediafile.objects.get(pk=pk)
         except Mediafile.DoesNotExist:
             raise ProjectorException(_('File does not exist.'))
-        if not (mediafile.is_presentable and mediafile.filetype == 'application/pdf'):
```
normanjaeckel: What if the mediafile object isn't a PDF and therefor isn't presentable. We should raise ProjectorException in this case again, shouldn't we?

openslides/mediafiles/serializers.py
```diff
@@ -1,14 +1,38 @@
-from openslides.utils.rest_api import ModelSerializer, SerializerMethodField
+import mimetypes
 
+from django.db import models as dbmodels
+from rest_framework.fields import FileField
+
+from ..utils.rest_api import ModelSerializer, SerializerMethodField
```
normanjaeckel: @ostcar I think this style `..` is better than ours but do `from openslides.utils...` all the time. What do you think?

openslides/mediafiles/serializers.py
```diff
@@ -1,14 +1,38 @@
...
+        if data == '':
+            return None
+        return super(AngularCompatibleFileField, self).to_internal_value(data)
+
+    def to_representation(self, value):
+        if not value:
+            return None
+        return {
+            'name': value.name,
+            'type': mimetypes.guess_type(value.path)[0]
```
normanjaeckel: It looks like that the filetype is already in the dataset, see fields in serializer below.

openslides/mediafiles/serializers.py
```diff
@@ -1,14 +1,38 @@
...
+
+
 class MediafileSerializer(ModelSerializer):
     """
     Serializer for mediafile.models.Mediafile objects.
     """
     filesize = SerializerMethodField()
 
+    def __init__(self, *args, **kwargs):
+        super(MediafileSerializer, self).__init__(*args, **kwargs)
```
normanjaeckel: Please add a short docstring or comment about the customization here.

openslides/mediafiles/serializers.py
```diff
@@ -1,14 +1,38 @@
-from openslides.utils.rest_api import ModelSerializer, SerializerMethodField
+import mimetypes
 
+from django.db import models as dbmodels
+from rest_framework.fields import FileField
```
normanjaeckel: For unknown reason we import all rest_framework stuff from our utils.rest_api so we can change this globally later if required. So you have to add it there and change the import. Maybe use `# noqa` to disabled flake8 over there.

normanjaeckel:
Some remarks.

openslides/mediafiles/projector.py
boehlke: I think we agreed that the client handles this. The server now only saves the files.

openslides/mediafiles/serializers.py
boehlke: Sorry, I don't understand. What exactly is the problem here?

openslides/mediafiles/serializers.py
boehlke: What do you mean? This Field costomizes the serialization of FileFields. I thought, this would be a nice, reusable way to implement a serialization which the angular upload module likes.

openslides/mediafiles/projector.py
normanjaeckel: Ok, so we say the server may provide every file to be presentable.

openslides/mediafiles/serializers.py
normanjaeckel: See `openslides/utils/rest_api.py`. All openslides apps import rest_framework stuff from this file rather than from the rest_framework module directly. So we can override these classes later if necessary. We want to do this only for rest_framework stuff at the moment because we do not know where the travel goes to ... :-)

openslides/mediafiles/serializers.py
normanjaeckel: Ok but then we don't need the filetype field in models.py line 33. Also the customized save method can be removed.

openslides/mediafiles/serializers.py
ostcar: +1

emanuelschuetze:
@boehlke ping

emanuelschuetze:
If you're done with this pull request please reassign to @normanjaeckel . Thx.

boehlke:
Hi @normanjaeckel,
I think/hope, now the request is fine for now.
Regards,
André

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
```diff
@@ -45,24 +45,25 @@ <h1 translate>Files</h1>
         </i>
       <th os-perms="mediafiles.can_manage core.can_manage_projector" class="minimum">
         <translate>Actions</translate>
-  <tbody>
+        <tbody>
```
ostcar: Can you remove this tab?

emanuelschuetze:
@boehlke So finally, please rebase to one commit and set it on top of master.
Note: Please always reassign pull requests with a short comment that we can see who "has the ball". Thx.

boehlke:
@normanjaeckel @ostcar Please final review!

openslides/mediafiles/migrations/0002_auto_20150906_1246.py
ostcar: @boehlke I know that this lines are added by django. But it is only necessary for the compatibility with python 2. OpenSlides needs python 3, so this line can be deleted.

openslides/mediafiles/serializers.py
```diff
@@ -1,14 +1,41 @@
...
+class AngularCompatibleFileField(FileField):
+
+    def to_internal_value(self, data):
+        if data == '':
+            return None
+        return super(AngularCompatibleFileField, self).to_internal_value(data)
+
+    def to_representation(self, value):
+        if not value:
+            return None
```
ostcar: Are you sure that you want to test for False and not for None.?
```python
if value is None:
    return None
```

openslides/mediafiles/static/templates/mediafiles/mediafile-form.html
```diff
@@ -9,23 +9,26 @@ <h1 ng-if="!mediafile.id" translate>New file</h1>
...
+    <label for="inputTitle" translate>Title</label>
+    <input type="text" ng-model="mediafile.title" class="form-control" name="inputTitle">
   </div>
 
+  <span ng-show="mediafile.mediafile.result">Upload Successful</span>
+  <span class="err" ng-show="errorMsg">{{ errorMsg }}</span>
+
+  <i ng-show="mediafile.file.$error.required">*required</i><br>
+  <i ng-show="mediafile.file.$error.maxSize">File too large
+    {{ picFile.size / 1000000|number:1}}MB: max {{ mediafile.mediafile.$errorParam}}</i>
```
ostcar: Proposal: Add a space before the closing ```}}``` or remove the space after the opening ```{{```

openslides/mediafiles/migrations/0002_auto_20150906_1246.py
boehlke: Thanks, done.

openslides/mediafiles/serializers.py
boehlke: Thanks, you are right. Changed that.

openslides/mediafiles/serializers.py
boehlke: Done.

ostcar:
@normanjaeckel please review or merge

PR #1630, opened 2015-09-08
Removed imports from openslides.users in other apps

normanjaeckel:


openslides/agenda/models.py
```diff
@@ -372,14 +380,14 @@ class Speaker(RESTModelMixin, models.Model):
 
     objects = SpeakerManager()
 
-    user = models.ForeignKey(User)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL)
```
ostcar: Are you absolutely sure you can use the setting? This can be a string. See https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#substituting-a-custom-user-model

openslides/motions/pdf.py
```diff
@@ -278,15 +278,18 @@ def motion_poll_to_pdf(pdf, poll):
 
     # set number of ballot papers
     if ballot_papers_selection == "NUMBER_OF_DELEGATES":
-        # TODO: get this number from users
         try:
-            if Group.objects.get(pk=3):  # TODO: Find a better way
-                number = User.objects.filter(groups__pk=3).count()
-        except Group.DoesNotExist:
+            from openslides.users.models import Group
```
ostcar: This will only fail when the user app is deleted, but not, if another user app is used in the settings.

openslides/agenda/models.py
normanjaeckel: `settings.AUTH_USER_MODEL` is always a string but this is valid, see example in the box [here](https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#referencing-the-user-model).

openslides/motions/pdf.py
normanjaeckel: Changed. See next commit.

openslides/agenda/models.py
ostcar: ok

ostcar:
Please rebase and merge

Issue #1633, opened 2015-09-12
Add zoom and scroll function for projector controls


emanuelschuetze:
like in OpenSlides 1.x:
- zoom up/down
- scroll up/down
- reset zoom & scroll

normanjaeckel:
Hm. What about just saving `{"zoom": "42"}` into the projector element config field? This would not be a global zoom but a zoom for one projector element.

ostcar:
I think we need a global zoom. Maybe each projector model could have one more db-field for global config. This could also be useful when we start the multiprojector to give each projector a name, a room or another human readable identifier

PR #1634, opened 2015-09-12
Projector elements controls

emanuelschuetze:


emanuelschuetze:
this pull request is currently just for discussion and core "preview"

openslides/core/static/js/core/core.js
```diff
@@ -605,6 +605,46 @@ angular.module('OpenSlidesApp.core.site', [
...
         );
     };
+
+    // *** countdown ***
+    $scope.countdowns = [];
+    $scope.messages = [];
+    _.forEach(Projector.get(1).elements, function(element) {
+        if (element.name == 'core/countdown') {
+            $scope.countdowns.push(element);
+        }
```
emanuelschuetze: @normanjaeckel I need the countdown uuid within of each element so that I can push the whole element to may array 'countdowns'. Later in template I want to access the id by countdown.id. Can you change it?

openslides/core/static/js/core/core.js
```diff
@@ -605,6 +605,46 @@ angular.module('OpenSlidesApp.core.site', [
...
+        if (element.name == 'core/message') {
+            $scope.messages.push(element);
+        }
+    });
+    $scope.addCountdown = function () {
+        $http.post(
+            '/rest/core/projector/1/activate_elements/',
+            [{name: 'core/countdown', status: 'stop', visible: true,
+                countdown_time: 42, stable: false}]
+            // TODO countdown_time: config auslesen
```
emanuelschuetze: How can I get the countdown_time config from projector here?

openslides/core/static/js/core/core.js
```diff
@@ -605,6 +605,46 @@ angular.module('OpenSlidesApp.core.site', [
...
+    var stopped;
+    $scope.startCountdown = function (countdown) {
+        stopped = $timeout(function() {
+            countdown.countdown_time--;
+            $scope.startCountdown(countdown);
+        }, 1000);
+    };
+    $scope.stopCountdown = function(){
+       $timeout.cancel(stopped);
+    };
```
emanuelschuetze: Currently I can start and stop the countdown locally (without any notice to the projector). What and when the projector need the update request?

openslides/core/static/templates/core/customslide-list.html
```diff
@@ -48,15 +48,97 @@ <h1 translate>Custom slides</h1>
...
+
+      <!-- projector live view -->
+      <h3 translate>Projector</h4>
       <a ui-sref="projector" target="_blank">
         <div id="iframewrapper">
           <iframe id="iframe" src="/projector" frameborder="0"></iframe>
           <div id="iframeoverlay"></div>
         </div>
       </a>
+      <!-- TODO: add functions for projector control buttons -->
```
emanuelschuetze: I added already the (disabled) projector control buttons. Please add the function on server-side (see #1633).

openslides/core/static/templates/core/customslide-list.html
```diff
@@ -48,15 +48,97 @@ <h1 translate>Custom slides</h1>
...
+      <div os-perms="core.can_manage_projector">
+        <a os-perms="core.can_manage_projector"
+            class="btn btn-default btn-sm disabled"
+            title="{{ 'Edit current slide' | translate}}">
```
emanuelschuetze: I want to edit the current slide (if use hase manage_projector permissions). How can I get the right link?

openslides/core/static/templates/core/customslide-list.html
```diff
@@ -48,15 +48,97 @@ <h1 translate>Custom slides</h1>
...
+            title="{{ 'Add countdown' | translate}}">
+          <i class="fa fa-plus"></i> <translate>Add new countdown</translate>
+        </a>
+      </div>
+
+       <!-- messages -->
+      <div os-perms-lite="core.can_manage_projector">
+        <h3 translate>Messages</h3>
+        <ul ng-repeat="message in messages" class="message list-group">
+          <li class="list-group-item">{{message}}</li>
```
emanuelschuetze: @normanjaeckel I can't test the projector messages because the rest framework view shows always "Authentication credentials were not provided.".

emanuelschuetze:
@normanjaeckel ( and maybe @ostcar): I added some questions which blocks me to finish this pull request. Hope you can give me some hints. Thanks.

openslides/core/static/js/core/core.js
normanjaeckel: What about not using an array but an object. In the forEach loop you can also get access to the dict key and so use it.

openslides/core/static/js/core/core.js
normanjaeckel: You take it from OpenSlides config, not from projector model. See openslides.core.signals.py

openslides/core/static/js/core/core.js
normanjaeckel: The scenario is like this:

* The projector entry says "status": "stop". This means the client should not have any `$timeout` loop running.
* Then someone clicks an start button.
* The client sends one request to the server. That's all. Don't start any `$timeout` loop.
* A while later the server sends an autoupdate info to the client telling them that "status" is now "running". This means the client should have such a `$timeout` loop running until it gets another autoupdate info.

openslides/core/static/templates/core/customslide-list.html
normanjaeckel: This is not easy for we might have multiple projector elements activated. Loop over all active projector elements and add all links according to the name of the elements. Filter non editable elements like clock and countdown (but not message).

openslides/core/static/templates/core/customslide-list.html
normanjaeckel: I propose looping over all elements, filter elements like clock and countdown and add control buttons for the rest. If one clicks such a button this element gets a new "zoom" or "scroll" entry.

openslides/core/static/templates/core/customslide-list.html
normanjaeckel: Hm. There should not be any difference between messages and any other projector element. There is only one REST framework viewset for them all. Please give me a hint to reproduce the but.

normanjaeckel:
All hints given.

openslides/core/static/templates/core/customslide-list.html
emanuelschuetze: Done. Rest framework works now.

openslides/core/static/js/core/core.js
```diff
@@ -606,6 +633,55 @@ angular.module('OpenSlidesApp.core.site', [
...
+        var data = {};
+        data[uuid] = {
+            "visible": !visible
+        };
+        $http.post('/rest/core/projector/1/update_elements/', data);
+    };
+    $scope.startCountdown = function (countdown) {
+        var data = {};
+        var serverOffset = Date.parse(new Date().toUTCString()) - countdown.context.server_time;
+        var time = Date.now() - serverOffset + countdown.countdown_time;
```
emanuelschuetze: @normanjaeckel Here I need your help how it should work correctly. I use your calculation from docstring: "countdown_time = now - serverTimeOffset + duration". But that seems to be not correct in my start method.

normanjaeckel: You must not calculate the server offset if someone clicks start but immendiatelly after a new projector element comes in via autoupdate or after initial loading.

normanjaeckel: I think the start stop buttons have to be part of a CountdownDetailCtrl.

openslides/core/static/js/core/core.js
```diff
@@ -620,6 +620,91 @@ angular.module('OpenSlidesApp.core.site', [
...
+        $http.post('/rest/core/projector/1/deactivate_elements/', [countdown.uuid]);
+    };
+    $scope.showCountdown = function (visible, uuid) {
+        var data = {};
+        data[uuid] = { "visible": !visible };
+        $http.post('/rest/core/projector/1/update_elements/', data);
+    };
+    $scope.startCountdown = function (countdown) {
+        var data = {};
+        var serverOffset = Date.parse(new Date().toUTCString()) - countdown.context.server_time;
```
normanjaeckel: `var serverOffset = new Date().getTime() / 1000 - countdown.context.server_time;`

openslides/core/static/js/core/core.js
normanjaeckel: Don't use milliseconds but seconds: `Date.now() / 1000`

emanuelschuetze:
@normanjaeckel Can you look at the travis failure in test_invalid_slide_on_default_projector please? I can't se why this test failed now.

openslides/core/static/js/core/core.js
```diff
@@ -620,6 +646,125 @@ angular.module('OpenSlidesApp.core.site', [
...
+                    // TODO: save timeout object that other client can use it for cancel $interval
+                    var timeout = $interval( function() { $scope.callCountdownTimer(value); }, 10000);
+                    if (value.timeout != '') {
+                        console.log(timeout);
+                        var data = {};
+                        data[value.uuid] = { "timeout": timeout.intervalId };
+                        $http.post('/rest/core/projector/1/update_elements/', data);
+                    }
+                } else {
+                    $interval.cancel(value.timeout);
```
emanuelschuetze: @normanjaeckel Any idea how we can save the $interval object that (maybe an other client) can load and cancel this $interval (countdown)? I tried to push the whole timeout object to elements, which is still empty. And only the "intervalID" is not enough.

normanjaeckel:
@emanuelschuetze The tests fail because you added the UUID to the content of the projector element objects. You also have to add in the expected test result.

openslides/core/static/js/core/core.js
normanjaeckel: Hm. I don't understand what you try to do here. It is not possible to send some data through the server in order to *control* other clients that way. I think you should not try to send anything in this case. Just let the countdown run forever. The timeout may be destroyed (and rebuild) if the projector gets an autoupdate.

emanuelschuetze:
@normanjaeckel Integration tests are fixed. Thx for the hint.

emanuelschuetze:
Ready for first review!

openslides/core/static/js/core/core.js
```diff
@@ -592,7 +618,7 @@ angular.module('OpenSlidesApp.core.site', [
 })
 
 // Customslide Controller
-.controller('CustomslideListCtrl', function($scope, Customslide) {
+.controller('CustomslideListCtrl', function($scope, $http, Customslide) {
```
normanjaeckel: Please use the new style with dependency strings.

openslides/core/static/js/core/core.js
```diff
@@ -620,6 +646,207 @@ angular.module('OpenSlidesApp.core.site', [
     };
 })
 
+// Projector Control Controller
+.controller('ProjectorControlCtrl', function($scope, $http, $interval, $state, Config, Projector) {
```
normanjaeckel: Here the same.

openslides/core/static/js/core/core.js
```diff
@@ -787,5 +1024,34 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
     // Attention! Each object that is used here has to be dealt on server side.
     // Add it to the coresponding get_requirements method of the ProjectorElement
     // class.
-    $scope.serverOffset = Date.parse(new Date().toUTCString()) - $scope.element.context.server_time;
+    $scope.serverOffset = Date.now() / 1000 - $scope.element.context.server_time;
+})
+
+.controller('SlideCountdownCtrl', function($scope, $interval) {
```
normanjaeckel: Here the same.

openslides/core/static/js/core/core.js
```diff
@@ -787,5 +1024,34 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+        interval = $interval( function() { 
+            $scope.restTime = Math.floor( $scope.element.countdown_time - Date.now() / 1000 + $scope.serverOffset );
+        }, 1000);
+    } else {
+         $scope.restTime = $scope.element.countdown_time;
+    }
+
+})
+
+.controller('SlideMessageCtrl', function($scope) {
```
normanjaeckel: Here the same.

emanuelschuetze:
Fixed all comments. Clean up code. All done. Ready for review.

emanuelschuetze:
@normanjaeckel I want to delete the default countdown because a use can just add a new countdown if needed. So how I can delete the predefined countdown?

normanjaeckel:
The default countdown is added in a core/migrations file. The correct way would be to create a new file and delete it manually. The smart way would be to remove all migrations files and write them new so the countdown isn't created during migration. The question is how we want to be consistent with our alpha release.

openslides/core/static/js/core/core.js
```diff
@@ -176,6 +182,32 @@ angular.module('OpenSlidesApp.core', [
     });
 }])
 
+/* Converts number of seconds into string "hh:mm:ss" or "mm:ss" */
+.filter('osSecondsToTime',function() {
```
normanjaeckel: Add one space behind the comma.

openslides/core/static/js/core/core.js
```diff
@@ -770,22 +982,65 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+        // Add it to the coresponding get_requirements method of the ProjectorElement
+        // class.
+        $scope.servertime = ( Date.now() / 1000 - $scope.serverOffset ) * 1000;
+    }
+])
+
+.controller('SlideCountdownCtrl', [
+    '$scope',
+    '$interval',
+    function($scope, $interval) {
```
normanjaeckel: Somewhere in this function you have to add an interval cancel for the case the the Ctrl goes down. See note box here: https://docs.angularjs.org/api/ng/service/$interval

openslides/core/static/js/core/core.js
```diff
@@ -592,33 +628,208 @@ angular.module('OpenSlidesApp.core.site', [
...
+            $.each(Projector.get(1).elements, function(key, value) {
+                if (value.name != 'core/clock' &&
+                    value.name != 'core/countdown' &&
+                    value.name != 'core/message' ) {
+                    $state.go(value.name.replace('/', '.')+'.detail.update', {id: value.id });
+                }
+            });
+        };
+    }
+])
```
normanjaeckel: I do not know whether we need to cancel the intervals on destruction of this controller too.
@ostcar What happens if the user is here (projector tab) and clicks on e. g. Motion. Is this controller destroyed and rebuild if here comes back later?
If yes @emanuelschuetze has to implement a listener on the destruction event to cancel the intervals so the browser memory keeps clean.

openslides/core/static/js/core/core.js
emanuelschuetze: Done.

openslides/core/static/js/core/core.js
emanuelschuetze: Done. Thx for the example. It works.

openslides/core/static/js/core/core.js
emanuelschuetze: I tested it: If user changed page to e.g. Motions the interval timer is still running. So I added the same  "$scope.$on('$destroy', function() {" function to cancel all intervals. It works now. Ok?

emanuelschuetze:
All done. Also added a new migrations file to reset the countdown.
@normanjaeckel ready for next review.
@ostcar Please start review so that we can merge it this weekend. Thx.

openslides/core/static/js/core/core.js
```diff
@@ -592,33 +628,217 @@ angular.module('OpenSlidesApp.core.site', [
...
+                if ( $scope.countdowns[i].interval )
+                    $interval.cancel($scope.countdowns[i].interval);
+            }
+
+        });
+
+        // *** countdown functions ***
+        $scope.calculateCountdownTime = function (countdown) {
+            countdown.seconds = Math.floor( countdown.countdown_time - Date.now() / 1000 + $scope.serverOffset );
+            console.log(countdown.uuid + " " +countdown.seconds);
```
normanjaeckel: Please remove this line.

normanjaeckel:
To fix the test_begin_speech... test you have to add something like this to the begin of the test:
```python
projector = Projector.objects.get(pk=1)
projector.config['.....<add random uuid here>.....'] = {'name': 'core/countdown', .... <add all required properties here>}
projector.save()
```

openslides/core/static/js/core/core.js
emanuelschuetze: Done

emanuelschuetze:
All done (I added some fixes for countdown index position and an improvement for description on projector). Ready for review.

normanjaeckel:
Ok.

openslides/core/static/js/core/core.js
```diff
@@ -176,6 +182,32 @@ angular.module('OpenSlidesApp.core', [
     });
 }])
 
+/* Converts number of seconds into string "hh:mm:ss" or "mm:ss" */
+.filter('osSecondsToTime', function() {
```
ostcar: Proposal: Break the line before ```function``` and add a space between ```function``` and ```()```

openslides/core/static/js/core/core.js
```diff
@@ -176,6 +182,32 @@ angular.module('OpenSlidesApp.core', [
...
+    return function(totalseconds) {
```
ostcar: Proposal: Add a space between ```function``` and ```(totalseconds)```

openslides/core/static/js/core/core.js
ostcar: Proposal. Put the function inside a list, so it is easier to add dependencies in the future.

openslides/core/static/js/core/core.js
```diff
@@ -592,33 +628,231 @@ angular.module('OpenSlidesApp.core.site', [
...
+    'Projector',
+    function($scope, $http, $interval, $state, Config, Projector) {
+         // bind projector elements to the scope, update after projector changed
+        $scope.$watch(function () {
+            return Projector.lastModified(1);
+        }, function () {
+            // stop ALL interval timer
+            for (var i=0; i<$scope.countdowns.length; i++) {
+                if ( $scope.countdowns[i].interval )
+                    $interval.cancel($scope.countdowns[i].interval);
```
ostcar: Please add ```{``` and ```}``` so the if-block can be easier changed in the future

openslides/core/static/js/core/core.js
```diff
@@ -592,33 +628,231 @@ angular.module('OpenSlidesApp.core.site', [
...
+            }
+            // rebuild all variables after projector update
+            $scope.rebuildAllElements();
+        });
+        $scope.$on('$destroy', function() {
+            // Cancel all intervals if the controller is destroyed
+            for (var i=0; i<$scope.countdowns.length; i++) {
+                if ( $scope.countdowns[i].interval )
+                    $interval.cancel($scope.countdowns[i].interval);
```
ostcar: Here the same

openslides/core/static/js/core/core.js
```diff
@@ -592,33 +628,231 @@ angular.module('OpenSlidesApp.core.site', [
...
+        };
+        $scope.removeCountdown = function (countdown) {
+            var data = {};
+            var delta = 0;
+            // rebuild index for all countdowns after the selected (deleted) countdown
+            for (var i=0; i<$scope.countdowns.length; i++) {
+                if ( $scope.countdowns[i].uuid == countdown.uuid ) {
+                    delta = 1;
+                } else {
+                    if (delta > 0) {
```
ostcar: You can use ```else if``` there

ostcar:
Some proposals. Please fix them if you like, then rebase and merge

emanuelschuetze:
Fixed all comments.

emanuelschuetze:
All done. Rebased.

PR #1638, opened 2015-09-16
Configure js-data for projector elements to "replace" on inject.

### LINKED ISSUE #1632: Overlapping projector slides
There seems to be something wrong if you activate e.g. the agenda slide and some item detail slides or custom slides. The 'prune_elements' post seems work correctly but the old slide on the projector will not clear. After reload the projector the new slide was shown correctly (without any old slid in background).
So the client-side autoupdate process for slides seems to be broken.

ostcar:
fixes #1632

### DIFF
--- a/openslides/core/static/js/core/core.js
+++ b/openslides/core/static/js/core/core.js
@@ -158,9 +158,21 @@ angular.module('OpenSlidesApp.core', [
     });
 }])
 
+/* Model for a projector.
+ *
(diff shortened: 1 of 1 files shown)


normanjaeckel:
Easy. :-)

normanjaeckel:
Optional: Please add a short comment why this replace is used there so we can keep in mind why it ist there and not anywhere else.

For the rest: Ready to merge.

ostcar:
I added the comment

PR #1642, opened 2015-09-17
Cleaned up users app.

normanjaeckel:


openslides/utils/utils.py
```diff
@@ -1,25 +1,10 @@
...
-    Deletes the permissions, django creates by default for the admin.
-    """
-    # TODO: Find a way not to create the permissions in the first place.
-    #       Meta.default_permissions does not work, because django will
-    #       nevertheless create permissions for its own models like "group"
-    for p in Permission.objects.all():
-        if (p.codename.startswith('add') or
-                p.codename.startswith('delete') or
-                p.codename.startswith('change')):
-            p.delete()
```
ostcar: Are you sure this can be deleted? How do you plan to delete the django-default passwords?

normanjaeckel: This code is not used at the moment. Someone has to reimplement a new solution in the users app. See #1643

ostcar:
ok. Please rebase and merge

PR #1648, opened 2015-09-26
Refactored view to provide OpenSlides JavaScript files.

normanjaeckel:
All JavaScript files are concated now. Also users app JS is refactored.

@ostcar Please have a look at this. For now everything is loaded with one request. I can add comments with the filenames of the concated files to the output if you like.


openslides/core/urls.py
```diff
@@ -16,9 +16,9 @@
         views.VersionView.as_view(),
         name='core_version'),
 
-    url(r'^angular_js/(?P<openslides_app>site|projector)/$',
-        views.AppsJsView.as_view(),
-        name='core_apps_js'),
+    url(r'^openslides_javascript/(?P<realm>site|projector)/$',
```
ostcar: I don't like the ur "openslides_javascript" because it is obvious that openslides give you openslides code, and we could create an alternative client in the future (query mobile?) that is also javascript.

So I prefer somethink with angular and without openslides in it

normanjaeckel: What about `url(r'^webclient/(?P<realm>site|projector)/$'`

normanjaeckel: This URL does not only provide AngularJS related stuff but also our third party libs and the like.

normanjaeckel:
See #1654. At the moment we do not merge everything together. Maybe later ...

PR #1656, opened 2015-10-21
Split and cleaned up JS files.

normanjaeckel:


openslides/agenda/static/js/agenda/base.js
```diff
@@ -0,0 +1,122 @@
+(function () {
```
ostcar: Why do you added this closure? We work with angular modules which do the same thing. I don't think that we need double closures.

normanjaeckel: We do not always just call an angular function. Sometimes there are extra functions defined in a file. I would say using a closure is always best practice to protect the global namespace.

ostcar: This other function should be coded as angular services.

I don't see the advantage to use two techincs to solve one problem.

normanjaeckel:
Anything important open here? So we can merge this.

PR #1658, opened 2015-10-26
Removed bleach.

normanjaeckel:


ostcar:
Yesterday on the telephone we sad, that we will no change the server side because we do not know yet, how we will handle malicious html in reportlab.

We will probably need bleach for that, if we can't get rid of the server side pdf creation.

normanjaeckel:
At the moment Reportlab just strips unknown or malicious html without even using bleach so we actually don't need bleach any more. (I do not find it in Reportlab Docs at the moment.)
Beautifulsoup4 is still required.

PR #1661, opened 2015-10-28
Added CKEditor to custom slide create and edit form.

normanjaeckel:
A full rerun of Gulp is required.

@emanuelschuetze Please test and check the options. Maybe ask @rolandgeider to help to fix the toolbar and the options object and to reinsert 'insertpre' if neccessary.

### DIFF
--- a/openslides/core/static/templates/core/customslide-form.html
+++ b/openslides/core/static/templates/core/customslide-form.html
@@ -14,8 +14,8 @@
     <input type="text" ng-model="customslide.title" class="form-control" name="inputTitle" required>
   </div>
   <div class="form-group">
-    <label for="textareaDesciption" translate>Text</label>
-    <textarea ng-model="customslide.text" class="form-control" name="textareaText" />
+    <label for="customSlideTextCKEditor" translate>Text</label>
+    <div id="customSlideTextCKEditor" ckeditor="CKEditorOptions" ng-model="customslide.text"></div>
   </div>
 
   <button type="submit" ng-click="save(customslide)" class="btn btn-primary" translate>
--- a/openslides/core/static/templates/index.html
+++ b/openslides/core/static/templates/index.html
@@ -9,6 +9,7 @@
 <link rel="stylesheet" href="static/css/openslides-libs.css">
 <link rel="stylesheet" href="static/css/app.css">
 <script src="static/js/openslides-libs.js"></script>
+<script src="static/ckeditor/ckeditor.js"></script>
 
 <!-- Navbar -->
 <nav id="header" class="navbar navbar-inverse">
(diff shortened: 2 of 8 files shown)


openslides/core/static/templates/index.html
```diff
@@ -9,6 +9,7 @@
 <link rel="stylesheet" href="static/css/openslides-libs.css">
 <link rel="stylesheet" href="static/css/app.css">
 <script src="static/js/openslides-libs.js"></script>
+<script src="static/ckeditor/ckeditor.js"></script>
```
ostcar: Is there a special reason why the ckeditor can not be part of openslides-libs.js?

openslides/core/static/templates/core/customslide-form.html
```diff
@@ -15,7 +15,7 @@ <h1 ng-if="!customslide.id" translate>New custom slide</h1>
   </div>
   <div class="form-group">
     <label for="textareaDesciption" translate>Text</label>
-    <textarea ng-model="customslide.text" class="form-control" name="textareaText" />
+    <div ckeditor="CKEditorOptions" ng-model="customslide.text"></div>
```
emanuelschuetze: we need a new id attribute for the <div> which has the same name as defined in <label for="..">. And please fix the wrong name in "for" attribute.

FYI: "The for attribute of the <label> tag should be equal to the id attribute of the related element to bind them together." (see http://www.w3schools.com/tags/tag_label.asp)

openslides/core/static/templates/index.html
normanjaeckel: It is an unknown reason. Please try to add
```json
"overrides": {
  "ckeditor": {
    "main": [
      "ckeditor.js"
    ]
  }
}
```
to bower.json. Then remove the line 12 in the index.html, run gulp and the ckeditor.js is now included our libs. But for unknown reason CKEditor now tries to load more files like icon images and language files from the wrong path instead of the global static path ... and crashes.

openslides/core/static/templates/core/customslide-form.html
normanjaeckel: Done.

openslides/core/static/templates/index.html
ostcar: It works if we:

1. Add this  code before openslides-lib.js is loaded:
```<script>
var CKEDITOR_BASEPATH= "/static/ckeditor/";
</script>
```
2. Add a lot of files from the ckeditor folder to our static folder.

Im not sure what to do.

normanjaeckel: At the moment I use an extra gulp task to copy the files, see gulpfile.js above. I think this would be enough. Integrating only one file more to the openslides-libs.js does not bring salvation to us.

ostcar: ok

emanuelschuetze:
@normanjaeckel I added a custom ckeditor config in pull request https://github.com/normanjaeckel/OpenSlides/pull/9. After that please rebase on top of master.

normanjaeckel:
Rebased. Ready to review and merge.

PR #1663, opened 2015-11-03
Motions

emanuelschuetze:
Full rework of motions list, detail and form template.
Added motion polls in detail view, quickedit mode in list view.

TODO (for later motion pull requests): motion csv import, versions overview table, list of speakers


ostcar:
I opend a pull request on your pr to fix travis: https://github.com/emanuelschuetze/OpenSlides/pull/6

The rest can be merged.

I have no idea how we should rebase this. @normanjaeckel any suggestions?

openslides/motions/static/templates/motions/motion-form.html
```diff
@@ -55,34 +44,45 @@ <h1 ng-if="!motion.id" translate>New motion</h1>
...
       <ui-select-choices repeat="user.id as user in users | filter: $select.search">
-        <div ng-bind-html="user.get_short_name() | highlight: $select.search"></div>
-        <small ng-bind-html="user.structure_level | highlight: $select.search"></small>
+        <div ng-bind-html="user.get_full_name() | highlight: $select.search"></div>
       </ui-select-choices>
     </ui-select>
   </div>
-  <!-- TODO: preselect default workflow -->
+  <!-- TODO: preselect default workflow
+  does not work: ng-init="motion.workflow = config('motions_workflow')"-->
```
emanuelschuetze: @ostcar Any hint how I can fix this TODO?

openslides/motions/static/templates/motions/motion-list.html
```diff
@@ -21,78 +21,173 @@ <h1 translate>Motions</h1>
...
+          class="btn btn-primary btn-sm form-control">
+        <i class="fa fa-trash fa-lg"></i>
+        <translate>Delete selected motions</translate>
+      </a>
+      <!-- TODO: add filters -->
+      &nbsp;
+      <select ng-model="stateFilter" class="form-control" id="stateFilter">
+        <option value="" translate>--- Select state ---</option>
+        <!-- TODO: use default workflow id from config('motions_workflow') -->
+        <option ng-repeat="state in workflows[0].states" value="{{ state.id }}">{{ state.name }}</option>
```
emanuelschuetze: @ostcar Any hint how can i use the default workflow to get all states for filter? Currently I use the first workflow (0).

ostcar: You can get the id of the default workflow from the config. But this does not give you all states. To get all stats you need to write a function that goes through all workflows and returns all stats

ostcar: Another method would be to get to all motions and return motion.state for each motion. That would only return the stats that are used right now

openslides/motions/models.py
```diff
@@ -409,6 +409,7 @@ def workflow(self):
         """
         Returns the id of the workflow of the motion.
         """
+        # TODO: Rename to workflow_id
```
normanjaeckel: @ostcar What does this TODO mean. Can it be fixed easily?

openslides/motions/models.py
```diff
@@ -763,6 +764,7 @@ class State(RESTModelMixin, models.Model):
     This behavior can be changed by the form and view, e. g. via the
     MotionDisableVersioningMixin.
     """
+    # TODO: preferred_for = ChoiceField
```
normanjaeckel: Please remove this TODO and add an issue about this idea.

openslides/motions/serializers.py
```diff
@@ -58,19 +57,20 @@ class Meta:
             'versioning',
             'leave_old_version_active',
             'dont_set_identifier',
-            'next_states',)
+            'next_states',
+            'workflow')
```
normanjaeckel: I think this is redundant. Or is it required because of JSData?

openslides/motions/serializers.py
```diff
@@ -97,6 +97,7 @@ class Meta:
         model = MotionPoll
         fields = (
             'id',
+            'motion',
```
normanjaeckel: Here the same?!

openslides/motions/serializers.py
```diff
@@ -219,7 +223,7 @@ def create(self, validated_data):
         motion.reason = validated_data.get('reason', '')
         motion.identifier = validated_data.get('identifier')
         motion.category = validated_data.get('category')
-        motion.reset_state(validated_data.get('workflow', int(config['motions_workflow'])))
+        motion.reset_state(validated_data.get('workflow_id'))
```
normanjaeckel: Why not using the config if the client does not send an workflow_id? The field is not required so we need a fallback here.

openslides/motions/serializers.py
```diff
@@ -241,9 +245,9 @@ def update(self, motion, validated_data):
                 setattr(motion, key, validated_data[key])
 
         # Workflow.
-        workflow = validated_data.get('workflow')
-        if workflow is not None and workflow != motion.workflow:
-            motion.reset_state(workflow)
+        workflow_id = validated_data.get('workflow_id')
+        if workflow_id is not None and workflow_id != motion.workflow:
```
normanjaeckel: I am not sure whether `workflow_id != motion.workflow` is correct or not. Maybe it has to be `workflow_id != motion.workflow_id`

openslides/motions/static/templates/motions/motion-list.html
emanuelschuetze: What I want: All available states only from the default workflow (nobody should change the workflow during event). So I have the workflow id by `config('motions_workflow')`. But how can I get all these states in template? workflows is an array. I have not the index but the id of the default workflow. What I search is something like `workfows[id=config('motions_workflow')]`...

normanjaeckel: I think you have to loop over all workflows and add all states here. Or you loop over all motions as @ostcar wrote.

normanjaeckel: During an event motions might be in different workflows. So your filter has to loop over all possible states.

emanuelschuetze: > During an event motions might be in different workflows. So your filter has
> to loop over all possible states.

Then the filter is not very useable because to many (unordered) state entries. I 
prefer to use only the selected (default) workflow. I can NOT imagine that a 
manager change the workflow during event. (I know it is possible in OS!)

openslides/motions/static/templates/motions/motion-form.html
ostcar: I would fix this in the MotionCreateCtrl

$scope.motion = {workflow: config('motions_workflow')}

openslides/motions/models.py
ostcar: The methods name is "workflow" but it returns the id of a workflow. This is missleading.

This TODO can be fixed by seaching for all places, where workflow is called and change it to workfow_id. Feed free to do it.

openslides/motions/serializers.py
ostcar: Yes it is redundant. I did not find another way in JSData without this redundant field.

openslides/motions/serializers.py
ostcar: See above.

openslides/motions/serializers.py
ostcar: The method "reset_state" uses the default workflow, if it is called with "None").

This works. See the tests.

openslides/motions/serializers.py
ostcar: motion.workflow does not return a workflow object, but a workflow-id. See by todo above.

emanuelschuetze:
I get a "405 METHOD NOT ALLOWED" if I try to update a poll. Any idea?

emanuelschuetze:
Same for DELETE poll request.

normanjaeckel:
@emanuelschuetze Please rewrite/check all commit messages an may be rebase if there is more than one commit of the same author in series.

emanuelschuetze:
@normanjaeckel: I need a motion.get_allowed_actions check like in OpenSlides 1.7, e.g. for checking if support/unsupport action is allowed https://github.com/OpenSlides/OpenSlides/blob/stable/1.7.x/openslides/motion/templates/motion/motion_detail.html#L313
Is this already possible via REST api?

normanjaeckel:
The REST API does not have this at the moment. I can implement it (or tell you how to do it yourself) but it is a little bit “quick and dirty”.

emanuelschuetze:
All done. Ready for review and tests. Needs full run of bower install and gulp.

emanuelschuetze:
There is still a problem: motion.allowed_actions is empty (only on client-side) after reload the page! The rest api contains still filled values for allowed_actions. @normanjaeckel any idea?

normanjaeckel:
@emanuelschuetze This has something to do with the DS I think. May be the auto-fetch does not detect this extra field for whatever reason. @ostcar Any idea?

ostcar:
Actually I don't know how this works on the agenda list view.

This is not a necessary part of this pull request. So please create an issue for it.

openslides/core/static/js/core/site.js
```diff
@@ -299,6 +293,23 @@ angular.module('OpenSlidesApp.core.site', [
     editableOptions.theme = 'bs3';
 })
 
+// angular formly config options
+.run(function(formlyConfig) {
```
ostcar: Please write:
```
.run([
    'formlyConfig',
    function (formlyConfig) {
...
```
with a list and a white space after function

openslides/motions/static/js/motions/motions.js
```diff
@@ -225,92 +385,483 @@ angular.module('OpenSlidesApp.motions.site', ['OpenSlidesApp.motions'])
...
+    function($scope, Motion, Category, Workflow, Tag, User, motion, $http) {
+        Motion.bindOne(motion.id, $scope, 'motion');
         Category.bindAll({}, $scope, 'categories');
         Workflow.bindAll({}, $scope, 'workflows');
         Tag.bindAll({}, $scope, 'tags');
-        Mediafile.bindAll({}, $scope, 'mediafiles');
+        User.bindAll({}, $scope, 'users');
+        Motion.loadRelations(motion, 'agenda_item');
+        var state = motion.state;
+        state.getNextStates()
```
ostcar: Why do you call this line?

openslides/motions/static/js/motions/motions.js
```diff
@@ -225,92 +385,483 @@ angular.module('OpenSlidesApp.motions.site', ['OpenSlidesApp.motions'])
...
+                .then(function(success) {
+                    $state.go('motions.motion.list');
+                })
+                .catch(function(fallback) {
+                    console.log(fallback);
+                });
+        };
+
+        // define all update form fields
+        $scope.formFields = [
```
ostcar: You already wrote this in the CreateCtr.

Please find a method that is not redundant. In other case it is to hard to fix the form.

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -36,24 +35,188 @@
...
+    <!-- log -->
+    <button type="button" class="btn btn-default spacer" ng-click="isCollapsed = !isCollapsed" translate>
+      Show log
+    </button>
+    <div uib-collapse="isCollapsed">
+      <div class="well well-sm">
+        <ul class="list-unstyled">
+          <li ng-repeat="message in motion.log_messages">
+            <small> {{ message.message }}
+            </small>
```
ostcar: Proposal: Please put this line at the end of line 51

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -36,24 +35,188 @@
...
       <div ng-repeat="submitter in motion.submitters">
         {{ submitter.get_full_name() }}<br>
       </div>
 
+      <!-- supporters -->
+      <div ng-if="config('motions_min_supporters') > 0">
+        <h3 translate>Supporters</h3>
+        <ol>
+          <li ng-repeat="supporters in motion.supporters">
+          {{ supporters.get_full_name() }}
```
ostcar: Please add to whitespaces to show that this line is inside the <li>

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -36,24 +35,188 @@
...
+      <span class="label" ng-class="motion.state.css_class ? 'label-'+motion.state.css_class : 'label-primary'">
+        {{ motion.state.name | translate }}
+      </span>
+      <div ng-if="motion.allowed_actions.change_state" class="spacer">
+        <div class="btn-group-vertical spacer" role="group">
+          <button ng-repeat="state in motion.state.getNextStates()" ng-click="update_state(state)"
+              class="btn btn-default btn-sm">
+            {{state.action_word}}
+          </button>
+          <button ng-if="motion.allowed_actions.reset_state" ng-click="reset_state()" class="btn btn-danger btn-xs">
```
ostcar: This line is to long

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -36,24 +35,188 @@
...
+          <translate>Vote</translate>
+          <button os-perms-lite="motions.can_manage" ng-click="poll.isEditMode=true;"
+              class="btn btn-default btn-xs">
+            <i class="fa fa-pencil"></i>
+          </button>
+          <button os-perms="motions.can_manage" ng-click="delete_poll(poll)"
+              class="btn btn-default btn-xs">
+            <i class="fa fa-times"></i>
+          </button>
+          <br>
```
ostcar: Is this line-break really necessary? Can it be removed using <div>-blocks?

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -36,24 +35,188 @@
...
       {{ motion.category.name }}</a>
 
-      <h3 translate>Voting result</h3>
-      -
+      <h3 translate>Tags</h3>
+      <span ng-repeat="tag in motion.tags">
+        <span class="label label-default">
+          {{ tag.name }}
+        </span>
+        &nbsp;
```
ostcar: Is this line really necessary? Can it be fixed by a css-rule for the span above? (margin-right:5px)

openslides/motions/static/templates/motions/motion-list.html
```diff
@@ -21,78 +21,184 @@ <h1 translate>Motions</h1>
...
+          os-perms="motions.can_manage" ng-click="delete()"
+          class="btn btn-primary btn-sm form-control">
+        <i class="fa fa-trash fa-lg"></i>
+        <translate>Delete selected motions</translate>
+      </a>
+      <!-- state filter -->
+      &nbsp;
+      <select ng-model="stateFilter" class="form-control" id="stateFilter">
+        <option value="" translate>--- Select state ---</option>
+        <!-- TODO: use default workflow id from config('motions_workflow') -->
```
ostcar: Can this be fixed? If not, then please create an issue for it

openslides/motions/views.py
```diff
@@ -66,6 +66,16 @@ def check_view_permissions(self):
...
 
+    def retrieve(self, request, *args, **kwargs):
+        """
+        Customized view endpoint to retrieve a motion.
+
+        Adds the allowed actions for the motion.
+        """
+        response = super().retrieve(request, *args, **kwargs)
+        response.data['allowed_actions'] = self.get_object().get_allowed_actions(request.user)
+        return response
```
ostcar: I don't like this method. When I am looking for all fields for the motion, then I look in serlializers.py. Therefore this functionality is hard to debug.

If we need this method in python, then it should be a read only field on the motion serializer. But I would prefer that we move this function to js.

normanjaeckel: @emanuelschuetze asked me about a quick and dirty hack to get the allowed actions from the server. I said that we might want to hold the logic on client side but we do not have too much time at the moment. Read-only field in the serializer is not easy because we do not have the request user there. You can try ...

openslides/motions/models.py
```diff
@@ -729,14 +730,14 @@ class State(RESTModelMixin, models.Model):
...
+    workflow = models.ForeignKey('Workflow', related_name='states')
     """A many-to-one relation to a workflow."""
 
     next_states = models.ManyToManyField('self', symmetrical=False)
     """A many-to-many relation to all states, that can be choosen from this state."""
 
-    icon = models.CharField(max_length=255)
-    """A string representing the url to the icon-image."""
+    css_class = models.CharField(max_length=255)
+    """A css class string. """
```
normanjaeckel: What is the use case of his css class? The icon? Please add a short note in the docstring.

openslides/motions/views.py
ostcar: I can live with this method, if someone creates an issue, that this function has to be recreated in javascript and deleted from this place

openslides/motions/models.py
```diff
@@ -729,14 +730,14 @@ class State(RESTModelMixin, models.Model):
...
-    workflow = models.ForeignKey('Workflow')
+    workflow = models.ForeignKey('Workflow', related_name='states')
     """A many-to-one relation to a workflow."""
 
     next_states = models.ManyToManyField('self', symmetrical=False)
     """A many-to-many relation to all states, that can be choosen from this state."""
 
-    icon = models.CharField(max_length=255)
-    """A string representing the url to the icon-image."""
+    css_class = models.CharField(max_length=255)
```
normanjaeckel: I propose to add the option `default='default'` so you can remove some of the definitions in the signals.py

openslides/motions/signals.py
```diff
@@ -189,13 +189,16 @@ def create_builtin_workflows(sender, **kwargs):
...
     state_1_3 = State.objects.create(name=ugettext_noop('rejected'),
                                      workflow=workflow_1,
-                                     action_word=ugettext_noop('Reject'))
+                                     action_word=ugettext_noop('Reject'),
+                                     css_class='danger')
     state_1_4 = State.objects.create(name=ugettext_noop('not decided'),
                                      workflow=workflow_1,
-                                     action_word=ugettext_noop('Do not decide'))
+                                     action_word=ugettext_noop('Do not decide'),
+                                     css_class='default')
```
normanjaeckel: Remove the default setting if you use the model option default, see above.

This is here and many times in this file, see below.

openslides/motions/static/js/motions/motions.js
normanjaeckel: Maybe use a service like
```javascript
.factory('MotionFormFieldFactory', [
    'User',
    'gettext',
    'motion'
    function (User, gettext, motion) {
        return {formFields: [ ... ] }
    }
])
```

emanuelschuetze:
@normanjaeckel: MotionPoll allows -1 and -2 as special values (like in 1.7). Can you give me via rest api the defined string (e.g. "majority") instead of -1 so that I don't iterate over each value on client (template) side? That would be nice.

emanuelschuetze:
@normanjaeckel an other wish: Can you accept "null" for yes/no/abstain in PUT request and set "-2" by default for these null values (or directly the string "undocumented")?

openslides/motions/models.py
emanuelschuetze: docstring for css_class added (it is used for colored label, not any more for icon which was unused)

openslides/motions/models.py
emanuelschuetze: I added `default='primary'` because the blue labels are the default color. `default` is set for undecided end states, rejected for accept, green for accepted.

openslides/motions/signals.py
emanuelschuetze: Not changed because new default (defined in model and not in template) is now `primary`. see above.

openslides/motions/views.py
emanuelschuetze: Issue #1668 created..

openslides/motions/static/templates/motions/motion-list.html
emanuelschuetze: The fix can wait until we rework the workflow api for 2.1. Then only one workflow with all states is shown in the select box. For 2.0 we can use this filter which contains all states group-by workflows. So done here.

emanuelschuetze:
Issue #1669 created for @normanjaeckel for special motionpoll values.

openslides/motions/static/js/motions/motions.js
emanuelschuetze: Done! MotionFormFieldFactory added for update and create view. Still open: permission check for manager fields (will create new issue for that)

emanuelschuetze:
All done. please review again.

ostcar:
ok

@normanjaeckel

normanjaeckel:
One proposal, then rebase review commits and merge.

emanuelschuetze:
Rebased. Ready for merge.

emanuelschuetze:
Rebased on top of current master.

PR #1665, opened 2015-11-06
Added possibility to reset password. Work on user app in general.

normanjaeckel:


openslides/users/static/js/users/site.js
```diff
@@ -338,168 +353,258 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
...
+    '$scope',
+    '$state',
+    '$http',
+    'user',
+    function($scope, $state, $http, user) {
+        $scope.user = user;  // autoupdate is not activated
+        $scope.save = function (user) {
+            if ($scope.newPassword != $scope.newPassword2) {
+                $scope.newPassword = $scope.newPassword2 = '';
+                $scope.formError = 'Password confirmation does not match.'
```
ostcar: Please add a semicolon at the end of the line

openslides/users/static/js/users/site.js
```diff
@@ -338,168 +353,258 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
...
+                    {'old_password': $scope.oldPassword, 'new_password': $scope.newPassword}
+                ).then(
+                    function(data) {
+                        $state.go('users.user.list');
+                    },
+                    function(data) {
+                        $scope.oldPassword = $scope.newPassword = $scope.newPassword2 = ''
+                        $scope.formError = data;
+                    }
+                );
```
ostcar: You have to solve the case when the request raises ValidationError. (I don't know which status code this is.) This happens when the old password is wrong

openslides/users/views.py
```diff
@@ -232,6 +238,22 @@ def get_context_data(self, **context):
...
+    """
+    Users can set a new password for themselves.
+    """
+    http_method_names = ['post']
+
+    def post(self, request, *args, **kwargs):
+        if request.user.check_password(request.data['old_password']):
+            user = request.user
+            user.set_password(request.data['new_password'])
+            user.save()
```
ostcar: Proposal:

Either move line 249 before line 248 and use user in 248 as well, or remove this line and work with request.user in line 250 and 251.

ostcar:
It would be easier for the review process, when you did the code cleanup in another commit then your real changes. It was really hard to know which functions I had to look on.

openslides/users/static/js/users/site.js
normanjaeckel: Done.

openslides/users/static/js/users/site.js
normanjaeckel: I added a comment. This case is solved in line 450 and 451. ValidationError returns HTTP 400.

openslides/users/views.py
normanjaeckel: Solved.

normanjaeckel:
All done.

openslides/users/static/js/users/site.js
ostcar: Ups, sry. I missed that.

I also missed the missing semicolon in line 450. Can you add it?

ostcar:
One last small command. Then you can rebase and merge it.

PR #1646, opened 2015-09-21
Added first chatbox draft.

normanjaeckel:
@emanuelschuetze You may build your template stuff on top of this.

PR #1679, opened 2015-11-16
Chatbox

emanuelschuetze:
Added chatbox template stuff on top of @normanjaeckel 's pull request (#1646).

@normanjaeckel Please review.

### DIFF
--- a/openslides/core/static/js/core/base.js
+++ b/openslides/core/static/js/core/base.js
@@ -101,8 +101,17 @@ angular.module('OpenSlidesApp.core', [
             // Loads all projector data
             Projector.findAll();
 
-            // Loads all chat messages data
-            ChatMessage.findAll();
+            // Loads all chat messages data and their user_ids
+            // TODO: add permission check if user has required chat permission
+            // error if include 'operator' here:
+            // "Circular dependency found: loadGlobalData <- operator <- loadGlobalData"
+            //if (operator.hasPerms("core.can_use_chat")) {
+                ChatMessage.findAll().then( function(chatmessages) {
+                    angular.forEach(chatmessages, function (chatmessage) {
+                        ChatMessage.loadRelations(chatmessage, 'user');
+                    });
+                });
+            //}
 
             // Loads server time and calculates server offset
             $http.get('/core/servertime/').then(function(data) {
@@ -172,32 +181,41 @@ angular.module('OpenSlidesApp.core', [
     }
 ])
 
-.factory('Tag', ['DS', function(DS) {
-    return DS.defineResource({
-        name: 'core/tag',
-    });
-}])
+.factory('Tag', [
+    'DS',
+    function(DS) {
+        return DS.defineResource({
+            name: 'core/tag',
+        });
+    }
+])
 
-.factory('Config', ['DS', function(DS) {
-    return DS.defineResource({
-        name: 'core/config',
-        idAttribute: 'key',
-    });
-}])
+.factory('Config', [
+    'DS',
+    function(DS) {
+        return DS.defineResource({
+            name: 'core/config',
+            idAttribute: 'key',
+        });
+    }
+])
 
-.factory('ChatMessage', ['DS', function(DS) {
-    return DS.defineResource({
-        name: 'core/chatmessage',
-        relations: {
-            belongsTo: {
-                'users/user': {
-                    localField: 'user',
-                    localKey: 'user_id',
+.factory('ChatMessage', [
+    'DS',
+    function(DS) {
+        return DS.defineResource({
+            name: 'core/chatmessage',
+            relations: {
+                belongsTo: {
+                    'users/user': {
+                        localField: 'user',
+                        localKey: 'user_id',
+                    }
                 }
             }
-        }
-    });
-}])
+        });
+    }
+])
 
 /* Model for a projector.
  *
--- a/openslides/core/static/js/core/base.js
+++ b/openslides/core/static/js/core/base.js
@@ -80,9 +80,10 @@ angular.module('OpenSlidesApp.core', [
 .factory('loadGlobalData', [
     '$rootScope',
     '$http',
+    'ChatMessage',
     'Config',
     'Projector',
-    function ($rootScope, $http, Config, Projector) {
+    function ($rootScope, $http, ChatMessage, Config, Projector) {
         return function () {
             // Puts the config object into each scope.
             Config.findAll().then(function() {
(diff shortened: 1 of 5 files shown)


normanjaeckel:
Some errors:
* If a user is not allowed to chat the button should be disabled.
* The JSData does not work like expected. Sometimes I have to reload the page or move around to get the usernames.

How to get the number of unread messages? Try the following:
* Set a global counter (use a .value() service for this) to 0 when the user opens the base template controller which controlls the Chat button.
* Increase this counter when a message comes in.
* Disabled the counter when the user opens the chat controller.
* Reset and enable the counter when the user closes the chat window.

normanjaeckel:
@ostcar Any idea about this described JSData thing?

emanuelschuetze:
All done:
- Fixed missing permission check.
- Fixed missing user in JS-Data by adding loadRelations for each chatmessage.
- Added unread counter.

Rebased on top of master.

openslides/core/static/js/core/base.js
```diff
@@ -257,7 +279,15 @@ angular.module('OpenSlidesApp.core', [
     removeButtons: 'Anchor,SpecialChar,Subscript,Superscript,Styles,RemoveFormat,HorizontalRule'
 })
 
+
```
normanjaeckel: Remove this line.

openslides/core/static/js/core/site.js
```diff
@@ -752,6 +753,47 @@ angular.module('OpenSlidesApp.core.site', [
...
+                ($.inArray(newVal[newVal.length-1].id, NewChatMessages) == -1)) {
+                console.log(newVal[newVal.length-1].id);
+                NewChatMessages.push(newVal[newVal.length-1].id);
+                $scope.unreadMessages = NewChatMessages.length;
+            }
+        })
+    }
+])
+
+// counter of new (unread) chat messages
```
normanjaeckel: Proposal: Better move this service before the controller.

openslides/core/static/js/core/site.js
```diff
@@ -752,6 +753,47 @@ angular.module('OpenSlidesApp.core.site', [
...
+            .success(function () {
+                $scope.newMessage = '';
+            });
+        };
+        // increment unread messages counter for each new message
+        $scope.$watch('chatmessages', function (newVal, oldVal) {
+            // add new message id if there is really a new message which is not yet tracked
+            if ((oldVal[oldVal.length-1].id != newVal[newVal.length-1].id) &&
+                ($.inArray(newVal[newVal.length-1].id, NewChatMessages) == -1)) {
+                console.log(newVal[newVal.length-1].id);
```
normanjaeckel: Remove this line.

normanjaeckel:
One remark and two proposals. Then ready to rebase and merge.

openslides/core/static/js/core/base.js

openslides/core/static/js/core/site.js
emanuelschuetze: Done

openslides/core/static/js/core/site.js
emanuelschuetze: Done

emanuelschuetze:
All done. Rebased. Ready for merge.

normanjaeckel:
Ok.

openslides/core/static/js/core/base.js
```diff
@@ -100,6 +101,13 @@ angular.module('OpenSlidesApp.core', [
             // Loads all projector data
             Projector.findAll();
 
+            // Loads all chat messages data and their user_ids
+            ChatMessage.findAll().then( function(chatmessages) {
```
ostcar: Please add a permission check here. If the user is not a manager, then he gets a 403 what is not a problem, but it saves one request that is only needed for managers

openslides/core/static/js/core/base.js
```diff
@@ -181,6 +189,20 @@ angular.module('OpenSlidesApp.core', [
     });
 }])
 
+.factory('ChatMessage', ['DS', function(DS) {
```
ostcar: Please use our new style here

openslides/core/static/js/core/site.js
```diff
@@ -752,6 +753,45 @@ angular.module('OpenSlidesApp.core.site', [
...
+            NewChatMessages = [];
+            $scope.unreadMessages = NewChatMessages.length;
+        }
+        $scope.sendMessage = function () {
+            $http.post(
+                '/rest/core/chatmessage/',
+                {message: $scope.newMessage}
+            )
+            .success(function () {
+                $scope.newMessage = '';
```
ostcar: If you have a poor connection to the server, then the message in the text disappears not until the request returned a success message. In this case the user could hit the send button multiple times and the message would be send to the server more then once. We need a solution for this. For example, the server should skip multipe messages with the same content or you should disable the send button until the request returns a success or error message

openslides/core/static/templates/index.html
```diff
@@ -146,16 +179,17 @@ <h3 class="modal-title" translate>Please sign in!</h3>
...
-          <a ui-sref="version">Version</a>
-        </small>
-      </footer>
     </div><!--/#content-->
   </div><!--/.row-->
+
+  <hr>
+  <footer>
+    <small>
+      &copy; Copyright 2011-2015 |
```
ostcar: I never understood why we need the years. I am against it, because it has no use and we have to change it every year

openslides/core/static/js/core/site.js
emanuelschuetze: Done.

openslides/core/static/js/core/base.js

openslides/core/static/js/core/base.js
emanuelschuetze: I added a permission check:
 `if (operator.hasPerms("core.can_use_chat")) ...`
But I get this error if I add `operator` to `loadGlobalData`:
"Circular dependency found: loadGlobalData <- operator <- loadGlobalData"
@ostcar How can we fix this?

emanuelschuetze:
All done. One open question.

emanuelschuetze:
I added my open question again (copied from outdated diff):
I added a permission check:
`if (operator.hasPerms("core.can_use_chat")) ...`
But I get this error if I add operator to loadGlobalData:
"Circular dependency found: loadGlobalData <- operator <- loadGlobalData"
@ostcar How can we fix this?

ostcar:
This is hard to fix, the problem is, that the operator is in the users app, but is requested from the core app. We have to solve this by moving the operator to the core app and make it independent from the user object. I have no idea how we can do this easily. We have to check the permission after the beta1.

So please rebase

emanuelschuetze:
Rebased.

PR #1690, opened 2015-11-19
Reset default password of an user in UserUpdateCtrl. (Fixes #412)

### LINKED ISSUE #412: Improve password reset process
It is confusing for new users. Insert a new button which does the two steps in one.

emanuelschuetze:


openslides/users/static/templates/users/user-form.html
```diff
@@ -44,7 +44,15 @@ <h1 ng-if="!user.id" translate>New participant</h1>
   <div class="form-group row">
     <div class="col-xs-6">
       <label for="inputDefaultPassword" translate>Default password</label>
-      <input type="text" ng-model="user.default_password" class="form-control" name="inputDefaultPassword">
+      <div class="input-group">
```
normanjaeckel: Is it self-explaining the you first have to submit the update form and then come again to this place and click reset?

emanuelschuetze: Yes! It's like in 1.x. We can improve it later if somebody ask for improvement.

Issue #1692, opened 2015-11-20
Agenda content object requirements


emanuelschuetze:
Some change requests for related agenda item information provided by rest api (if possible):
* I need the content object real name in singular (e.g. "motion", "election", "agenda item") to show it in delete confirm message ("Do you really want to delete the Motion/Election/agenda item?"). I want to provide a delete link in agenda for the related items.
* `item.content_object.collection` should get a string with "." instead of "/" separator; e.g. "motions.motion" instead of "motions/motion"
* string of (JS-Data) Object name, e.g. "Agenda", "Motion", "Assignment". I want to use it for auto delete command `$ObjectName.destroy(item.content_object.id)`

normanjaeckel:
I added a new field element `display_name`. See #1695.

Why do you want a string with `.` instead of `/`? We defined the collection name used for autoupdate stuff with `/`.

The server does not know the JSData string (i. e. Angular factory service name). The mapping between collection names and these string should be done in the client. -> @ostcar

emanuelschuetze:
> Why do you want a string with . instead of /? We defined the collection name used for autoupdate 
> stuff with /.

I need it for ui-sref links, e.g. `motions.motion.detail`:
https://github.com/OpenSlides/OpenSlides/pull/1698/files#diff-3f7cf361bcdcee23fe42ded55a85d33bR119
But I can use the replace command if other code needs the `/` separator.

emanuelschuetze:
We can close this. I don't need any more special content object updates...

PR #1695, opened 2015-11-20
Added display_name to agenda item content object serializer field. Se…

normanjaeckel:
…e #1692.


emanuelschuetze:
Ok. Please fix travis and merge asap so that I can use it in my pull request. Thanks.

ostcar:
I don't like this solution. The RelatedItemRelatedField should only contain the information needed to get another object. Each information containing this object can be get from the object. For example use the attribute agendaSupplement from the DS-Model. (We have to remove the bracelets.) Or add another string to the DS-Model or send it with the related object.

This is important, because we could use this field in other cases, too. This field is useful for each generic relation not only agenda items.

normanjaeckel:
@emanuelschuetze Please let us know whether to close this pull request.

emanuelschuetze:
Yes we can close this pull request and should discuss it later.

PR #1698, opened 2015-11-22
Improvements for related agenda items

emanuelschuetze:
- Added QuickEdit mode for related agenda items
- support hidden items
- Added ng-dialog for modal create/update dialogs of customslides
- use generic links for list of speakers, edit, delete

Some template style issues are still open. I'll fix it with my next pull request which brings up a new base template. So please merge this asap. Thanks!

### DIFF
--- a/openslides/agenda/static/js/agenda/site.js
+++ b/openslides/agenda/static/js/agenda/site.js
@@ -92,26 +92,82 @@ angular.module('OpenSlidesApp.agenda.site', ['OpenSlidesApp.agenda'])
     '$scope',
     '$http',
     '$state',
+    'ngDialog',
     'Agenda',
     'AgendaTree',
+    'Customslide',
     'Projector',
-    function($scope, $http, $state, Agenda, AgendaTree, Projector) {
+    function($scope, $http, $state, ngDialog, Agenda, AgendaTree, Customslide, Projector) {
         // Bind agenda tree to the scope
         $scope.$watch(function () {
             return Agenda.lastModified();
-
         }, function () {
             $scope.items = AgendaTree.getFlatTree(Agenda.getAll());
         });
+        $scope.alert = {};
 
-        // open detail view link
-        $scope.openDetail = function (id) {
-            $state.go('agenda.item.detail', {id: id});
+        // project related item (content object)
+        $scope.project = function (item) {
+            item.getContentResource().find(item.content_object.id).then(
+                function(object) {
+                    object.project();
+                }
+            );
         };
 
-        // save changed item
-        $scope.save = function (item) {
-            Agenda.save(item);
+        // open new customslide dialog
+        $scope.newDialog = function () {
+            ngDialog.open({
+                template: 'static/templates/core/customslide-form.html',
+                controller: 'CustomslideCreateCtrl',
+                className: 'ngdialog-theme-default wide-form'
+            });
+        };
+        // detail view of related item (content object)
+        $scope.open = function (item) {
+            $state.go(item.content_object.collection.replace('/','.')+'.detail',
+                    {id: item.content_object.id});
+        };
+        // edit view of related item (content object)
+        $scope.edit = function (item) {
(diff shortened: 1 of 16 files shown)


normanjaeckel:
Looks ok.

openslides/agenda/static/js/agenda/site.js
```diff
@@ -134,8 +191,19 @@ angular.module('OpenSlidesApp.agenda.site', ['OpenSlidesApp.agenda'])
...
+                            Customslide.destroy(item.content_object.id);
+                            break;
+                        case 'motions/motion':
+                            Motion.destroy(item.content_object.id);
+                            break;
+                        case 'assignments/assignment':
+                            Assignment.destroy(item.content_object.id);
+                            break;
+                    }
+                }
```
ostcar: I don't like this code, because the agenda app should be independent from Motion and Assignment. But it is not needed anyway, because the Customslide-, Motion- or Assignmentobject is deleted from the server when the related item is deleted.

openslides/users/static/js/users/site.js
```diff
@@ -620,6 +621,59 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
...
+    'Config',
+    function ($scope, $http, operator, gettext, Config) {
+        $scope.alerts = [];
+
+        // TODO: add welcome message only on first time (or if admin password not changed)
+        $scope.alerts.push({
+            type: 'success',
+            msg: gettext("Installation was successfully.<br>" +
+                "Use <strong>admin</strong> and <strong>admin</strong> for first login." +
+                "<p>Important: Please change your password! ")
```
ostcar: Proposal: Try to remove the html-tags from the string, so it is not part of the translation.

openslides/agenda/static/js/agenda/site.js
emanuelschuetze: I changed it back to `Agenda.destroy(item.id)`.
But now, this code does not allowed me to delete an agenda item (Customslide, Motion, Assignment)! I get a `403 Forbidden` for the DELETE request (e.g. `DELETE http://localhost:8000/rest/agenda/item/10/`).

emanuelschuetze:
Rebased on top of master. Added review commit. Still a open question about not working DELETE request for agenda items.

normanjaeckel:
It was a design decision not to allow DELETE requests for agenda items because they should be fully related to the obejct behind it (motion, custom slide etc). Deleting an agenda item would break the datastructure at this point and open some more questions to handle.

emanuelschuetze:
@ostcar So, how can we solve quickly the DELETE topic today? Should I add my old code (with Motions,Assignments etc.) again?

emanuelschuetze:
I just added two template fixes.

ostcar:
@emanuelschuetze I have no problem with your code for the customslide. You can add this. But the agenda list view should not have buttons to delete a motion or assignment. My suggestion is, that you check in the template if the item is a custom slide item and only then show the delete button.

emanuelschuetze:
Rebased on top of master.
Added 'Review2' commit which deletes only customslides (not motions/assignments) in ItemListCtrl.
One last thing: How can I project related items (not only the agenda items)? ìtem.project()`is showing the title of agenda item only. I search something like `Customslide.find(item.content_object.id).project();`which does not work...
@oskar?

emanuelschuetze:
Project function now works with 'review 3' commit.
@ostcar Thanks!
So please do a last review. After that I will rebase.

emanuelschuetze:
If you have a quick solution for this pull request:
`<div ng-bind-html="customslide.title"></div>`shows nothing on customslide slide template. I can't see why. So I comment it out in review3 for this pr and I'll fix it later.

ostcar:
ok. Please rebase and merge

emanuelschuetze:
All done. Rebased and fixed also the ng-bind-html by moving ngSanitize to core/base.js.
So ready for merge!

PR #1701, opened 2015-11-23
Added REST API viewpoint to number the agenda.

normanjaeckel:


emanuelschuetze:
Ok. I just added a pull request for you to merge it directly in your PR: 
https://github.com/normanjaeckel/OpenSlides/pull/10

normanjaeckel:
Ready to review and merge.

openslides/agenda/models.py
```diff
@@ -281,15 +281,16 @@ def calc_item_no(self):
...
         """
-        return Item.objects.filter(
-            parent=self.parent,
-            type=self.AGENDA_ITEM,
-            weight__lte=self.weight).count()
+        result = 0
+        for item in Item.objects.filter(parent=self.parent, weight__lte=self.weight):
+            if not item.is_hidden():
+                result += 1
+        return result
```
ostcar: What was the problem with the old implementation? It is a faster db-query

normanjaeckel: The old query is wrong.

ostcar: Then you should fix the count query instead of fetching all data from the db and counting in python.

You problably have to add .exclude(type=self.HIDDEN_ITEM)

normanjaeckel: Sorry. Please tell me how to fix this. I think it is not possible because we defined hidden items in our method `is_hidden()`. See there for details.

ostcar:
rebase and merge please

emanuelschuetze:
@normanjaeckel You can rebase my commit into your (single) commit if you like.

PR #1702, opened 2015-11-23
Split motions.js into base, site and projector

emanuelschuetze:
just split up the current motions.js into 3 parts


normanjaeckel:
Ok.

Issue #1707, opened 2015-11-26
Autoupdate does not work after delete a candidate from list (assignment detail view)


emanuelschuetze:
autoupdate does not work after delete (other) candidate from list (assignment detail view), reload helps

emanuelschuetze:
Also auto update of deleted assignmentpolls does not work.

emanuelschuetze:
see #1711

emanuelschuetze:
Fixed.

PR #1709, opened 2015-11-26
Assignments template improvements

emanuelschuetze:
- list view with quickview
- ngDialog for create/update view
- candidates
- polls

Ready for first review.

### DIFF
--- a/openslides/assignments/static/templates/assignments/assignment-detail.html
+++ b/openslides/assignments/static/templates/assignments/assignment-detail.html
@@ -24,18 +24,98 @@
   </a>
 </div>
 
-Agenda Item: {{ assignment.agenda_item }}
-
 
 <h3 translate>Description</h3>
 <div class="white-space-pre-line">{{ assignment.description }}</div>
 
 <h3 translate>Candidates</h3>
-<ul>
+<ol>
   <li ng-repeat="related_user in assignment.assignment_related_users" ng-if="related_user.status == 1">
-    User: {{ related_user.user.get_full_name() }}
-  </li>
-</ul>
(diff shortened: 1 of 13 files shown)


openslides/assignments/static/templates/assignments/assignment-detail.html
```diff
@@ -24,18 +24,98 @@
...
+        <translate>Unpublish result</translate>
+      </button>
+      <a ng-click="deleteBallot(poll)" class="btn btn-default btn-sm">
+        <i class="fa fa-times"></i>
+        <translate>Delete</translate>
+      </a>
+    </div>
+    <div class="results">
+      <div ng-repeat="option in poll.options">
+        <strong>User#{{ option.candidate_id }}</strong>
```
emanuelschuetze: @ostcar How is the best way to get the full user name of candidate_id?

emanuelschuetze:
@ostcar Please look also at #1707. Maybe you can give me a quick fix for it which I can implement into this pull request.

openslides/assignments/static/templates/assignments/assignment-detail.html
ostcar: I think we have to create DS-Store classes for option objects. See the assignment_related_user as an example.

ostcar:
I would like that we merge this right now so no "super-pr" like the one from motions arises.

@normanjaeckel please check an merge

ostcar:
@normanjaeckel emanuel and I promise to fix your remarks in another pull request, as we would in this one. But we need this merged to work on it this evening.

normanjaeckel:
Ok.No remarks at the moment.

PR #1712, opened 2015-11-27
Angular: downgrade ui-tree to fix the agenda sort view

ostcar:


emanuelschuetze:
Successfully tested.

normanjaeckel:
Please add somewhere a short note which version works which not and want to do if a new one is released. Then rebase and merge.

ostcar:
That is done in the bower file. If someone wants to update to a newer version, this person has to test if the newer version works.

normanjaeckel:
Hm. There is not mentioned that version 2.10.x does not work. (Like `"~2.x,!2.10") I can merge this nevertheless. We can keep this in mind in issue #1717.

PR #1741, opened 2015-12-07
Updated English translation strings.

emanuelschuetze:


ostcar:
ok for me, after travis is fixed

emanuelschuetze:
Fixed and rebased.

openslides/assignments/static/templates/assignments/assignment-detail.html
```diff
@@ -114,14 +114,18 @@ <h3 translate>Election result</h3>
...
       <div class="results">
         <div ng-repeat="option in poll.options">
-          <strong>{{ option.candidate.get_full_name() }}</strong>
-          <div ng-if="option.votes.length > 0">
-            <div ng-repeat="vote in option.votes">
-              {{ vote.value}}: {{ vote.weight}}
-            </div>
+          <div ng-if="poll.yesnoabstain && option.votes.length > 0">
+            <strong>{{ option.candidate.get_full_name() }}</strong><br>
+            <translate>Yes</translate>: {{ option.votes[0].weight }}<br>
```
normanjaeckel: Ah this is tricky. We assume an array like this:
```json
[
  {
    "value": "Yes",
    "weight": "42"
  },
  {
    "value": "No",
    "weight": "-1"
  },
  {
    "value": "Abstain",
    "weight": "2"
  }
]
```
But it may also be:
```json
[
  {
    "value": "No",
    "weight": "-1"
  },
  {
    "value": "Yes",
    "weight": "42"
  },
  {
    "value": "Abstain",
    "weight": "2"
  }
]
```
Then the template output would be sliently wrong. At the moment the server always sends "Yes" as first element but this is not "protected" very well.

Maybe add a comment here?

emanuelschuetze:
I go back to a more secure solution: `{{ option.votes[0].value | translate }}` but this requires that the string "Yes" gets a `gettext(..)` in Javascript... What do you think?

normanjaeckel:
You can do so: Just add something like
```javascript
// Just mark some vote value strings for translation.
gettext('Yes'), gettext('No'), gettext('Abstain');
```
to the detail controller.

openslides/assignments/static/templates/assignments/assignment-detail.html
```diff
@@ -114,14 +114,18 @@ <h3 translate>Election result</h3>
...
         <div ng-repeat="option in poll.options">
-          <strong>{{ option.candidate.get_full_name() }}</strong>
-          <div ng-if="option.votes.length > 0">
-            <div ng-repeat="vote in option.votes">
-              {{ vote.value}}: {{ vote.weight}}
-            </div>
+          <div ng-if="poll.yesnoabstain && option.votes.length > 0">
+            <strong>{{ option.candidate.get_full_name() }}</strong><br>
+            {{ option.votes[0].value | translate }}: {{ option.votes[0].weight }}<br>
+            {{ option.votes[1].value | translate }}: {{ option.votes[0].weight }}<br>
```
normanjaeckel: Typo: `option.votes[1].weight`

emanuelschuetze:
Fixed and rebased. Ready for merge.

PR #1744, opened 2015-12-07
Readme

emanuelschuetze:


CHANGELOG
```diff
@@ -53,7 +56,8 @@ Other:
 - Used setup.cfg for development tools.
 - Removed code for Windows portable version with GUI. Used new repository for
   this.
-
+Translations:
+- Added DE and FR translations.
```
normanjaeckel: Please add a newline here so we have two lines before the next chapter, see below.

PR #1753, opened 2015-12-09
Project related object (Fixes #1704)

### LINKED ISSUE #1704: project related object
If the clients sends a request to the server to project an agenda item, the server should not save the item in the database but the related item.

emanuelschuetze:
- Override generic project and isProjected function in agenda.
- Fix project and isProjected function for list of speakers and move
  to jsDataModel of Agenda.

### DIFF
--- a/openslides/agenda/static/js/agenda/base.js
+++ b/openslides/agenda/static/js/agenda/base.js
@@ -22,12 +22,13 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
 ])
 
 .factory('Agenda', [
+    '$http',
     'DS',
     'Speaker',
     'jsDataModel',
     'Projector',
     'gettextCatalog',
-    function(DS, Speaker, jsDataModel, Projector, gettextCatalog) {
+    function($http, DS, Speaker, jsDataModel, Projector, gettextCatalog) {
         var name = 'agenda/item';
         return DS.defineResource({
             name: name,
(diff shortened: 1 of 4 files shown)


emanuelschuetze:
@ostcar I does not move projectAgenda() and isAgendaProjected() to jsDataModel, right?

ostcar:
Yes. The methods of the datastore are only for its instances.

Looks good.

openslides/agenda/static/js/agenda/base.js
```diff
@@ -66,6 +75,27 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
                                element.id == self.content_object.id;
                     };
                     return typeof _.findKey(projector.elements, predicate) === 'string';
+                },
+                // project list of speakers
+                projectListOfSpeakers: function() {
+                    return $http.post(
+                        '/rest/core/projector/1/prune_elements/',
+                        [{name: 'agenda/item', id: this.id, list_of_speakers: true}]
```
normanjaeckel: Here I have the question if we want to project the list of speakers of an agenda item or directly of the related object. Example: `{"name": "motions/motion", "id": "5", "list_of_speakers": true}`

emanuelschuetze: I don't know what's the best way. @ostcar ?

ostcar: Never mind what we should write in the db.

But the template code should be in the agenda app. In other case we would have to add templates in any app that can be a related item.

@normanjaeckel you should choose what is the easiest way on the server side.

normanjaeckel: Ok then @emanuelschuetze please do not use `name: 'agenda/item'` but `{name: this.content_object.collection, id: this.content_object.id, "list_of_speakers": true}`

emanuelschuetze: I tested this: This does not work because the list of speakers is projected as agenda/item and needs `name: agenda/item`. I tested e.g.  a motion: this project only the motion but NOT the list of speakers slide of the motion. Therefor you have to send post `{name: "agenda/item", id: 5, list_of_speakers: true}
The current code works good. So why not merge this?

normanjaeckel: Ok, we can go this way if you like, but some changes should be also done. I do them in an extra pull request.

normanjaeckel: See #1767.

PR #1756, opened 2015-12-09
Redirect to login dialog if user is not logged in (Fixes #1750)

### LINKED ISSUE #1750: Proposal: New login page before loading Angular site
What do you think about creating a general login page which will be shown before the user can enter OpenSlides at all. We can use a standard Django view without AngularJS and this stuff. The advantage would be that we could clearly manage the first data fetching after login (just the first loading of AngularJS) and we can avoid a lot of useless HTTP requests. At the moment the first start generates 12 requests that get HTTP 403.

# Case 1: Guest disabled
- If anonymous user calls any url he gets the login view. We can solve this with a middleware or a smart filter in IndexView and optionally ProjectorView class.
- If a registered user calls an url he gets AngularJS just as in 2.0b1.
- In the top navigation bar there is no Login dialog but only a logout button. If the user clicks this, he posts the logout request and we force a full reload of the page.

# Case 2: Guest enabled
- Any user who calls an url gets AngularJS just as in 2.0b1.
- In the top navigation bar there is also a login button and dialog (but without the button "continue as guest", this is not neccesary any more). The client may force reload after login.

emanuelschuetze:


ostcar:
ok

Issue #1762, opened 2015-12-09
Login button does not work


normanjaeckel:
See #1756.

Click Continue as guest to close the model. The try to click "Login" ... nothing happens.

PR #1763, opened 2015-12-09
Added guest enabled info to WhoAmI view.

normanjaeckel:
@emanuelschuetze Please check this out and try to control the "Continue as guest" button depending on the flag I submit to you in the who am I view with this pull request. I do not know how to do this. Maybe combine with #1762.

### DIFF
--- a/tests/integration/users/test_views.py
+++ b/tests/integration/users/test_views.py
@@ -12,7 +12,9 @@ class TestWhoAmIView(TestCase):
         response = self.client.get(self.url)
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.content, b'{"user_id":null}')
+        self.assertEqual(
+            json.loads(response.content.decode('utf-8')),
+            {'user_id': None, 'guest_enabled': False})
 
     def test_get_authenticated_user(self):
         self.client.login(username='admin', password='admin')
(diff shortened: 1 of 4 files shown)


tests/integration/users/test_views.py
```diff
@@ -12,7 +12,9 @@ def test_get_anonymous(self):
         response = self.client.get(self.url)
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.content, b'{"user_id":null}')
+        self.assertEqual(
+            json.loads(response.content.decode('utf8')),
```
normanjaeckel: Here we get the "Quittung". In Python 3.3 and 3.4 the result is `b'{"user_id":null,"guest_enabled":false}'` but in Python 3.5 it is `b'{"guest_enabled":false,"user_id":null}'` for whatever reason, so this test always failed. We should not use bytestring tests any more if we want to check JSON content.

tests/integration/users/test_views.py
```diff
@@ -12,7 +12,9 @@ def test_get_anonymous(self):
         response = self.client.get(self.url)
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.content, b'{"user_id":null}')
+        self.assertEqual(
+            json.loads(response.content.decode('utf-8')),
```
normanjaeckel: Here we get the "Quittung". In Python 3.3 and 3.4 the result is `b'{"user_id":null,"guest_enabled":false}'` but in Python 3.5 it is `b'{"guest_enabled":false,"user_id":null}'` for whatever reason, so this test always failed. We should not use bytestring tests any more if we want to check JSON content.

ostcar: ok. But you can remove the string 'utf-8'. It is the default value of the argument: https://docs.python.org/3.5/library/stdtypes.html#bytes.decode

emanuelschuetze:
see #1765.

normanjaeckel:
Closed with respect of #1765.

Issue #1643, opened 2015-09-18
Delete Django default permissions


normanjaeckel:

ostcar:
The old code was:

```python
def delete_default_permissions(**kwargs):
    """
    Deletes the permissions, django creates by default for the admin.
    """
    # TODO: Find a way not to create the permissions in the first place.
    #       Meta.default_permissions does not work, because django will
    #       nevertheless create permissions for its own models like "group"
    for p in Permission.objects.all():
        if (p.codename.startswith('add') or
                p.codename.startswith('delete') or
                p.codename.startswith('change')):
            p.delete()
```

emanuelschuetze:
Norman can you do it for 2.0b1? Otherwise let milestone on 2.0b2.

emanuelschuetze:
This is very important for 2.0b2.

normanjaeckel:
#1746 solves some cases but some are still open. See there.

emanuelschuetze:
So still open here (see #1764): Delete the permissions of Django apps auth, contentypes and session.
We should fix this for 2.0b2. @normanjaeckel can you make a pull request for this?

PR #1764, opened 2015-12-09
Deleted Django's default permissions.

normanjaeckel:
See #1643. New database build required, not only migrate an old one.

Still open: Permissions of Django apps auth, contentypes and session. @ostcar Any idea how to get rid of them?


ostcar:
You could create a signal, that delets the permissions :)

Or we create a filter in the permission list view that does not return the django permission to the rest api.

Why is it needed to create a new database? It should be possible to create a migration file, that delets the old permission from the database.

normanjaeckel:
Ok. Please merge this. The filter is something extra. Creating an extra migrations file only for the development (we will renew any migration file before final release) can be done by anyone who wants to do this.

Issue #1727, opened 2015-11-30
Template stuff and other bugs


normanjaeckel:
To do:
- [x] Z-index of messages and countdown are not always consistent
- [x] Autoupdate of config form fields does not work if you change the field and click reset. It works, if you first reload and then click reset.
- [x] Remove old Logo from login modal.
- [x] Change password error messages are bad.
- [x] New election form: we need a default for numbers to be elected field.
- [x] typo: users/static/templates/users/user-detail-profile.html:3
- [x] Use "election" not "assignment" in assignments/static/templates/assignments/assignment-list.html:108
- [x] Change string in agenda/static/templates/agenda/item-import.html:43 and users/static/templates/users/user-import.html:47. Don't say it twice (header == first line).
- [x] Enhance translation of `h` in agenda/static/templates/agenda/item-list.html:158
- [x] Missing favicon at projector base template.
- [x] Agenda item numbers disappear and appear twice on agenda list view depending on active slide.
- [x] Zoom does not work on agenda main slide

emanuelschuetze:
all done here, some open tasks moved to #1840 and #1861

PR #1775, opened 2015-12-11
Fixed some template bugs (#1727)

emanuelschuetze:
- Fixed z-index of messages and countdowns.
- Sidebar: Live view open by default. Improve section style.
- Set default value for open_posts in assignments create form.
- Redefine default values in motions create form.


ostcar:
ok

PR #1778, opened 2015-12-19
Fix publish assignment poll function. (Fixes#1718)

emanuelschuetze:
Set 'votes' to required=False in AssignmentAllPollSerializer.


normanjaeckel:
Ok.

PR #1782, opened 2016-01-03
Added a search on the server side.

ostcar:
Please review and test this PR. There could be erros in it but also some open design decisions.

Use localhost:8000/search/?q=foo to search for foo 

Stil open:
- [ ] manage command to recreate the index with existing instances
- [x] support umlauts in the search query string
- [x] isort and flake8

### DIFF
--- a/openslides/core/views.py
+++ b/openslides/core/views.py
@@ -2,6 +2,7 @@ import re
 import uuid
 from collections import OrderedDict
 from operator import attrgetter
+from urllib.parse import unquote
 
 from django.apps import apps
 from django.conf import settings
@@ -27,6 +28,7 @@ from openslides.utils.rest_api import (
     ViewSet,
     detail_route,
 )
+from openslides.utils.search import search
 
 from .config import config
 from .exceptions import ConfigError, ConfigNotFound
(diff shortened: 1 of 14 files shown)


ostcar:
Hit to myseld. The speakers have to be indexed to each related item

emanuelschuetze:
Tested:
I created a motion and added a submitter "Max Mustermann". The search finds the motion if I search for "max" or "mustermann" but not for "muster". 
=> Search should also work for substrings.

openslides/core/views.py
```diff
@@ -561,3 +563,17 @@ def get_context_data(self, **context):
                 'description': get_plugin_description(plugin),
                 'version': get_plugin_version(plugin)})
         return result
+
+
+class SearchView(utils_views.APIView):
+    """
+    Accepts a search string and returns a list of objects where each object
+    is a dictonary with the keywords collection and id.
+    """
```
normanjaeckel: Proposal: Add a link to the search string syntax.

openslides/utils/search.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def create_index(self):
+        """
+        Creats the whoosh index. Delets an existing index if exists.
+
+        Returns the index.
+        """
+        path = self.get_index_path()
+        if path == 'ram':
+            self.storage = RamStorage().create_index(get_schema())
```
normanjaeckel: Proposal: Add a comment that this case is only for testing.

openslides/utils/models.py
```diff
@@ -39,3 +39,11 @@ def get_root_rest_url(self):
         root_instance = self.get_root_rest_element()
         rest_url = '%s-detail' % type(root_instance)._meta.object_name.lower()
         return reverse(rest_url, args=[str(root_instance.pk)])
+
+    def get_collection_string(self):
+        """
+        Returns the string representing the name of the collection.
+        """
+        # TODO: find a way not to use the url
```
normanjaeckel: Please add an issue and remove this TODO.

openslides/core/views.py

openslides/utils/models.py
ostcar: #1791

openslides/utils/search.py
ostcar: See #1792

PR #1802, opened 2016-01-09
Removed possibility to block candidates. Fixes #1708.

### LINKED ISSUE #1708: Remove "blocked candidates" feature
I think we can drop this old (ESG) feature which is not really important for all(?) OpenSlides users...

Discussed with @normanjaeckel - suggestions:
- rename "assignment_related_users" to "candidates"
- each candidate has a attribute "elected" (true|false)
- remove old "status" (1|2|3) attribute

We want to change it before 2.0.

@ostcar Is this suggestion ok for you?

normanjaeckel:
Please review carefully.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@ Agenda:
 - Removed mptt.
 Assignments:
 - Renamed app from assignment to assignments.
+- Removed possibility to block candidates.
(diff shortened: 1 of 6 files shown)


ostcar:
looks good. Except of travis...

PR #1811, opened 2016-01-09
Squished all migrations.

normanjaeckel:


normanjaeckel:
Please have a special look on core migration.

PR #1824, opened 2016-01-09
Fix tornado

ostcar:


openslides/utils/autoupdate.py
```diff
@@ -21,6 +21,9 @@
 
 from .rest_api import get_collection_and_id_from_url
 
+RUNNING_PORT = None
+RUNNING_HOST = None
```
normanjaeckel: Please switch these lines.

openslides/utils/autoupdate.py
```diff
@@ -127,6 +131,11 @@ def run_tornado(addr, port, *args, **kwargs):
 
     It runs in one thread.
     """
+    # Save the port and the addr in a global var
+    global RUNNING_PORT, RUNNING_HOST
```
normanjaeckel: Here the same.

PR #1862, opened 2016-01-14
Modify user in form dialog without updating list/detail view.

emanuelschuetze:
Create deep copy of user object before modify user in form dialog.
So list/detail view is not updated while editing.
Before saving inject the changed user (copy) object back into DS
store. Refresh user from DB is save request fails.

### DIFF
--- a/openslides/users/static/js/users/site.js
+++ b/openslides/users/static/js/users/site.js
@@ -519,8 +519,11 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
     'user',
     function($scope, $state, $http, User, UserForm, Group, user) {
         Group.bindAll({where: {id: {'>': 2}}}, $scope, 'groups');
-        // set initial values for form model
-        $scope.model = user;
+        $scope.alert = {};
+        // set initial values for form model by create deep copy of user object
+        // so list/detail view is not updated while editing
+        $scope.model = angular.copy(user);
+
         // get all form fields
         $scope.formFields = UserForm.getFormFields();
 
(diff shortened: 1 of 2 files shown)


emanuelschuetze:
@ostcar Please review carefully. Is this a good solution for our discussed problem?

emanuelschuetze:
BTW: Works as expected! :-) Please give me an ok - than I will do this solution also for motions, assignments and customslides.

normanjaeckel:
Ok.

openslides/users/static/js/users/site.js
```diff
@@ -529,9 +531,18 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
...
+            // save change user object on server
             User.save(user).then(
                 function(success) {
                     $scope.closeThisDialog();
+                },
+                function(error) {
+                    // save error: revert all changes by restore
+                    // (refresh) original user object from server
+                    User.refresh(user);
+                    $scope.closeThisDialog();
```
ostcar: How does the user now in this case, that there was an error?

emanuelschuetze: Right, we should show the error in the dialgo and NOT close the dialog... I'll fix it.

emanuelschuetze:
@ostcar Fix remark. Please review and merge.

openslides/users/static/js/users/site.js
```diff
@@ -529,9 +532,22 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
...
                 user.groups = [];
             }
+            // inject the changed user (copy) object back into DS store
+            User.inject(user);
+            // save change user object on server
             User.save(user).then(
                 function(success) {
                     $scope.closeThisDialog();
+                },
+                function(error) {
```
ostcar: Proposal. Add a space between function an (error)

openslides/users/static/js/users/site.js
```diff
@@ -529,9 +532,22 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
...
+                    // save error: revert all changes by restore
+                    // (refresh) original user object from server
+                    User.refresh(user);
+                    var message = '';
+                    for (var e in error.data) {
+                        message += e + ': ' + error.data[e] + ' ';
+                    }
+                    $scope.alert = { type: 'danger', msg: message, show: true };
```
ostcar: Proposal: Remove the spaces after { and before }

ostcar:
Two proposals. You can rebase and merge if you don't want to fix them

openslides/users/static/js/users/site.js
emanuelschuetze: Fixed.

openslides/users/static/js/users/site.js
emanuelschuetze: Fixed.

emanuelschuetze:
Fixed remarks and rebased. Ready for merge.

PR #1869, opened 2016-01-14
Translated state names. Fixed #1864.

### LINKED ISSUE #1864: Motion state names are not translated
The rest api should send **translated** motion state names which are already marked as translated in django.po (and already translated into German). So why they not send as translated by serializer?

normanjaeckel:
@ostcar: Do you know a smarter solution for this? Something generic? If not we have to use this ugly hack.

### DIFF
--- a/openslides/motions/serializers.py
+++ b/openslides/motions/serializers.py
@@ -43,6 +43,9 @@ class StateSerializer(ModelSerializer):
     """
     Serializer for motion.models.State objects.
     """
+    name = SerializerMethodField()
+    action_word = SerializerMethodField()
+
     class Meta:
         model = State
         fields = (
@@ -59,6 +62,12 @@ class StateSerializer(ModelSerializer):
             'dont_set_identifier',
             'next_states',
             'workflow')
+
+    def get_name(self, obj):
+        return _(obj.name)
+
+    def get_action_word(self, obj):
+        return _(obj.action_word)
 
 
 class WorkflowSerializer(ModelSerializer):


ostcar:
I strongly recommend to translate the string on the client side. When we start caching the rest api, then we will become a lot of errors on assemblies where users use different languages.

@emanuelschuetze do you have any idea how we could add the state string (that is part of the workflow) into the javascript translation file?

emanuelschuetze:
No, I don't know a good solution. If we move the translations to client side we have to "define" all state names in JavaScript with gettext, like in https://github.com/OpenSlides/OpenSlides/blob/master/openslides/assignments/static/js/assignments/site.js#L351
That's not a good solution, especially when we add more states on server. So please translate on server side.

ostcar:
How do you fix the caching problem? Or do we decide now, that we will not cache the rest api in the future? This could mean, that OpenSlides can not be used on big assemblies.

emanuelschuetze:
For 2.0 we have no caching for rest api. But later (for 2.1+) we want it of course!  So what's the best solution now for this translation issue?

ostcar:
My solution is to send english strings from the server and translate them on the client.

The problem is, that the state_names are example content that any user could change. Normally there is only user generated content in the rest api.

Another solution would be, that the language has to be chosen when the stats are generated (in python manage.py migrate) then the translated strings would be in the database. Then it is the same problem like the config strings.

normanjaeckel:
Ping. ?!

emanuelschuetze:
I could make a new pull request where all strings are defined in JavaScript (with gettext). Ok?

normanjaeckel:
Ok.

PR #1872, opened 2016-01-14
Motion template fixes

emanuelschuetze:
- Add has_votes to motion serializer and add poll.has_votes check in motion detail view.
- Fix motion meta box 3 column layout.
- Added missing intents
- Show motionPoll edit form in dialog. Moved into new template.
- Clean up assignmentpoll form (like motionPoll)

### DIFF
--- a/openslides/motions/static/templates/motions/motion-detail.html
+++ b/openslides/motions/static/templates/motions/motion-detail.html
@@ -51,187 +51,142 @@
     </div>
   </div>
   <div class="detail" uib-collapse="isMeta">
-  <div class="row">
-    <div class="col-sm-4">
-        <!-- submitters -->
-        <h3 translate>Submitters</h3>
-        <div ng-repeat="submitter in motion.submitters">
-          {{ submitter.get_full_name() }}<br>
-        </div>
+    <div class="row">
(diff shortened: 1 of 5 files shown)


openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -51,187 +51,142 @@
...
+                      <span class="result_label"><translate>Votes cast</translate>:</span>
+                      <span class="result_value">{{ poll.votescast }} {{ poll.getVotesCastPercent() }}</span>
+                </table>
+          </ol>
+          <button ng-if="motion.isAllowed('create_poll')" ng-click="create_poll()" class="btn btn-default btn-sm">
+            <i class="fa fa-bar-chart fa-lg"></i>
+            <translate>New vote</translate>
+          </button>
+      </div>
+      <div class="col-sm-1"></div>
```
normanjaeckel: Why do we have an empty cell here?

normanjaeckel:
Ok.

openslides/motions/static/templates/motions/motion-detail.html
emanuelschuetze: fixed.

PR #1875, opened 2016-01-17
Template fixes

emanuelschuetze:


emanuelschuetze:
@ostcar @normanjaeckel ping, please review!

openslides/agenda/static/js/agenda/base.js
```diff
@@ -52,11 +52,11 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
...
                         title = this.title;
                     }
-                    if (this.getContentResource().verboseName) {
-                        title = gettextCatalog.getString(this.getContentResource().verboseName) + ' ' + title;
+                    if (this.getContentResource().agendaSupplement) {
+                        title = gettextCatalog.getString(this.getContentResource().agendaSupplement) + ' ' + title;
                     }
                     if (this.item_number) {
-                        title = this.item_number + ' ' + title;
+                        title = this.item_number + ' · ' + title;
```
ostcar: I am not sure if we want to force the dot here. The item_number can be anything, for examle 'a)'

ostcar:
Only one proposal, that could be "solved" by creating an issue.

openslides/agenda/static/js/agenda/base.js
emanuelschuetze: This is not a simple point ("."). This is a middle dot ("&middot;") which is used as separator between item_number and title. The symbol should separate the automated number from the custom title. And the middle dot works for all forms of item_number, see e.g. "TOP 1 &middot; Title" or "a) &middot; Title"
So I won't fix it here.

ostcar: ok

normanjaeckel:
Ok. Should I merge at once or do you want to rebase it?

emanuelschuetze:
In this pull request I prefered to keep the single commits because some commits are useful to show the diff of the single changes. So please merge if you can live with these commits.

PR #1889, opened 2016-01-25
Fixed requirements.

normanjaeckel:


emanuelschuetze:
Why we havn't Tornado listed in our `requirements_production.txt`. I know it is a requirement from sockjs-tornado. But it could be useful that we control the delivered version of Tornado. What do you think?

normanjaeckel:
Why should we want to unlock the Tornado requirement from the sockjs package. I don't see any advantage for this.

PR #1901, opened 2016-01-27
Translate permission strings. (Fixes #1895)

### LINKED ISSUE #1895: Change display name of permissions
Change the display name of all permissions to simpler string. Use don't know about our app-model system. So please use the third string only - like in 1.7.

Example:
"agenda | item | Can manage agenda" ->  "Can manage agenda"

Maybe we keep an additional 'group display name' string, e.g. 'Agenda', for better grouping in a matrix later.

And how we should translate these strings? Should the client do this?

emanuelschuetze:
Remove unused gettextCatalog content provider.

### DIFF
--- a/openslides/core/models.py
+++ b/openslides/core/models.py
@@ -2,7 +2,6 @@ from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.core.urlresolvers import reverse
 from django.db import models
-from django.utils.translation import ugettext_noop
 from jsonfield import JSONField
 
 from openslides.mediafiles.models import Mediafile
@@ -64,9 +63,9 @@ class Projector(RESTModelMixin, models.Model):
         """
         default_permissions = ()
         permissions = (
-            ('can_see_projector', ugettext_noop('Can see the projector')),
-            ('can_manage_projector', ugettext_noop('Can manage the projector')),
-            ('can_see_frontpage', ugettext_noop('Can see the front page')))
+            ('can_see_projector', 'Can see the projector'),
+            ('can_manage_projector', 'Can manage the projector'),
+            ('can_see_frontpage', 'Can see the front page'),)
 
     @property
     def elements(self):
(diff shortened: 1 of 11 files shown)


openslides/users/static/js/users/site.js
```diff
@@ -872,8 +871,47 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
             $scope.closeThisDialog();
         };
     }
+])
+
+// Mark all permission strings for translation in JavaScript.
+// (see models.py of each djang app)
```
normanjaeckel: `Django`

normanjaeckel:
Do you want to remove the translation tags for the permissions on server side (ugettext_noop in every models.py)?

ostcar:
I don't get it? Do you want to get rid of the django translation file? What is the goal of this pull request?

openslides/users/static/js/users/site.js
emanuelschuetze: done

emanuelschuetze:
Removed all ugettext_noop from models.py.
All done. Ready for merge.

normanjaeckel:
Travis failed. Otherwise looks good.

emanuelschuetze:
@normanjaeckel Any idea what's wrong with the travis check?

openslides/core/models.py
```diff
@@ -184,7 +183,7 @@ class Meta:
         ordering = ('name',)
         default_permissions = ()
         permissions = (
-            ('can_manage_tags', ugettext_noop('Can manage tags')),)
+            ('can_manage_tags', 'Can manage tags'))
```
normanjaeckel: Add comma before last closing bracket.

openslides/core/models.py
```diff
@@ -204,7 +203,7 @@ class ConfigStore(models.Model):
     class Meta:
         default_permissions = ()
         permissions = (
-            ('can_manage_config', ugettext_noop('Can manage configuration')),)
+            ('can_manage_config', 'Can manage configuration'))
```
normanjaeckel: Here the same.

openslides/core/models.py
```diff
@@ -230,7 +229,7 @@ class ChatMessage(RESTModelMixin, models.Model):
     class Meta:
         default_permissions = ()
         permissions = (
-            ('can_use_chat', ugettext_noop('Can use the chat')),)
+            ('can_use_chat', 'Can use the chat'))
```
normanjaeckel: Here the same.

emanuelschuetze:
Fixed and rebased.

Issue #1853, opened 2016-01-14
Show red bar on top if server is offline


emanuelschuetze:
@ostcar As discussed, please add a boolean variable to $rootscope.

Then, I can show a popup on client with a refresh link e.g..

PR #1907, opened 2016-01-27
Reconnect at once. Do not wait 5 seconds. Add a connected flag to the…

### LINKED ISSUE #1854: Autoupdate does not work - check reconnect
Projector loses autoupdate sometimes (reconnect does not work). Projector have to check reconnect.

ostcar:
… root scope

Fixes #1854

See also #1853


emanuelschuetze:
Tested: looks good! Thanks!

PR #1922, opened 2016-01-28
fix tornado on windows

### LINKED ISSUE #1921: Autoupdate broken under Windows
Tested 2.0b2-portable with Firefox/Chrome under Windows 7 and 10:
All autoupdates does not work! E.g. add new countdown or activate slide. Only after reload countdown is visible or slide is activated.

Gets some errors in JS console:
* `Error: motions/workflowstate.inject: 'attrs' must be an object or an array!`
* `Error: Projector.get(...) is undefinded` ($scope.rebuildAllElements -> core/site.js:612)

Gets an traceback in server console:
```
File "E:\src\openslides-portable\dist\openslides-2.0b2-portable\lib\site-packages\openslides\utils\autoupdate.py", line 88, in forward_rest_response
        'data': json.loads(response.body.decode())}
AttributeError: 'NoneType' object has not attribute 'decode'
```

ostcar:
fixes #1921

### DIFF
--- a/openslides/utils/autoupdate.py
+++ b/openslides/utils/autoupdate.py
@@ -97,9 +97,15 @@ class OpenSlidesSockJSConnection(SockJSConnection):
         object_url.
         """
         # Join network location with object URL.
+        if RUNNING_HOST == '0.0.0.0':
+            # Windows can not connect to 0.0.0.0. So connect to localhost instead
+            local_host = 'localhost'
+        else:
+            local_host = RUNNING_HOST
+
         wsgi_network_location = (
             settings.OPENSLIDES_WSGI_NETWORK_LOCATION or
-            'http://{}:{}'.format(RUNNING_HOST, RUNNING_PORT))
+            'http://{}:{}'.format(local_host, RUNNING_PORT))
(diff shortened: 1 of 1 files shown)


openslides/utils/autoupdate.py
```diff
@@ -97,9 +97,15 @@ def send_object(cls, object_url):
         object_url.
         """
         # Join network location with object URL.
+        if RUNNING_HOST == '0.0.0.0':
+            # Windows can not connect to 0.0.0.0. So connect to localhost instead
+            local_host = 'localhost'
```
normanjaeckel: Please rename this variable to `running_host` or something like this because it is not always 'localhost' or '127.0.0.1'.

ostcar: you can do it if you like. And it is not localhost but the host where the local application runs. If the application runs on 192.168.0.1, then local_host is the local host.

emanuelschuetze:
I tested your fix: Autoupdate works without problems under windows (tested with Chrome and Firefox). Nice! So please merge that I can publish soon a new 2.0b3!

PR #1926, opened 2016-01-29
Fix the start command not to reload if debug is False

### LINKED ISSUE #1920: Manage.py start opens two browser windows instead of one.

ostcar:
fixes #1920


emanuelschuetze:
Tested successfully on Linux with `python manage.py start` (Debug=false and true) -> Only one brwoser tab is opened. Nice! Thanks!
So please merge soon ;-)

PR #1927, opened 2016-01-29
Refactor reconnect function of autoupdate.

emanuelschuetze:


openslides/core/static/js/core/base.js
```diff
@@ -39,40 +39,41 @@ angular.module('OpenSlidesApp.core', [
...
+        $rootScope.connected = true;
+
+        var newConnect = function () {
+            socket = new SockJS(location.origin + "/sockjs");
+            clearInterval(recInterval);
+            socket.onopen = function () {
+                $rootScope.connected = true;
+            };
+            socket.onclose = function () {
+                $rootScope.connected = false;
```
emanuelschuetze: How can I prevent that $rootScope.connected is set to false after reload of the page? I don't want to show the red bar on reload - only for connection failed during event.

emanuelschuetze: I don't understand why `onclose()` is called while reload...

emanuelschuetze:
Ping, please merge today if possible... (The mentioned improvements of hide red bar in startup is not reaaly important and can do later if to complicated)

openslides/core/static/js/core/base.js
```diff
@@ -39,40 +39,41 @@ angular.module('OpenSlidesApp.core', [
...
-                    $rootScope.connected = false;
-                    autoupdate.connect()
-
-                };
             },
             on_message: function (receiver) {
-                this.message_receivers.push(receiver);
+                this.messageReceivers.push(receiver);
             },
             reconnect: function () {
```
normanjaeckel: Is there still any use of this `reconnect`?

normanjaeckel:
One question. Rest is ok.

openslides/core/static/js/core/base.js
emanuelschuetze: @ostcar can you answer? I am not sure.

ostcar: Yes. As you can see if you type "reconnect" in the search field on top of this page.

openslides/core/static/js/core/base.js
ostcar: onclose is called when the connection is closed. There is no difference if the connection is closed by the server, by the client or by some other reasons.

You can add a css-animation, so the red bar is shown 1 second after connected is set to false. So the red bar keeps hidden, if the connection lost is only for 1 second.

emanuelschuetze: 1 second delay (in css animation) is not the best solution (because in some large list views a reload need more than 1 second.
Currently, this "hide red bar on startup" feature is not really important - I like it even if the blue loading bar overrides the red bar. So "want fix" this red bar issue.

Please merge.

PR #1931, opened 2016-01-30
Refactored logic in autoupdate.py concerning wsgi network connection …

### LINKED ISSUE #1929: Change variable name in autoupdate.py
See #1922.

normanjaeckel:
…variables. Fixed #1929."

### DIFF
--- a/openslides/utils/autoupdate.py
+++ b/openslides/utils/autoupdate.py
@@ -97,15 +97,14 @@ class OpenSlidesSockJSConnection(SockJSConnection):
         object_url.
         """
         # Join network location with object URL.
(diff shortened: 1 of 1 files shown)


ostcar:
The variable has nothing to do with wsgi. Wsgi is a specific protocol, but in this function the server is connected using normal http. But you can name it as you like, and therefore merge this pull request if you like it

normanjaeckel:
Hm. Then the settings var is also wrong. Suggestions?

ostcar:
I have no suggestions. The current code is ok for me. If you want to search for another name, then please add an issue

PR #1932, opened 2016-02-01
Fix the autoupdate

ostcar:


normanjaeckel:
The code is hard to understand. With this pull request, it becomes much harder. See #1934 and please to just say: "It's not my thing ...". :-)

PR #1933, opened 2016-02-01
Optimize agenda tree save method on client side.

### LINKED ISSUE #1660: Optimise OpenSlides for at least 100 Agena Items
At the moment OpenSlides can not handle 100 Agenda Items. When one item is changed in the sort view, then all items are sent to all clients. We need a function to send more then one item at the same time to the clients

emanuelschuetze:
with ui-tree
Fixes #1660


ostcar:
ok

PR #1936, opened 2016-02-02
Add Pagination and improve import.

emanuelschuetze:
- Use Pagination for users and item list and users import table.
- Improve agenda import: Allow to define agenda item stuff like
  duration, comment, type.


normanjaeckel:
Ok.

PR #1944, opened 2016-02-05
Updated README.

### LINKED ISSUE #1382: Use pyvenv instead of virtualenv
Python 3.3+ contains pyvenv which works like virtualenv:
https://docs.python.org/3/library/venv.html
We should change the OpenSlides install instruction (in README and manual?) to switch to pyvenv.

Note: Python3.3 does not install pip automatically. Python 3.4 does it.

normanjaeckel:
Fixed #1382.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,8 +14,8 @@ II. Requirements
 ================
 
 OpenSlides runs everywhere where Python is running (for example on
-GNU/Linux, Mac or Windows (XP or newer)). On each client you need only an
-actual webbrowser.
+GNU/Linux, Mac or Windows (XP or newer)). On each client you need only a
+current version of a webbrowser.
 
 
 III. Installation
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -91,24 +90,18 @@ portable version you should observe the following install steps.*
...
+   You can setup a virtual Python environment to install OpenSlides as
+   non-root user.
 
    Create your OpenSlides directory, change to it, setup and activate the
    virtual environment::
 
        > md OpenSlides
        > cd OpenSlides
-       > virtualenv .virtualenv
+       > python3 -m venv .virtualenv
```
emanuelschuetze: Do you have test this command on windows with python3.4?

normanjaeckel: Not yet. :-)

emanuelschuetze: We need a Windows Test.

emanuelschuetze: I tested it with Python 3.4 (32bit) under Windows 7. Please use the following lines in README:
```
python -m venv .virtualenv
.virtualenv\Scripts\activate.bat
```

README.rst
```diff
@@ -119,7 +112,7 @@ portable version you should observe the following install steps.*
 
        > easy_install openslides-x.x.tar.gz
```
emanuelschuetze: We should recommend `pip` instead of `easy_install`. I am not sure but I think first we have to install pip:
`easy_install pip`. And then: `pip install openslides-x.x.tar.gz`

emanuelschuetze: Yes, pip already available after Python 3.4 installation on windows. See https://docs.python.org/3/whatsnew/3.4.html#whatsnew-pep-453
So you can replace `easy_install`to `pip`

normanjaeckel:
All done.

emanuelschuetze:
Ok

Issue #1955, opened 2016-02-08
Gulp production bugs


normanjaeckel:
* [x] Problem 1: the gulp tasks 'js-libs', 'fonts-libs' and 'ckeditor' runs very long! (>3min).
* [x] Problem 2: The two fonts (Roboto and Open Sans) are not installed correctly. You can't use it in template.

emanuelschuetze:
Problem 1 is fixed with #1958.

emanuelschuetze:
@normanjaeckel Any idea how we can fix problem 2? Looks like this was introduces with cssnano(?)...

normanjaeckel:
Problem 23 fixed by #1960

PR #1958, opened 2016-02-08
Updated gulp-uglify.

normanjaeckel:
I can't reproduce the time bug in #1955 any more.


emanuelschuetze:
ok. Works for me too.

PR #1961, opened 2016-02-08
Refactored MotionPoll.

emanuelschuetze:
Use new getVotes function for vote value and percent value.
Remove 'to_representation' function in motion serializer.py to get
special values in numbers (-1,-2) instead of translated strings.


normanjaeckel:
Ok.

PR #1964, opened 2016-02-09
Assignmentpoll slide

emanuelschuetze:


emanuelschuetze:
Ready for review and merge.

openslides/assignments/projector.py
```diff
@@ -20,6 +20,11 @@ def get_context(self):
             # Detail slide.
             if not Assignment.objects.filter(pk=pk).exists():
                 raise ProjectorException('Election does not exist.')
+        poll_id = self.config_entry.get('poll')
```
normanjaeckel: Indent this line and the following because they are only required for detail slides.

openslides/assignments/projector.py
```diff
@@ -20,6 +20,11 @@ def get_context(self):
...
+        if poll_id is not None:
+            # Detail slide.
```
normanjaeckel: Change comment: `# Poll slide`

openslides/assignments/static/js/assignments/base.js
```diff
@@ -183,6 +191,25 @@ angular.module('OpenSlidesApp.assignments', [])
...
+                        '/rest/core/projector/1/prune_elements/',
+                        [{name: 'assignments/assignment', id: this.id, poll: poll_id}]
+                    );
+                },
+                // override isProjected function of jsDataModel factory
+                isProjected: function (poll_id) {
+                    // Returns true if there is a projector element with the name
+                    // 'assignments/assignment'.
+                    var projector = Projector.get(1);
+                    if (typeof projector === 'undefined') return false;
```
normanjaeckel: Please do not call a return inside a function. Use an if-else-clause for example.

openslides/assignments/static/js/assignments/base.js
```diff
@@ -183,6 +191,25 @@ angular.module('OpenSlidesApp.assignments', [])
...
+                    var self = this;
```
normanjaeckel: This line is not used.

openslides/assignments/projector.py
```diff
@@ -2,8 +2,8 @@
 from openslides.core.views import TagViewSet
 from openslides.utils.projector import ProjectorElement, ProjectorRequirement
 
-from .models import Assignment
-from .views import AssignmentViewSet
+from .models import Assignment, AssignmentPoll
+from .views import AssignmentPollViewSet, AssignmentViewSet
```
normanjaeckel: AssignmentPollViewSet is not used.

openslides/assignments/projector.py
emanuelschuetze: Done.

openslides/assignments/projector.py
emanuelschuetze: Done

openslides/assignments/static/js/assignments/base.js
emanuelschuetze: We override the isProjected function from [core/base.js](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/core/static/js/core/base.js#L260), so we copy 1:1 the function and just modified line 210.
Wan't  fix for consistency.

openslides/assignments/static/js/assignments/base.js
emanuelschuetze: see above.

openslides/assignments/projector.py
emanuelschuetze: Done.

emanuelschuetze:
Fixed remarks. Rebased.

openslides/assignments/static/js/assignments/base.js
normanjaeckel: Both functions are really ugly. There is no self and the `return` inside the code makes it really hard to understand. If we implement this here for consistency we will have much more work in the future.
But the solution is easy. The general function looks like this:
```javascript
function () {
    // Returns true if there is a projector element with the same
    // name and the same id.
    var projector = Projector.get(1);
    if (typeof projector !== 'undefined') {
        var predicate = function (element) {
            return element.name == self.getResourceName() &&
                   typeof element.id !== 'undefined' &&
                   element.id == self.id;
        };
        var isProjected = typeof _.findKey(projector.elements, predicate) === 'string';
    } else {
        var isProjected = false;
    }
    return isProjected
};
```
Don't forget to adapt the comment.

emanuelschuetze: Done for all functions (in core, agenda and assignments).

emanuelschuetze:
ready for review

normanjaeckel:
One last comment.

emanuelschuetze:
All done

PR #1968, opened 2016-02-11
Fixed error in poll default_votes_validator. Fixed #1967.

### LINKED ISSUE #1967: Error if poll default votes are empty
Pull request #1965 has introduced a new error:
If you let the poll fields (valid/invalid/cast) empty you get an (very long und ugly!) error message that all values have to be >=-2. Please test it with current master.
This error has to be fixed for 2.0. These poll fields should be are optional (you should get an error only if value <-2.

normanjaeckel:


emanuelschuetze:
ok

PR #1969, opened 2016-02-11
Template fixes

emanuelschuetze:


emanuelschuetze:
ping, ready for merge

openslides/motions/signals.py
```diff
@@ -68,7 +68,7 @@ def setup_motion_config(sender, **kwargs):
         subgroup=ugettext_lazy('General'))
 
     # Amendments
-
+    # Amendments currently not implemented. (TODO: Implement it like in OpenSlides 1.7.)
```
normanjaeckel: If we do not have amendments we might disable these two config variables, don't we?

normanjaeckel:
You can check my comment or just merge it.

openslides/motions/signals.py
emanuelschuetze: How we can disable the configs that they are defined but are not visible? Currently we can't just remove it because several tests and motion model required these options.

normanjaeckel:
See #1971. Merge it first and then rebase your branch on top of it.

emanuelschuetze:
rebased.

PR #1984, opened 2016-02-17
Updated CHANGELOG

emanuelschuetze:


emanuelschuetze:
Only some changelog strings. I'll merge it now (for 2.0b4).

CHANGELOG
```diff
@@ -13,7 +13,7 @@ Agenda:
 - Changed API of related objects. All assignments, motions and custom slides
   are now agenda items and can be hidden.
 - Removed mptt.
-- Added attachments to custom slides.
+- Added attachments to agenda items.
```
normanjaeckel: Ah that's not correct. See two lines before: Assignments are now agenda items but they have not attachments. I propose to revert this change.

emanuelschuetze: Mmh that's not really from user view - because user know only agenda items, motions and assignments. No custom slides.So maybe we should change the item to lines before...

PR #1994, opened 2016-02-23
Fixed error if a PDF file was deleted from the server. Fixed #1988.

### LINKED ISSUE #1988: FileNotFoundError if mediafile is deleted on file system
If you delete an uploaded file from /media/file/ and then open the mediafile list view you get a FileNotFoundError which should be catched. (Otherwise you can't access to the the mediaflie list view any more.) 
```
[..]
openslides/mediafiles/serializers.py", line 27, in to_representation
        result['pages'] = PdfFileReader(open(value.path, 'rb')).getNumPages()
    FileNotFoundError: [Errno 2] No such file or directory:
```

normanjaeckel:


emanuelschuetze:
ok

PR #1996, opened 2016-02-23
Added language header to internal server requests. Fixed #1981.

### LINKED ISSUE #1981: Translation of motion log missing (sometimes)
Motion log is switched to English if something was changed (e.g. motion state changed), reload helps

normanjaeckel:


emanuelschuetze:
ok.

PR #2000, opened 2016-02-24
Optimized number of queries for motion serializing. Fixed #1655.

### LINKED ISSUE #1655: Motion poll serializer gets the poll data three times from the db
At the moment the motion poll serializer gets all poll data for the yes, no and abstain field. But this is the same db_query.

See #1653

normanjaeckel:


emanuelschuetze:
ok

PR #2003, opened 2016-02-25
Reload after login and logout (Fixes #1820)

### LINKED ISSUE #1820: first load whoami and then load all data
when an anonymous gets some data and logs on later, then he could have castrated data

emanuelschuetze:
Not very nice solution but with DS.clear() we have to do a lot of more
work with high effort. Can improved later.

### DIFF
--- a/openslides/users/static/js/users/site.js
+++ b/openslides/users/static/js/users/site.js
@@ -129,10 +129,6 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
                     showClose: $stateParams.guest_enabled,
                     closeByEscape: $stateParams.guest_enabled,
                     closeByDocument: $stateParams.guest_enabled,
-                    preCloseCallback: function() {
-                        $state.go('home');
-                        return true;
-                    }
                 });
             }]
         });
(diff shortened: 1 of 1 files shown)


openslides/users/static/js/users/site.js
```diff
@@ -912,6 +907,9 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
                     // Success: User logged in.
                     operator.setUser(response.data.user_id);
                     $scope.closeThisDialog();
+                    setTimeout(function(){
```
normanjaeckel: Proposal: Close the dialog after the 1000 ms waiting time, i. e. move this call into the timeout.

normanjaeckel:
One idea. Then merge.

openslides/users/static/js/users/site.js
emanuelschuetze: I tested it but it does not work nicely because we have to wait until the dialog is closed and then reload  the page. Otherwise you has unattractiv effects if you reload if the dialog is still open.
So: won't fix!

normanjaeckel: ok.

PR #2011, opened 2016-02-26
Updated requirements_production.txt

ostcar:
reportlab 3.3 can be used
sockjs-tornado 1.0.0 is not usable. We need at least 1.0.1

### DIFF
--- a/requirements_production.txt
+++ b/requirements_production.txt
@@ -5,9 +5,9 @@ djangorestframework>=3.2.0,<3.4.0
 html5lib>=0.9,<1.0
 jsonfield>=0.9.19,<1.1
 natsort>=3.2,<4.1
-reportlab>=3.0,<3.3
+PyPDF2>=1.25.0,<1.26
+reportlab>=3.0,<3.4
 roman>=2.0,<2.1
 setuptools>=2.2,<20.0
-sockjs-tornado>=1.0,<1.1
+sockjs-tornado>=1.0.1,<1.1
 Whoosh>=2.7.0,<2.8
-PyPDF2==1.25.1


requirements_production.txt
```diff
@@ -5,9 +5,9 @@ djangorestframework>=3.2.0,<3.4.0
...
 natsort>=3.2,<4.1
-reportlab>=3.0,<3.3
+PyPDF2>=1.25.0,<1.26
+reportlab>=3.0,<3.4
 roman>=2.0,<2.1
 setuptools>=2.2,<20.0
-sockjs-tornado>=1.0,<1.1
+sockjs-tornado>=1.0.1,<1.1
 Whoosh>=2.7.0,<2.8
-PyPDF2==1.25.1
```
emanuelschuetze: @normanjaeckel Was there a reason why pyPDF2 currently use the explicit version 1.25.1?

requirements_production.txt
```diff
@@ -5,9 +5,9 @@ djangorestframework>=3.2.0,<3.4.0
 html5lib>=0.9,<1.0
 jsonfield>=0.9.19,<1.1
 natsort>=3.2,<4.1
-reportlab>=3.0,<3.3
+PyPDF2>=1.25.0,<1.26
+reportlab>=3.0,<3.4
```
emanuelschuetze: @ostcar Do you tested that all reportlab pdfs works with 3.3? Is there a reason to upgrade to 3.3 in our late beta phase (could bring us new regressions) - and we will drop out reportlab for OpenSlides 2.1 (by replacing with jsPDF).

ostcar: I tested our pdf with reportlab 3.3.

The reason for the upgrade is, that debian stretch comes with reportlab 3.3. I installed openslides without the need to compile reportlab by installing python3-reportlab. But this was only possible after I changed the supported version to reportlab<3.4.

So in conclusion, to update the version makes the installation a lot easier.

emanuelschuetze: ok

requirements_production.txt
normanjaeckel: I don't think that there is a reason.

PR #2012, opened 2016-02-27
Changed name of '--development' option and name of the respective dir…

### LINKED ISSUE #2001: Renaming --development option
I think about if the option `--development` (of `openslides start --development`) has the right wording... It's not really a development feature to keep all settings/database etc. in the same directory and not in .config and .local. 

So my suggestion: rename this option (and the directory!) and improve the help text (`command for development purpose`) which is not really helpful to understand what this option make.

(split out this issue from #1998)

normanjaeckel:
…ectory. Fixed #2001 

Please check very carefully.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,8 +12,8 @@
 node_modules/*
 bower_components/*
 
-# Development user data (settings, database, media, search index, static files)
-development/*
+# Local user data (settings, database, media, search index, static files)
+personal_data/*
 openslides/static/*
 
 # Package building/IDE
--- a/openslides/__main__.py
+++ b/openslides/__main__.py
@@ -10,8 +10,8 @@ from openslides.utils.main import (
     ExceptionArgumentParser,
     UnknownCommand,
     get_default_settings_path,
-    get_development_settings_path,
-    is_development,
+    get_local_settings_path,
+    is_local_installation,
     setup_django_settings_module,
     start_browser,
     write_settings,
@@ -32,8 +32,8 @@ def main():
 
     if unknown_command:
         # Run a command, that is defined by the django management api
-        development = is_development()
-        setup_django_settings_module(development=development)
+        local_installation = is_local_installation()
+        setup_django_settings_module(local_installation=local_installation)
         execute_from_command_line(sys.argv)
     else:
         # Run a command that is defined here
@@ -45,7 +45,7 @@ def get_parser():
     """
     Parses all command line arguments.
     """
-    if len(sys.argv) == 1 and not is_development():
+    if len(sys.argv) == 1 and not is_local_installation():
         sys.argv.append('start')
 
     # Init parser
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -12,8 +12,8 @@
 node_modules/*
 bower_components/*
 
-# Development user data (settings, database, media, search index, static files)
-development/*
+# Local user data (settings, database, media, search index, static files)
+openslides_data/*
```
emanuelschuetze: I think the new directory name `openslides_data` is not very ideal because the `openslides` directory in a git checkout is very similar (and we have to open this `openslides` directory very often in command line during development). So my suggestion: rename to `data`.

openslides/__main__.py
```diff
@@ -104,9 +104,9 @@ def get_parser():
         help='The used settings file. The file is created, if it does not exist.')
     subcommand_start.set_defaults(callback=start)
     subcommand_start.add_argument(
-        '--development',
+        '--local-installation',
         action='store_true',
-        help='Option for development purposes.')
+        help='Store settings and user files in a local directory.')
```
emanuelschuetze: I think this option is only for *create* the directory (by strating openslides on first time) - not for further start commands. Suggestionn for help text: `Create database and settings file in a local subdirectory of this current directory.`

emanuelschuetze:
Tested successfully. Two comments.

openslides/__main__.py
normanjaeckel: No: You have to use this option if you want to start OpenSlides and use the settings over there. Else OpenSlides tries to find the settings deep inside the user path.

.gitignore
normanjaeckel: What about `personal_data` or `local_data`?

openslides/__main__.py
emanuelschuetze: Really? I tested the following:
1. `python manage.py start --local-installation` -> `openslides_data` created, server starts with this new db.
2. stop server and start again: `python manage.py start` -> same db is used (from `openslides_data`

So it seems that I need the `--local-installation` option for create only... isn't it?

.gitignore
emanuelschuetze: Both suggestion are ok for me. `local_data` matchs a bit better for `local-installation` option. But `personal_data`explains better that this directory contains *personal* user data. So choose one of them...

openslides/__main__.py
normanjaeckel: No: If you use `manage.py` then OpenSlides will always use the local folder (development, openslides_data ot whatever we call it). If you use `manage.py` with the flag `--local-installation` this does the same. But if you use `openslides start` after you installed a tar.gz version then OpenSlides will install to `~/.config/openslides` and thelike. See #2013.

.gitignore
normanjaeckel: I take `personal_data` and rebase.

openslides/__main__.py
normanjaeckel: See also [here](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/utils/main.py#L310).

emanuelschuetze: Ok.

emanuelschuetze:
One (really) last thing: Can you add a note about the new option in README? Could be very useful for users who want to test 2.0 but have already an old .local/share/openslides and .config/openslides directory.

normanjaeckel:
Ready for review.

normanjaeckel:
Changed README text in another way. Ready to merge.

emanuelschuetze:
Ok. Fine!

PR #2018, opened 2016-03-01
Optimization

normanjaeckel:
@emanuelschuetze Please test this branch (test functionality and compare performance with numerous clients). Some details are still missing:

- [x] default password handling
- [x] handling of hidden agenda items
- [x] removing of one old rest api method
- [x] Rebase on top of squished migrations because of error during database creation.

~~moving of translations~~ See #2025


emanuelschuetze:
Can you rebase the squihsed migrations please? I can't test this with a fresh db...

emanuelschuetze:
Translations checked in /rest/: only motion logs are important. config labels comes with option request (not via auto update). All other apps (motions/assingments/mediafiles/agenda) has no special translations in auto update data. So you can create a new issue (for 2.1) to fix the motion log translation after auto update (that not the browser language is used from user who triggers the auto update).

emanuelschuetze:
As discussed: I tested it with 100 clients (50 open tabs with opened projector sidebar) which opens /agenda with 20 items.
I changed one item. Now only a PUT request is send. No 100 GET requests as before (on master). 
So it is a good performance improvement!

emanuelschuetze:
Code looks ok for me after quick code review. And works as expected...

@ostcar Can you make also a quick review because these are important changes for our auto update process. Thx.

openslides/utils/autoupdate.py
```diff
@@ -72,78 +72,58 @@ def on_open(self, info):
...
+            if access_permissions.get_dispatch_uid() == container.get('dispatch_uid'):
+                break
         else:
-            if RUNNING_HOST == '0.0.0.0':
-                # Windows can not connect to 0.0.0.0, so connect to localhost instead.
-                wsgi_network_location = 'http://localhost:{}'.format(RUNNING_PORT)
-            else:
-                wsgi_network_location = 'http://{}:{}'.format(RUNNING_HOST, RUNNING_PORT)
-        url = ''.join((wsgi_network_location, object_url))
+            raise ValueError('Invalid container. A valid dispatch_uid is missing.')
```
ostcar: Are you sure you want to raise an Error here? I am not sure, but i think in this case openslides stops. It is different then in the django part, where only a 500er is send. You should test this

openslides/utils/autoupdate.py
```diff
@@ -72,78 +72,58 @@ def on_open(self, info):
...
+                    'id': container['rest_pk'],
+                    'data': data}
+            elif container.get('action') == 'deleted':
+                output = {
+                    'status_code': 404,  # TODO: Refactor this. Use strings like 'change' or 'delete'.
+                    'collection': container['collection_string'],
+                    'id': container['rest_pk']}
             else:
-                headers.parse_line('Accept-Language: ' + languages)
+                raise ValueError('Invalid container. A valid action is missing.')
```
ostcar: Here the same

ostcar:
looks promising! ok

openslides/utils/autoupdate.py
normanjaeckel: I tested this error: If it is raised the client gets a HTTP 500 but the server continues running even with tornado debug set to False. So I think this is ok. I expect this error only if a developer writes bad code so that wrong data are submitted to our autoupdate system.

openslides/utils/autoupdate.py
normanjaeckel: See below. I think this is the same here.

emanuelschuetze:
Are this PR ready for last test? Can you rebase it?

normanjaeckel:
Ready for last test. I rebase it later.

openslides/utils/autoupdate.py
ostcar: ok

emanuelschuetze:
I make again some performance tests:
Opened ~40 /agenda/ tabs, imported ~10 agenda items and numbering all items.
* with current master: a huge number of GET requests goes to server. CPU load goes up to 100%!
* with this branch: Only some post request for import/numbering goes to server, no longer a huge number of GET requests. CPU load goes up to ~50% (while import) and only ~20% (while numbering).

Summarized a very good performance improvement! Good work!
Please rebase and merge!

PR #2024, opened 2016-03-02
Fixed gulp jshint task.

normanjaeckel:
@emanuelschuetze Please check the implementation in projector.js. I think the `if (tree[0])` clause might be strange.


emanuelschuetze:
Your change looks ok for me.

PR #2037, opened 2016-03-06
Use specific tag of tinymce-18n repo (Fixes #1975).

### LINKED ISSUE #1975: Tag OpenSlides/tinymce-i18n repo or use commit hash

emanuelschuetze:


normanjaeckel:
Why does our repo has a tag named `4.3.2`? I now that this is the current releas ot TinyMCE but the languages files do not contain to this released. They have their own revision dates as it can be seen [here](http://archive.tinymce.com/i18n3x/index.php?ctrl=lang&act=index&pr_id=1). So I prefer not to use version tags that look like releases. We can use git commit hashes instead.

normanjaeckel:
If we are sure that we do not change we can also tag with the number of the current OpenSlides release but then we have to add new tags even if the languages do not change.

emanuelschuetze:
I use the last stable version as tag name because all language files contains the last strings of this 4.x release. Of course we could use also a commit hash... It makes no difference to me. I see our tinymce-i18n repo only for temporary use until TinyMCE has set up their own bower repo for i18n. So don't spend to much time for this issue...
Let me know what you like. I prefer the 4.3.2 tag name...

emanuelschuetze:
I don't like tag names correspond to our OpenSlides releases - it should be independed that also other tinyMCE user can use our repo. And it should contains a state corresponding to the current last stable release. So keep this current tag name also in our bower file.

normanjaeckel:
The problem of tag names that correlate to the releases of TinyMCE is that the guys over there might add or change some languages without releasing a new version of the editor. Then we can't take these changes without breaking our tag naming system.

What about a combined version number like `4.3.2-1.0.0`?

emanuelschuetze:
> So I prefer not to use version tags that look like releases. We can use git commit hashes instead.

Ok. Then let us remove the tag and us git commit hashes. I think it is to confused to use any tag number which is similar to the tinymce release (BTW: 4.3.7 is currently last stable version, see [releases](https://github.com/tinymce/tinymce/releases)).

So, I don't want to spend to much time in this discussion. Is it ok for you to use commit has in our bower.json?

emanuelschuetze:
I just removed the tag and updated my commit by using the git commit hash. Ready for merge.

normanjaeckel:
Please rebase and merge.

emanuelschuetze:
Rebased.

PR #2045, opened 2016-03-17
Use angular-chosen instead of ui-select for better performance.

### LINKED ISSUE #2006: Fix empty users dropdown fields
To reproduce:
* import >50 users
* create a new motion and select the last user as submitter (is not in scoll list becase only the first 50 users are displayed)
* open motion update form 
-> submitter field shows empty user (because this user is filtered by 'limitTo')

emanuelschuetze:
The jQuery select field from angular-chosen is used for all ui-select
fields. See https://github.com/leocaseiro/angular-chosen/

Use input delay on users filter field with 'debounce'.

Fixed #2006.


gulpfile.js
```diff
@@ -68,6 +68,12 @@ gulp.task('fonts-libs', function() {
         .pipe(gulp.dest(path.join(output_directory, 'fonts')));
 });
 
+// Catches img file for angular-chosen.
```
normanjaeckel: Is there more than one file? If yes, type `// Catches image files for angular-chosen.`

normanjaeckel:
One comment, rest looks good.

gulpfile.js
emanuelschuetze: Done.

emanuelschuetze:
All done. I added a new commit which improved some translation strings. Please review 2nd commit. If ok I'll rebase and merge.

normanjaeckel:
Ok. Rebase and merge.

PR #2051, opened 2016-03-18
Template improvements

emanuelschuetze:


openslides/users/static/js/users/site.js
```diff
@@ -58,6 +58,26 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
                 }
             }
         })
+        // redirects to motion detail and opens motion edit form dialog, uses edit url,
```
normanjaeckel: I do not understand this comment. Please write one sentence more.

PR #2052, opened 2016-03-20
Use the title of the custom slide directly in the slide

### LINKED ISSUE #2050: Agenda item title of new customslide is missing on projector after autoupdate
To reproduce:
* open projector in new window and project agenda
* create new customslide with title and text -> customslide title is shown in agenda list on projector
* project created customslide -> only customslide text is shown on projector (the title is missing!)

It seems that in core/projector.js the related agenda item is not loaded correctly by loadRelation:
`Customslide.loadRelations(customslide, 'agenda_item');`

If you reload the projector the related agenda item is loaded and the title is shown correctly.

@ostcar I don't know whats wrong here. Any idea?

ostcar:
Fixes #2050

### DIFF
--- a/openslides/core/static/templates/core/slide_customslide.html
+++ b/openslides/core/static/templates/core/slide_customslide.html
@@ -1,4 +1,4 @@
 <div ng-controller="SlideCustomSlideCtrl" class="content scrollcontent">
-  <h1>{{ customslide.agenda_item.getTitle() }}</h1>
+  <h1>{{ customslide.title }}</h1>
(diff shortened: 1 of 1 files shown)


emanuelschuetze:
We need the Agenda title (which contains also the item number) not only the customslide title...

ostcar:
In this case you have to load the item and use the old code. The item is not loaded and therefore the title can not be shown

emanuelschuetze:
Closed. See #2056

PR #2056, opened 2016-03-20
Fix #2050

### LINKED ISSUE #2050: Agenda item title of new customslide is missing on projector after autoupdate

ostcar:


emanuelschuetze:
Tested successfully: #2050 works now.

PR #2057, opened 2016-03-20
Use specific host and port for 'openslides start' (Fixes #1987)

### LINKED ISSUE #1987: 'openslides start' needs port option
If you install 2.0b4 (`pip install openslides-2.0b4.tar.gz`) you can run OpenSlides server via `openslides start` only. There is no `manage.py runserver` command in release build. The server will run on `0.0.0.0:8000`.

With the start script you can not change host and port. At least changing the port is important for production use (e.g. adminstrators typically wants to change the port to `80`).

See also #1532.

emanuelschuetze:


openslides/__main__.py
```diff
@@ -98,6 +98,14 @@ def get_parser():
         action='store_true',
         help='Do not launch the default web browser.')
     subcommand_start.add_argument(
+        '--host',
+        action='store',
+        help='IP address to listen on. Default is 0.0.0.0.')
+    subcommand_start.add_argument(
+        '--port',
+        action='store',
+        help='Port to listen on. Default is 8000.')
```
normanjaeckel: You can use `default` here: https://docs.python.org/3.5/library/argparse.html#default so you don't need the if clausees below.

openslides/__main__.py
```diff
@@ -165,12 +173,23 @@ def start(args):
...
+        host = "0.0.0.0"
+    if args.port:
+        port = args.port
+    else:
+        port = "8000"
+
+    # Open browser always on localhost (0.0.0.0 is not supported on Windows)
     if not args.no_browser:
-        start_browser('http://localhost:8000')
+        start_browser('http://localhost:%s' % port)
```
normanjaeckel: The browser should open at the given host. Only if it is `0.0.0.0`, it should be changed to `localhost`.

emanuelschuetze:
Rebased. Ready for merge.

PR #2070, opened 2016-04-06
Used exit method from sys library. Fixed #2065.

### LINKED ISSUE #2065: Unknown function exit() in __main__.py
There is an error in the last line of __main__.py. The statement exit(main()) does not work in this way, because exit is not imported. 
Suggested solution:: sys.exit(main()) An import for sys ist already there

normanjaeckel:


emanuelschuetze:
ok

PR #2079, opened 2016-04-13
Support setuptools 20

### LINKED ISSUE #2078: Support setuptools 20
At the moment we only support setuptools<20.0 but in Ubuntu 16.04 there is setuptools 20.3 in the packages. To make the creation of a ubuntu package possible after the 2.0 release, we should change the requirements_production.txt to support setuptools 20.

http://packages.ubuntu.com/xenial/python3-setuptools

ostcar:
Fixes #2078

I did not test this. This has to be tested when a new release is created because setuptools is only used by the "openslides" command.


emanuelschuetze:
Successfully tested with setuptools 20.8.1 on ubuntu 14.04 and windows portable.

PR #2082, opened 2016-04-14
Use always user.groups_id (Fixes #2081)

### LINKED ISSUE #2081: Delegate group permissions are ignored
Create new delegate user and login as this user.
-> All delegate specific permissions does not work. E.g. there is no new motion button! (for stuff users it works). 

Group permissions of delegate user in /rest/users/group are ok (e.g. motions.can_support). So it seems to be a client problem with group delegate.

emanuelschuetze:
- Fix rest api: send always groups_id (instead of groups).
- Fix JS-Data-Store: Add hasMany relations for user.groups.
- Fix templates: use field 'groups_id' instead of 'groups'.

### DIFF
--- a/openslides/users/serializers.py
+++ b/openslides/users/serializers.py
@@ -4,8 +4,8 @@ from django.utils.translation import ugettext as _
 from django.utils.translation import ugettext_lazy
 
 from ..utils.rest_api import (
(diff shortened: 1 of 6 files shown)


normanjaeckel:
Ok, but not tested.

PR #2101, opened 2016-05-10
Change the z-index of `.fullscreen` to 199.

### LINKED ISSUE #2096: Messages and timers during file presentation
When starting a presentation from a file (PDF), I can't seem to show any other elements, like messages or timers. It seems that there is no way to change the presentation from fullscreen to non-fullscreen.

lesteenman:
Z-Index 199 places the fullscreen content just below the various
projector overlays, such as messages and timers.

Fixes #2096.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,3 +18,4 @@ Authors of OpenSlides in chronological order of first contribution:
     René von der Haar <rene.vonderhaar@medienweite.de> (Logo design)
     Jörn Bensch <bensch@triagonale.de> (Template design)
     John Felipe Urrego Mejia <ingenierofelipeurrego@gmail.com> (Spanish translation)
+	Erik Steenman <eriksteenman@gmail.com>
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,4 +18,4 @@ Authors of OpenSlides in chronological order of first contribution:
     René von der Haar <rene.vonderhaar@medienweite.de> (Logo design)
     Jörn Bensch <bensch@triagonale.de> (Template design)
     John Felipe Urrego Mejia <ingenierofelipeurrego@gmail.com> (Spanish translation)
-	Erik Steenman <eriksteenman@gmail.com>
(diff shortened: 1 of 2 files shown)


normanjaeckel:
Suggestion: Use `100` or `150` so that plugins may add additional content between PDF and Countdown (e. g. a custom countdown background).

emanuelschuetze:
@lesteenman And with your first contribution to OpenSlides core you are allowed to add your name to AUTHORS file: https://github.com/OpenSlides/OpenSlides/blob/master/AUTHORS ;-)
(I'll add you to the list of github developers to assign issues/pr to you.)

lesteenman:
@normanjaeckel: That sounds like a good idea, actually. Am I correct that there are no other z-indices in use between <10 and 300?

lesteenman:
@emanuelschuetze: Should I include that in a PR, make a separate one, or will you do it on the next release?

normanjaeckel:
I we don't have any other z-indexes in this range. Countdown has 200, messages has 300 and 301 and the content has -1 and 0.

You can add you changes to the AUTORS file right now in this pull request. You may even use just one commit for all (using git rebase).

lesteenman:
Rebase to squash both commits and then force-push, you mean?

normanjaeckel:
Yes, this how we mostly work. But it is no obligation.

normanjaeckel:
Ok.

AUTHORS
```diff
@@ -18,3 +18,4 @@ Authors of OpenSlides in chronological order of first contribution:
     René von der Haar <rene.vonderhaar@medienweite.de> (Logo design)
     Jörn Bensch <bensch@triagonale.de> (Template design)
     John Felipe Urrego Mejia <ingenierofelipeurrego@gmail.com> (Spanish translation)
+	Erik Steenman <eriksteenman@gmail.com>
```
emanuelschuetze: Please use same indent like other Autors.

emanuelschuetze:
One remark. Rest ok.

PR #2117, opened 2016-05-12
Initial attempt at support for image and video files.

### LINKED ISSUE #2107: feature request: possibility to project image files
Currently one can only display PDF files via the menu Files. It should also be possible to display PNG and JPG/JPEG files because in a meeting it is common to use blackboards etc. for hand-written sheets. Images of these sheets should usually eventually be projected to summarize the results.

lesteenman:
This is an initial attempt to incorporate support for image and video files in OpenSlides. It seems to work on my side, but testing by others is encouraged.

Fixes #2107.

### DIFF
--- a/openslides/core/static/templates/projector.html
+++ b/openslides/core/static/templates/projector.html
@@ -40,6 +40,12 @@
       margin-top: {{scroll}}px !important;
       font-size: {{scale}}%;
     }
+    .mediascrollcontent {
+      margin-top: {{scroll/2}}em !important;
+      -webkit-transform: scale({{scale/100}});
+      -ms-transform: scale({{scale/100}});
+      transform: scale({{scale/100}});
+    }
   </style>
   <div ng-repeat="element in elements | orderBy:'index'">
     <div ng-include="element.template"></div>
--- a/openslides/mediafiles/static/js/mediafiles/projector.js
+++ b/openslides/mediafiles/static/js/mediafiles/projector.js
@@ -19,8 +19,22 @@ angular.module('OpenSlidesApp.mediafiles.projector', ['OpenSlidesApp.mediafiles'
     function($scope, Mediafile) {
         // load mediafile object
         var mediafile = Mediafile.get($scope.element.id);
-        $scope.pdfName = mediafile.title;
-        $scope.pdfUrl = mediafile.mediafileUrl;
+        $scope.mediafile = mediafile;
+
+        // Allow the elements to render properly
+        setTimeout(function() {
+            if ($scope.mediafile.is_pdf) {
+                $scope.pdfName = mediafile.title;
+                $scope.pdfUrl = mediafile.mediafileUrl;
+            } else if ($scope.mediafile.is_video) {
+                var player = angular.element.find('#video-player')[0];
+                if ($scope.element.playing) {
+                    player.play();
+                } else {
+                    player.pause();
+                }
+            }
+        }, 0);
     }
 ]);
 
--- a/openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
+++ b/openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
@@ -26,7 +26,7 @@
       <div class="col-md-12">
         <div ng-repeat="presentedMediafile in presentedMediafiles">
           <h3>{{ getTitle(presentedMediafile) }}</h3>
-          <nav ng-class="getNavStyle(scroll)" class="form-inline">
+          <nav ng-show="getType(presentedMediafile) === 'pdf'" ng-class="getNavStyle(scroll)" class="form-inline">
             <div class="btn-group">
               <button class="btn btn-default" ng-click="mediafileGoToPage(presentedMediafile.page - 1)"
                   ng-class="{ 'disabled': (presentedMediafile.page - 1) < 1 }"
@@ -63,9 +63,42 @@
               </button>
             </div>
           </nav>
+          <nav ng-show="getType(presentedMediafile) === 'image'" ng-class="getNavStyle(scroll)" class="form-inline">
+            <div class="btn-group">
+              <button class="btn btn-default" ng-click="mediafileToggleFullscreen()" title="{{ 'Toggle fullscreen' | translate }}"
+                  ng-class="presentedMediafile.fullscreen ? 'btn-primary' : 'btn-default'">
+                <i class="fa fa-arrows-alt"></i>
+              </button>
+            </div>
+            <div class="btn-group">
+              <button class="btn btn-default" ng-click="mediafileRotate()" title="{{ 'Rotate clockwise' | translate }}">
+                <i class="fa fa-repeat"></i>
+              </button>
+            </div>
+          </nav>
+          <nav ng-show="getType(presentedMediafile) === 'video'" ng-class="getNavStyle(scroll)" class="form-inline">
+            <div class="btn-group">
+              <button class="btn btn-default" ng-click="mediafileToggleFullscreen()" title="{{ 'Toggle fullscreen' | translate }}"
+                  ng-class="presentedMediafile.fullscreen ? 'btn-primary' : 'btn-default'">
+                <i class="fa fa-arrows-alt"></i>
+              </button>
+            </div>
+            <div class="btn-group">
+              <button class="btn btn-default" ng-click="setPlaying(false)" title="{{ 'Stop' | translate }}"
+                  ng-class="presentedMediafile.playing ? 'btn-default' : 'btn-primary'">
+                <i class="fa fa-stop"></i>
+              </button>
+            </div>
+            <div class="btn-group">
+              <button class="btn btn-default" ng-click="setPlaying(true)" title="{{ 'Play' | translate }}"
+                  ng-class="presentedMediafile.playing ? 'btn-primary' : 'btn-default'">
+                <i class="fa fa-play"></i>
+              </button>
+            </div>
+          </nav>
         </div>
         <div ng-show="!presentedMediafiles.length" class="spacer">
-          <i translate>No PDF file projected.</i>
+          <i translate>No media file projected.</i>
         </div>
       </div>
     </div>
@@ -163,9 +196,9 @@
         <td ng-show="!isDeleteMode"
             os-perms="core.can_manage_projector">
           <a class="btn btn-default btn-sm"
-              ng-if="mediafile.mediafile.type == 'application/pdf'"
+              ng-if="mediafile.is_presentable"
               ng-class="{ 'btn-primary': mediafile.isProjected() }"
-              ng-click="showPdf(mediafile)"
+              ng-click="showMediafile(mediafile)"
               title="{{ 'Project mediafile' | translate }}">
             <i class="fa fa-video-camera"></i>
           </a>
--- a/openslides/mediafiles/static/templates/mediafiles/slide_mediafile.html
+++ b/openslides/mediafiles/static/templates/mediafiles/slide_mediafile.html
@@ -1,6 +1,18 @@
-<div ng-controller="SlideMediafileCtrl" class="content fullscreen">
-    <ng-pdf template-url="/static/templates/mediafiles/slide_mediafile_partial.html"
-        ng-attr-scale="{{ element.scale }}"
-        ng-attr-page="{{ element.page }}">
+<div ng-controller="SlideMediafileCtrl" class="content" ng-class="{'fullscreen': element.fullscreen, 'video-container': element.is_video}">
+    <!-- PDF -->
+    <ng-pdf ng-if="mediafile.is_pdf" template-url="/static/templates/mediafiles/slide_mediafile_partial.html"
+         ng-attr-scale="{{ element.scale }}"
+         ng-attr-page="{{ element.page }}">
     </ng-pdf>
+
+    <!-- Image -->
+    <img ng-if="mediafile.is_image && !element.fullscreen" class='projector-image rotate{{element.rotate}}'
+        ng-src='{{mediafile.mediafileUrl}}'></img>
+    <div ng-if="mediafile.is_image && element.fullscreen" class='projector-image rotate{{element.rotate}}'
+        style='background-image: url("{{mediafile.mediafileUrl}}")'></div>
+
+    <!-- Video -->
+    <div ng-if="mediafile.is_video" class='video-container'>
+        <video id='video-player' ng-src="{{mediafile.mediafileUrl}}"></video>
+    </div>
 </div>
--- a/openslides/core/static/css/projector.css
+++ b/openslides/core/static/css/projector.css
@@ -430,3 +430,32 @@ tr.elected td {
 .motion-text.line-numbers-none .os-line-number {
     visibility: hidden;
 }
+
+/*** Video and Image projection ***/
+img.projector-image {
+    width: 100%;
+}
+
+div.projector-image {
+    width: 100%;
+    height: 100%;
+    background-size: contain;
+    background-repeat: no-repeat;
+    background-position: 50% 50%;
+}
+
+.video-container {
+    width: 100%;
+    height: 100%;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.video-container > * {
+    flex: 1 1 auto;
+    max-width: 100%;
+    max-height: 100%;
+    width: 100%;
+    height: 100%;
+}
--- a/openslides/mediafiles/static/js/mediafiles/base.js
+++ b/openslides/mediafiles/static/js/mediafiles/base.js
@@ -38,13 +38,22 @@ angular.module('OpenSlidesApp.mediafiles', [])
                 },
             },
             computed: {
-                is_presentable: ['filetype', function (filetype) {
-                    var PRESENTABLE_FILE_TYPES = ['application/pdf'];
-                    return _.contains(PRESENTABLE_FILE_TYPES, filetype);
+                is_pdf: ['filetype', function (filetype) {
+                    var PDF_FILE_TYPES = ['application/pdf'];
+                    return _.contains(PDF_FILE_TYPES, filetype);
                 }],
                 is_image: ['filetype', function (filetype) {
                     var IMAGE_FILE_TYPES = ['image/png', 'image/jpeg', 'image/gif'];
                     return _.contains(IMAGE_FILE_TYPES, filetype);
+                }],
+                is_video: ['filetype', function (filetype) {
+                    var VIDEO_FILE_TYPES = [ 'video/quicktime', 'video/mp4', 'video/webm',
+                        'video/ogg', 'video/x-flv', 'application/x-mpegURL', 'video/MP2T',
+                        'video/3gpp', 'video/x-msvideo', 'video/x-ms-wmv', 'video/x-matroska' ];
+                    return _.contains(VIDEO_FILE_TYPES, filetype);
+                }],
+                is_presentable: ['is_pdf', 'is_image', 'is_video', function (is_pdf, is_image, is_video) {
+                    return is_pdf || is_image || is_video;
                 }],
                 mediafileUrl: [function () {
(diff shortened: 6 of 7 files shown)


openslides/core/static/templates/projector.html
```diff
@@ -37,6 +37,12 @@
       margin-top: {{scroll}}em !important;
       font-size: {{scale}}%;
     }
+    .mediascrollcontent {
```
emanuelschuetze: This css class is not used in your code.

emanuelschuetze:
I tested your branch. It's really great that also images can be projected now! Very nice! Also videos works for me.

Some minor observations (hope you can fix it):
* It's not possible to scroll projected images (tested with png/jpg) in non-fullscreen mode. For big images this would be helpful and should be possible.
* A playing video (tested with mp4) always restarts if you modify the projector (e.g. switch video to fullscreen, zoom/scroll, add/project a countdown or message)
* Have you tested to play a long video (e.g. > 15min)? I am not sure if the auto update makes problem (e.g. the video could be restarts after x min). I observe it if I project a pdf file. After ca. 10min the projector flickers... maybe this has effect on a playing video. You should test this again.

Rest looks nice. Good work! @normanjaeckel should still look a the code.

emanuelschuetze:
@lesteenman One wish: Please add the fullscreen mode (used by images and videos) also for pdf files. This would be more consistently.

lesteenman:
Thank you all for the feedback. After using this feature in Practice at a large conference 2 weeks back, I've found some more issues myself. I will be looking into them, but not within a month due to deadlines for finishing my study. 

If someone else feels like picking it up as well, feel welcome, although I fully intend to do so after finishing my studies.

emanuelschuetze:
@lesteenman Ok, we are looking forward to hearing from you with an updated pull request.

emanuelschuetze:
@lesteenman when you can finish with this pr?

lesteenman:
Probably in July, at the soonest. I'm finishing my thesis at the moment.

I would have time for small fixes, but there are some larger issues with this PR I want to address.

emanuelschuetze:
@lesteenman Any news here?

lesteenman:
Not yet, I'm afraid. I'm having issues really getting back into development. If anyone else feels like picking it up, feel very welcome.

lesteenman:
@emanuelschuetze:

> A playing video (tested with mp4) always restarts if you modify the projector (e.g. switch video to fullscreen, zoom/scroll, add/project a countdown or message)

Switching to fullscreen indeed resets. I'm not sure if this can be prevented, however, as you would need to have quite good time synchronization between server and client... And it would still cause jumps in current position if you force the time as kept by the server. Since clients will first buffer before starting to play, you don't even have control over all clients playing simultaneously.

Zoom/Scroll on a page should just not be allowed for video files.

Changing countdowns or messages is a legitimate issue, however. Do you have any idea on how that might be fixed? It seems that this would need to be quite a fundamental change to how the slides are rendered, as the main elements should not be re-rendered when one of the other elements (e.g. timers) are changed. I'm not sure how to approach this at this point.

lesteenman:
Django error does not seem to be caused by me:

> ............................................../home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/django/template/utils.py:37: RemovedInDjango110Warning: You haven't defined a TEMPLATES setting. You must do so before upgrading to Django 1.10. Otherwise Django will be unable to load templates.

emanuelschuetze:
@lesteenman The travis error is caused by a JS syntax error found by jshint:
`openslides/mediafiles/static/js/mediafiles/site.js: line 234, col 10, Missing semicolon.`

emanuelschuetze:
I'll test your PR again and talk with @normanjaeckel about open (projector) issues.

emanuelschuetze:
@lesteenman I retested your branch.
* Can you find a way to use a "real" fullscreen for videos (my test video is placed on top of black page in full width; it would be nice if you can center it also vertically).
* stll open is the issue "replay after switch to fullscreen" - maybe a minor issue which can split out from this pull request @normanjaeckel ?
* please change the order of the play/stop buttons of video controls - use the same order like in countdowns: Stop - play/pause
* Can you implement a pause function?
* image: scrolling (up down) should be used by the global projector scroll buttons (see pdf presentation), maybe also the zoom buttons

Please ping me here if I can retest something. Thanks in advance! I want to get these media controls for 2.1 ;-)

lesteenman:
> Can you find a way to use a "real" fullscreen for videos (my test video is placed on top of black page in full width; it would be nice if you can center it also vertically).

This is just styling, that should be possible.

> stll open is the issue "replay after switch to fullscreen" - maybe a minor issue which can split out from this pull request @normanjaeckel ?

This is a bit more than a minor issue, as this would (I think) require time syncing between clients. Or the clients themselves have to keep track of it between page redraws. Either way, I'd split it from this ticket, as it's still very usable as long as you switch to fullscreen immediately if you want to.

> Can you implement a pause function?

Can do, but I feel it might work weirdly. Since videos will only start playing once they have been buffered, all clients watching will pause at a different time. That's why I kept controls like that (also forward/backward) from this patch for now.

> image: scrolling (up down) should be used by the global projector scroll buttons (see pdf presentation), maybe also the zoom buttons

In that case, it's probably a different feature request, right?

lesteenman:
> Can you find a way to use a "real" fullscreen for videos (my test video is placed on top of black page in full width; it would be nice if you can center it also vertically).

Would an implementation that works on all modern browsers with the current situation as a graceful fallback be acceptable for you? (Flexbox)

emanuelschuetze:
Yes.

lesteenman:
> Can you implement a pause function?

Just tried again, and I remember what the issue was: The whole projector screen refreshes when a mediafile command is sent, which causes the video position to reset. Any ideas for that?

emanuelschuetze:
> Just tried again, and I remember what the issue was: The whole projector screen refreshes when a 
> mediafile command is sent, which causes the video position to reset. Any ideas for that?

@lesteenman Maybe you can save the timestamp of the current pause position to projector element?

@normanjaeckel Any more ideas how to handle videos on projector?

emanuelschuetze:
@lesteenman I retested your branch:
* Please change play/stop video control buttons (see countdown control buttons).
* My small mp4 video is now centered in fullscreen but not scaled to full width (or height)...

lesteenman:
> @lesteenman Maybe you can save the timestamp of the current pause position to projector element?

Will this work? I thought the page was refreshed before the command got in, and only the actual page element knows at which position it is.

> Please change play/stop video control buttons (see countdown control buttons).

I have those in the patch for play/pause.

> My small mp4 video is now centered in fullscreen but not scaled to full width (or height)...

Very good point, I forgot about that bit... Had not tested with smaller videos. Will send in a patch asap.

emanuelschuetze:
> Will this work? I thought the page was refreshed before the command got in, and only the actual 
> page element knows at which position it is.

My idea: If you click on pause button send a post request to update the projector element (use a uuid of the video element and save the pause position). See coutdown stop function:
https://github.com/OpenSlides/OpenSlides/blob/master/openslides/core/static/js/core/site.js#L1260
If pause is set in projector element all other clients have to stop the video on given time stamp.

lesteenman:
> If pause is set in projector element all other clients have to stop the video on given time stamp.

That's actually quite an excellent idea. However, I think it would cause confusion if the 2 buttons, which are quite close to each other (I suppose you mean the one in the expandable dialog with the projector preview?), cause different actions (pause at my point / pause at your current point).

> My idea: If you click on pause button send a post request to update the projector element (use a uuid of the video element and save the pause position). See coutdown stop function:

Will check that out later this week.

emanuelschuetze:
> I suppose you mean the one in the expandable dialog with the projector preview?

Yes, I mean you the countdown controls buttons start/stop (if you expand the projector sidebar and click on "Countdowns", create one if no available).

> I think it would cause confusion if the 2 buttons, which are quite close to each other
> , cause different actions (pause at my point / pause at your current point).

I don't understand "pause at my point / pause at your current point"...
The admin (user with projector_can_manage permission) has a play button which can click again to pause the video. Now all norman users (with projector_can_see permission) see the video paused on same position, get it from projector element). There is no special play/pause/stop buttons for a normal user. So it's not confused for me...

lesteenman:
> The admin (user with projector_can_manage permission) has a play button which can click again to pause the video. Now all normal users (with projector_can_see permission) see the video paused on same position, get it from projector element). There is no special play/pause/stop buttons for a normal user. So it's not confused for me...

So the projector element (which is serverside, I think?) has to keep track of their own time track? In that case, as all clients will only start playing once they're buffered, which means they will always be behind on what the server expects, clients will miss parts of the video.

Thus, I assumed you meant it would take the time of the projector preview that you press 'pause' on as the 'master time'. This would still cause issues, however...

The point is: I don't think it's possible to really sync all clients, unless we (can) listen for the bufferready event, and then just fast-forward the client video to what the server thinks the position should be, which means you will always miss the first few seconds of a video (assuming a reasonably fast network).

normanjaeckel:
> @normanjaeckel Any more ideas how to handle videos on projector? 

I do not know how I can help here. I think the idea is that you save a timestamp in the projector config (via update_elements). The problem is that the controlling client is on angular site module and has no open projector controller. Only within this controller you may know how much time the video was running since start. You may only use some extra buttons inside the projector controller. The admin would have to open an extra projector window to control at least the pause feature of the video. Another approach would be that the admin can only stop and restart the video and has an input field additionally where he could enter xy seconds/minutes. The projector should skip this value after restart.

lesteenman:
I think that the current approach would do for now, as implementing a feature that feels broken at start, rather than missing some features, feels wrong to me. Hopefully, the community will have enough feedback as to how this should be done afterwards.

openslides/mediafiles/models.py
```diff
@@ -23,6 +23,9 @@ class Mediafile(RESTModelMixin, models.Model):
     title = models.CharField(max_length=255, unique=True)
     """A string representing the title of the file."""
 
+    fullscreen = False
```
normanjaeckel: These attributes are not used in your code, are they?

lesteenman: Don't think so. This is probably a remainder of an earlier version...

emanuelschuetze:
@lesteenman Can you finish this pr in the next days? Some remarks are still open. We want to merge your feature (without any new logic for synchronize videos, stop on special timestamps could be in a new issue for later work) but therefor we have to fix all remarks.

emanuelschuetze:
And please rebase on top of master. There is a conflict in projector.js. 
Let me know if you need help (or if you have no time to do it). Thanks!

lesteenman:
I'll get to it later today, missed the open remarks for a bit. Am I right that these are:

- the 'Fullscreen' attribute not being used
- Order of the stop/play controls
- `image: scrolling (up down) should be used by the global projector scroll buttons (see pdf presentation), maybe also the zoom buttons`

lesteenman:
I think I fixed most of the points. However, it seems that the projector controls don't work with PDF files either. I removed my custom handling of the zoom/scroll controls, and as they're using the same name, the scrolling/zooming using the projector controls should work once it gets fixed for PDF.

lesteenman:
Same here: Can this branch be merged as it is now? If so, I'll merge the patches.

openslides/core/static/templates/projector.html
emanuelschuetze: ping, please remove this unused class!

openslides/mediafiles/static/templates/mediafiles/slide_mediafile.html
```diff
@@ -1,6 +1,18 @@
-<div ng-controller="SlideMediafileCtrl" class="content fullscreen">
-    <ng-pdf template-url="/static/templates/mediafiles/slide_mediafile_partial.html"
-        ng-attr-scale="{{ element.scale }}"
-        ng-attr-page="{{ element.page }}">
+<link rel="stylesheet" href="static/css/mediafiles/projector.css" />
```
emanuelschuetze: Remove this line and add all your mediafile css classed to core/static/css/projector.css (currently we use only one css file for projector, we will split it later)

openslides/mediafiles/static/css/mediafiles/projector.css
```diff
@@ -0,0 +1,27 @@
...
+	justify-content: center;
+}
+
+.video-container > * {
+	flex: 1 1 auto;
+	max-width: 100%;
+	max-height: 100%;
+	width: 100%;
+	height: 100%;
+}
```
emanuelschuetze: remove this file and move content to core/static/css/projector.css. Note: Use only 4 spaces indentation.

openslides/mediafiles/static/js/mediafiles/projector.js
```diff
@@ -19,8 +19,22 @@ angular.module('OpenSlidesApp.mediafiles.projector', ['OpenSlidesApp.mediafiles'
     function($scope, Mediafile) {
         // load mediafile object
         var mediafile = Mediafile.get($scope.element.id);
-        $scope.pdfName = mediafile.title;
-        $scope.pdfUrl = mediafile.mediafileUrl;
+		$scope.mediafile = mediafile;
```
emanuelschuetze: Please fix the indentation here and in the following lines!

openslides/mediafiles/static/js/mediafiles/projector.js
```diff
@@ -19,8 +19,22 @@ angular.module('OpenSlidesApp.mediafiles.projector', ['OpenSlidesApp.mediafiles'
...
+
+		// Allow the elements to render properly
+		setTimeout(function() {
+			if ($scope.mediafile.is_pdf) {
+				$scope.pdfName = mediafile.title;
```
emanuelschuetze: Only 4 spaces for indentation please.

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
```diff
@@ -63,9 +63,63 @@ <h1 translate>Files</h1>
               </button>
             </div>
           </nav>
+          <nav ng-show="getType(presentedMediafile) === 'image'" ng-class="getNavStyle(scroll)" class="form-inline">
+            <div class="btn-group">
+              <button class="btn btn-default" ng-click="mediafileToggleFullscreen()" title="{{ 'Toggle fullscreen' | translate }}"
+				ng-class="presentedMediafile.fullscreen ? 'btn-primary' : 'btn-default'">
```
emanuelschuetze: Fix indentation here (use 2x 4 spaces for this line, beginning from `<button ...`.

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
```diff
@@ -63,9 +63,63 @@ <h1 translate>Files</h1>
...
+                <i class="fa fa-arrows-alt"></i>
+              </button>
+            </div>
+            <div class="btn-group">
+              <button class="btn btn-default" ng-click="mediafileRotate()" title="{{ 'Rotate clockwise' | translate }}">
+                <i class="fa fa-repeat"></i>
+              </button>
+            </div>
+            <div class="btn-group">
```
emanuelschuetze: This zooming and scrolling buttons does not work for projected images. For me: you can remove these buttons. Use just the fullscreen and rotate button for images. This is enough.

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
```diff
@@ -63,9 +63,63 @@ <h1 translate>Files</h1>
...
+              <input type="number" class="form-control" style="width: 80px" ng-model="presentedMediafile.scroll" ng-change="mediafileScroll(presentedMediafile.scroll)"></input>
+              <button class="btn btn-default" ng-click="mediafileScroll(presentedMediafile.scroll + 1)" title="{{ 'Scroll down' | translate }}">
+                <i class="fa fa-arrow-down"></i>
+              </button>
+            </div>
+          </nav>
+          <nav ng-show="getType(presentedMediafile) === 'video'" ng-class="getNavStyle(scroll)" class="form-inline">
+            <div class="btn-group">
+              <button class="btn btn-default" ng-click="mediafileToggleFullscreen()" title="{{ 'Toggle fullscreen' | translate }}"
+				ng-class="presentedMediafile.fullscreen ? 'btn-primary' : 'btn-default'">
```
emanuelschuetze: fix indendation

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
```diff
@@ -63,9 +63,63 @@ <h1 translate>Files</h1>
...
+                <i class="fa fa-arrows-alt"></i>
+              </button>
+            </div>
+            <div class="btn-group">
+              <button class="btn btn-default" ng-click="setPlaying(false)" title="{{ 'Stop' | translate }}"
+				ng-class="presentedMediafile.playing ? 'btn-default' : 'btn-primary'">
```
emanuelschuetze: here the same

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
```diff
@@ -63,9 +63,63 @@ <h1 translate>Files</h1>
...
+                <i class="fa fa-stop"></i>
+              </button>
+            </div>
+            <div class="btn-group">
+              <button class="btn btn-default" ng-click="setPlaying(true)" title="{{ 'Play' | translate }}"
+				ng-class="presentedMediafile.playing ? 'btn-primary' : 'btn-default'">
```
emanuelschuetze: here the same

openslides/mediafiles/static/templates/mediafiles/slide_mediafile.html
```diff
@@ -1,6 +1,18 @@
...
+<div ng-controller="SlideMediafileCtrl" class="content" ng-class="{'fullscreen': element.fullscreen, 'video-container': element.is_video}">
+    <!-- PDF -->
+    <ng-pdf ng-if="mediafile.is_pdf" template-url="/static/templates/mediafiles/slide_mediafile_partial.html"
+         ng-attr-scale="{{ element.scale }}"
+         ng-attr-page="{{ element.page }}">
     </ng-pdf>
+
+    <!-- Image -->
+    <img ng-if="mediafile.is_image && !element.fullscreen" class='projector-image rotate{{element.rotate}}' ng-src='{{mediafile.mediafileUrl}}'></img>
+    <div ng-if="mediafile.is_image && element.fullscreen" class='projector-image rotate{{element.rotate}}' style='background-image: url("{{mediafile.mediafileUrl}}")'></div>
```
emanuelschuetze: please add line breaks for these two long lines.

emanuelschuetze:
@lesteenman I added some remarks - mostly coding style issues. Please fix this and rebase your commits to one commit. Then we will do a last review and merge your pr. - Works really nice. I want it for OpenSlides 2.1 ;-).

Just for note: For PDF presenation you can use the scroll buttons (up/down) of the projector controls (in sidebar). The zoom buttons from the sidebare does not yet works, right. Maybe you want to fix/improve this later?

emanuelschuetze:
Please ping me if you finished your work.

emanuelschuetze:
@lesteenman ping, when you can finish your work here (next week 2016-09-30) we will reorder all JS files. Then you get a lot of merge conflicts. So let us merge this PR in the next days!

lesteenman:
@emanuelschuetze busy few days, I'll probably finish it tonight.

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
lesteenman: Same with PDFs at the moment, I believe. This seems to be the same bug. I'll remove the buttons for now though.

lesteenman:
Should be done. The indentation issues are quite annoying as my editor doesn't properly show the difference it seems, but I hope I got them right this time.

emanuelschuetze:
Ok, fine. Looks good now for me. 
@normanjaeckel Ready for your review and merge.

PR #2128, opened 2016-05-14
Add a 'private' flag to mediafiles.

### LINKED ISSUE #2116: Allow hidden/private mediafiles
In our case, we want to upload mediafiles beforehand, but not allow anonymous users to view/play these (yet). A checkmark 'public' or similar, with specific permissions related to it, would allow this.

lesteenman:
Only users with the 'mediafiles.can_see_private' permission can create
and view these private mediafiles.

Fixes #2116.


lesteenman:
This is a check that fails in a generated migration file; Is that intended?

openslides/mediafiles/views.py
```diff
@@ -15,6 +15,13 @@ class MediafileViewSet(ModelViewSet):
     access_permissions = MediafileAccessPermissions()
     queryset = Mediafile.objects.all()
 
+    def get_queryset(self):
```
normanjaeckel: I think you also have to add some logic to the file `access_permissions.py` and use the `check_object_permissions()` method. See agenda access_permissions and views for example: There we have the same case. Some people may see hidden agenda items, others don't. We need this different logic to ensure that only authorized users get this data even via our websocket autoupdate process.

openslides/mediafiles/views.py
```diff
@@ -15,6 +15,13 @@ class MediafileViewSet(ModelViewSet):
...
+        user = self.request.user
+        if user.has_perm('mediafiles.can_see_private'):
+            return Mediafile.objects.all()
```
normanjaeckel: Please use something like `queryset = super().get_queryset()` so that line 16 is still in effect.

openslides/users/signals.py
```diff
@@ -118,6 +118,7 @@ def create_builtin_groups_and_admin(**kwargs):
         'core.can_use_chat',
         'mediafiles.can_manage',
         'mediafiles.can_see',
+        'mediafiles.can_see_private',
```
normanjaeckel: This permission should also be added to the staff group, see line 178.

openslides/mediafiles/models.py
```diff
@@ -41,6 +44,7 @@ class Meta:
         default_permissions = ()
         permissions = (
             ('can_see', 'Can see the list of files'),
+            ('can_see_private', 'Can see privately listed files'),
```
emanuelschuetze: Please add the permission string to users/site.js to make the string translateable:
https://github.com/OpenSlides/OpenSlides/blob/master/openslides/users/static/js/users/site.js#L986

lesteenman:
Thank you all for the feedback. After using this feature in Practice at a large conference 2 weeks back, I've found some more issues myself. I will be looking into them, but not within a month due to deadlines for finishing my study. 

If someone else feels like picking it up as well, feel welcome, although I fully intend to do so after finishing my studies.

lesteenman:
Hoping you are still watching this: I've picked up this issue again, but I'm running into some issues as I don't fully grasp how everything fits together.

The main issue at the moment for this merge request: Creating a new mediafile does not pick up on the given private setting. Editing it later and changing the 'private' field does work, however. Any thoughts? From what I could see from the `create` function, it should just work as-is.

normanjaeckel:
May be you have to add some logic to the MediafileForm factory (site.js line 298). The uploadFile method is called inside the MediafileCreateCtrl.

lesteenman:
It seems like it fails on installing versions of requirements...

emanuelschuetze:
@lesteenman Please rebase on top of master. I fixed the travis error caused by outdated requirements (see #2232).

For your problem with creating new mediafile: Maybe you run into this problem #2223?

emanuelschuetze:
@lesteenman please ping me here  if your pull request is ready for review.

lesteenman:
I just realize I completely forgot to ping you. I think it is ready for review.

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
```diff
@@ -177,6 +177,7 @@ <h1 translate>Files</h1>
           <strong><a ng-href="{{ mediafile.mediafileUrl }}" target="_blank">{{ mediafile.title_or_filename }}</a></strong>
           <br><small><i class="fa fa-file"></i> {{ mediafile.filetype }}</small>
           <br><small><i class="fa fa-database"></i> {{ mediafile.filesize }}</small>
+          <span ng-if="mediafile.private"><br><small><i class="fa fa-lock"></i> Private</small></span>
```
emanuelschuetze: Please mark string for translation: `<translate>Private</translate>`

openslides/mediafiles/models.py
emanuelschuetze: Suggestion for shorter translation string: `Can see private files`

openslides/users/static/js/users/site.js
```diff
@@ -989,6 +989,7 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
         gettext('Can see the list of files');
         gettext('Can upload files');
         gettext('Can manage files');
+        gettext('Can see privately listed files');
```
emanuelschuetze: Suggestion for shorter translation string: `Can see private files`

emanuelschuetze:
@lesteenman Some generally remarks:
* The new permission does not work for the list view: A user without this new permission should NOT be able to see private files! But all users can see private marked files... (e.g. guests)
* Please add `(Fixes #2116)` in commit message to closed the associated issue automatically after merge.

emanuelschuetze:
I retested: 
Now the new permission works well after reload: A guest can not see private files. Good.
But one use case does not work very good:
* upload a file (without private flag) -> a guest user can see the file
* now the admin edit the file and switch it to 'private' -> the guest user list view gets a "private" string, but the table row is still visible. Expected: row should be removed/invisible.

And please rebase your branch on top of master (there is a fix in master for handling static media files).
Pleese ping me here if I can retest your PR. Thanks!

lesteenman:
> now the admin edit the file and switch it to 'private' -> the guest user list view gets a "private" string, but the table row is still visible. Expected: row should be removed/invisible.

I was not aware something like this was (easily?) possible. I'll check it out!

emanuelschuetze:
You need autoupdate - like in all other list views. E.g. if you delete a motion - in an other list view the motion is gone. Same should be work with files which set to private. Aim: user should not reload manually the list view page.

lesteenman:
Yes, working on it now. I assume that it should pretend the file is deleted when it is changed to private. If that's not possible, deleting it from the list of mediafiles locally might work... But that seems iffy.

emanuelschuetze:
Just try to use a private filter in ng-repeat and something with os-perms="mediafiles.can_see_private", as e.g.:
``
<tr os-perms="!mediafiles.can_see_private" ng-repeat="mediafile in mediafilesFiltered = (mediafiles | ... | filter: {private: false})" ...
``
Maybe that helps...

lesteenman:
That should work as long as the user can't still access the file once it's private; Then it's safe enough to just filter it out clientside. However, I just noticed that you can still access the URL directly when it's private... Trying to fix that first.

lesteenman:
Ok, it seems like this is not that easy to implement at the moment, as mediafiles are served directly. I'll have to figure out how to check permissions on individual files. For now, I will filter it from the list, and send in a new PR.

lesteenman:
> Just try to use a private filter in ng-repeat and something with os-perms="mediafiles.can_see_private", as e.g.:

Do I understand correctly that this would require duplication of everything within that TR, as you'd need one for those who do have the permission, and one for those who don't?

lesteenman:
@emanuelschuetze, please check it again. I think this implementation is a good improvement on the current situation already, although being able to deny direct access to private mediafiles would be a future improvement.

openslides/mediafiles/views.py
normanjaeckel: Ping.

lesteenman: Forgot about that bit, added it. However, I'm getting errors when logged in, which aren't caused by my changes (checking out upstream/master causes the same issues). Are these known issues?

What I did was rebase on upstream/master, and run `./manage.py migrate`.

capture

normanjaeckel: I don't know but it loosk like that you do not have enough permissions. We changed the permissions system some weeks ago. A suitable migration file is still missing.

Try to remove your personal data directory and then run python manage.py start to create a new settings file and database file.

openslides/mediafiles/models.py
emanuelschuetze: This renaming to "Can see private files" is still open.

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
```diff
@@ -177,6 +177,7 @@ <h1 translate>Files</h1>
           <strong><a ng-href="{{ mediafile.mediafileUrl }}" target="_blank">{{ mediafile.title_or_filename }}</a></strong>
           <br><small><i class="fa fa-file"></i> {{ mediafile.filetype }}</small>
           <br><small><i class="fa fa-database"></i> {{ mediafile.filesize }}</small>
+          <span ng-if="mediafile.private"><br><small><i class="fa fa-lock"></i><translate>Private</translate></small></span>
```
emanuelschuetze: Add a whitespace between icon and `<translate>` tag.

emanuelschuetze:
I retested. Looks good. Thanks! I found a minor bug only. How to reproduce it: 
* Upload a new (public) file.
* Look with a guest user (without permission to see private files) to mediafile list view (see "1/1 files").
* Admin changed file to private.
* Guest user does not see private file in list but "0/1 files" is still shown. After reload it shows correctly "0/0 files".
Would be great if you can fix this.

If we cannot prevent the direct access to private files we should at least add a "important note" as help text to the private checkbox in mediafile form (see in users or motion form how to do it). Otherwise the admin could get a unpleasant surprise if he found out that files are not protected for direct access - especially if OpenSlides is used online.

@normanjaeckel  Please review again.

openslides/mediafiles/access_permissions.py
```diff
@@ -18,3 +18,16 @@ def get_serializer_class(self, user=None):
         from .serializers import MediafileSerializer
 
         return MediafileSerializer
+
+    # def get_restricted_data(self, full_data, user):
```
normanjaeckel: Do you want to use this code? How do you ensure, that only authorized users get private file data via autoupdate. Please have a look at #2357. This is under heavy development. :-(

openslides/mediafiles/static/js/mediafiles/site.js
```diff
@@ -331,12 +331,27 @@ angular.module('OpenSlidesApp.mediafiles.site', ['ngFileUpload', 'OpenSlidesApp.
...
+
+.filter('privateFilter', [
+    '$filter',
+    'operator',
+    function ($filter, operator) {
+        return function (array) {
+            if (operator.hasPerms('mediafiles.can_see_private')) {
+                return array;
+            }
+            return Array.prototype.filter.call(array, function (item) {
```
normanjaeckel: Just for me to learn something: Why not just `array.filter(function (item) {...})`?

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
```diff
@@ -155,8 +155,8 @@ <h1 translate>Files</h1>
           </i>
       </tr>
     <tbody>
-      <tr ng-repeat="mediafile in mediafilesFiltered = (mediafiles | osFilter: filter.search : getFilterString |
-          filter: {filetype: filter.showPDFs} | orderBy: sortColumn:reverse)"
+      <tr ng-repeat="mediafile in mediafilesFiltered = (mediafiles | privateFilter | osFilter: filter.search : getFilterString |
+	  filter: {filetype: filter.showPDFs} | orderBy: sortColumn:reverse )"
```
normanjaeckel: Unnecessary whitespace and the end.

Why did you change indention?

openslides/mediafiles/static/js/mediafiles/site.js
lesteenman: I think that should actually work in this case. I believe I took this approach from somewhere else in the codebase or online, and is usually what you want to use if you have array-likes rather than actual arrays. So if the files are always an array, that approach is possible.

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
lesteenman: It appears I finally forgot to change Vim's whitespace settings before pushing :/

openslides/mediafiles/access_permissions.py
normanjaeckel: See #2394 .

openslides/mediafiles/static/js/mediafiles/site.js
normanjaeckel: Ok.

normanjaeckel:
@lesteenman Do you need any help here? We want to refactor many files in about two weeks.

lesteenman:
I'll get to it later today. Am I right that the remaining open issue is:

- The `get_restricted_data` function to prevent others from receiving private files (using something from #2394)

lesteenman:
Not receiving private works, removing files from the list when turned private does not yet...

lesteenman:
@normanjaeckel I just noticed that the same issue occurs with Agenda items that become hidden. Is it safe to assume this is a bug in general?

lesteenman:
Assuming I can, and that these were the remaining issues: Should I merge all commits in this branch myself, or will that be done when it's merged?

openslides/mediafiles/models.py
normanjaeckel: Ping.

openslides/mediafiles/static/templates/mediafiles/mediafile-list.html
normanjaeckel: Ping

normanjaeckel:
Ok. This is a bug in general. We have to think about it later.

Please also change the permission string in the migrations file. Then rebase all commits to one and rebase it also on top of master.

Thanks. :-)

lesteenman:
Done!

emanuelschuetze:
One migration problem:
I can't see the new checkbox "private" if I migrate my database (created with current master). I cleared my full cache and I have already the right permissions...
If I start with a fresh db in this branch all works well (private checkbox is visible).

@lesteenman: Can you verify this? Please test the migration path. We want to make sure that users with an old database can use the new private feature.

lesteenman:
@emanuelschuetze what is the recommended way to clean an active working directory? I seem to get issues when rebasing on master every time with random issues. I run the server with either `./manage.py start` or `./manage.py runserver`. Just deleting `personal_data` didn't do the trick.

emanuelschuetze:
Just remove the `personal_data` directory and restart openslides with `python manage.py start`.

An example how I reproduced  the migration problem:
```
rm -r personal data
git checkout master
npm install
python manage.py start
# add some mediafiles
# stop server
git checkout <your-privateMediafiles-branch>
npm install
python manage.py start  # migrates database
# reload browser, clear cache
# edit a mediafile or create a new mediafile -> no 'private' field available in form
```

Hope this helps you.

lesteenman:
Thank you, that should help.

lesteenman:
Right, what is going wrong is that the default permissions are not migrated, so even admin users can not see private mediafiles (and thus the flag) by default. I'll check to see how this is set up properly in migration.

lesteenman:
@emanuelschuetze I'm trying to add the permission to the Staff group in [migration file 3](https://github.com/lesteenman/OpenSlides/blob/privateMediafiles/openslides/mediafiles/migrations/0003_auto_20160514_1347.py). However, it seems the permission with codename `can_see_private` does not yet exist here.

What is the recommended way to add a default permission in a migration? [The groups migration](https://github.com/lesteenman/OpenSlides/blob/privateMediafiles/openslides/users/migrations/0004_groups.py) has been the only migration file I've found so far which does something with migrations. It seems that in this case, there's an issue where the new permissions are not yet created when the migration runs, however...

normanjaeckel:
At the moment, it is not possible to add new permissions to groups during migration because Django creates the permission just **after** the migration process. Sorry.

lesteenman:
@normanjaeckel that explains it... So then what can be done? The permission needs to be added to staff somewhere.

normanjaeckel:
Add permission to staff in users/signals.py would do the thing for people who start with a fresh db. All others have to read a nice CHANGELOG message (add this!) and use the interface (users -> groups -> nice interface to control the permissions).

emanuelschuetze:
This is also ok for me.

lesteenman:
Ok. In that case, I think the current branch can be merged.

normanjaeckel:
@ostcar This pull request implies small changes in your cache branch. You have to check the mediafile app and delete the get_queryset() method in the viewset.

PR #2169, opened 2016-05-28
Order speaker on projector

### LINKED ISSUE #2138: Reordering speakers has no effect on projector
- create a speakers list
- move the last speaker to the top to be the next speaker

result: this works within OpenSlides but the change is not displayed in the projector or the live preview. One can update the browser window without success.

This is a regression to OpenSlides 1.5.1.

ostcar:
fixes #2138


emanuelschuetze:
Tested successfully. Thanks for the quick fix!

PR #2187, opened 2016-06-07
Fix linebreak on tags (Fixes #2147)

### LINKED ISSUE #2147: linebreaks can occur inside info blocks
The info blocks are not protected against linebreaks. In this example:
info-button-bug
the blocks for the tag "Soziales" is destroyed by a linebreak.

joelmacht:
Changed label class from display:inline to display:inline-block and added myself to AUTHORS


emanuelschuetze:
I tested your branch: Linebreaks looks good now. But we need more space _between_ labels (e.g. margin-left or right). See screenshot:
motion-tags

Issue #2182, opened 2016-06-07
Add new config option for 100 % base of voting results (yes+no without abstain)


emanuelschuetze:
Add to the config field "The 100 % base of a voting result consists of" (in section motions and elections) a new select option:
"Only YES and NO values"

Add a check for this new config option in [motions/base.js](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/motions/static/js/motions/base.js#L86) and [assignments/base.js#L38](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/assignments/static/js/assignments/base.js#L38) and [assignments/base.js#L107](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/assignments/static/js/assignments/base.js#L107).

matakuka:
For motions there is currently no way to alter the settings of "motions_poll_100_percent_base" -- should it mirror the settings of the election method (with invalid/without invalid/ without abstains) ?

normanjaeckel:
What do you mean by
> currently no way to alter the settings of "motions_poll_100_percent_base" 

? Is this a bug? Maybe you have to checkout the current master because we fixed a bug in the config recently.

matakuka:
my bad, I was looking at the wrong line. Works fine

normanjaeckel:
We should discuss the goal of this issue in the different use cases:

# Motions

We only support one voting method: You can fill out six fields per motion poll: yes, no, abstain, votes valid, votes invalid and vots casts.

This issue is about the new possibility that the percent values should base on the sum of yes and no votes instead of the field valid votes or votes cast. No problem here.

# Assignments

At the moment we support three voting methods:

## Config value `votes`

In this voting method you can fill out for one poll one field per candidate and three global fields (votes valid, votes invalid and vots casts).

Here is the question about what this issue is. I do not know and I can not imagine at the moment.

## Config value `yesnoabstain`

In this voting method you can fill out for one poll three fields per candidate (yes/no/abstaion) and three global fields (votes valid, votes invalid and vots casts).

Here is the question about what this issue is. Should the percent base the sum of all yes and no fields of all candidates?

## Config value `yesno`

In this voting method you can fill out for one poll three fields per candidate (yes/no/abstaion) and three global fields (votes valid, votes invalid and vots casts).

Here is the question about what this issue is. Should the percent base the sum of all yes and no fields of all candidates? Is there any difference to the behavior of case 2?

## Config value `auto`

[...]

PR #2189, opened 2016-06-08
adding option "yes/no for each candidate"

matakuka:
New option that differs from Yes/No/Abstain in that there is no "abstain" option when electing a candidate. There might be some redundancy to the yes/no/abstain option, but I tried to keep it as low as possible.

PR #2192, opened 2016-06-09
config "calculate % without abstains" (motions) (Issue #2182)

matakuka:
The 100% base in this option is calculated by (valid votes  - abstains), making it necessary to know the "abstains" beforehand (in case of assignments), or knowing which type of vote 'vote' is (in case of motions).

### DIFF
--- a/openslides/assignments/views.py
+++ b/openslides/assignments/views.py
@@ -510,7 +510,7 @@ class AssignmentPollPDF(PDFView):
                                  'yes': _("Yes"),
                                  'no': _("No"),
                                  'abstain': _("Abstain")},
-                               stylesheet['Ballot_option_circle_YNA']))
+                                stylesheet['Ballot_option_circle_YNA']))
                 else:
                     cell.append(Paragraph(
                         "&nbsp;", stylesheet['Ballot_option_suffix_YNA']))
--- a/openslides/motions/config_variables.py
+++ b/openslides/motions/config_variables.py
@@ -1,4 +1,5 @@
 from django.core.validators import MinValueValidator
+from django.utils.translation import ugettext_lazy
 
 from openslides.core.config import ConfigVariable
 from openslides.poll.models import PERCENT_BASE_CHOICES
@@ -23,6 +24,10 @@ def get_config_variables():
     papers' and 'PDF'. The generator has to be evaluated during app loading
     (see apps.py).
     """
+    percent_base_choices_motion = PERCENT_BASE_CHOICES
+    percent_base_choices_motion += ({
+        'value': "WITHOUT_ABSTAIN",
+        'display_name': ugettext_lazy('Yes and No votes')},)
     # General
     yield ConfigVariable(
         name='motions_workflow',
@@ -148,7 +153,7 @@ def get_config_variables():
         default_value='WITHOUT_INVALID',
         input_type='choice',
         label='The 100 % base of a voting result consists of',
-        choices=PERCENT_BASE_CHOICES,
+        choices=percent_base_choices_motion,
         weight=355,
         group='Motions',
         subgroup='Voting and ballot papers')
--- a/openslides/motions/static/js/motions/base.js
+++ b/openslides/motions/static/js/motions/base.js
@@ -71,7 +71,7 @@ angular.module('OpenSlidesApp.motions', [
             },
             methods: {
                 // returns object with value and percent
-                getVote: function (vote) {
+                getVote: function (vote, type) {
                     if (!this.has_votes) {
                         return;
                     }
@@ -89,13 +89,18 @@ angular.module('OpenSlidesApp.motions', [
                     }
                     // calculate percent value
                     var config = Config.get('motions_poll_100_percent_base').value;
-                    var percentStr, percentNumber;
+                    var percentStr;
+                    var percentNumber = null;
                     if (config == "WITHOUT_INVALID" && this.votesvalid > 0 && vote >= 0) {
                         percentNumber = Math.round(vote * 100 / this.votesvalid * 10) / 10;
                     } else if (config == "WITH_INVALID" && this.votescast > 0 && vote >= 0) {
                         percentNumber = Math.round(vote * 100 / (this.votescast) * 10) / 10;
+                    } else if (config == "WITHOUT_ABSTAIN" && this.votesvalid > 0 && vote >= 0){
+		                if (type == 'yes' || type == 'no') {
+                            percentNumber = Math.round(vote * 100 / (this.yes + this.no) * 10) / 10;
+                        }
                     }
-                    if (percentNumber) {
+                    if (percentNumber !== null) {
                         percentStr = "(" + percentNumber + "%)";
                     }
                     return {
@@ -103,7 +108,7 @@ angular.module('OpenSlidesApp.motions', [
                         'percentStr': percentStr,
                         'percentNumber': percentNumber
                     };
-                },
+                }
             }
         });
     }
--- a/openslides/motions/static/js/motions/site.js
+++ b/openslides/motions/static/js/motions/site.js
@@ -70,14 +70,15 @@ angular.module('OpenSlidesApp.motions.site', ['OpenSlidesApp.motions', 'OpenSlid
                     results = function() {
                         return motion.polls.map(function(poll, index) {
                             var id = index + 1,
-                                yes = poll.yes,
-                                yesRelative = (poll.yes) * 100 / (poll.votescast),
-                                no = poll.no,
-                                noRelative = (poll.no) * 100 / (poll.votescast),
-                                abstain = poll.abstain,
-                                abstainRelative = (poll.abstain) * 100 / (poll.votescast),
+                                yes = poll.yes ? poll.yes : 0, // if no poll.yes is given, set it to 0
+                                yesRelative = poll.getVote(poll.yes, 'yes').percentStr,
+                                no = poll.no ? poll.no : 0,
+                                noRelative = poll.getVote(poll.no, 'no').percentStr,
+                                abstain = poll.abstain ? poll.abstain : 0,
+                                abstainrelativeGet = poll.getVote(poll.abstain, 'abstain').percentStr,
+                                abstainRelative = abstainrelativeGet? abstainrelativeGet : "",
                                 valid = poll.votesvalid,
-                                validRelative = (poll.votesvalid) * 100 / (poll.votescast),
+                                validRelative = poll.getVote(poll.votesvalid, 'votesvalid').percentStr,
                                 number = {
                                     text: id + ".",
                                     width: "5%"
@@ -100,7 +101,7 @@ angular.module('OpenSlidesApp.motions.site', ['OpenSlidesApp.motions', 'OpenSlid
                                     var indexColumn = converter.createElement("text");
                                     var nameColumn = converter.createElement("text", "" + name);
                                     var valueColumn = converter.createElement("text", "" + value);
-                                    var relColumn = converter.createElement("text", "(" + "" + relValue + "%)");
+                                    var relColumn = converter.createElement("text", relValue);
                                     valueColumn.width = "40%";
                                     indexColumn.width = "5%";
                                     valueColumn.width = "5%";
--- a/openslides/motions/static/templates/motions/motion-detail.html
+++ b/openslides/motions/static/templates/motions/motion-detail.html
@@ -151,7 +151,7 @@
                   <tr>
                     <td class="icon">
                       <i class="fa fa-thumbs-up fa-2x"></i>
-                    <td ng-init="voteYes = poll.getVote(poll.yes)">
+                    <td ng-init="voteYes = poll.getVote(poll.yes, 'yes')">
                       <span class="result_label"><translate>Yes</translate>:</span>
                       <span class="result_value">
                         {{ voteYes.value }} {{ voteYes.percentStr }}
@@ -163,7 +163,7 @@
                   <tr>
                     <td class="icon">
                       <i class="fa fa-thumbs-down fa-2x"></i>
-                    <td ng-init="voteNo = poll.getVote(poll.no)">
+                    <td ng-init="voteNo = poll.getVote(poll.no, 'no')">
                       <span class="result_label"><translate>No</translate>:</span>
                       <span class="result_value" >
                         {{ voteNo.value }} {{ voteNo.percentStr }}
@@ -175,7 +175,7 @@
                   <tr>
                     <td class="icon">
                       <strong style="font-size: 26px">&empty;</strong>
-                    <td ng-init="voteAbstain = poll.getVote(poll.abstain)">
+                    <td ng-init="voteAbstain = poll.getVote(poll.abstain, 'abstain')">
                       <span class="result_label"><translate>Abstain</translate>:</span>
                       <span class="result_value">
                         {{ voteAbstain.value }} {{ voteAbstain.percentStr }}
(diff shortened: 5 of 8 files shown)


openslides/motions/static/js/motions/base.js
```diff
@@ -67,7 +67,7 @@ angular.module('OpenSlidesApp.motions', ['OpenSlidesApp.users'])
             },
             methods: {
                 // returns object with value and percent
-                getVote: function (vote) {
+                getVote: function (vote,type) {
```
emanuelschuetze: please add a blank after ",".

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -145,7 +145,7 @@ <h3 ng-if="motion.polls.length > 0" translate>Voting result</h3>
                   <tr>
                     <td class="icon">
                       <i class="fa fa-thumbs-up fa-2x"></i>
-                    <td ng-init="voteYes = poll.getVote(poll.yes)">
+                    <td ng-init="voteYes = poll.getVote(poll.yes,'yes')">
```
emanuelschuetze: add a blank after ","

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -169,7 +169,7 @@ <h3 ng-if="motion.polls.length > 0" translate>Voting result</h3>
                   <tr>
                     <td class="icon">
                       <strong style="font-size: 26px">&empty;</strong>
-                    <td ng-init="voteAbstain = poll.getVote(poll.abstain)">
+                    <td ng-init="voteAbstain = poll.getVote(poll.abstain,'abstain')">
```
emanuelschuetze: add a blank after ","

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -190,7 +190,7 @@ <h3 ng-if="motion.polls.length > 0" translate>Voting result</h3>
                   <tr>
                     <td class="icon">
                       <i class="fa fa-ban fa-lg"></i>
-                    <td ng-init="votesInvalid = poll.getVote(poll.votesinvalid)">
+                    <td ng-init="votesInvalid = poll.getVote(poll.votesinvalid,'votesinvalid')">
```
emanuelschuetze: add a blank after ","

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -181,7 +181,7 @@ <h3 ng-if="motion.polls.length > 0" translate>Voting result</h3>
                   <tr>
                     <td class="icon">
                       <i class="fa fa-check fa-lg"></i>
-                    <td ng-init="votesValid = poll.getVote(poll.votesvalid)">
+                    <td ng-init="votesValid = poll.getVote(poll.votesvalid,'votesvalid')">
```
emanuelschuetze: add a blank after ","

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -202,7 +202,7 @@ <h3 ng-if="motion.polls.length > 0" translate>Voting result</h3>
                   <tr class="total">
                     <td class="icon">
                       <strong style="font-size: 16px">&sum;</strong>
-                    <td ng-init="votesCast = poll.getVote(poll.votescast)">
+                    <td ng-init="votesCast = poll.getVote(poll.votescast,'votescast')">
```
emanuelschuetze: add a blank after ","

emanuelschuetze:
After we have merged #2189 I'll make a test with this pr.

emanuelschuetze:
Please rebase this pull request on top of current master (I want to test it with the merged pr #2189)

openslides/motions/static/js/motions/base.js
```diff
@@ -90,7 +90,18 @@ angular.module('OpenSlidesApp.motions', ['OpenSlidesApp.users'])
...
+                        console.log(this.abstain);
+			            percentNumber = Math.round(vote * 100 /(this.votesvalid - this.abstain) * 10) / 10;
+                        } else if (type == 'abstain'){
+                          percentNumber = "";
+                        } else if (type == 'valid'|| type=='invalid'){
+                          percentNumber = Math.round(vote * 100 /(this.votescast) * 10) / 10;
+                        }
                     }
+                    
+
```
emanuelschuetze: please remove these 2 empty lines.

openslides/poll/models.py
```diff
@@ -70,6 +70,7 @@ def print_weight(self, raw=False):
 PERCENT_BASE_CHOICES = (
     {'value': 'WITHOUT_INVALID', 'display_name': ugettext_lazy('Only all valid votes')},
     {'value': 'WITH_INVALID', 'display_name': ugettext_lazy('All votes cast (including invalid votes)')},
+    {'value': 'WITHOUT_ABSTAIN', 'display_name':  ugettext_lazy('Only YES and NO values')},
```
emanuelschuetze: use better: "Only Yes and No votes" for the new option.

Change the string "Only all valid votes" to "All valid votes (Yes/No/Abstain)". This should better explain the difference between the options.

And please change the order (descending):
* "Only Yes and No votes"
* "All valid votes (Yes/No/Abstain)"
* "All votes cast (including invalid votes)"
* "Disabled (no percents)"

emanuelschuetze:
Retested:
* motion slide does not show percent values if new 'Yes/No'-config-option is selected.
* motion pdf creation fails if new 'Yes/No'-config-option is selected (error: "MotionPoll" object has no attribute 'abstain')

matakuka:
The pdf printouts (assigment/ motions) still need work. Motion shows the total percent including abstains, assignment shows nothing.

matakuka:
I looked wrong. Assignment pdf printout works as expected. Motion pdf printout is still wrong, but I'm lost in where the pdf gets its percents from.
Here is where I lose track:
[motions/pdf.py:111](https://github.com/OpenSlides/OpenSlides/blob/28584c62f6b4b582e6168df24371f16bbbd54a43/openslides/motions/pdf.py#L111) --> [poll/models.py:173-177](https://github.com/OpenSlides/OpenSlides/blob/28584c62f6b4b582e6168df24371f16bbbd54a43/openslides/poll/models.py#L173-L177) --> ?

openslides/poll/models.py
```diff
@@ -66,10 +66,13 @@ def print_weight(self, raw=False):
...
         return print_value(self.weight, percent_base)
 
-
 PERCENT_BASE_CHOICES = (
-    {'value': 'WITHOUT_INVALID', 'display_name': ugettext_lazy('Only all valid votes')},
-    {'value': 'WITH_INVALID', 'display_name': ugettext_lazy('All votes cast (including invalid votes)')},
+    {'value': 'WITHOUT_ABSTAIN', 'display_name':  ugettext_lazy(
+        'Only Yes and No votes')},
+    {'value': 'WITHOUT_INVALID', 'display_name': ugettext_lazy(
+        'Only all valid votes (Yes/No/Abstain)')},
```
emanuelschuetze: As I suggested in my last comment: Please remove "Only" from this string: "All valid votes (Yes/No/Abstain)" 
Should be consistent with the "All votes cast (..)" option

emanuelschuetze:
Retested:
* Motion slide works now.
* Motion pdf is created now. For the right percent values (if yes/no config option is selected) you should look at the [print_value function in poll/models.py](https://github.com/OpenSlides/OpenSlides/blob/28584c62f6b4b582e6168df24371f16bbbd54a43/openslides/poll/models.py#L256). There, I think, you have to modifiy the percent output. Try a bit with debug outputs.

matakuka:
I disabled the wrong percent values on the printouts and just print the raw numbers. 
@normanjaeckel: is there an easy way to access the poll's "abstains" at [poll/models.py](https://github.com/OpenSlides/OpenSlides/pull/2192/files#diff-63a1c59fe8a052b345cbb1d36a2d7ccaR134)? Votesvalid etc. are there, but I couldn't get "yes/no/abstain" (I want to subtract the abstains from the valid votes)

openslides/poll/models.py
```diff
@@ -119,10 +122,19 @@ def print_votescast(self):
...
             base = 100 / float(self.votesvalid)
-        elif self.get_percent_base_choice() == "WITH_INVALID" and self.votescast and self.votescast > 0:
+        elif self.get_percent_base_choice() == "WITH_INVALID" and\
+                self.votescast and self.votescast > 0:
             base = 100 / float(self.votescast)
+        # elif self.get_percent_base_choice() == "WITHOUT_ABSTAIN" and\
+        #     self.votesvalid and self.votesvalid > 0:
+        #     try:
+        #         base = 100 / float(self.votesvalid - abstain)
+        #         #TODO get abstains from server
```
emanuelschuetze: Ok, now, the motion/assignment pdf has no percent values (only absolute votes). Not a really nice workaround but ok for (as long as we move the pdfs to client side - then we can use the JS code for percent values).

@normanjaeckel What do you think? Maybe do you have a quick solution to get the abstain value on  server side?

normanjaeckel: The `CollectDefaultVotesMixin` is a mixin that should work for different poll/vote class systems in the respective apps. In this case you can not access the abstain votes in this abstract layer because there is not guarantee that the concrete poll class (MotionPoll e. g.) has abstention votes at all. What you have to do is to override the method `get_percent_base` in both the `MotionPoll` class and the `AssignmentPoll` class. Then you can substract the abstention easily in motion and you have top decide what should happen in the different assignment cases.

normanjaeckel:
@matakuka Please check whether overriding `get_percent_base()` in `MotionPoll` may help.

normanjaeckel:
@matakuka What is the status of this pull request?

matakuka:
Still on my list, but with lower priority. I'll check it in the next days/this weekend.

openslides/assignments/views.py
```diff
@@ -497,32 +497,31 @@ def append_to_pdf(self, story):
                     cell.append(Paragraph(
                         "(%s)" % candidate.structure_level,
                         stylesheet['Ballot_option_suffix_YNA']))
+                if self.poll.yesnoabstain:
```
matakuka: the change of indentation on the following lines fixes an issue that haven't been seen before: if a candidate had a structure_level, they were printed without ballot options on the paper.

matakuka:
It should work now as desired, the pdf printouts include the percentages as well.
However, as I dabbled in several files and on several data model levels and the "proper thinking about this issue" was two months ago, I'm not entirely sure if I didn't accidentally inserted bad or unnecessary code (especially maybe duplicating "abstain/votesabstain" variables) . Also, the combination of a new mode of votes and this new config option leave quite a lot of test cases, and I may not have covered them all.
@normanjaeckel @emanuelschuetze

normanjaeckel:
Hm. I think we need a different approach. I can have a look at it, but not at the moment.

openslides/assignments/views.py
matakuka: - I moved this to a different PR #2254

normanjaeckel:
First we should discuss the goal of the issue. See #2182. I can specify my critics on the current proposal if necessary but I think we should focus on the different voting methods first.

emanuelschuetze:
As discussed: for motion only...

matakuka:
Done.

I reworked it for motions only.

matakuka:
@emanuelschuetze for testing.
Note: pdf creation has incorrect percentage calculations in master/motions/site.js, this PR fixes it.

matakuka:
ping @emanuelschuetze

openslides/motions/config_variables.py
```diff
@@ -1,4 +1,6 @@
 from django.core.validators import MinValueValidator
+from django.utils.translation import ugettext_lazy
+
```
emanuelschuetze: please remove this line

openslides/motions/static/js/motions/site.js
```diff
@@ -1113,7 +1114,7 @@ angular.module('OpenSlidesApp.motions.site', ['OpenSlidesApp.motions', 'OpenSlid
             }
         }
 
-        // save motion
+        // save motion.
```
emanuelschuetze: please remove this "."

openslides/motions/static/templates/motions/slide_motion.html
```diff
@@ -23,36 +23,37 @@ <h3 ng-if="motion.polls.length > 0" translate>Voting result</h3>
...
-              <td ng-init="voteNo = poll.getVote(poll.no)">
+              <td ng-init="voteNo = poll.getVote(poll.no, 'no')">
                 <span class="result_label"><translate>No</translate>:</span>
                 <span class="result_value" >
                   {{ voteNo.value }} {{ voteNo.percentStr }}
                 </span>
-                <div ng-if="voteNo.percentNumber">
-                  <uib-progressbar value="voteNo.percentNumber" type="danger"></uib-progressbar>
+                <div ng-if="voteNo.percentNumber >= 0">
+                  <uib-progressbar value="voteNo.percentNumber"  type="danger"></uib-progressbar>
```
emanuelschuetze: remove the new blank

openslides/motions/static/templates/motions/slide_motion.html
```diff
@@ -23,36 +23,37 @@ <h3 ng-if="motion.polls.length > 0" translate>Voting result</h3>
...
                 <strong style="font-size: 26px">&empty;</strong>
-              <td ng-init="voteAbstain = poll.getVote(poll.abstain)">
+              <td ng-init="voteAbstain = poll.getVote(poll.abstain, 'abstain')">
                 <span class="result_label"><translate>Abstain</translate>:</span>
                 <span class="result_value">
                   {{ voteAbstain.value }} {{ voteAbstain.percentStr }}
                 </span>
-                <div ng-if="voteAbstain.percentNumber">
+                <div ng-if="config('motions_poll_100_percent_base') != 'WITHOUT_ABSTAIN' && 
+                         voteAbstain.percentNumber >= 0">
```
emanuelschuetze: is this config check really required? voteAbstain.percentNumber should be null/undefined if WITHOUT_ABSTAIN is selected.

openslides/poll/models.py
```diff
@@ -66,10 +66,11 @@ def print_weight(self, raw=False):
...
 
-
 PERCENT_BASE_CHOICES = (
-    {'value': 'WITHOUT_INVALID', 'display_name': ugettext_lazy('Only all valid votes')},
-    {'value': 'WITH_INVALID', 'display_name': ugettext_lazy('All votes cast (including invalid votes)')},
+    {'value': 'WITHOUT_INVALID', 'display_name': ugettext_lazy(
+        'All valid votes (Yes/No/Abstain)')},
+    {'value': 'WITH_INVALID', 'display_name': ugettext_lazy(
+        'All votes cast (including invalid votes)')},
     {'value': 'DISABLED', 'display_name': ugettext_lazy('Disabled (no percents)')})
```
emanuelschuetze: Please reformat it in one line. The lines are not yet to long.

matakuka:
as requested, some minor changes & a rebase @emanuelschuetze

openslides/motions/config_variables.py
```diff
@@ -23,6 +24,10 @@ def get_config_variables():
     papers' and 'PDF'. The generator has to be evaluated during app loading
     (see apps.py).
     """
+    percent_base_choices_motion = PERCENT_BASE_CHOICES
```
emanuelschuetze: please use capital letters for constants

normanjaeckel:
Discussed some changes with @emanuelschuetze. Rest is ok.

openslides/motions/config_variables.py
emanuelschuetze: done in #2331

emanuelschuetze:
I created a new PR #2331 which makes last fixes for this feature. So I closed this PR here.

Issue #2193, opened 2016-06-09
Config values are created always in English


emanuelschuetze:
A bug was introduces with 2221b23 (Redesign of the config variables.) which should fix the big mode:
After db creation all config values are always in English (also if my browser language is German). See e.g. title and text of welcome page after new db creation. Before this commit the config value language depends on the browser language (correct).

@ostcar I think it's your turn because you introduced the bug ;).

ostcar:
@emanuelschuetze is this still an issue? see #2194

emanuelschuetze:
#2261 will fix this issue.

PR #2194, opened 2016-06-09
Config translations moved to client side (Issue #2093)

matakuka:
probably needs a rebase/ update, as it affects other pending PR

PR #2195, opened 2016-06-09
Configure config default value with ugettext_lazy.

ostcar:
In 2.0 the config value was initialized per request and therefore with
the browser language. Now the config value is initialized at startup.

Fixes: #2193

### DIFF
--- a/openslides/assignments/config_variables.py
+++ b/openslides/assignments/config_variables.py
@@ -1,5 +1,4 @@
 from django.core.validators import MinValueValidator
-from django.utils.translation import ugettext as _
 from django.utils.translation import ugettext_lazy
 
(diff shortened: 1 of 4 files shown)


emanuelschuetze:
@ostcar Please look at #2194 which is in conflict with your PR. @matakuka has moved all config translation strings to client side. I'm not sure if #2194 fixes already #2193... Then we does not need your PR. What do you think?

ostcar:
This pull request only gives the behaviour of the 2.0. #2194 is a improvement.

I suggest we leave this pull request open until #2194 is merged, then we close this without merging it.

emanuelschuetze:
Yes, let us merge #2194 first. I'll test if it and then I close this pr.

PR #2199, opened 2016-06-14
sets language to en if no browser language is set

matakuka:
When no language is set in the browser's setting, OpenSlides fails to start with a type error (at least in firefox). This PR fixes it by setting it to 'en'

### DIFF
--- a/openslides/core/static/js/core/base.js
+++ b/openslides/core/static/js/core/base.js
@@ -106,10 +106,14 @@ angular.module('OpenSlidesApp.core', [
             // get detected browser language code
             getBrowserLanguage: function () {
                 var lang = navigator.language || navigator.userLanguage;
-                if (lang.indexOf('-') !== -1)
-                    lang = lang.split('-')[0];
-                if (lang.indexOf('_') !== -1)
-                    lang = lang.split('_')[0];
+                if (!navigator.language && !navigator.userLanguage) {
+                    lang = 'en';
+                } else {
+                    if (lang.indexOf('-') !== -1)
+                        lang = lang.split('-')[0];
+                    if (lang.indexOf('_') !== -1)
+                        lang = lang.split('_')[0];
+                }
                 return lang;
             },
             // set current language and return updated languages object array


emanuelschuetze:
Succeccfully tested. Looks good. Good catch!

openslides/core/static/js/core/base.js
```diff
@@ -106,10 +106,14 @@ angular.module('OpenSlidesApp.core', [
             // get detected browser language code
             getBrowserLanguage: function () {
                 var lang = navigator.language || navigator.userLanguage;
-                if (lang.indexOf('-') !== -1)
-                    lang = lang.split('-')[0];
-                if (lang.indexOf('_') !== -1)
-                    lang = lang.split('_')[0];
+                try {
```
normanjaeckel: I do not understand why we have to use a try-catch here. Which line in the old code raises a typeError in which cases?

The expected value of `navigator.language` is something like `de-DE` or `en_US` or thelike or just `undefinded`, isn't it?. So we should implement at first a check if the language attribute is set (or just undefined), then we can parse the language string.

matakuka:
I have a browser with no entries in "preferred languages" (setting intl.accept_languages=""). Thus, upon starting OpenSlides, I get static/js/core/base.js:109: 'TypeError: lang is undefined'. 
I would get the same error at 111, but I'm not reaching this line, yet. 

The behavior as of PR is:
1) if there is no string (not "undefined" !), my new try/catch sets the lang to "en".
2) If it is a string with underscore/hyphen, the existing statements take care of stripping it to, e.g. "en_US" -> "en"
3) If it was an "en" to begin with, it just passes through the conditions unchanged.
--> In all cases, the result is lang="en".

We could also do a 
```
if not navigator.language and not navigator.userLanguage:
    lang = en
else:
    (old code)
```
which would do the same, with the possible complication of one of these settings (but not both) existing, but being empty. 
Wrapping a try/catch around the problematic code seemed the easiest way.

normanjaeckel:
Hm. It is just about my idea that we should use exceptions as little as possible and if we want to use them we should wrap as little code as possible. In your code it is not obvious where the type error is expected to be raised because we have four different lines with four different function calls and two array accesses.

Second point: Is it possible that the browser gives a string like `de`? Then the old and the new code would fail.

emanuelschuetze:
> Second point: Is it possible that the browser gives a string like de? Then the old and the new code 
> would fail.

I tested it with browser language `de` (and no other languages). It works without errors.

matakuka:
I changed the switch to if/else now; thus it should be more cleanly defined. 
As for the second point, the behavior is valid for all supported languages and their strings (de, en, es ...). 
What I noted (and may tackle later on) is, that when the browser setting is something like "de_de, en_US", openslides takes the *second* option (en), although browsers sort this list by preference (i.e. german first, then english) so openslides should prefer the first entry.

openslides/core/static/js/core/base.js
```diff
@@ -106,10 +106,14 @@ angular.module('OpenSlidesApp.core', [
             // get detected browser language code
             getBrowserLanguage: function () {
                 var lang = navigator.language || navigator.userLanguage;
-                if (lang.indexOf('-') !== -1)
-                    lang = lang.split('-')[0];
-                if (lang.indexOf('_') !== -1)
-                    lang = lang.split('_')[0];
+                if (!navigator.language && !navigator.userLanguage) {
+                lang = 'en';
```
normanjaeckel: This line needs four whitespaces indention.

normanjaeckel:
One last remark. For the last point (browsers have multiple languages in order) you can open an issue if you like.

normanjaeckel:
Ok. Rebase everything to one commit, then I merge. Thanks for your work.

matakuka:
travis-ci failed with:
`/home/travis/build.sh: /home/travis/virtualenv/python3.5.2/bin/pip: /home/travis/virtualenv/python3.5.0/bin/python: bad interpreter: No such file or directory`
I guess this is (was?) a misconfiguration in travis-ci (#2222 failed with the same error). Rerunning the travis-ci build should fix it (can only be done by someone with write-access to the repo?)

PR #2216, opened 2016-06-30
Added new user field is_committee and new default group Committees.

normanjaeckel:


openslides/users/signals.py
```diff
@@ -6,12 +6,13 @@
...
+    Committees.
 
     Creates the builtin user: admin.
     """
-    # Check whether the group pks 1 to 4 are free
-    if Group.objects.filter(pk__in=range(1, 5)).exists():
+    # Check whether the group pk's 1 to 5 are free.
+    if Group.objects.filter(pk__in=range(1, 6)).exists():
         # Do completely nothing if there are already some of our groups in the database.
         return
```
emanuelschuetze: Can you explain what this check should do? Why you filter from 1 to 6 and not from pk=1 to 5?

emanuelschuetze:
@normanjaeckel I created a pull request for your branch that you can merge #2216 together with my client-side changes: https://github.com/normanjaeckel/OpenSlides/pull/12

openslides/users/signals.py
normanjaeckel: `range(1,6)` returns `[1,2,3,4,5]`

PR #2230, opened 2016-07-27
Remove unused assignment config to publish winner results only.

### LINKED ISSUE #2145: election option "publish election result only for elected candidate" has no effect
In the settings of OpenSlides under elections there is an option "publish election result only for elected candidate". But setting/unsetting this has no effect on existing elections or on new elections.

emanuelschuetze:
Fixes #2145

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,6 +18,9 @@ Core:
 Motions:
 - Added origin field.
 
+Assignments:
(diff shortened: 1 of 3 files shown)


normanjaeckel:
Please add short note to CHANGELOG. Then merge.

emanuelschuetze:
Changelog entry added.

Issue #2211, opened 2016-06-29
New list of speakers overlay


emanuelschuetze:
The current list of speakers (of projected agenda item) should be shown on the projector in a new "overlay" - a box on the bottom right corner of the projector slide. See https://docs.openslides.org/de/_images/Agenda_04.png for an example in OpenSlides 1.7. (This feature is still missing in 2.0.)

The overlay should show the last speaker (in grey), the current speaker (in bold) and the next max. 5 speakers. The overlay should be hide/transparent if the list of speaker is empty or no list of speakers is available for the current slide (e.g. if you project an participant).

To show/hide this overlay add a new section 'List of speakers' to the projector sidebar (maximize it by clicking on projector button on the right side of the webinterface view). There place the new section below countdowns and messages. In the new section should be a project button to enable/disable the overlay on projector - similar like for countdown or messages.

**Optional:**
Add the two control buttons ("Begin next speach" / "End next speach") from list of speakers control view (agenda/item-detail.html) to the overlay control section (below the show/hide projector button). Then an administrator can control the main functions of a list of speakers from projector sidebar (enable/disable, start/stop).

PR #2241, opened 2016-08-01
List of speakers overlay (issue #2211) - old version

matakuka:
My approach for issue #2211 is this: 

[done] defining an element "core/listofspeakers", which mainly contains information if visible or not, and behaves similar to "core/messages".

[?] Creating such an element in 0001_initial.py, where a clock is already been created.
 --> ? This element does _not_ appear in the list of elements ? Perhaps 0001_initial is never triggered, not even by creating a new database?

[done] Display this newly created object on the /projector slide if it's property "visible" is true, by including a  /templates/core/listofspeakers.html, managed by the controller SlideSpeakerListCtrl.

[done] Define a css id "speakerbox" where the design part of the overlay will be stored.

[  ] Fill the template with the information from agenda/item/speaker; if there is an agenda item with agenda.isProjected == true and speaker is not empty. Else, don't display the overlay.

I'm sorry for posting semi-works, but my last approaches were as time-intensive as they were fruitless. Maybe I could do with a prod, and a second pair of eyes may easily encounter a fundamental error I'm doing


openslides/core/static/js/core/projector.js
```diff
@@ -54,6 +54,10 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
         slidesProvider.registerSlide('core/message', {
             template: 'static/templates/core/slide_message.html',
         });
+        
+        slidesProvider.registerSlide('core/listofspeakers', {
+            template: 'static/templates/core/listofspeakers.html',
```
emanuelschuetze: slide_listofspeakers.html

openslides/core/static/templates/core/listofspeakers.html
```diff
@@ -0,0 +1,17 @@
+<div ng-controller="SlideSpeakerListCtrl">
```
emanuelschuetze: please add comment: "overlay for list of speakers"

openslides/core/static/js/core/site.js
```diff
@@ -822,6 +822,26 @@ angular.module('OpenSlidesApp.core.site', [
                 }
             });
         };
+
+        //*** List of speakers on slide***
```
emanuelschuetze: list of speakers projector overlay

openslides/core/static/templates/core/projector-controls.html
```diff
@@ -211,6 +211,30 @@ <h4 translate>Messages</h4>
...
+        <a class="btn btn-default btn-sm"
+                ng-model="listofspeakers"
+                ng-click="listOfSpeakers()"
+                ng-class="{ 'btn-primary': listofspeakers.visible }"
+                title="{{ 'Project list of speakers' | translate }}">
+              <i class="fa fa-video-camera"></i>
+         </a>
+         <button ng-click="beginSpeech()"
+            class="btn btn-sm btn-default">
+            <i class="fa fa-microphone"></i>
```
emanuelschuetze: <translate>Start</translate>

openslides/core/static/templates/core/projector-controls.html
```diff
@@ -211,6 +211,30 @@ <h4 translate>Messages</h4>
...
+            </button>
+         <button ng-click="endSpeech()"
+            class="btn btn-sm btn-default">
+            <i class="fa fa-microphone-slash"></i>
```
emanuelschuetze: `<translate>Stop</translate>`

normanjaeckel:
> [?] Creating such an element in 0001_initial.py, where a clock is already been created.
> --> ? This element does not appear in the list of elements ? Perhaps 0001_initial is never triggered, not even by creating a new database?

Can you please explain the problem? Where does the clock *not appear*? Creating a new database and opening OpenSlides shows the clock on projector and on `/rest/core/projector/1`. If you want to add a new default projector element to the (default) projector, you have to add a new migrations file. Editing existing files like `0001_...py` or `0002_....py` wont help. Try `python manage.py makemigrations --empty core`.

openslides/core/migrations/0003_auto_20160803_1058.py
```diff
@@ -0,0 +1,34 @@
...
+    Adds a list of speaker overlay to the existing projector id=1
+    """
+    projector = Projector.objects.get(id=1)
+    soverlay = {}
+    soverlay_uuid = uuid.uuid4().hex
+    soverlay[soverlay_uuid] = {
+        'name': 'core/speakeroverlay',
+        'stable' : True,
+        'uuid' : soverlay_uuid}
+    projector.elements.update(soverlay)
```
matakuka: The projector object (id=1) has the entry `elements` which is a dict of dicts, at the time of this call holding the entry for core/clock. I want to append the new entry (soverlay) and save the projector object. However, `/rest/core/projector/1` reports only core/clock as being part of the element. I guess I am confused by similarities/differences between Django, json and python.

matakuka: After @normanjaeckel s comment I tried a different approach. I could produce a second projector with the speakeroverlay, but no clock. Now I'm trying to append the speakeroverlay to the dict of elements (see above). However, I am stuck again, with syntax I and the consulted web pages consider correct. As it doesn't work, my thinking appears to be wrong.

normanjaeckel: Try this:
```python
# -*- coding: utf-8 -*-
# Generated by Django 1.9.7 on 2016-08-03 10:58
from __future__ import unicode_literals

import uuid

from django.db import migrations, models


def add_speaker_overlay(apps, schema_editor):
    """
    Adds a list of speakers overlay to the existing projector id=1.
    """
    # We disconnect the inform_changed_data_receiver to disabled the
    # autoupdate mechanism during migrations.
    models.signals.post_save.disconnect(
        dispatch_uid='inform_changed_data_receiver')

    # We get the model from the versioned app registry;
    # if we directly import it, it will be the wrong version.
    Projector = apps.get_model('core', 'Projector')
    projector = Projector.objects.get(id=1)
    projector.config[uuid.uuid4().hex] = {
        'name': 'core/speakeroverlay',
        'stable' : True}
    projector.save()


class Migration(migrations.Migration):

    dependencies = [
        ('core', '0002_session'),
    ]

    operations = [
        migrations.RunPython(add_speaker_overlay)
    ]
```

normanjaeckel: Just do describe the main points:
- You can not save something to projector.elements but to projector.config. See core/models.py.
- You should not import Projector in global scope but use the model from the versioned app registry.
- You have to disable autoupdate otherwise you would get an error. Here we can think about a better solution: Disable autoupdate globally during migrations, e. g. in a customized migrate command.

openslides/core/static/js/core/projector.js
```diff
@@ -140,11 +156,48 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+                var currentmotion = Motion.find(element.id);
+                //TODO displayeditem = currentmotion.agenda_item;
+            } else if (element.name == "core/customslide") {
+                displayeditem = element.id;
+            } else if (element.name == "assignments/assignment") {
+                
+                var currentassign = Assignment.find(element.id);
+                
+                //TODO I want to get the 'value.agenda_item_id' from this currentassign object.
+
```
matakuka: @emanuelschuetze:

Here I am in the projector/speakeroverlay element. I want to connect to the currently displayed agenda item, to get their list of speakers. `currentassign` is the currently displayed assignment item, which holds in itself the agenda information.

However, I fail miserably at accessing its' `agenda_item_id` value (which *is* existant and accessible for the user in question, as the values are sent somewhere nested in the answer - I can see them in Inspektor). Below are some of the accesses I tried. I may be missing something really simple.

openslides/core/static/js/core/projector.js
```diff
@@ -54,6 +54,25 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
             template: 'static/templates/core/slide_message.html',
         });
+        slidesProvider.registerSlide('core/speakeroverlay', {
+            template: 'static/templates/core/slide_speakeroverlay.html',
+            resolve: {
+                motions: function(Motion) {
+                    return Motion.findAll().then(function(motions) {
+                        angular.forEach(motions, function(motion) {
+                            Motion.loadRelations(motion, 'agenda_item');
+                            });
```
emanuelschuetze: remove one indentation

openslides/core/static/js/core/projector.js
```diff
@@ -134,6 +153,60 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+    '$scope',
+    'Motion',
+    'Assignment',
+    'Agenda',
+    'Projector',
+    function($scope, Motion, Assignment, Agenda, Projector) {
+        // Attention! Each object that is used here has to be dealt on server side.
+        // Add it to the coresponding get_requirements method of the ProjectorElement
+        // class.
+        //TODO: speed it up. Resolving?
```
emanuelschuetze: What do you mean here?

openslides/core/static/js/core/site.js
```diff
@@ -197,6 +197,23 @@ angular.module('OpenSlidesApp.core.site', [
...
+                        });
+                    },
+                    assignments: function(Assignment) {
+                        return Assignment.findAll().then(function(assignments) {
+                            angular.forEach(assignments, function(assignment) {
+                                Assignment.loadRelations(assignment, 'agenda_item');
+                            });
+                        });
+                    }
+                //TODO: resolve custonslides
```
emanuelschuetze: still open?

openslides/core/static/js/core/site.js
```diff
@@ -827,6 +844,25 @@ angular.module('OpenSlidesApp.core.site', [
...
         };
+        //*** List of speakers overlay on slide***
+        $scope.speakeroverlay = function() {
+            var elements = Projector.get(1).elements;
+            for (var element in elements) {
+                if (elements[element].name == 'core/speakeroverlay') {
+                    return elements[element];
+                }
+            }
+            return null;//TODO: error
```
emanuelschuetze: error still open?

openslides/core/static/js/core/projector.js
```diff
@@ -134,6 +153,60 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+                    .then(function(item) {
+                        $scope.AgendaItem = item.agenda_item;
+                    });
+                });
+            } else {
+                $scope.AgendaItem = null;
+            }
+        };
+        $scope.$watch($scope.$parent.elements, $scope.displayedElement());
+
```
emanuelschuetze: remove line

emanuelschuetze:
* Please use projector button without text (like in countdown/messages).
* Add a link '>> Manage list of speakers'  (which redirects to item-detail view of current list)

matakuka:
done; ready for testing

emanuelschuetze:
Retested:
* Project a customslide. Then click on edit button below projector live view. Gets  error: "Could not resolve 'core.speakeroverlay.detail.update' from state 'agenda.item.list'"

Rest works good.

emanuelschuetze:

openslides/core/static/js/core/site.js
```diff
@@ -827,6 +853,46 @@ angular.module('OpenSlidesApp.core.site', [
...
+            var data = {};
+            if ($scope.speakeroverlay().visible) {
+                data[$scope.speakeroverlay().uuid] = { "visible": false };
+            } else {
+                data[$scope.speakeroverlay().uuid] = { "visible": true };
+            }
+            $http.post('/rest/core/projector/1/update_elements/', data);
+        };
+        $scope.goToListofSpeakers = function() {
+            $.each(Projector.get(1).elements, function(key, value) {
```
normanjaeckel: Can we use angular.forEach or lodash?

normanjaeckel:
One remark.

matakuka:
remark done

matakuka:
@normanjaeckel done

openslides/core/projector.py
```diff
@@ -139,3 +139,10 @@ class Message(ProjectorElement):
     def check_data(self):
         if self.config_entry.get('message') is None:
             raise ProjectorException('No message given.')
+
+
+class ListOfSpeakersOverlay(ProjectorElement):
+    """
+    List of speakers (last, current, next) on the projector. Rendered as overlay.
+    """
+    name = 'core/speakeroverlay'
```
normanjaeckel: Hm. I think the `get_requirements()` is missing. Please test to view the projector with a user which has only the permission to see the projector, nothing else (no permission to see agenda, motions, users, ...). Then the access to all these objects should be restricted but not if the element is on the projector. See utils/projector.py for more info.

openslides/core/static/js/core/projector.js
```diff
@@ -134,6 +160,65 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+                var ee = $scope.$parent.elements[e];
+                if (ee.name == 'motions/motion') {
+                    displayedElement = ['motion', ee.id];
+                } else if (ee.name == 'core/customslide') {
+                    displayedElement = ['agenda', ee.id];
+                } else if (ee.name == 'assignments/assignment') {
+                    displayedElement = ['assignment', ee.id];
+                }
+            }
+            if (displayedElement[0] == 'motion') {
```
normanjaeckel: Hm. The motion was resolved in the slidesProvider so why do we have to find it again?

openslides/core/static/js/core/site.js
```diff
@@ -197,6 +197,29 @@ angular.module('OpenSlidesApp.core.site', [
                 data: {extern: true},
                 onEnter: function($window) {
                     $window.location.href = this.url;
+                },
+                resolve: {
```
normanjaeckel: Why do we have to resolve this stuff here. It is already resolved in the slidesProvider.

openslides/core/static/js/core/site.js
```diff
@@ -815,15 +841,54 @@ angular.module('OpenSlidesApp.core.site', [
...
+                data[$scope.speakeroverlay().uuid] = { "visible": false };
+            } else {
+                data[$scope.speakeroverlay().uuid] = { "visible": true };
+            }
+            $http.post('/rest/core/projector/1/update_elements/', data);
+        };
+        $scope.goToListofSpeakers = function() {
+            angular.forEach(Projector.get(1).elements, function(element) {
+                if (element.name == 'motions/motion') {
+                    Motion.find(element.id).then(function(motion){
```
normanjaeckel: Please check whether this can be done either by a resolve or by Motion.get(element.id).

normanjaeckel:
I am sorry but I saw some more things to discuss. Maybe we have to find a general solution for the so called resolve-find-get problem. @emanuelschuetze ?

openslides/core/projector.py
matakuka: Can't test this; can_see_projector gets too much permissions 
(I hesitate to open yet another issue when I may just have forgotten something)

---**details:**---
I did:
```
git checkout master
rm -r personal-data/
pip install -r requirements.txt
npm install
cleared browser cache
restarted browser
log in as admin
create group "onlyprojector(6)"  "permissions": ["core.can_see_projector"]
create user "John", groups_id: [6]
create some agenda and motions
log out
delete cache again
log in as John
```
I can see what I should not see: the agenda, each agenda_item, motions/categories, 
motions in detail, all workflows, all assignments in detail

openslides/core/static/js/core/projector.js
matakuka: Testing:
After line 192: console.log(motion) : motion *without* agenda_item
After line 194: console.log(motion) : motion *with* agenda_item

What indeed seems unneccessary is the resolving at lines 59 pp. Even with those lines disabled, speakeroverlay gets the information needed

openslides/core/projector.py
normanjaeckel: At the moment, the user gets also all permissions from the Registered group. But in a few seconds/hours we will merge the Pull request of @FinnStutzenstein with the permission matrix and some changes on the built-in groups. Then your workflow should work as expected. So just wait on #2260 .

openslides/core/static/js/core/projector.js
normanjaeckel: Yes but try to add `motions` to the dependency injection list above and then console log motions just after line 171. There you should get the resolved motions.

openslides/core/static/js/core/site.js
matakuka: As it seems, this section is not even triggered (a console log in onEnter never fires), not when opening the projector directly, and neither when opening it through the core app.

matakuka:
I keep running into problems. As I understand it, the causes behind the symptoms are:

1.  *resolve* only resolves on loading the state (?). The projector is only loaded *once*, as is the state where the projector controls reside. They keep being present, but idle, so their resolve gets out of sync, and new items won't appear.
2. The /projector has no working resolve in itself¹. Such a resolve would need to monitor far too much, as there are users, customslides, agenda_items, motions, agendas; which change more often than the projector is reloaded.
3.  The same problem applies for each projector element. motions/js/projector.js, for example, just  `find()` the current element than to get it from a findAll(). The overlay of list of speakers has it worse; it can either load all resolves from assignment, motions, customslides, and potentially running into (1), or just find() the one element that has just changed.


¹just a non-working; in core/site.js:201

openslides/core/projector.py
matakuka: Okay, with the get_requirements I have a major problem.

option 1: getRequirements (all ) -- too much permission

option 2: removing speakerOverlay, and re-adding it as a `<div>` on the projector templates: complete rewrite, still needs to getRequirements(all usernames¹).

option3: option 2 + adding a field "user_name" to agenda_item.speakers and ...poll.options.

¹as every other projector slide displaying usernames, because the names will change before the projector is reloaded.

normanjaeckel: Hm. If you want to display potentially all usernames on your projector element then you just have to leave them all out. I do not see any other possibility at the moment. Maybe we can later implement a lazy loading of the projector if some other stuff than the projector itself changes.

matakuka: The problem is not just with my list of speakers overlay, but also with other, existing slides (which, as of now, are trying to load all users).

An example is elections. As an user without "can_see_user_names" you cannot see the candidates of an election, nor the results.

matakuka: The *idea* of the "cannot see users", in my opinion, is *not* that I cannot see who's elected. I just should not get the list of all names.

I think putting a "name" field nexrt to the user_id in poll.option and agenda_item.speakers would be a good workaround. 
Pro: you only get the names of the users in question; you don't always have to query Users if all you want is a name (we can avoid some monsters à la `User.find(assignment.agenda_item.speaker.id).get_full_name` ),
Con: The name can get out of sync with the user (which is already the case with resolve); you'll add one more variable to "poll" and "agenda", which already carry around quite some stuff.

normanjaeckel: See https://github.com/OpenSlides/OpenSlides/issues/2302. For short you may just release all users. We later can change the restrictions again.

matakuka:
I changed the approach (in the other PR #2306). I'll keep *this* branch because it is very different.

matakuka:
closing this PR because the new one is preferred.

Issue #2248, opened 2016-08-03
autoupdate mechanism is broken after introduction of django channels


ThomasJunk:
Introduced with commit fe64941aabd0228fc8090ca21f2f99df5204ee20

Bug 1)
2016-08-03 12:39:00,523 - ERROR - worker - Error processing message with consumer 
…
  File "/home/thomas/python/OpenSlides/openslides/utils/autoupdate.py", line 89, in send_data
    instance = Model.objects.get(pk=message_pk)
…
openslides.motions.models.DoesNotExist: Motion matching query does not exist.

Steps to reproduce:
Create a new motion

Always reproducible.

Bug 2) 
2016-08-03 12:53:44,475 - ERROR - worker - Error processing message with consumer openslides.utils.autoupdate.send_data:
Traceback (most recent call last):
  File "/home/thomas/python/env/lib/python3.5/site-packages/channels/worker.py", line 111, in run
    consumer(message, **kwargs)
  File "/home/thomas/python/OpenSlides/openslides/utils/autoupdate.py", line 93, in send_data
    'id': instance.get_rest_pk(),
UnboundLocalError: local variable 'instance' referenced before assignment

Steps to reproduce:
upload media.
delete media.

Always reproducible.

The according code is: 

```
def send_data(message):
…
    if not message['is_deleted']:
        Model = get_model_from_collection_string(message['collection_string'])
        instance = Model.objects.get(pk=message['pk'])
…
    for user in itertools.chain(get_logged_in_users(), [AnonymousUser()]):
    …
            'id': instance.get_rest_pk(),
    
        if not message['is_deleted']:
   …
```

If a message is in the state of 'is_deleted' the `instance` is never defined. Therefore instance.get_rest_pk() is called before defined. 

As far as I can does `instance.get_rest_pk()` equal `message['pk']`. That would fix (2).

But in case, a 'create' message is processed `Model.objects.get(pk=message['pk'])` should fail, since there isn't an object to retrieve. 

Is there any way to distinguish messages per type other than `is_deleted`?
Where are (possible) messages defined? 

As a Newb I need some help :]

@ostcar @normanjaeckel

normanjaeckel:
Added fix for case 2. In case 1 Model.objects.get(pk=message['pk']) should not fail because the object is already created before the autoupdate process starts to distribute the message to the clients.

normanjaeckel:
Creating works for users, custom slides and thelike but not for motions. So I think this is a special motion problem.

[...]

PR #2250, opened 2016-08-03
Fixed error in autoupdate. See #2248.

normanjaeckel:


emanuelschuetze:
Tested. Fixed successfully the UnboundLocalError that 'instance' referenced before assignment.
So ready to merge...

PR #2254, opened 2016-08-05
Repairs assignment ballots for candidates with structure_level

matakuka:
Moved an if/else - block to the correct position. It affected the ballot papers for assignments, in that candidates with structure level were only printed, but had no fields to vote for them.

Old behaviour: if-Block (structure level **OR** if-Block (voting method) )
New behaviour: if-Block (structure-Level) **THEN** if-Block (voting method)


normanjaeckel:
Looks good. Ready to merge.

emanuelschuetze:
Good catch! I tested it successfully.

PR #2255, opened 2016-08-05
Added search filtering by category (fixes #2019, fixes #2239)

### LINKED ISSUE #2019: Add help text how to search partial words with whoosh
The global search with whoosh allows to search for full words but also for partial words. 

Example: If you have an agenda item title "foobar" you can search for `foobar` or `foo*` to get the item as search result.

There should be at least a help text to explain how to find partial words. Or improve whoosh to support partial word searching without wildcard char (*).

### LINKED ISSUE #2239: Filter for search results
The search results should be filtered by a selected OpenSlides app: agenda item, motions, elections, participants, files. Each filter should show as checkbox between search bar and search results. - This was a feature from OpenSlides 1.7 which is still missing in 2.0.

matakuka:
*The search now uses "partial matches" as default; added "search for full words only" as option, added filtering per category on search results page*

1) Added a javascript search variable 'fullword'. If this variable is false or not defined, the `string` will be transformed to `*string*`, allowing for partial matches, before the request is sent to the server (#2019).
2) Added checkboxes on the search result page. If checked, the elements that match the search string *and* the checked categories will be shown (#2239).

**Future goals on search (not included)**
- Sub-options, like "participants: only those present"
- more advanced search options, like AND, OR, correction of misspelling, non-case-sensitive
- indicate somehow WHY the search results are listed.

(edit: new summary, TODOs are done)

### DIFF
--- a/openslides/core/static/js/core/site.js
+++ b/openslides/core/static/js/core/site.js
@@ -944,40 +944,89 @@ angular.module('OpenSlidesApp.core.site', [
 .controller('SearchBarCtrl', [
     '$scope',
     '$state',
-    function ($scope, $state) {
-        $scope.search = function(query) {
-            $scope.query = '';
+    '$sanitize',
+    function ($scope, $state, $sanitize) {
+        $scope.search = function() {
+            var query = _.escape($scope.querybar);
+            $scope.querybar = '';
             $state.go('search', {q: query});
         };
     }
 ])
+
 // Search Controller
 .controller('SearchCtrl', [
     '$scope',
     '$http',
     '$stateParams',
+    '$location',
+    '$sanitize',
     'DS',
-    function ($scope, $http, $stateParams, DS) {
+    function ($scope, $http, $stateParams, $location, $sanitize, DS) {
+        $scope.fullword = false;
+        $scope.filterAgenda = true;
+        $scope.filterMotion = true;
+        $scope.filterAssignment = true;
+        $scope.filterUser = true;
+        $scope.filterMedia = true;
+
         // search function
-        $scope.search = function(query) {
-            $http.get('/core/search_api/?q=' + query).then(function(success) {
-                var elements = success.data.elements;
-                $scope.results = [];
-                angular.forEach(elements, function(element) {
-                    DS.find(element.collection, element.id).then(function(data) {
-                        data.urlState = element.collection.replace('/','.')+'.detail';
-                        data.urlParam = {id: element.id};
-                        $scope.results.push(data);
+        $scope.search = function() {
+            var query = _.escape($scope.query);
+            if (query !== '') {
+                var lastquery = query;
+                // attach asterisks if search is not for full words only
+                if (!$scope.fullword) {
+                    if (query.charAt(0) != '*'){
+                        query = "*" + query;
+                    }
+                    if (query.charAt(query.length - 1) != '*'){
+                        query = query + "*";
+                    }
+                }
+                $scope.query = lastquery;
+                $http.get('/core/search_api/?q=' + query).then(function(success) {
+                    $scope.results = [];
+                    var elements = success.data.elements;
+                    angular.forEach(elements, function(element) {
+                        DS.find(element.collection, element.id).then(function(data) {
+                            data.urlState = element.collection.replace('/','.')+'.detail';
+                            data.urlParam = {id: element.id};
+                            $scope.results.push(data);
+                        });
                     });
                 });
-            });
+                $location.url('/search/?q=' + lastquery);
+            }
         };
 
-        // run search with get parameter from url
+        //get search string from parameters submitted from outside the scope
         if ($stateParams.q) {
-            $scope.search($stateParams.q);
             $scope.query = $stateParams.q;
+            $scope.search();
         }
+
+        // returns element if part of the current search selection
+        $scope.filterresult = function() {
+            return function(result) {
+                if ($scope.filterUser && result.urlState == 'users.user.detail') {
+                    return result;
+                }
+                if ($scope.filterMotion && result.urlState == 'motions.motion.detail') {
+                    return result;
+                }
+                if ($scope.filterAgenda && result.urlState == 'core.customslide.detail') {
+                    return result;
+                }
+                if ($scope.filterAssignment && result.urlState == 'assignments.assignment.detail') {
+                    return result;
+                }
+                if ($scope.filterMedia && result.urlState== 'mediafiles.mediafile.detail') {
+                    return result;
+                }
+                return;
+            };
+        };
     }
 ])
 
--- a/openslides/core/static/templates/index.html
+++ b/openslides/core/static/templates/index.html
@@ -139,9 +139,9 @@
       </ul>
       <!-- Search bar -->
       <div class="searchbar pull-right" ng-controller="SearchBarCtrl">
-        <form ng-submit="search(query); closeMenu()">
+        <form ng-submit="search(); closeMenu()">
           <div class="input-group">
-            <input ng-model="query" class="form-control" type="text" placeholder="{{ 'Search' | translate}}">
+            <input ng-model="querybar" class="form-control" type="text" placeholder="{{ 'Search' | translate}}">
             <span class="input-group-btn">
               <button type="submit" class="btn btn-default">
                 <i class="fa fa-search"></i>
--- a/openslides/core/static/templates/search.html
+++ b/openslides/core/static/templates/search.html
@@ -5,16 +5,43 @@
 </div>
 
 <div class="details">
-    <form class="input-group" ng-submit="search(query)">
+    <form class="input-group" ng-submit="search()">
       <input type="text" ng-model="query" class="form-control">
       <span class="input-group-btn">
         <button type="submit" class="btn btn-default" translate>Search</button>
       </span>
     </form>
-
+    <div class="searchfilter spacer-top">
+      <label class="checkbox-inline">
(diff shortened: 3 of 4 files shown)


normanjaeckel:
You should add the model `fullword` to the scope inside the controller:
```javascript
.controller('SearchCtrl', [
   ...
   function (...) {
      $scope.fullword = true;
      ...
    }
])
```
Then you do not need to put it as argument into the search function. Even the query should not be an argument of the search function. This is more AngularJS like. You may correct this bad style in the old code.
The same for the result filter.

matakuka:
Changes:
- `fullword` is now inside the controller, `query` is moved as well.
- the result filter changed a bit. It does not return an array of objects to be rendered by the html, but instead checks if an object rendered by the html is visible. Reason: In the first version it had rely on the results of the search arriving first. Now, it can start filtering even if not all results are available.

emanuelschuetze:
I added some template improvements in pull request https://github.com/matakuka/OpenSlides/pull/2. Please merge it into this pr. Thx.

Another open issue: Enter a string in the search bar (in site header) and try to search. The query is not sent to the searchCtrl. Please try to fix it. Thx

matakuka:
@emanuelschuetze I rebased it on top of current master, integrated your PR and repaired the page header search bar

openslides/core/static/js/core/site.js
```diff
@@ -526,32 +526,79 @@ angular.module('OpenSlidesApp.core.site', [
     '$scope',
     '$state',
     function ($scope, $state) {
-        $scope.search = function(query) {
-            $scope.query = '';
+        $scope.search = function() {
```
emanuelschuetze: please clear the search bar after searching - keep it empty for entering new search string.

matakuka: both search bars?

matakuka: I think then we should change the `Results` heading to `results for %searchstring` .

matakuka:
I integrated the clearance of the search bars and a heading reading `Search results for: (String)`. One test remaining (I will test/repair it tomorrow morning:) If there already are results and the "search for full words only" setting is changed - does this trigger an empty search?

matakuka:
Tested, made another change/revert upon demand: 
- The search bar in the search result page is *not* reset anymore (to allow for correction of search strings)
- the "results for" line was removed again (as you see your query in the search bar)
- the url changes to `/search/?q=Searchstring`, so that there is an option to easily share a search query between users.

(In the meantime I had a problem of the search returning users twice, but I cannot replicate it anymore. I guess it was a corrupted database).

@emanuelschuetze for testing

emanuelschuetze:
Nice. Works very good now. Please rebase for a last review and merge.

matakuka:
another thing, that may need attention (later? In this PR?):
- Searching for `Jim&x=maliciouscode` --> two query variables (`q: *Jim`, `x:maliciouscode*`)
- Searching for `Jim=beautiful` --> shorted query (`q:*Jim`)

So some escaping is not done properly here.

matakuka:
Added sanitizing of search strings and rebased. Ready for last review @emanuelschuetze

matakuka:
Search for `");console.log('Hello');` results in query been sent: `*`. It is not dangerous, but a query of * should be avoided. So another commit is coming.

matakuka:
I think the problem here is `"`, which is changed to `&#34` by $sanitize. However, upon sending the query, somehow `*&#34 ...` changes into `*`, so the double quote is active once again. What happens with the rest of the line, I'm not sure (which should trigger my paranoia).

matakuka:
Tested again and it is **not** dangerous. It "just" shortens/destroys queries, so it's a lower priority and something for another issue.

emanuelschuetze:
@normanjaeckel Ready for last review and merge!

openslides/core/static/js/core/site.js
```diff
@@ -525,33 +525,87 @@ angular.module('OpenSlidesApp.core.site', [
...
     '$scope',
     '$state',
-    function ($scope, $state) {
-        $scope.search = function(query) {
-            $scope.query = '';
+    '$sanitize',
+    function ($scope, $state, $sanitize) {
+        $scope.search = function() {
+            var query = $sanitize($scope.query);
+            $scope.query = "";
```
normanjaeckel: Proposal: Use single quotes (as we do most of the time).

openslides/core/static/js/core/site.js
```diff
@@ -525,33 +525,87 @@ angular.module('OpenSlidesApp.core.site', [
...
-        $scope.search = function(query) {
-            $http.get('/core/search_api/?q=' + query).then(function(success) {
-                var elements = success.data.elements;
-                $scope.results = [];
-                angular.forEach(elements, function(element) {
-                    DS.find(element.collection, element.id).then(function(data) {
-                        data.urlState = element.collection.replace('/','.')+'.detail';
-                        data.urlParam = {id: element.id};
-                        $scope.results.push(data);
+        $scope.search = function(q) {
```
normanjaeckel: Is there any usecase for the function argument `q`? It looks like that we always use `$scope.query` and so we can omit some lines of code here.

openslides/core/static/js/core/site.js
```diff
@@ -525,33 +525,87 @@ angular.module('OpenSlidesApp.core.site', [
...
+            var query;
+            if (q) {
+                query = $sanitize(q);
+            } else {
+                query = $sanitize($scope.query);
+            }
+            if (query !== "") {
```
normanjaeckel: Proposal: Use single quotes.

normanjaeckel:
Some remarks.

openslides/core/static/js/core/site.js
matakuka: > Is there any usecase for the function argument q? It looks like that we always use $scope.query and so we can omit some lines of code here.

If you're coming from the search bar (upper right), you mostly (always?) are in a *different* scope than the search results page, and after a state change, this first scope's `$scope.query` will be lost, so it is sent as argument `q`.

matakuka:
I changed the style inconsistencies, but *not* the argument `q` part (see line comment) @normanjaeckel

openslides/core/static/js/core/site.js
normanjaeckel: I do not understand because I think in this case there should be something like `search(<something here>)` somewhere in a template. I can not find such a string in one of our templates. If there is some hidden magic for this function call then we can merge this. No problem.

openslides/core/static/js/core/site.js
```diff
@@ -1155,3 +1210,4 @@ angular.module('OpenSlidesApp.core.site', [
 ]);
 
 }());
+
```
normanjaeckel: Remove blank line.

openslides/core/static/js/core/site.js
```diff
@@ -525,33 +525,88 @@ angular.module('OpenSlidesApp.core.site', [
 .controller('SearchBarCtrl', [
     '$scope',
     '$state',
-    function ($scope, $state) {
-        $scope.search = function(query) {
+    '$sanitize',
+    function ($scope, $state, $sanitize) {
+        $scope.search = function() {
+            var query = $sanitize($scope.query);
```
matakuka: `$scope.query` comes from the SearchBar. Upon `search()`, we go to the SearchCtrl, and will have a different `$scope`, thus we are not able to access this `$scope.query` any more, but a *different* `$scope.query`, which will be empty if we don't arrive from the `.search` state.

matakuka: I'll rename the SearchBar query to queryBar, so it's less confusing

openslides/core/static/js/core/site.js
```diff
@@ -525,33 +525,88 @@ angular.module('OpenSlidesApp.core.site', [
...
-    function ($scope, $state) {
-        $scope.search = function(query) {
-            $scope.query = '';
+    '$sanitize',
+    function ($scope, $state, $sanitize) {
+        $scope.search = function() {
+            var query = $sanitize($scope.querybar);
+            $scope.querybar = '';
             $state.go('search', {q: query});
         };
```
matakuka: upon calling this `search()`, we go to a different $scope, where `$scope.query` is undefined yet. So we need to pass the content of `$scope.querybar` to the *other* $scope. That's when `q` is used.
@normanjaeckel

normanjaeckel: As discussed please try to change to code in
```
        // run search with get parameter from url
        if ($stateParams.q) {
            $scope.search($stateParams.q);
            $scope.query = $stateParams.q;
        }
```
to 
```
        // run search with get parameter from url
        if ($stateParams.q) {
            $scope.query = $stateParams.q;
            $scope.search();
        }
```
Then you can remove the attribute in the search function.

matakuka:
done as requested @normanjaeckel

normanjaeckel:
Ok. Please rebase.

matakuka:
@normanjaeckel done, ready for merge

normanjaeckel:
@emanuelschuetze All done. Ready to merge.

emanuelschuetze:
Retested. I found a minor bug:
* Create a customslide with umlaut ("Begrüßung").
* Enter search string "Begrüßung" into search bar.
* Search view shows correct search result but the search string in the big input field is cropped: "begr"

matakuka:
done @emanuelschuetze

emanuelschuetze:
Ok, retested. Please rebase it for merging.

matakuka:
@emanuelschuetze ready for merge

PR #2256, opened 2016-08-08
Only do an autoupdate after an ongoing transaction.

ostcar:
This needs at least django 1.9

Fixes: #2248

### DIFF
--- a/openslides/utils/autoupdate.py
+++ b/openslides/utils/autoupdate.py
@@ -5,6 +5,7 @@ from asgiref.inmemory import ChannelLayer
 from channels import Channel, Group
 from channels.auth import channel_session_user, channel_session_user_from_http
 from django.apps import apps
+from django.db import transaction
 from django.utils import timezone
 
 from ..users.auth import AnonymousUser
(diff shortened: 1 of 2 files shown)


openslides/utils/autoupdate.py
```diff
@@ -108,14 +109,19 @@ def inform_changed_data(instance, is_deleted=False):
...
-                'collection_string': root_instance.get_collection_string(),
-                'pk': root_instance.pk,
-                'is_deleted': is_deleted and instance == root_instance,
-                'dispatch_uid': root_instance.get_access_permissions().get_dispatch_uid()})
-        except ChannelLayer.ChannelFull:
-            pass
+        # If currently there is an open database transaction, then the following
+        # function is only called, when the transaction is commited. If there
+        # is currently no transaction, then the function is called immediately.
+        def call_channel():
```
normanjaeckel: Suggestion: Rename this to `call_autoupdate_channel` or something similar.

normanjaeckel:
Nice. One suggestion to @ostcar.

@ThomasJunk Can you please test if this fix works for you?

ThomasJunk:
Does the trick :+1:

emanuelschuetze:
I tested successfully this pull request:
After creating new motion there are no `openslides.motions.models.DoesNotExist` any more as reported in #2248. 

@ostcar Thanks. Please look at Norman's suggestion. Then we can merge it.

openslides/utils/autoupdate.py
ostcar: done (with other name)

normanjaeckel:
Please rebase.

PR #2257, opened 2016-08-08
No crash without a browser (fixes #2249)

### LINKED ISSUE #2249: 'manage.py start' fails on computers without a browser
When running `manage.py start`, it do not start the webserver if no runnable browser is found while giving this error:

```
Traceback (most recent call last):
  File "manage.py", line 11, in <module>
    exit(main())
  File "/home/finn/OpenSlides/OpenSlides/openslides/__main__.py", line 41, in main
    known_args.callback(known_args)
  File "/home/finn/OpenSlides/OpenSlides/openslides/__main__.py", line 182, in start
    start_browser('http://localhost:%s' % args.port)
  File "/home/finn/OpenSlides/OpenSlides/openslides/utils/main.py", line 264, in start_browser
    browser = webbrowser.get()
  File "/usr/lib/python3.4/webbrowser.py", line 51, in get
    raise Error("could not locate runnable browser")
webbrowser.Error: could not locate runnable browser
```
This happend to me on a remote server with only a console.

Maybe simple solution: Catch error, skip the start and maybe give a message on the console.

FinnStutzenstein:
Just catching the error raised in webbrowser.get and print a remainder message.

### DIFF
--- a/openslides/utils/main.py
+++ b/openslides/utils/main.py
@@ -261,15 +261,19 @@ def start_browser(browser_url):
     Launches the default web browser at the given url and opens the
     webinterface.
     """
-    browser = webbrowser.get()
+    try:
+        browser = webbrowser.get()
+    except webbrowser.Error:
+        print('Could not locate runnable browser: Skipping start')
+    else:
 
-    def function():
-        # TODO: Use a nonblocking sleep event here. Tornado has such features.
-        time.sleep(1)
-        browser.open(browser_url)
+        def function():
+            # TODO: Use a nonblocking sleep event here. Tornado has such features.
+            time.sleep(1)
+            browser.open(browser_url)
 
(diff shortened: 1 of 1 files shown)


normanjaeckel:
Thanks for this pull request.

I would prefer that we do not use a general `except` but a `except webbrowser.Error` to be explicit about which error we expect here and what to do if the error occurs. Otherwise it can be very hard to debug. See [here](https://wiki.python.org/moin/HandlingExceptions#General_Error_Catching) for more about this.

The second point is that I would prefer to have only one statement inside a try-block because then it is clear where exactly the error is expected. So here I would put the try-block inside the start_browser function around `browser = webbrowser.get()`

normanjaeckel:
Ok. Ready to merge.

emanuelschuetze:
Ok

Issue #2245, opened 2016-08-02
Unnecessary amount of traffic and lag on editing config texts


matakuka:
When writing text in the /config form, for example, 'general_event_welcome_text', after every keystroke the field is sent as a PUT request, creating traffic, lag and server usage.

emanuelschuetze:
#2261 fixes this issue. I added a 'wait period' of 1 sec (`ng-model-options="{debounce:1000}"`) before a put request was send.

PR #2261, opened 2016-08-09
Fixed missing translations of config options (Fixes #2193)

### LINKED ISSUE #2193: Config values are created always in English

emanuelschuetze:
Fixes also #2245

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -11,7 +11,7 @@
     "angular-csv-import": "~0.0.29",
     "angular-formly": "~7.3.9",
     "angular-formly-templates-bootstrap": "~6.2.0",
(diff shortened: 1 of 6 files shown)


normanjaeckel:
Ok. Travis failed. Fix this, then merge.

emanuelschuetze:
Travis fixed.

Issue #2238, opened 2016-07-29
JS Error: this.getDoc(...) is undefined


normanjaeckel:
I got this error in the console when opening the motion create or edit dialog. The dialog works however.

FinnStutzenstein:
It seems to be a [known Issue](https://github.com/angular-ui/ui-tinymce/issues/264) with tinyMCE 4.3.13. There are some possibilities:

I could specify the version if tinymce-dist in the `bower.json` to 4.3.12. Or I could use the fixed version from mortizpflaume (see the link above) and somehow change the source of tinyMCE to [this modified code](https://cdn.rawgit.com/moritzpflaum/ui-tinymce/master/src/tinymce.js). 

But moritzpflaume pointed out, that this only happens in production. There is no error on demo.openslides.org (Does the code have changed?), so does this need to have a fix?

emanuelschuetze:
@FinnStutzenstein Can you check if tinyMCE 4.4.x works with OpenSlides and ui-tinymce - and fixed this reported issue?

FinnStutzenstein:
@emanuelschuetze Ho do I change the version? This doesn't work (part of the bower.json):
```
 "tinymce-dist": {
      "version": "~4.4.0",
      "main": [
          "tinymce.js",
          "themes/modern/theme.js",
          "plugins/*/plugin.js"
      ]
    }
```
I've ask on [Stackoverflow](http://stackoverflow.com/questions/38831455/how-to-specify-the-version-in-bower-override) yesterday, but I didn't get any answer..

FinnStutzenstein:
@normanjaeckel @emanuelschuetze I would consider one of there two options:

1. Downgrade from 4.3.13 to 4.3.12 --> no error, but maybe some more bugs ([see changelog](https://github.com/tinymce/tinymce-dist/blob/master/changelog.txt#L23))
2. Stay at 4.3.13 and ignore the error and check, whether it appears in a non production environment.

emanuelschuetze:
I want to upgrade to tinyMCE 4.4.x later (for OpenSlides 2.1). But we should wait until this bug is fixed in upstream. - Up to that time I prefer option 1 (downgrade to 4.3.12). Please make a pull request.

normanjaeckel:
This issue remains open so we do not forget to fix this problem finally. Temporary fix is merged.

PR #2266, opened 2016-08-11
Downgrade tinymce-dist to 4.3.12 to avoid Issue #2238

FinnStutzenstein:
It should be a temporary fix until this issue is fixed in a higher version of tinymce-dist.


emanuelschuetze:
Ok. Ready for merge.

PR #2281, opened 2016-08-18
Extended autoupdate API for plugins.

normanjaeckel:


normanjaeckel:
@ostcar What about this implementation?

normanjaeckel:
All example code is to show how plugins can use this exended API. I will remove the example code before merge.

normanjaeckel:
We need a new approach because of the changes made by autoupdate API refactoring and caching system.

PR #2283, opened 2016-08-19
Added generic fields for comments for motions.

normanjaeckel:


emanuelschuetze:
This branch has conflicts. Can you rebase it? Or should I wailt until #2291 is merged?

normanjaeckel:
Rebased ready for test and review and templating.

openslides/motions/config_variables.py
```diff
@@ -151,6 +151,19 @@ def get_config_variables():
...
 
+    # Comments
+
+    yield ConfigVariable(
+        name='motions_comments',
+        default_value='Comment',
+        input_type='text',
+        label='Comment fields for motions',
+        help_text='Use comma separated list of field names for public '
+                  'fields or use special JSON. See manual for syntax.',
```
emanuelschuetze: I prefere it to give an json example with internal and public fiels here.

Can you give me such a json example for testing? I tested with this code which does not work:
```
{{"name": "internal comment", "public": false},{"name": "public comment", "public": true}}
```

normanjaeckel: Your string isn't valid JSON:
Correct should be `[{"name": "internal comment", "public": false},{"name": "public comment", "public": true}]`

openslides/motions/migrations/0003_auto_20160819_0925.py
```diff
@@ -0,0 +1,35 @@
...
+        migrations.AlterModelOptions(
+            name='motion',
+            options={
+                'default_permissions': (),
+                'ordering': ('identifier',),
+                'permissions': (
+                    ('can_see', 'Can see motions'),
+                    ('can_create', 'Can create motions'),
+                    ('can_support', 'Can support motions'),
+                    ('can_see_and_manage_comments', 'Can see and manage comments'),
```
emanuelschuetze: This new permission 'Can see and manage comments' have to add to the staff group in a migrate file. Otherwise an admin cannot edit a  comment field.

openslides/motions/config_variables.py
emanuelschuetze: Please add this short example to the help text (we should use it until we replace it by an improved UI to select public/private via checkbox): 
```
[{"name": "foo", "public": false},{"name": "bar", "public": true}]
```

emanuelschuetze:
Important remark: 
A guest user can see my private comment field (in template and in rest api). I tested with the json example (see above) and an old DB (hope that migrations works).

openslides/motions/config_variables.py
normanjaeckel: Changed.

openslides/motions/migrations/0003_auto_20160819_0925.py
normanjaeckel: It is not possible to add the new permission to a group during migration because the new permission itself is added to the database (django.contrib.auth.models.Permissin) after all migrations are finished.

normanjaeckel:
Tried to log in as guest. REST API does not show the hidden comments:

```
     "comments": [
            null,
            "public content"
        ],
```

For  manager it is:

```       
     "comments": [
            "hidden content",
            "public content"
        ],
```

emanuelschuetze:
Ok, I can confirm it with a fresh database:  private comments are only visible for users with extra comment permission.
But currently the name of private comment fields are printed in motion detail template. Please hide field if they are private.

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -370,3 +370,10 @@ <h3 translate>Version history</h3>
 
   </div>
 </div>
+
+<div class="details">
+    <h3>Motion Comments</h3>
+    <p ng-repeat="field in commentsFields">
+        {{ field.name }}: {{ motion.comments[$index] }}
```
emanuelschuetze: Here we need a check: show privte field names (and values) only if user has right permission.

emanuelschuetze:
One remark. rest looks good.

openslides/motions/static/templates/motions/motion-detail.html
normanjaeckel: Done.

openslides/motions/config_variables.py
```diff
@@ -151,6 +151,20 @@ def get_config_variables():
...
 
+    # Comments
+
+    yield ConfigVariable(
+        name='motions_comments',
+        default_value='Comment',
+        input_type='text',
+        label='Comment fields for motions',
+        help_text='Use comma separated list of field names for public '
+                  'fields or use special JSON. Example: [{"name": "Hidden Comment ", '
```
emanuelschuetze: Remove the whitespace after "Hidden Comment" please.

emanuelschuetze:
One last remark. Then you can rebase and merge!

openslides/motions/config_variables.py
normanjaeckel: Ups ...

normanjaeckel: Done.

PR #2290, opened 2016-08-20
Remove our session backend

ostcar:
The OpenSlidesSessionBackend was added, because there was no way to get
all users from a channel-group. But now channels supports that, so we do
not need our session backend anymore.

I am not sure if this pull request is an improvement. On the pro side, we can again support other session-backends then the db-session-backend. Other session-backends can be a lot faster.

On the con side, with this pull request touches the session backend on every autoupdate for each open websocketconnection. So if there are 1000 open connections and one autoupdate per second, there will be 1000 requests to the session backend per second.

Another con is, that access_permissions.get_restricted_data() is now called for each open websocket-connection. Before this pull request, it is only called once per user. This could be a big difference if there is a user with a lot of open websocket-connections, like the anonymous user.

So in conclusion I would prefer not to merge this. But I realise it after I did the work, so I would like to start this discussion anyway.


emanuelschuetze:
@ostcar is this pull request still valid? Should we close this?

emanuelschuetze:
@ostcar ping: Can we close this pr?

PR #2285, opened 2016-08-19
Line numbers and inline editing

CatoTH:
New features:
- Line numbers in the motion detail view (including projector view). The maximum line length is configurable.
- Inline editing of motion text (if the "update"-permission is present and the most recent version of the motion is selected)
- Unit-tests based on Karma

The functions in diff.js are in preparation for the next feature (amendments)

PR #2292, opened 2016-08-20
Line numbers and inline editing

CatoTH:
New features:

- Line numbers in the motion detail view (including projector view). The maximum line length is configurable.
- Inline editing of motion text (if the "update"-permission is present and the most recent version of the motion is selected)
- Unit-tests based on Karma

The functions in diff.js are in preparation for the next feature (amendments).
This commit is the merged version of all commits in https://github.com/OpenSlides/OpenSlides/pull/2285 .

### DIFF
--- a/DEVELOPMENT.rst
+++ b/DEVELOPMENT.rst
@@ -96,3 +96,12 @@ To setup and activate the virtual environment in step c. use::
     > .virtualenv\Scripts\activate.bat
 
 All other commands are the same as for GNU/Linux and Mac OS X.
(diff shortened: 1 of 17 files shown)


emanuelschuetze:
Ok, ready to merge!

donovaly:
Hi CatoTH, many, many thanks for this great new feature! This is a huge improvement for the workflow. Also the feature to go to a certain line to highlight it is great. That is what we need in party conventions almost every time.

Issue #2302, opened 2016-08-23
Permissions to see objects on the projector.


normanjaeckel:
See discussion [here](https://github.com/OpenSlides/OpenSlides/pull/2241/files#diff-6c3a4eee52612ead0a8e34f29168811cR148). Then question is whether we want to move down restrictions for **all** users if a slide needs to display some speakers, candidates, submitters and thelike. Or if we want to implement some kind of reload of the projector if some user changes. Or if we want to add some more data like the name of the user to our models (speaker model, poll option, ...).

normanjaeckel:
@ostcar What do you think?

ostcar:
I like our current way. If you add permission for some users then the only Thing we need is a way to request some users to the rest api. For example: give the users with the ids 5, 46 and 50. Dango Rest framework and js-data support this.

matakuka:
The problems I am having on slides with "can_see_projector" only¹:
- Getting a subitem (Assignment -> agenda_item) only works upon reloading the page (F5). Changes in agenda_item are not updated, unless I have can_see_agenda. This applies both to Agenda.find() and Assignment.loadRelations()
- Displaying speaker lists (in the new overlay or in the existing agenda list of speakers) requires can_see_usernames. A get_requirements of item.speakers.all() does not work, and item.speakers will change before I access the projector.py again.

¹setup: load the non-privileged user in a different browser or in "private mode" so you don't share cache with the projector managing browser

emanuelschuetze:
was done by #2394

PR #2306, opened 2016-08-24
list of speakers overlay (fixes #2211)

### LINKED ISSUE #2211: New list of speakers overlay

matakuka:
This is a different approach from #2241, touching the same issue #2211. 
- it does not create a projector element, just a projector variable "speakeroverlay: true|false"
- the overlay gets the permissions the slide gets. So permission granting issues are avoided.

- only remaining issue from my view: if not user_can_see_names, name lists will be empty.


normanjaeckel:
Thanks for this approach. I am not so happy with implementing non generic features directly into the projector model. I can see some advantages but I do not think they are worth breaking up our (maybe only my) current idea of what the projector is and what the projector elements are.

The permission problem is not completly solved. We have to add (or simply check) permissions to see agenda items to all object (motion, assignment, customslide) slides. Then everybody can see the list of speakers of a motion if this motion is displayed on the projector even if the list of speakers flag is set to False.

What about a more or less lazy reloading of DS objects by the active slide controller if something changes?

In case of MotionSlideCtrl
```javascript
$scope.$watch(function () {
  return Agenda.lastModified() + Motion.lastModified() + Category.lastModified() + User.lastModified() + ...;
}, function () {
    // Now just find, findAll and/or loadReleations of what is necessary here.
});
```

matakuka:
Done without the issues discussed in #2302.

The permission for the agenda_item is already there. 

The problem lies in that agenda_item needs to load a relation, too (agenda_item.user), and thus we need permissions for Users (whom, we don't know on page load, as it changes).

matakuka:
see also #2321, where the issue for the user names will be solved (will need some minor rework in this PR)

normanjaeckel:
My objection is still valid. See above.

@ostcar Some opinion?

ostcar:
I don't understand the question.

matakuka:
@normanjaeckel I need help resolving.

For this (or #2241) I cannot get resolve to work, and I am lost.
- Is a resolve for a single new slide via slidesProvider possible?
- For a resolve upon (re-) loading the projector, it does not seem to work. Probably I am doing it wrong.
As I understand, it is 
(core/site.js)
```
.state('projector', {
                url: '/projector',
                templateUrl: 'static/templates/projector-container.html',
                data: {extern: true},
                onEnter: function($window) {
                    $window.location.href = this.url;
                },
                resolve: {
                    motions: function(Motion) {
                        return Motion.findAll().then(function(motions) {
                            angular.forEach(motions, function(motion) {
                                Motion.loadRelations(motion, 'agenda_item');
                            });
                        });
                    },
                })
```
Then, in the ProjectorCtrl/ProjectorContainerCtrl:
```
.controller('ProjectorCtrl', [
    'motions',
    function(motions) {
    do_something(motions);
 }
```
However, motions is not injected there (core/projector.js):

```
"Error: [$injector:unpr] Unknown provider: motionsProvider <- motions <- ProjectorContainerCtrl
```

normanjaeckel:
You can not use a resolve in site.js to get some vars to the projector.js because these are two different angular apps. The state definition in site.js is only a redirect, see onEnter ... window.location .... There the browser reloads the projector page one time.

To use something like resolve in the slidesProvider is not implemented yet. This would be nice.
Nevertheless I think we should follow either my approach in #2357 or a new proposal of @ostcar. We will discuss this in about 10 days. Before it is not worth doing anything here. But feel free ...

normanjaeckel:
New approach in #2371 .

PR #2308, opened 2016-08-24
Handle duplicates during user import (fixes #1213)

### LINKED ISSUE #1213: New Checkbox for csv participant import: 'really override existing participants'
If you import participants from CSV file multiple times (e.g. you add a line or correct something) you get multiple participants with the same name. Currently, you have to delete duplicated participants manually. So, a new import option (like in motion csv import view!) would be great: "Override existing participants (with the same name)" (default: disabled)

FinnStutzenstein:
Some duplicates-checking and selection how to handle these duplicates.
Pleas test this and suggest some changes.

### DIFF
--- a/openslides/users/static/js/users/site.js
+++ b/openslides/users/static/js/users/site.js
@@ -98,6 +98,9 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
             resolve: {
                 groups: function(Group) {
                     return Group.findAll();
+                },
+                users: function(User) {
+                    return User.findAll();
                 }
             }
         })
@@ -646,10 +649,12 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
 
 .controller('UserImportCtrl', [
     '$scope',
+    '$q',
     'gettext',
+    'gettextCatalog',
     'User',
     'Group',
-    function($scope, gettext, User, Group) {
+    function($scope, $q, gettext, gettextCatalog, User, Group) {
         // import from textarea
         $scope.importByLine = function () {
             $scope.usernames = $scope.userlist[0].split("\n");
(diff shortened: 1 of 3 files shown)


matakuka:
Maybe we can color the buttons differently, if something is chosen? Like "btn-warning" if the existing user is to be overwritten, and "btn-success" if a duplicate will be made. All in red can be confusing if I work the list through 200 entries.

FinnStutzenstein:
Ok from my side. @normanjaeckel, @emanuelschuetze?

openslides/users/static/js/users/site.js
```diff
@@ -782,20 +796,94 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
...
+                                user.duplicate_info += ' ';
+                            if (user_.last_name)
+                                user.duplicate_info += user_.last_name;
+                            user.duplicate_info += ' (';
+                            if (user_.number)
+                                user.duplicate_info += gettext('number') + ': ' + user_.number + ', ';
+                            if (user_.structure_level)
+                                user.duplicate_info += gettext('structure level') + ': ' + user_.structure_level + ', ';
+                            user.duplicate_info += gettext('username') + ': ' + user_.username + ') '+
+                                gettext('exists already.');
```
emanuelschuetze: typo: "already exists."
You have to use "gettextCatalog.getString()" to translate the concatenated string parts.

openslides/users/static/js/users/site.js
```diff
@@ -707,9 +711,17 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
         $scope.pageChanged = function() {
             $scope.limitBegin = ($scope.currentPage - 1) * $scope.itemsPerPage;
         };
+        $scope.duplicateActions = [
+            'keep original',
+            'override new',
+            'create duplicate'
```
emanuelschuetze: Default should be the action "override new" because the most important use case whiche shows duplicates comes if someone want to overide existing users. So please change it.

Wish: Changing all duplicates to an other action with a global select could be helpful.

emanuelschuetze:
Tested. Looks good. Some remarks.

FinnStutzenstein:
@emanuelschuetze ping

emanuelschuetze:
Ok. Successfully tested. Please rebase/stash to one commit.

emanuelschuetze:
@normanjaeckel Ready for your last review and merge.

PR #2310, opened 2016-08-24
Sort supporters in a motion (wip)

FinnStutzenstein:
I'm trying to make it possible to sort submitters in a motion. For this reason, I have to modify the many to many relationship between the users and motions to provide something like a `sort_id`. But the migration fails.

@normanjaeckel Could you give me a hint with the backend database handling?


normanjaeckel:
`ValueError: Cannot alter field motions.Motion.submitters into motions.Motion.submitters - they are not compatible types (you cannot alter to or from M2M fields, or add or remove through= on M2M fields)`

So I think this change is not possible. See http://stackoverflow.com/questions/29450333/django-migration-error-involving-m2m-field. This is not the same case but might be adaptable.

FinnStutzenstein:
@normanjaeckel Please have a look on it. The backend is mostly finished. The important change is the new SubmittersRelationship between motions and users. The migration works and tests on the backend seems to work, too.
I think I have managed the new relationship to get it accessible in the REST framework but there and on the client side are some problems.

normanjaeckel:
Please rebase on top of master and use new autoupdate/permission system stuff: can_retrieve is renamed to check_permissions. Also you may use [this](https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ManyToManyField.through).

FinnStutzenstein:
@normanjaeckel The migration works, this is not the problem.
If you have time, please just play around with this. There are problems to get all this through REST and how to make the DS structure.

emanuelschuetze:
@FinnStutzenstein A rebase on top of master would be useful for more tests. Can you explain how @normanjaeckel can help you exactly?

normanjaeckel:
Ping. Is this valid for 2.1?

normanjaeckel:
I propose taht we discuss this in January on code sprint meeting. Please save your changes and close this pull request.

PR #2313, opened 2016-08-25
Updated translations for DE, ES, PT (used by JS, Django).

emanuelschuetze:


emanuelschuetze:
I'll merge this updated translations now. 
@normanjaeckel Please note the second commit where I removed the old (failed) translation test.

normanjaeckel:
The removing of this test was not correct. The test has nothing to do with translation of config labels but of config default values. See #2320.

PR #2314, opened 2016-08-25
new change password view (fixes #2137)

### LINKED ISSUE #2137: blocker: user login system is broken
- create a new user and give him the username "johndoe"
- set his password to "test"
- try to login

result: one cannot login with johndoe, but OpenSlides created his login info correctly:
johndoe-bug

The whole user login system seems to be broken because for many users I get the same problem. This makes OpenSlides 2.0 quite unusable.

FinnStutzenstein:


openslides/users/static/js/users/site.js
```diff
@@ -312,26 +322,6 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
                     }
                 },
                 {
-                    key: 'default_password',
```
normanjaeckel: If this field is remove, how can an admin reset or change the password of another user?

FinnStutzenstein: in the users list is a new link "change password" in the hover menu. It is a suggestion to solve it like this, but there are some reasons:

- the default password is a little bit confusing, so is the reset button. 
- changing a password is not trivial - there should be a seperate place to do this, because the administrator should be aware of, what he is doing.
- the behavior of the reset button is not clear for an unexperienced user. Why do you have to click reset and could not simply save?

@emanuelschuetze has some more arguments ;)

normanjaeckel:
When I create a new user I just want to set may favourite password immediately. This is not possible if the form field is removed from the create view. See also my old arguments and proposals in #412.

FinnStutzenstein:
So adding the passwordfield if a user is created and hide it if a user is edited is ok?

FinnStutzenstein:
@normanjaeckel I've added the default password field if the user is created.

openslides/users/static/js/users/site.js
```diff
@@ -316,20 +326,9 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
                     type: 'input',
                     templateOptions: {
                         label: gettextCatalog.getString('Default password'),
```
normanjaeckel: What about renaming this field (now) to `First password` or `Initial password` or something like this?

normanjaeckel:
Looks good. Not tested.

openslides/users/static/js/users/site.js
FinnStutzenstein: Yes, this sounds good.

emanuelschuetze:
@FinnStutzenstein As discussed...

* change sting in PDF from "Password:" to "Initial password:"
[see pdf.py](https://github.com/OpenSlides/OpenSlides/blob/b3582f462e032b0f5f855447f7bfcf80e706f65f/openslides/users/pdf.py#L118)
* password change template:
 * change string in  from "Current default passwort" to "Initial password".
 * Add "(can not be changed)" behind the initial password.
 * Add more space below input field. 
 * Change generate icon to 'magic'
 * Change button string from "Change" to "Change password"
 * Add fa-exclamation-triangle icon for warning sentence. Remove "may" from sentence.

FinnStutzenstein:
@emanuelschuetze ping

emanuelschuetze:
Looks very good. 
@normanjaeckel Ready for last review and merge.

openslides/users/static/templates/users/user-change-password.html
```diff
@@ -0,0 +1,50 @@
...
+    </div>
+    <h1 translate>Change password for {{ user.get_short_name() }}</h1>
+  </div>
+</div>
+
+<div class="details">
+  <form name="userForm">
+    <h3 class="text-danger" style="margin-top: 0;">
+      <i class="fa fa-exclamation-triangle"></i>
+      <translate>You override a personally set password from</translate> {{ user.username }}!
```
normanjaeckel: `<translate>You override the personally set password of</translate> {{ user.get_short_name() }}!`

openslides/users/views.py
```diff
@@ -159,12 +159,17 @@ def reset_password(self, request, pk=None):
         """
         View to reset the password using the requested password.
         """
-        user = self.get_object()
-        if request.data.get('password'):
-            user.default_password = request.data['password']
-        user.set_password(user.default_password)
-        user.save()
-        return Response({'detail': _('Password successfully reset.')})
+        # check manager perms
```
normanjaeckel: The permissions are already checked in the `check_permission` method.

openslides/users/views.py
```diff
@@ -159,12 +159,17 @@ def reset_password(self, request, pk=None):
...
+        if request.user.has_perm('users.can_manage'):
+            user = self.get_object()
+            if isinstance(request.data.get('password'), str):
+                user.set_password(request.data.get('password'))
```
normanjaeckel: I still do not understand why the default password is immutable. Why can't we use the new set password dialog to set a new password and also to set the initial password to this value. Then the admin may print a user access data sheet if he likes. @emanuelschuetze ?

tests/integration/users/test_viewset.py
```diff
@@ -172,18 +172,6 @@ def test_reset(self):
         self.assertTrue(User.objects.get(pk=user.pk).check_password(
             'new_password_Yuuh8OoQueePahngohy3_new'))
 
-    def test_reset_to_default(self):
```
normanjaeckel: Please reuse the test for the case that no string is sent to the view.

openslides/users/views.py
emanuelschuetze: Let me explain our idea (from Finn and me): 
* The initial password is set onetime with creation of a new user. Then this initial password is fixed and not changeable.
* The access PDF prints always the initial password. This PDF is always valid. It's not possible that the initial password changed and so the PDF becomes invalid.
* The admin can change the personal password of an user only - like the user can do it in its own password change view. There are two use cases how a admin can do it:
 * The user stands near the admin PC and the user enter his new password.
 * The admin reset the current password to the initial password (by coping of the printed initial password) and says to user: You can reuse the already printed initial password from PDF.

Our idea is to reduce the confusion of resting the user password. There should be only ONE password which can changed. The initial password is for onetime login and printing on PDF.

Hope this is clear now. Otherwise ask me personally.

FinnStutzenstein:

normanjaeckel:
Ok, please rebase.

FinnStutzenstein:
@normanjaeckel It's rebased

PR #2318, opened 2016-08-26
Control the resolution of the projector

FinnStutzenstein:
This feature provides the same projectorview and resolution on every device. One can set a resolution for the main projector and the projector scales (if needed) to fit this resolution. The preview in the sidebar stays at a fixed width, so the height of the preview changes with the current resolution.

The old projector is now available under `/real-projector`. The `/projector` shows the real projector in an iframe to control the display and resolution of the real projector.

I've added the fields `width` and `height` to the projector model and added a route `set_resolution`, because with the multiprojector one should be able to set different resolutions per projector. For now, these fields are unused, because I put the resolution in the global settings in fact that OpenSlides do not have a separate `manage-projector` view. But if someone switches to multiprojector, he hasn't to change much of the code; Just delete the config and get the resolution from the `Projector` instead of `Config`.


openslides/core/config_variables.py
```diff
@@ -155,3 +155,19 @@ def get_config_variables():
...
+        default_value='1024_768',
+        input_type='choice',
+        label='Projector Resolution',
+        choices=(
+            {'value': '800_600', 'display_name': '800x600 (4:3)'},
+            {'value': '1024_768', 'display_name': '1024x768 (4:3)'},
+            {'value': '1280_1024', 'display_name': '1280x1024 (5:4)'},
+            {'value': '1360_768', 'display_name': '1360x768 (16:9)'},
+            {'value': '1680_1050', 'display_name': '1680x1050 (8:5)'},
+            {'value': '1920_1080', 'display_name': '1920x1080 (16:9)'}),
```
emanuelschuetze: It should also bei possible to add customized resolution, e.g. 16:10.

openslides/core/static/js/core/projector.js
```diff
@@ -57,6 +57,60 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+    'Config',
+    function($scope, Config) {
+        // watch for changes in Config
+        $scope.$watch(function () {
+            return Config.lastModified();
+        }, function () {
+            var split = Config.get('projector_resolution').value.split('_');
+            $scope.projectorWidth = parseInt(split[0]);
+            $scope.projectorHeight = parseInt(split[1]);
+            $scope.recalculateIframe();
```
emanuelschuetze: Try to recalculate only if the special config is changed - not if any config changed.

openslides/core/config_variables.py
FinnStutzenstein: How do you handle custom fields now? Like just adding a file somewhere and read and parse it in the config?

emanuelschuetze: Maybe just a textarea/Input with a Special Input Element like Color picker

FinnStutzenstein:
@emanuelschuetze Do you have some good limits for the resolution? I would say the minimum is 800x600 and maximum FullHD? Or even 4K?

openslides/core/urls.py
```diff
@@ -30,6 +30,9 @@
     # View for the projectors are handelt by angular.
     url(r'^projector.*$', views.ProjectorView.as_view()),
 
+    # Original view without resolutioncontrol for the projectors are handelt by angular.
```
normanjaeckel: Typo `handled`

openslides/core/views.py
```diff
@@ -315,6 +329,42 @@ def clear_elements(self, request, pk):
...
+        Example:
+
+        {
+            "width": "1024",
+            "height": "768"
+        }
+        """
+        if not isinstance(request.data, dict):
+            raise ValidationError({'detail': 'Data must be a dictionary.'})
+        if (request.data['width'] <= 0 or
```
normanjaeckel: Here you have to validate that someone does not send something like `{"foo": "bar"}`. At the moment you would get a KeyError.

normanjaeckel:
Please add comments to all places where config related changes (resolution field, template, ...) have to be reverted when we change to multiprojector.

normanjaeckel:
@ostcar may have a look at this.

ostcar:
I don't understand the part on the client side. But if it works I am happy :)

emanuelschuetze:
Ok. Looks very good! - @normanjaeckel Ready for last review and merge.

openslides/core/static/js/core/projector.js
```diff
@@ -57,6 +57,64 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
     }
 ])
 
+// Projector Container Controller
+.controller('ProjectorContainerCtrl', [
+    '$scope',
+    'Config',
+    function($scope, Config) {
+        // watch for changes in Config
+        var last_conf;
```
normanjaeckel: Where is this last_conf set and used?

openslides/core/static/js/core/site.js
```diff
@@ -1076,6 +1087,21 @@ angular.module('OpenSlidesApp.core.site', [
 
         });
 
+        // watch for changes in Config
+        var last_conf;
```
normanjaeckel: Here the same.

openslides/core/static/templates/core/projector-controls.html
```diff
@@ -9,9 +9,31 @@
         <h4 translate>Live view</h4>
       </a>
       <div uib-collapse="isLiveViewClosed" ng-cloak>
+        <style>
```
normanjaeckel: Please indend the inner stuff.

openslides/core/static/templates/projector-container.html
```diff
@@ -0,0 +1,45 @@
...
+<title>OpenSlides &ndash; Projector</title>
+<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0;">
+<link rel="stylesheet" href="static/css/openslides-libs.css">
+<link rel="stylesheet" href="static/css/projector.css">
+<link rel="icon" href="/static/img/favicon.png">
+<script src="static/js/openslides-libs.js"></script>
+
+<div ng-controller="ProjectorContainerCtrl" class="pContainer">
+  <div>
+    <style>
```
normanjaeckel: Here the same.

normanjaeckel:
Some stylistic remarks.

FinnStutzenstein:

PR #2321, opened 2016-08-27
autoupdate for non-privileged projectors

matakuka:
See issue #2302 and PR #2241.

- Fix 1: Autoupdate now checks the projector get_requirement permissions. Thus, these data will be updated after the projector is loaded.
- Fix 2: projector slides that watch user names will check if the users are  in the cache, if not, they will retreive them.

**Result**: Slides update with all information if user has can_see_projector, but information retrieval is restricted to the actual content.

### DIFF
--- a/openslides/agenda/projector.py
+++ b/openslides/agenda/projector.py
@@ -38,6 +38,11 @@ class ItemListSlide(ProjectorElement):
         yield ProjectorRequirement(
             view_class=ItemViewSet,
             view_action='list')
+        for speaker in self.speakers.all():
+            yield ProjectorRequirement(
+                view_class=speaker.user.get_view_class(),
+                view_action='retrieve',
+                pk=str(speaker.user_id))
 
 
 class ListOfSpeakersSlide(ProjectorElement):
@@ -73,4 +78,4 @@ class ListOfSpeakersSlide(ProjectorElement):
                     yield ProjectorRequirement(
                         view_class=speaker.user.get_view_class(),
                         view_action='retrieve',
-                        pk=str(speaker.user_id))
+                        pk=str(speaker.user.pk))
--- a/openslides/agenda/projector.py
+++ b/openslides/agenda/projector.py
@@ -38,11 +38,6 @@ class ItemListSlide(ProjectorElement):
         yield ProjectorRequirement(
             view_class=ItemViewSet,
             view_action='list')
-        for speaker in self.speakers.all():
-            yield ProjectorRequirement(
-                view_class=speaker.user.get_view_class(),
-                view_action='retrieve',
-                pk=str(speaker.user_id))
 
 
 class ListOfSpeakersSlide(ProjectorElement):
@@ -78,4 +73,4 @@ class ListOfSpeakersSlide(ProjectorElement):
                     yield ProjectorRequirement(
                         view_class=speaker.user.get_view_class(),
                         view_action='retrieve',
-                        pk=str(speaker.user.pk))
+                        pk=str(speaker.user_id))
--- a/openslides/assignments/projector.py
+++ b/openslides/assignments/projector.py
@@ -1,5 +1,6 @@
 from openslides.core.exceptions import ProjectorException
 from openslides.core.views import TagViewSet
+from openslides.agenda.views import ItemViewSet
 from openslides.utils.projector import ProjectorElement, ProjectorRequirement
 
 from .models import Assignment, AssignmentPoll
@@ -45,6 +46,15 @@ class AssignmentSlide(ProjectorElement):
                     view_class=AssignmentViewSet,
                     view_action='retrieve',
                     pk=str(assignment.pk))
+                yield ProjectorRequirement(
+                    view_class=ItemViewSet,
+                    view_action='retrieve',
+                    pk=str(assignment.agenda_item_id))
+                for speaker in assignment.agenda_item.speakers.all():
+                    yield ProjectorRequirement(
+                        view_class=speaker.user.get_view_class(),
+                        view_action='retrieve',
+                        pk=str(speaker.user))
                 for user in assignment.related_users.all():
                     yield ProjectorRequirement(
                         view_class=user.get_view_class(),
--- a/openslides/assignments/projector.py
+++ b/openslides/assignments/projector.py
@@ -1,6 +1,6 @@
+from openslides.agenda.views import ItemViewSet
 from openslides.core.exceptions import ProjectorException
 from openslides.core.views import TagViewSet
-from openslides.agenda.views import ItemViewSet
 from openslides.utils.projector import ProjectorElement, ProjectorRequirement
 
 from .models import Assignment, AssignmentPoll
--- a/openslides/assignments/projector.py
+++ b/openslides/assignments/projector.py
@@ -50,11 +50,6 @@ class AssignmentSlide(ProjectorElement):
                     view_class=ItemViewSet,
                     view_action='retrieve',
                     pk=str(assignment.agenda_item_id))
-                for speaker in assignment.agenda_item.speakers.all():
-                    yield ProjectorRequirement(
-                        view_class=speaker.user.get_view_class(),
-                        view_action='retrieve',
-                        pk=str(speaker.user))
                 for user in assignment.related_users.all():
                     yield ProjectorRequirement(
                         view_class=user.get_view_class(),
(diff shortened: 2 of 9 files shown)


openslides/agenda/projector.py
```diff
@@ -38,6 +38,11 @@ def get_requirements(self, config_entry):
         yield ProjectorRequirement(
             view_class=ItemViewSet,
             view_action='list')
+        for speaker in self.speakers.all():
```
normanjaeckel: Why do we need users on the item list slide?

openslides/agenda/projector.py
```diff
@@ -73,4 +78,4 @@ def get_requirements(self, config_entry):
                     yield ProjectorRequirement(
                         view_class=speaker.user.get_view_class(),
                         view_action='retrieve',
-                        pk=str(speaker.user_id))
```
normanjaeckel: I thought that `speaker.user_id` returns the id of the user without hitting the database about the user itself, so not for `speaker.user.pk`. @ostcar Am I right?

openslides/assignments/projector.py
```diff
@@ -45,6 +46,15 @@ def get_requirements(self, config_entry):
                     view_class=AssignmentViewSet,
                     view_action='retrieve',
                     pk=str(assignment.pk))
+                yield ProjectorRequirement(
+                    view_class=ItemViewSet,
+                    view_action='retrieve',
+                    pk=str(assignment.agenda_item_id))
+                for speaker in assignment.agenda_item.speakers.all():
```
normanjaeckel: Why do we need these users on the AssignmentSlide?

openslides/assignments/projector.py
```diff
@@ -45,6 +46,15 @@ def get_requirements(self, config_entry):
...
+                    yield ProjectorRequirement(
+                        view_class=speaker.user.get_view_class(),
+                        view_action='retrieve',
+                        pk=str(speaker.user))
```
normanjaeckel: `speaker.user_id`?

openslides/assignments/static/js/assignments/projector.js
```diff
@@ -22,11 +22,26 @@ angular.module('OpenSlidesApp.assignments.projector', ['OpenSlidesApp.assignment
         // Add it to the coresponding get_requirements method of the ProjectorElement
         // class.
         var id = $scope.element.id;
-        var poll = $scope.element.poll;
 
         // load assignemt object and related agenda item
         Assignment.find(id).then(function(assignment) {
             Assignment.loadRelations(assignment, 'agenda_item');
+            $scope.$watchGroup([assignment.assignment_related_users, $scope.element.poll.options],                                      function(newval, oldval) {
+                if (newval[0] != oldval[0]) {
```
normanjaeckel: Syntax is broken. A function is expected.

normanjaeckel: Later we have a User.findAll and User.bindAll in this controller. Are they still required?

openslides/core/projector.py
```diff
@@ -19,12 +19,23 @@ def check_data(self):
...
+            slide = CustomSlide.objects.get(pk=pk)
             yield ProjectorRequirement(
                 view_class=CustomSlideViewSet,
                 view_action='retrieve',
                 pk=str(pk))
+            yield ProjectorRequirement(
+                view_class=ItemViewSet,
+                view_action='retrieve',
+                pk=str(slide.agenda_item_id))
+            for speaker in slide.agenda_item.speakers.all():
```
normanjaeckel: Why do we need speakers on the custom slide slide?

openslides/core/projector.py
```diff
@@ -19,12 +19,23 @@ def check_data(self):
             raise ProjectorException('Custom slide does not exist.')
 
     def get_requirements(self, config_entry):
+        from openslides.agenda.views import ItemViewSet
```
normanjaeckel: Please add short comment about this "inline" import.

openslides/core/projector.py
```diff
@@ -19,12 +19,23 @@ def check_data(self):
...
         pk = config_entry.get('id')
         if pk is not None:
+            slide = CustomSlide.objects.get(pk=pk)
```
normanjaeckel: rename `slide` to `custom_slide`.

normanjaeckel: Please use the same style as in all other slide (catch DoesNotExist exception ...).

openslides/motions/projector.py
```diff
@@ -64,3 +69,8 @@ def get_requirements(self, config_entry):
                         view_class=TagViewSet,
                         view_action='retrieve',
                         pk=str(tag.pk))
+                for speaker in motion.agenda_item.speakers.all():
```
normanjaeckel: We do we need those users at motion slide?

openslides/motions/projector.py
```diff
@@ -64,3 +69,8 @@ def get_requirements(self, config_entry):
...
+                    yield ProjectorRequirement(
+                        view_class=speaker.user.get_view_class(),
+                        view_action='retrieve',
+                        pk=str(speaker.user))
```
normanjaeckel: `speaker.user_id`

openslides/utils/autoupdate.py
```diff
@@ -95,9 +97,16 @@ def send_data(message):
             'action': 'deleted' if message['is_deleted'] else 'changed'}
         if not message['is_deleted']:
             data = access_permissions.get_restricted_data(full_data, user)
+
             if data is None:
-                # There are no data for the user so he can't see the object. Skip him.
-                continue
+                # send data if user can see the projector and data is projected
```
normanjaeckel: This wont work because our current ProjectorRequirement approach relays on views and viewsets instead of models (root rest elements as we call them). Maybe we have to change these things here and in rest_api.py line 127.

openslides/agenda/projector.py
matakuka: this is a bit on confusion on my side. I will need them for the list of speakers overlay, but this belongs in the corresponding PR

openslides/assignments/projector.py
matakuka: you're right. Still I'm removing this part and reinsert it in the other PR

openslides/assignments/static/js/assignments/projector.js
matakuka: I don't think so, but I'm not sure

matakuka: I cannot see where syntax is broken. I compared $watchgroup syntax with the [doc](https://docs.angularjs.org/api/ng/type/$rootScope.Scope), and can't see what's wrong

openslides/utils/autoupdate.py
matakuka: As I tested it, it worked just fine. Get_all_requirements (core/projector.py:110), as I understand it, goes through all apps and fetches all requirements from all views currently active.

openslides/assignments/static/js/assignments/projector.js
normanjaeckel: `$watchGroup(watchExpressions, listener);`
watchExpressions: Array of expressions that will be individually watched
listener: Callback function.

Your listener is not a function but starts with `if (newval[0] ...` where newval is not defined e. g.

matakuka: ah, it was just an excess of whitespace before the "function", so the function probably was out of window  :)

openslides/utils/autoupdate.py
normanjaeckel: No. get_all_requirements fetches all requirements from all currently active projector elements. But then you have to go through these requirements and decide whether the current data match one of these. See utils/rest_api.py:127. The method is_currently_required does not work because it expects a view instance. and not a model instance.

When I created this feature/system there was no need to restrict autoupdate data separatly because we used to send also autoupdate date through an internal GET request. But now we don't so we have to refactor the whole ProjectorRequirements system.

openslides/agenda/projector.py
normanjaeckel: Ok. But care of my objection in the other pull request. Maybe it is not worth spending much time on that approach.

openslides/agenda/projector.py
ostcar: As far as I know: yes

openslides/utils/autoupdate.py
matakuka: Ah, I see now. I'm getting the requirements and then grant all permissions without checking if the requirement is fulfilled.

normanjaeckel: I will try something. See #2351.

normanjaeckel:
Done by #2394 .

PR #2323, opened 2016-08-28
Cache system for OpenSlides instances

ostcar:
System that caches instances.

The cache is automaticly updated through the autoupdate system.

It uses the django cache system, so the django cache has to be
configured.

The rest API should only use this cache system when it gets GET
requests.

TODO:
- [x] find a way to get the access_permission in the cache system
- [ ] change the viewsets to use the cache
- [ ] Test this

@normanjaeckel could you have a look at the two first todos?

### DIFF
--- a/openslides/utils/instance_cache.py
+++ b/openslides/utils/instance_cache.py
@@ -43,7 +43,7 @@ def get_instance_list(collection, user=None):
 
     # Get all instances out of the cache
     cache_keys = [get_instance_cache_key(collection, id) for id in cached_ids]
-    for cache_key, cached_instance in cache.get_many(cache_keys):
+    for cache_key, cached_instance in cache.get_many(cache_keys).items():
         # TODO: Test if cached_instance is None for instances, that do not exist in the cache
         if cached_instance is None:
             cached_instance = get_instance_from_db(
@@ -95,8 +95,9 @@ def update_instance(collection, id):
     # Add the id of the instance in the instance list
     # Do nothing if the ids are currently not in the cache
     cache_key = get_instance_list_cache_key(collection)
-    cached_ids = set(cache.get(cache_key))
+    cached_ids = cache.get(cache_key)
     if cached_ids is not None:
+        cached_ids = set(cached_ids)
         cached_ids.add(id)
         cache.set(cache_key, cached_ids)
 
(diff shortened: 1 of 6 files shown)


tests/integration/utils/instance_cache.py
```diff
@@ -0,0 +1,118 @@
...
+
+        with self.assertNumQueries(0):
+            instance_list = list(instances_generator('core/customslide'))
+        self.assertEqual(len(instance_list), 3)
+
+    def test_with_some_objects_in_the_cache(self):
+        CustomSlide.objects.create(title='test cs1')
+        CustomSlide.objects.create(title='test cs2')
+        list(instances_generator('core/customslide'))
+        CustomSlide.objects.create(title='test cs3')
```
ostcar: This test fails but will probably work in production.

CustomSlide.objects.create() sends a signal to the autoupdate system, which updates the cache. So in production the cache has all data. But in the test environment Channels does not work and therefore the cache system does not know about the third CustomSlide object.

One way to solve this is to not cache the list of ids in a collection but always get this list from the database. This will still be a big performance boost because the only db queries there will be are from Model.objects.value_list('id', flat=True), which are very fast.

There are other problems by caching the ids, because there can be more then one thread, that alters the list of ids at the same time. But on the other hand I would love to have a full cache. @normanjaeckel what do you think?

openslides/utils/instance_cache.py
```diff
@@ -0,0 +1,171 @@
...
+            user = AnonymousUser()
+
+        if isinstance(user, int):
+            # user is the id of an user instance.
+            # TODO: get_instance could be used here to get the cached version of
+            #       the user.
+            user = User.objects.get(pk=user)
+
+        Model = get_model_from_collection_string(collection)
+        # TODO: access_permissions is normaly get by get_access_permissions().
```
normanjaeckel: We can have a look whether switching to a classmethod is possible.

normanjaeckel:
Is it possible to retrieve all objects from the cache where the cache key start with a part of a string? Like: Give me all of `motions/motion:`. If yes, we can omit caching the list request but just retrieving all retrieve requests.

Anonther thing: Can we outsource the caching in writing threads to a special (unique) channel and therefor ensure that we do not get a race condition if two threads try to write to the same list cache?

normanjaeckel:
Changing viewsets should be easy: Just override `list()` and `retrieve` and double check customizations in out apps.

ostcar:
In some cache backends it is possible to work with key-prefixes but this ways are not recommended, see for example: http://redis.io/commands/KEYS
And the django cache api does not support this.

With a key-value-store you always need keys to get the value.

The thing with the writing thread could be possible, at least for the future. I have a look at this.

ostcar:
Another solution for the thread problem is to use special redis commands, if the redis backend is used, like http://redis.io/commands/append. On big assemblies we can recommend redis, because django channels will need redis anyway. On smaller assemblies, the threading problem is not relevant. It is no problem at all, when sqlite is used and therefore not more then one thread.

ostcar:
#2404 closes this

PR #2330, opened 2016-08-29
Fixed autoupdate for deleted instances. Fixed #2329.

### LINKED ISSUE #2329: Agenda Tree does not autoupdate correctly
**Problem:**
The agenda list (agenda/base.js:AgendaTree) keeps deleted entries. A page reload does not help if you have cache enabled.

**Probable cause:**
Autoupdate does not correctly send information on deleted data

The cause for this starts in utils/autoupdate.py: 119
```
                 'pk': root_instance.pk,
```
The root_instance calling the "send delete information" is queried for "which object is deleted?", but the object *is already deleted* at this point and can't be queried anymore. In the end, *pk:null* is sent to javascript.

AgendaTree currently seems to be the only function affected, because it uses the (correct, desired) *"resolve, then watch all updates sent per autoupdate"* model. 

All functions correctly using the resolve should have this problem. = most functions don't use resolve the way it is intended

normanjaeckel:
@emanuelschuetze Please test. I still get "Kein Element gefunden" messages in JS console and I do not know where they come from.

### DIFF
--- a/openslides/utils/autoupdate.py
+++ b/openslides/utils/autoupdate.py
@@ -109,19 +109,23 @@ def inform_changed_data(instance, is_deleted=False):
         # Instance has no method get_root_rest_element. Just ignore it.
         pass
     else:
+        message_dict = {
+            'collection_string': root_instance.get_collection_string(),
+            'pk': root_instance.pk,
+            'is_deleted': is_deleted and instance == root_instance,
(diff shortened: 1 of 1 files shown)


emanuelschuetze:
Tested successfully. Autoupdate for removed item works now. 
BTW: I get no "Kein Element gefunden" message in console...

emanuelschuetze:
You can merge it...

PR #2334, opened 2016-08-30
Migration for new permission matrix (fixes #2315)

### LINKED ISSUE #2315: Add migrations for new permission matrix
We need a migration file from v2.0 to v2.1. Therefor we need a concept how to handle old (maybe customized) and new groups.

FinnStutzenstein:
I've implemented @normanjaeckel suggestion for a migration. Additionally some changes in `create_or_reset_admin_user`: If the group with `pk=3` is not found, the group `Staff` will be created. Also the two important permissions for managing (`users.can_see_name` and `users.can_manage`) are added.


FinnStutzenstein:
I'll fix the travis error tomorrow; It is only some stylistic things.

The [personal_data_old_.tar.gz](https://github.com/OpenSlides/OpenSlides/files/444820/personal_data_old_.tar.gz) is an old version of the database. I've added a user `test` without any group, so he should join the group `oldRegistered`.

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,47 @@
...
+def migrate_groups_and_user_permissions(apps, schema_editor):
+    """
+    This function migrates the database to the new groupslogic:
+    - Rename Group 'Anonymus' (pk=1) to 'Default'
+    - Rename Group 'Registered users' (pk=2) to 'oldRegisteres'
+    - Add all users who are not in any group to this group (pk=2)
+    - Add all permissions of 'old'Registeres' to all other groups (exclude 'Default')
+
+    But ony do this migration if:
+    - there are users in the database
```
FinnStutzenstein: I mean groups...

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,47 @@
...
+    - the name of the first group is 'Guests'.
+    """
+    User = apps.get_model('users', 'User')
+
+    if Group.objects.all().count() != 0:
+        if Group.objects.get(pk=1).name == 'Guests':
```
normanjaeckel: Maybe we also have to change users/signals.py:14 so that we do not create additional groups if there are any groups in the db not just pk 1 to pk 4.

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.7 on 2016-08-01 14:54
+from __future__ import unicode_literals
+
+from django.contrib.auth.models import Group
```
normanjaeckel: Is it possible to get Group also from app registry and not by import? See line 20.

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,47 @@
...
+from django.db import migrations, models
```
normanjaeckel: The import of models seam to be unnecessary.

openslides/users/models.py
```diff
@@ -35,13 +37,17 @@ def create_or_reset_admin_user(self):
         """
         Creates an user with the username 'admin'. If such a user already
         exists, resets it. The password is (re)set to 'admin'. The user
-        becomes member of the group 'Staff' (pk=3).
+        becomes member of the group 'Staff'. This group are added the two
```
normanjaeckel: Typo: `The two permissions ... are added to this group.`

normanjaeckel:
Needs more tests but looks good.

openslides/users/migrations/0004_groups.py
FinnStutzenstein: But you can not delete the group with pk=1, so `.exists()` will always return true, if once groups are created. Maybe I could change the line just testing for group 1 to exists?

openslides/users/migrations/0004_groups.py
FinnStutzenstein: No, it's not, because in the `users/models.py` we don't specify a model `Group`, so OpenSlides uses the default group model from Django. When adding
```
g = apps.get_model('users', 'Group')
```
I got a `LookupError: App 'users' doesn't have a 'group' model`.

FinnStutzenstein:
@normanjaeckel I've removed the `UsersError` from `models.py`. But now, `UsersError` isn't used anymore, so shall I delete it from the `exceptions.py` or delete the whole file, because it is the only error-class in this file?

FinnStutzenstein:
The Travis error is weird. I added
```
print(Group.objects.get_or_create(name='Staff'))
```
before the `staff, _ = Group....`. I expect to get a tupel with the object and the created status, like described [here](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#get-or-create).

What I get with `python manage.py createsuperuser` (expected):
```
(<Group: Staff>, False)
Admin user successfully reset.
```

And with `coverage run ./manage.py test tests.unit.users`:
```
<MagicMock name='Group.objects.get_or_create()' id='3043999020'>
```
No tupel is returned, so the assignment to `staff, _` fails. But what for an object is there passed?
@normanjaeckel Do you have an idea what went wrong?

openslides/users/migrations/0004_groups.py
normanjaeckel: Maybe you can get the model from the `auth` app.

normanjaeckel:
See tests/unit/users/test_models.py:334.

openslides/users/migrations/0004_groups.py
normanjaeckel: Line 23 should test `exists()`, line should not use and uncatched `get` because it is possible that the groups are deleted in the database if someone manipulated it. Our migration file should not break in this case. Last point: My question was about the check in users/signals.py:14 not about this line.

FinnStutzenstein:
The mock is set up for the function `get_or_create`. But now, this function is called two times: Once with `self.get_or_create` and once with `Group.objects.create`. How can this be differentiate?

normanjaeckel:
It is a different get_or_create function. The group manager and not only the user manager should be mocked. `mock_group.objects.get_or_create = MagicMock(return_value=(expected_group, True))`

FinnStutzenstein:
@normanjaeckel ping

openslides/users/migrations/0004_groups.py
normanjaeckel: Ping.

openslides/users/migrations/0004_groups.py
normanjaeckel: Ping.

emanuelschuetze:
Successfully tested a migration from 2.0 to this branch:
1. Run fresh db with 2.0 and created new group and users:
* without group (registered)
* with staff group only
* with delegates group only
* with staff/delegates group
* with new group only

2. Migrate to current branch and look at group overview: Groups show the old names and permissions. All users has the right groups and can login.

So, it's ok for me. Ready for last review and merge.

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,49 @@
...
+# Generated by Django 1.9.7 on 2016-08-01 14:54
+from __future__ import unicode_literals
+
+from django.contrib.auth.models import Group
+from django.db import migrations
+
+
+def migrate_groups_and_user_permissions(apps, schema_editor):
+    """
+    This function migrates the database to the new groupslogic:
```
normanjaeckel: Missing whitespace between groups and logic.

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,49 @@
...
+    - Rename Group 'Anonymus' (pk=1) to 'Default'
```
normanjaeckel: Typo Anonymous

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,49 @@
...
+    - Rename Group 'Registered users' (pk=2) to 'oldRegisteres'
```
normanjaeckel: Typo oldRegistered. Maybe we can choose something smarter like "Previous group Registered" or something similar.

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,49 @@
...
+    - Add all users who are not in any group to this group (pk=2)
+    - Add all permissions of 'old'Registeres' to all other groups (exclude 'Default')
```
normanjaeckel: Typo.

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,49 @@
...
+
+    But ony do this migration if:
```
normanjaeckel: Typo only

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,49 @@
...
+    - the name of the first group is 'Guests'.
+    """
+    User = apps.get_model('users', 'User')
+
+    if Group.objects.all().count() != 0:
+        if Group.objects.get(pk=1).name == 'Guests':
+            Group.objects.filter(pk=1).update(name='Default')
+            Group.objects.filter(pk=2).update(name='oldRegistered')
+
+            group_old_registered = Group.objects.get(pk=2)
```
normanjaeckel: This line may raise DoesNotExist.

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,49 @@
...
+
+            users = User.objects.all()
+            for user in users:
+                if user.groups.all().count() == 0:
```
normanjaeckel: Here you can also uses exists()

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,49 @@
...
+
+    But ony do this migration if:
+    - there are groups in the database
+    - the name of the first group is 'Guests'.
+    """
+    User = apps.get_model('users', 'User')
+
+    if Group.objects.all().count() != 0:
+        if Group.objects.get(pk=1).name == 'Guests':
+            Group.objects.filter(pk=1).update(name='Default')
```
normanjaeckel: If you use a get query above then you can reuse the instance and do not have to filter it again.

openslides/users/migrations/0004_groups.py
```diff
@@ -0,0 +1,49 @@
...
+            Group.objects.filter(pk=2).update(name='oldRegistered')
```
normanjaeckel: Here the same. See below.

normanjaeckel:
Thanks for this work. Some new and some old remarks still open.

normanjaeckel:
Also please rebase commits and the result on top of master. Thanks.

FinnStutzenstein:
@normanjaeckel Ping

PR #2337, opened 2016-09-01
Add title to user.get_short_name().

emanuelschuetze:


normanjaeckel:
Hm. What if I want to have a list of all users by name and then want to sort. The title breaks the sorting and I have to use extra logic to concat last name and first name on client side (e. g. in a plugin). Therefor we should offer a different method short_name_with_title

emanuelschuetze:
Suggestion: I can implement such a new method only in JavaScript (users/base.js) which will be used in future by our new pdfmake generated documents. Then I'll close this pr. Ok?
 (Currently the ballot pdf from reportlab uses still the get_short_name methoed from users/model.py)

normanjaeckel:
Ok.

emanuelschuetze:
closed, see #2338

Issue #2342, opened 2016-09-03
Label for "commit a bill"


normanjaeckel:
In complex workflow we have a state "committed a bill" with the action word "Commit a bill". Is this correct? I think the word we need is "committed" / "commit" but the extra part "a bill" does not belong to the sense of this state because we do not deal about legislative bills in special but about motions in general.

emanuelschuetze:
My suggestion for a much better wording:
 "Refered to committee" / "Refer to committee" (action word) 
(In German we use already the right expression: "In Ausschuss verwiesen" / "In Ausschuss verweisen")

@normanjaeckel ok? If yes you can just change it in #2344.

emanuelschuetze:
and the recommendation could be:
"Referral to committee" (German: Verweisung/Überweisung in Ausschuss? - I'm not sure here)

normanjaeckel:
Ok. Changed.

PR #2344, opened 2016-09-03
Added recommendations for motions.

normanjaeckel:


normanjaeckel:
@emanuelschuetze You can use the new config variable in the template to set a string like "Antragsvorbereitungskommission". :-)

The connection the the new motion comment fields require merging of this feature first.

openslides/motions/models.py
```diff
@@ -59,6 +60,15 @@ class Motion(RESTModelMixin, models.Model):
     This attribute is to get the current state of the motion.
     """
 
+    recommendation = models.ForeignKey(
+        'State',
+        related_name='+',
```
emanuelschuetze: Are you sure to use '+' as related_name?

openslides/motions/config_variables.py
```diff
@@ -39,6 +39,15 @@ def get_config_variables():
         subgroup='General')
 
     yield ConfigVariable(
+        name='motions_recommendations_by',
+        default_value='Recommendation committee',
+        label='Name of recommendation committee',
+        weight=312,
+        group='Motions',
+        subgroup='General',
+        translatable=True)
```
emanuelschuetze: Move the config to the end of the general subgroup. 
How can we deactivate the Recommendation field in template? Maybe we need a enable config like for amendments? (because not all users want to use/see this special recommendation feature)

openslides/motions/models.py
normanjaeckel: `+` disableds backwards relation: https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ForeignKey.related_name

openslides/motions/config_variables.py
normanjaeckel: Moved.

What about disabling recommendation feature if the label (this config var) is empty?

openslides/motions/signals.py
```diff
@@ -66,16 +72,19 @@ def create_builtin_workflows(sender, **kwargs):
...
     state_2_7 = State.objects.create(name=ugettext_noop('not concerned'),
                                      workflow=workflow_2,
                                      action_word='Do not concern',
+                                     recommendation_label='No concernment',
                                      versioning=True,
                                      css_class='default')
     state_2_8 = State.objects.create(name=ugettext_noop('commited a bill'),
                                      workflow=workflow_2,
                                      action_word='Commit a bill',
+                                     recommendation_label='Commitment',
```
emanuelschuetze: see my suggestion in #2342

openslides/motions/static/templates/motions/motion-detail.html
```diff
@@ -130,6 +130,32 @@ <h3 ng-if="!motion.isAllowed('change_state')" class="heading" translate>State</h
             {{ motion.state.name | translate }}
           </div>
 
+          <!-- Recommendation -->
+          <h3 ng-if="!motion.isAllowed('change_recommendation')" class="heading" translate>Recommendation</h3>
+          <div ng-if="motion.isAllowed('change_recommendation')" class="heading">
+            <span uib-dropdown>
+              <a href id="recommendation-dropdown" class="drop-down-name" uib-dropdown-toggle>
+                <translate>Recommendation</translate>
+                <i class="fa fa-star"></i>
```
emanuelschuetze: Use "fa-cog" icon (as for states). Its a generic "change" icon.

openslides/motions/config_variables.py
emanuelschuetze: Yes, better. 
Or maybe: Disable recommendation field if not any recommendation_label is set in signals.py?? But by default we will deliver the signals with recommendation_labels... And a "normal" admin does not know about changing signals.py... So disabling with empty string is easier  - but should explained in a help text. As well as the hint to change the lables in workflow in signals.py.

emanuelschuetze:
Ok, some remarks.

openslides/motions/static/templates/motions/motion-detail.html
normanjaeckel: Done.

openslides/motions/signals.py
normanjaeckel: Done.

openslides/motions/config_variables.py
normanjaeckel: Ok. Recommendation section is now hidden if config value is empty. Help text added. The other "feature" looks a little bit overkill.

normanjaeckel:
All done. I had to rebase everything. Please test again.

emanuelschuetze:
Ok. looks good.

PR #2352, opened 2016-09-05
Fixing resolving of Config for projector resolution (fixes #2349)

### LINKED ISSUE #2349: Projector resolution: Config.get() is undefined
Since #2318 I'm getting "Config.get(...) is undefined (/static/js/core/projector.js:70:24)" when working on projector slides.
I think it must be `Config.find(...).then( function() {do something})` in order to work?

(also typo : last-conf.height in line 72 should be last_conf)

FinnStutzenstein:
Now resolves Config.

### DIFF
--- a/openslides/core/static/js/core/projector.js
+++ b/openslides/core/static/js/core/projector.js
@@ -67,13 +67,15 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
         $scope.$watch(function () {
             return Config.lastModified();
         }, function () {
-            var conf = Config.get('projector_resolution').value;
             // With multiprojector, get the resolution from Prjector.get(pk).{width; height}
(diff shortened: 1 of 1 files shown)


normanjaeckel:
I do not understand why the config should have to be find here. The config is load in core/base.js. Then is is available in the DS and you should always be able to get it via $scope.config(...).

normanjaeckel:
See #2349

FinnStutzenstein:
Now the controller accesses the config via `$scope.config` but skips if `$scope.config` is not a function. This is the case if `loadGlobalData` hasn't finished yet, because until then `$scope.config` is undefined.

normanjaeckel:
Ok, please rebase.

FinnStutzenstein:
@normanjaeckel Ping

normanjaeckel:
Ok.

emanuelschuetze:
Ok.

PR #2355, opened 2016-09-05
Bugfixes: line numbering

CatoTH:
Fixes two bugs:
- the line numbers where shown bold if they ocurred within bold text
- when a new inline element started near the end of a line and the length of the first word exceeds the remaining space of the line, the first word was broken into two parts. The better solution would be to break the line before the inline element starts.


normanjaeckel:
Ok.

emanuelschuetze:
Ok.

Issue #2322, opened 2016-08-27
(Restricted) data are not cleared on logout


matakuka:
Admin logs in on browser, logs out, Joe logs in on the same browser, has the cache of Admin. 

Joe won't receive any updates he is not allowed to receive, but can still access the (cached) data. 

The only thing stopping him from doing so is that the data is not _shown_. Still, it is a security concern, and I have had cases where I could see data I was not allowed to (user names).

`sessionStorage.clear();` on logout might resolve this; we will need to go fetching all the resolves Joe should have access to when Joe logs in, which might affect the user experience in terms of speed.

matakuka:
This will affect #2295, as with this, the "beginning" will not be on page load, but on user login.

normanjaeckel:
I can not reproduce this with current master.

PR #2357, opened 2016-09-05
Projector requirement

normanjaeckel:
Heavy changes in projector/autoupdate/permission related system. There are a lot of issues to do. Please check carefully the functionality of a non privileged projector user. I hope I fixed everything on server side here. There are some TODOs left. @emanuelschuetze or @matakuka If you have some time, please test everything here and let me know, what else is missing.

(Example: Some watch-last-modified user-find constructions are missing on motion detail, assignment details and assignment poll slide.)

Attention: This work won't be compatible with our current master of OpenSlides/openslides-votecollector.

### DIFF
--- a/openslides/agenda/access_permissions.py
+++ b/openslides/agenda/access_permissions.py
@@ -5,7 +5,7 @@ class ItemAccessPermissions(BaseAccessPermissions):
     """
     Access permissions container for Item and ItemViewSet.
     """
-    def can_retrieve(self, user):
+    def check_permissions(self, user):
         """
         Returns True if the user has read access model instances.
         """
@@ -19,14 +19,19 @@ class ItemAccessPermissions(BaseAccessPermissions):
 
         return ItemSerializer
 
-    def get_restricted_data(self, full_data, user):
+    def get_restricted_data(self, full_data, user, id):
         """
         Returns the restricted serialized data for the instance prepared
         for the user.
         """
-        if (self.can_retrieve(user) and
+        if (user.has_perm('agenda.can_see') and
             (not full_data['is_hidden'] or
              user.has_perm('agenda.can_see_hidden_items'))):
+            data = full_data
+        elif self.check_projector_requirements(user, id):
+            # In this case it is already checked that the projector user needs
+            # this item whether it is a hidden item or not. So we do not have
+            # to care about this.
             data = full_data
         else:
             data = None
--- a/openslides/agenda/projector.py
+++ b/openslides/agenda/projector.py
@@ -1,15 +1,14 @@
-from openslides.core.exceptions import ProjectorException
-from openslides.utils.projector import ProjectorElement, ProjectorRequirement
-
+from ..core.exceptions import ProjectorException
+from ..utils.projector import ProjectorElement, ProjectorRequirement
+from .access_permissions import ItemAccessPermissions
 from .models import Item
-from .views import ItemViewSet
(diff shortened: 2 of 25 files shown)


matakuka:
(user has permissions "can_see_projector only:)

1. at projector page load, *all* users are retrieved, with additional info (name, number, about_me, ...)
2. *all* users are retrieved when a displayed list of speakers is updated.
3. all items currently displayed receive updates
4. users that were updated when not shown, but after projector loading, will then have outdated names.
5. Queries for agenda/item, probably those related to loadRelations, result in 404 (not found) errors (not 403!). 
6. Election titles (from agenda_item) are not shown. On motions, titles *are* shown.
7. On motions, the "state" is missing.
8. The motion on motion slides is cached. If I change the motion text when it is not displayed, and display it again, the projector will show the *old* motion text.
9. When I reload the projector when a agenda/list-of-speakers is displayed, I get only an empty template until the first change in the list of speakers.

I think that at least for non-privileged users, resolving is not a good approach.The info cached is potentially seriously outdated. New items and changed items will only be received if they are changed while displayed. 
In order to get current info, non-privileged users need to get the item/info from the server the moment it is displayed (and then listen for autoupdate in case updates happen while displayed).

Resolving and then autoupdating only works correctly if the user has the rights to *always* see *all* changes. In other cases, old data is cached and data the user may not be allowed to see, (1. and #2322).

normanjaeckel:
> 1. at projector page load, all users are retrieved, with additional info (name, number, about_me, ...)

I can not confirm this: Tested with projected custom slide. No access to rest/users/user seen in server log.

> 2. all users are retrieved when a displayed list of speakers is updated with a user not yet cached.

Confirmed that all users are fetched. This is a bug in SlideListOfSpeakersCtrl where we use User.findAll() instead of just fetching the required users. Server looks ok.

> 3. all items currently displayed receive updates

So no problem here?

> 4. users that were updated when not shown, but after projector loading, will then have outdated names.

Confirmed for list of speakers slide. There is a watch expression missing in SlideListOfSpeakersCtrl.

> 5. Queries for agenda/item, probably those related to loadRelations, result in 404 (not found) errors (not 403!). 

This but I know. I have to care about it (later). I think it is only relevant for hidden agenda items.

> 6. Election titles (from agenda_item) are not shown. On motions, titles are shown.

I can see agenda numbering on assignment slide. But this numbering is only autoupdated if the slide is open. This is a main problem: Slides do not recall their `find` or `findAll` expressions if the projector manager switches between them. Motion and custom slide the same.

> 7. On motions, the "state" is missing.

I can not confirm. I can see state and autoupdate if the slide is open. If not we have the same problem as described in my comment to no. 6.

> 8. The motion on motion slides is cached. If I change the motion text when it is not displayed, and display it again, the projector will show the old motion text.

The same as no. 6 and 7.

> 9. When I reload the projector when a agenda/list-of-speakers is displayed, I get only an empty template until the first change in the list of speakers.

I can not confirm. I open a list of speakers slide and then press F5. The slide is correct afterwards.

matakuka:
I'll later retest the things you could not confirm.
My setup for the test was: fresh database, user managing the slides on firefox, user with can_see_projector only on chrome.

normanjaeckel:
Please test custom slide with my new commit. Maybe this is a valid approach.

normanjaeckel:
We use another approach, see #2388.

PR #2360, opened 2016-09-06
Highlighting and jumping to lines

FinnStutzenstein:
~~One problem:
The motion text is rendered in the `linenumbers.js`. I've added span-tags around each line and provided an id `L + lineNumber`. So every line has an id like `L3` for line 3. But these id's are not available in javascript. Trying to click on scroll should scroll to line 1. Additionally the context from `L1` should be changed (see `$scope.scroll` in `site.js`).
For testing I've added some text above. Each line has an id `LC + lineNumber`. Jumping to them works. Even replacing the content inside works.
Does someone know how to make the id's of the motion text accessible and known in javascript?~~

@emanuelschuetze Please test, should be functional.

### DIFF
--- a/openslides/core/views.py
+++ b/openslides/core/views.py
@@ -192,7 +192,8 @@ class ProjectorViewSet(ReadOnlyModelViewSet):
         elif self.action == 'metadata':
             result = self.request.user.has_perm('core.can_see_projector')
         elif self.action in ('activate_elements', 'prune_elements', 'update_elements',
-                             'deactivate_elements', 'clear_elements', 'control_view', 'set_resolution'):
+                             'deactivate_elements', 'clear_elements', 'control_view',
+                             'set_resolution', 'set_scroll'):
             result = (self.request.user.has_perm('core.can_see_projector') and
                       self.request.user.has_perm('core.can_manage_projector'))
         else:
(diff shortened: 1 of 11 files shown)


emanuelschuetze:
@FinnStutzenstein Works good!
As discussed some minor issues:
* Use Enter button to submit.
* Use "Line" instead of "Line ..." for placeholder text.

FinnStutzenstein:
@emanuelschuetze fixed issues.

emanuelschuetze:
Please rebase on top of master - and resolve the conflicts.

FinnStutzenstein:
Now, the 'also on projector' button toggles the highlighting. So you can remove this.
Another big problem: In normal and inside mode, the line numbers aren't the displayed line numbers from the display:
```
L1: 1 short text 2 text long text ..... bla
L2: bla bla 3 another text..
```
So motion line 3 is in displayed line 2. There is no possibility to get the display line number because this depends on the lengths of the previous lines. I can scroll to a displayed line, but not to a actual motion line in these both modes.

Just the outside mode works fine, because the display line is the same as the motion line: 
```
L1: 1 short text
L2: 2 text long text ...... bla bla bla
L3: 3 another text
```

My workaround for this is now just to scroll on projector if the mode is outside.

FinnStutzenstein:
@emanuelschuetze I have now a different approach. Please check the new behavior.

emanuelschuetze:
Works very good now (tested with outside/inside line numbers)!
@FinnStutzenstein Just a minor wording suggestion: Use projector icon instead of long "Also on projector" checkbox text and add a hover tooltip "Show highlighted line also on projector." 

@normanjaeckel Ready for your review!

FinnStutzenstein:
@emanuelschuetze Done.

emanuelschuetze:
Looks good. Ready for merge.

openslides/core/views.py
```diff
@@ -428,6 +429,25 @@ def control_view(self, request, pk):
...
+        /rest/core/projector/<pk>/set_scroll/ with a new value for scroll.
+        """
+        if not isinstance(request.data, int):
+            raise ValidationError({'detail': 'Data must be an int.'})
+
+        projector_instance = self.get_object()
+        projector_instance.scroll = request.data
+
+        projector_instance.save()
+        message = 'setting scroll to {scroll} was successful.'.format(
```
normanjaeckel: `message = 'Setting scroll to {scroll} was successful.'.format(`

FinnStutzenstein:
@normanjaeckel Done. I will check this checkbox next time.

PR #2364, opened 2016-09-07
Move inline editing methods into a separate service

thoessl:
This moves about a hundred lines from the MotionDetailCtrl into a separate service in a new file.
Extra testing if everything still works as intended would be appreciated :-)

The PDF-function would be another candidate for a separate service in this file.

### DIFF
--- a/openslides/motions/static/js/motions/motion-services.js
+++ b/openslides/motions/static/js/motions/motion-services.js
@@ -1,0 +1,121 @@
+(function () {
+
+"use strict";
+
+angular.module('OpenSlidesApp.motions.motionservices', ['OpenSlidesApp.motions', 'OpenSlidesApp.motions.lineNumbering'])
+
+.factory('MotionInlineEditing', [
+    'Editor',
+    'Motion',
+    'Config',
+    '$timeout',
+    function (Editor, Motion, Config, $timeout) {
+        var obj = {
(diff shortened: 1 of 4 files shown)


emanuelschuetze:
@thoessl Please rebase and resolve the conflicts. Then I can test your pr with current master.

openslides/motions/static/js/motions/motion-services.js
```diff
@@ -0,0 +1,121 @@
+(function () {
+
+"use strict";
+
+angular.module('OpenSlidesApp.motions.motionservices', ['OpenSlidesApp.motions', 'OpenSlidesApp.motions.lineNumbering'])
+
+    .service('MotionInlineEditing', [
```
normanjaeckel: This indention is not necessary, I think.

normanjaeckel: Is there a reason why you use a service and not a factory? The difference is, I think, that a service uses the returned object as constructor and uses `new` whereas a factory just returns the object. We do not need the `new` construction here, do we?

thoessl:
Ok, I made a factory out of the service and reduced the indentation.

emanuelschuetze:
Ok, inline editing successfully tested. Ready for merge.

PR #2366, opened 2016-09-07
Fixed bug in templateHook directive. Fixed #2362.

### LINKED ISSUE #2362: Cannot read property 'map' of undefined
In current master:

- Create an assignment
- add a user to candidates
- set state to 'start voting'
- Click on 'new ballot'

This error is printed on the console. Happens in `core/base.js: 280`. Maybe just a single check is needed if there are any hooks.

normanjaeckel:


emanuelschuetze:
Ok, successfully tested.

PR #2367, opened 2016-09-08
Fixes and template improvements for motion comments.

emanuelschuetze:
Fixed erros while creating motion as anonymous user.

### DIFF
--- a/openslides/users/static/js/users/site.js
+++ b/openslides/users/static/js/users/site.js
@@ -1311,6 +1311,7 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
         gettext('Can create motions');
         gettext('Can support motions');
         gettext('Can manage motions');
+        gettext('Can see and manage comments');
         // users
         gettext('Can see names of users');
         gettext('Can see extra data of users (e.g. present and comment)');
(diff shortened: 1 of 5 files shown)


openslides/motions/views.py
```diff
@@ -109,7 +109,7 @@ def create(self, request, *args, **kwargs):
         if (not request.user.has_perm('motions.can_see_and_manage_comments') and
                 request.data.get('comments')):
             # Some users are not allowed to send comments data.
-            self.permission_denied(request)
+            del request.data['comments']
```
normanjaeckel: Why this? I think we should hide the field in the form (See MotionComment.getFormFields in site.js I think). Then the user should not submit comments if he is not authorized.

openslides/users/static/js/users/site.js
```diff
@@ -1311,6 +1311,7 @@ angular.module('OpenSlidesApp.users.site', ['OpenSlidesApp.users'])
         gettext('Can create motions');
         gettext('Can support motions');
         gettext('Can manage motions');
+        gettext('Can see and manage motions');
```
normanjaeckel: This string does not exist on serverside. Its something with comments I think.

normanjaeckel:
Some remarks.

emanuelschuetze:
All done. Rebased.

PR #2368, opened 2016-09-08
moved MotionComment factory to base.js

matakuka:
Projector slides don't find the factory if it is in site.js, thus projector fail.


emanuelschuetze:
Mmh, I also think about to move the factory to base.js but I came to the decision that "MotionComment" is correct in site.js because comments are used in web interface (site) only - not in projector.
@normanjaeckel What do you  mean?

matakuka:
As soon as I open the projector, I get:
`Error: [$injector:unpr] Unknown provider: MotionCommentProvider <- MotionComment <- Motion`

I don't even need anything projected.

Every time I update a motion, I get a `Error: motions/motion is not a registered resource!` on the projector page

After this move to base.js these errors are gone and everything is working again.

FinnStutzenstein:
> because comments are used in web interface (site) only - not in projector.
But why does the projector fails? A `MotionCommentProvider` is searched, so somewhere is the `MotionComment` factory used.

normanjaeckel:
MotionComment has to be moved to base.js. I think it had been there a few days ago but then I had to solve lots of merge conflicts .... :-(

PR #2371, opened 2016-09-08
New speaker list overlay (on projector) (fixes #2211) - Version 3

### LINKED ISSUE #2211: New list of speakers overlay

matakuka:
This is a rebased, cleaned-up version of #2241, as requested by @emanuelschuetze. See also #2306.

The permission problems are still not solved. The slides' requirements now include the items needed by the slide, including the speakers. A better approach would only allow access to the list of speakers if the speakeroverlay is *visible*.

 This PR is mainly for testing/reviewing purposes and might need changes when permissions/autoupdate is changed.


emanuelschuetze:
First review looks better. @normanjaeckel Is this the right way for you?

normanjaeckel:
Please rebase on top of master. We have refactored the autoupdate and permission system for the projector. Core changes: Don't use `find()` or `findAll()` in the projector.js any more. Implement requirements in `ProjectorElement.get_requirements()` in a new way (`ProjectorRequirement` is not used any more). You can have a look into our core apps.

emanuelschuetze:
Tested: Works good. Only some remarks:
* Rename  the manage button tooltip "Go to current list of speackers" -> "Mangae current list of speakers"
* Project agenda item (customslide) and click on "Go to current list of speakers" button ->  gets JS error (ReferenceError: Customslide is not defined) -> Wait until #2406 is merged and fix it for Topic model. But same problem also for Motion and Assignment (server shows AttributeError: 'Motion' object has no attribute 'agenda_itemspeakers')
* Rename the button tooltip "Project list of speakers" -> "Project current list of speakers as overlay"

emanuelschuetze:
Retested: Rename of buttons ok. Manage list button works now.
@matakuka Please rebase on top of master and use the new topic model (#2406 is merged).

matakuka:
I rebased it on top of master, with a new commit for the topic changes. @emanuelschuetze

openslides/assignments/projector.py
```diff
@@ -27,3 +28,9 @@ def get_requirements(self, config_entry):
             for poll in assignment.polls.all().prefetch_related('options'):
                 for option in poll.options.all():
                     yield option.candidate
+                for speaker in assignment.agenda_item.speakers.filter(end_time=None):
```
normanjaeckel: Do not yield speakers on this element. Yield them at the ListOfSpeakersOverlay projector element.

openslides/core/migrations/0006_auto_20160922_1126.py
```diff
@@ -0,0 +1,38 @@
...
+
+    # We get the model from the versioned app registry;
+    # if we directly import it, it will be the wrong version.
+    projector = Projector.objects.get(id=1)
+    soverlay_uuid = uuid.uuid4().hex
+    projector.config[soverlay_uuid] = {
+        'name': 'core/speakeroverlay',
+        'stable': True}
+    projector.save()
+
```
normanjaeckel: You have to reconnect autoupdate. See users/migrations/0004_groups.py for example.

openslides/core/projector.py
```diff
@@ -118,3 +118,10 @@ class Message(ProjectorElement):
     def check_data(self):
         if self.config_entry.get('message') is None:
             raise ProjectorException('No message given.')
+
+
+class ListOfSpeakersOverlay(ProjectorElement):
+    """
+    List of speakers (last, current, next) on the projector. Rendered as overlay.
+    """
+    name = 'core/speakeroverlay'
```
normanjaeckel: Here you should add a get_requirements() method which yields the speakers you want to show with this element.

openslides/core/static/js/core/projector.js
```diff
@@ -185,6 +188,45 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+    function($scope, Motion, Assignment, Agenda, Topic, Projector, User) {
+        // Attention! Each object that is used here has to be dealt on server side.
+        // Add it to the coresponding get_requirements method of the ProjectorElement
+        // class.
+
+        $scope.visible = $scope.element.visible;
+        $scope.displayedElement = function() {
+            var displayedElement = [];
+            angular.forEach($scope.$parent.elements, function(element){
+                if (element.name == 'motions/motion') {
```
normanjaeckel: Here you should use a lodash find or something similar. The loop should abort if you found any matching projector element.

openslides/core/static/js/core/projector.js
```diff
@@ -185,6 +188,45 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+                    displayedElement = ['motion', element.id];
+                } else if (element.name =='topics/topic') {
+                    displayedElement = ['topic', element.id];
+                } else if (element.name == 'assignments/assignment') {
+                    displayedElement = ['assignment', element.id];
+                }
+            });
+            if (displayedElement[0] == 'motion') {
```
normanjaeckel: What about a switch case block? What happens if displayedElement is an empty list? Motion.get(undefined).agenda_item looks like an error ...

openslides/core/static/js/core/site.js
```diff
@@ -1125,8 +1125,11 @@ angular.module('OpenSlidesApp.core.site', [
     '$state',
     'Config',
     'Projector',
-    function($scope, $http, $interval, $state, Config, Projector) {
-         // bind projector elements to the scope, update after projector changed
+    'Topic',
```
normanjaeckel: You do not have to add these factories here. See below.

openslides/core/static/js/core/site.js
```diff
@@ -1315,15 +1321,37 @@ angular.module('OpenSlidesApp.core.site', [
...
+                data[$scope.speakeroverlay.uuid] = { "visible": false };
+            } else {
+                data[$scope.speakeroverlay.uuid] = { "visible": true };
+            }
+            $http.post('/rest/core/projector/1/update_elements/', data);
+        };
+        $scope.goToListofSpeakers = function() {
+            angular.forEach(Projector.get(1).elements, function(element) {
+                if (element.name == 'motions/motion') {
+                    $state.go('agenda.item.detail', {id: Motion.get(element.id).agenda_item_id});
```
normanjaeckel: You may use DS.get(element.name, element.id).agenda_item_id. But before you run $state.go it would be more readable to finish the loop. E. g. you can use _.find or something similar to check if a related slide is on the projector. Then you use DS.get(result.name, result.id) ...

openslides/motions/projector.py
```diff
@@ -25,3 +26,9 @@ def get_requirements(self, config_entry):
             yield motion.state.workflow
             yield from motion.submitters.all()
             yield from motion.supporters.all()
+            for speaker in motion.agenda_item.speakers.filter(end_time=None):
```
normanjaeckel: Here the same.

openslides/topics/projector.py
```diff
@@ -22,3 +23,9 @@ def get_requirements(self, config_entry):
         else:
             yield topic
             yield topic.agenda_item
+            for speaker in topic.agenda_item.speakers.filter(end_time=None):
```
normanjaeckel: Here the same.

openslides/assignments/projector.py
matakuka: The ListOfSpeakersOverlay does not know which other projector elements exist, so they cannot easily get `agenda_item.speakers`. As the list of speakers will be dependant on any projector (not always the one with the overlay), I think it is better to wait for multiprojector here. I'll discuss it with @FinnStutzenstein.

emanuelschuetze:
@matakuka As discussed: wait until #2379 is merged and then rebase and use the same (generic) logic as current list of speaker slide (implemented in multiprojector pr).

normanjaeckel:
You can rebase (or try a new branch by copiing relevant code snippets). We changed a lot in the projector and autoupdate system.

matakuka:
@normanjaeckel 
I am trying it  on a new branch. I am almost ready, but I have two problems left.

**(1)**  The task is to make the getting of agenda_item in agenda/base.js generic.

DS.get() in [agenda/base.js](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/agenda/static/js/agenda/base.js#L291-L339) DS.get returns 'undefined' when I tried to refactor it and using the factory from from [core/site.js](https://github.com/matakuka/OpenSlides/blob/issue2211-multiprojector/openslides/core/static/js/core/site.js#L1543).

Maybe it is an easy error on my part. Please take a look on the approach I think of.
```
.factory('CurrentListOfSpeakersItem', [
    'DS',
    'Projector',
    'Agenda',
    function (DS, Projector, Agenda) {
        return {
            getItem: function (projectorId) {
                _.forEach(Projector.get(projectorId).elements, function (element) {
                    if(element.hasOwnProperty("agenda_item_id")){
                        return Agenda.get(DS.get(element.name, element.id).agenda_item_id);
                    }
                })
            }
        }
    }
])
```
(the currently existing approach works, so this is no blocker, but I understand you wanted it generic)

**(2)**  Yielding the speakers in [agenda/projector.py](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/agenda/projector.py#L73-L99). 
Here the current approach in master seems not to work in class ListOfSpeakersSlide. 

Currently, as I understand it, it gets the *agenda_item(projector_id)*, which may only work in tests if both are 1. I think that there is a need to get the Projector, THEN the item on the projector.

However, I am unsure how to get the Projector(projector_currentListOfSpeakers_reference) in python.
I thought it was 
`ProjectorManager.get_full_queryset().filter(pk=id)`, but it does not work.

Is this correct? If yes, how to handle it?

matakuka:
New PR see #2489.

emanuelschuetze:
see #2489

normanjaeckel:
(1): At the moment: Slides which provide a list of speakers item (motions/motion e. g. in contrast to users/user) do not keep this info in the projector element attribute. So element.hasOwnProperty("agenda_item_id") looks wrong at the moment. May be you have to add some logic to the update_data() of the motion, topic and assignment ProjectorElement. Then you can get the item id into the projector element data and there you can fetch using your factory.

(2) Projector.objects.get(pk=pk) returns the projector instance you want to use.

PR #2373, opened 2016-09-08
Updated settings files. Fixed timezone support.

normanjaeckel:
I renamed the SQLite file to Django's default. Is this a problem? I think not. Users with an old db **and** settings won't have problems. Users with a new settings should look into it and change entry or filename. What do you think?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -34,6 +34,7 @@ Users:
 - Added new matrix-interface for managing groups and their permissions.
 
 Other:
+- Django 1.10 is now supported. Dropped support for Django 1.8 and 1.9.
 - Removed config cache to support multiple threads or processes.
 - Fixed bug, that the last change of a config value was not send via autoupdate.
 - Added template hooks for plugins (in item detail view and motion poll form).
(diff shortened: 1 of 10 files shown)


normanjaeckel:
At the moment, we have a settings that forces to use Django 1.10.x. Should we revert some changes to support Django 1.9? I think we do not need this support but the change would not be very hard to do, of course.

emanuelschuetze:
Tested:
1)  branch with fresh personal_data dir: ok

2) Run 2.0 to create a last stable personal_data dir (with settings.py). Then switch to this branch. Upgrade requirements. Start server (which migrates the db). Alle done successfully. Only a django warning while starting:
```
?: (1_8.W001) The standalone TEMPLATE_* settings were deprecated in Django 1.8 and the
TEMPLATES dictionary takes precedence. You must put the values of the following settings
into your default TEMPLATES dict: TEMPLATE_DEBUG.
```
Just for notice: This warning can be removed be removing the unused lines in settings.py: `TEMPLATE_DEBUG` and `TEMPLATE_DIRS`.

Manually tests in OpenSlides 2.1-dev works successfully. So ready for merge.

PR #2374, opened 2016-09-08
Changed loading of JavaScript files.

normanjaeckel:
This is for first review and discussion. Open question: Do we want to concat openslides-libs.js and openslides.js? Do we want to use sourcemaps also for our libs? Is there a better way to load JS for plugins?

(A gulp watch task for smarter development is still missing.)


emanuelschuetze:
>  Do we want to use sourcemaps also for our libs?

Yes, please add sourcemaps also for the external libs.

emanuelschuetze:
Successfully tested. 

For the record the results of our discussion: 
Now (with this pull request) custom template files are no longer searched and loaded from OPENSLIDES_USER_DATA_PATH - with the exception of `projector.html`, `projector_container.html` and `index.html`.

To override other templates (for projector or site) you can insert a JavaScript snippet where you have to remove this template (which should be override) from `$templateCache` (see openslides-template.js). 

Example to override `slide_clock.html`:

```html
<!-- projector.html -->
...
<script src="static/js/openslides-templates.js"></script>
<script>
  (function (){
    angular.module('OpenSlidesApp-templates')
    .run([
      '$templateCache',
      function ($templateCache) {
        $templateCache.remove('static/templates/core/slide_clock.html');
      }
    ]);
  })();
</script>
...
```

Copy `slide_clock.html` to `OPENSLIDES_USER_DATA_PATH/static/templates/core/slide_clock.html`. Now  you can modify your custom clock template.

PR #2379, opened 2016-09-12
Multiprojector

FinnStutzenstein:
This feature is mainly ready. I have completely implemented this [suggestion (de)](https://github.com/OpenSlides/OpenSlides/wiki/De%3AKonzept-Multiprojektor-Vorschlag1). Some most obvious changes:

- Support of multiple projectors (with name and projection defaults)
- Updated every project-button (hopefully)
- New view: `/manage-projectors`
- upgraded sidebar
- projectors now reachable with `/projector/{id}`
- blanking, broadcasting and identifying of projectors

In the code I have completely changed the message and countdown system.

@normanjaeckel I ask you in sincerity, to write some test for adding and deleting projectors, messages and countdowns, in fact that i am not really familiar with this tests. If you haven't enough time to do this, I will try to do it my best.

@emanuelschuetze and others, please test this and make some suggestions for improvements.


emanuelschuetze:
I tested your branch. A lot of remarks:

* Projector sidebar: Hide projector dropdown element and broadcast button if only one projector available.
* Fix uib-dropdown-menu (caused by ui-bootstrap upgrade): Use css class `dropdown-menu` instead of old `uib-dropdown-menu` (see e.g. https://github.com/OpenSlides/OpenSlides/commit/852b175f3ea6d43e2629db786a0696c904521514#diff-ca46f8ef3bb02af362d744b7d5778c99L57)
* Projector button: Hide dropdown list if only one projector available.
* ng-bootbox-prompt text for new projector is empty.
* List of speakers reference dropdown does not show selected item.
* Projector manage view: TypeError if you broadcast a projector: Cannot read property 'elements' of undefined  at Object.getElements (projector.js:27)
* Projection defaults menu: Use the following ordering and wording: Agenda, Agenda item, List of speakers, Current list of speakers, Motions, Elections, Participants, Files
* Edit a projector via pen: Projector resolutions fields are empty
* Current list of speakers does not show any speakers.
* After adding new projector all iframes are reloaded. Expected: Only new projector iframe should be loaded.
* Change projection defaults sends no autoupdate. Therefor a client has to reload to get the updated defaults. Otherwise the old defaults are set in all project buttons.

FinnStutzenstein:
@emanuelschuetze I think, I've fixed everything.

> ng-bootbox-prompt text for new projector is empty.
Current list of speakers does not show any speakers.

I have text and speakers...

> Projector button: Hide dropdown list if only one projector available.

This works fine on my machine.. (It is implemented)

> After adding new projector all iframes are reloaded. Expected: Only new projector iframe should be loaded.

This is a bug of angular. The `$scope.projectors` variable has changes so all ngSrc's have changed, too. The result is a complete reload.

emanuelschuetze:
Retested:
All remarks are fixed. Looks good! Thanks! Find some minor issues:
* Countdown: Create and start new countdown -> many JS error "iterations reached" comes up.
* Wrong default projection: Set 'motions' as projection default for projector 2. Create a motion. Go to agenda and project the motion. Motion is projected to no 1 - expected: no 2. (If you go to motion list view and project motion there, it was successfully projected to no. 2. So looks like a motion is interpreted as agenda item in agenda - technically it is correct, but user see it as motion not as agenda item. Maybe you can change it easily...) 
* Current list of speakers view: drowndown-menu is brocken

emanuelschuetze:
Retested:
* Hide dropdown menu of projector button for countdown and messages if only one projector available.
* topic: projection default variable is missing
* add translate tag in projection defaults
* problem with autoupdate and broadcast (e.g. projector 1 shows agenda, click broadcast for no. 1, then projector 2  gets no agenda items via autoupdate)

emanuelschuetze:
@normanjaeckel Ready for your first review (only some minor changes left).

CHANGELOG
```diff
@@ -19,6 +19,7 @@ Core:
 - Added support for big assemblies with lots of users.
 - Added HTML support for messages on the projector.
 - Moved custom slides to own app "topics". Renamed it to "Topic".
+- Added the Multiprojector feature
```
normanjaeckel: `- Added support for multiple projectors.`

openslides/agenda/static/js/agenda/base.js
```diff
@@ -112,61 +112,89 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
...
-                                    typeof element.id !== 'undefined' &&
-                                    element.id == self.id;
-                            }
-                            return value;
-                        };
-                        isProjected = typeof _.findKey(projector.elements, predicate) === 'string';
-                    } else {
-                        isProjected = false;
+                isProjected: function (tree) {
+                    // Returns the projector id if there is a projector element with the same
```
normanjaeckel: This returns the first projector or all projectors where the element is shown?

openslides/agenda/static/js/agenda/site.js
```diff
@@ -193,34 +202,78 @@ angular.module('OpenSlidesApp.agenda.site', ['OpenSlidesApp.agenda'])
             $scope.uncheckAll();
         };
 
+        /** Project functions **/
+        // get ProjectionDefault for item
+        $scope.getPrDef = function (item) {
```
normanjaeckel: Can you please do not use an abbreviation here?

openslides/agenda/static/js/agenda/site.js
```diff
@@ -403,50 +464,64 @@ angular.module('OpenSlidesApp.agenda.site', ['OpenSlidesApp.agenda'])
...
+            var itemPromise = CurrentListOfSpeakersItem.getItem($scope.currentListOfSpeakersReference);
+            if (itemPromise) {
+                itemPromise.then(function(item) {
+                    $scope.AgendaItem = item;
                 });
             }
-        );
+        };
+        // Project current list of speakers
+        $scope.projectClos = function (projectorId) {
```
normanjaeckel: What is `Clos`? May be use `CLOS` or `CurrentLoS` oder `CurrentListofSpeakers`

openslides/agenda/static/js/agenda/site.js
```diff
@@ -403,50 +464,64 @@ angular.module('OpenSlidesApp.agenda.site', ['OpenSlidesApp.agenda'])
...
+            if (isClosProjectedId > 0) {
+                // Deactivate
+                $http.post('/rest/core/projector/' + isClosProjectedId + '/prune_elements/', []);
+            }
+            if (isClosProjectedId != projectorId) {
+                $http.post('/rest/core/projector/' + projectorId + '/prune_elements/',
+                    [{name: 'agenda/current-list-of-speakers'}]);
+            }
+        };
+        $scope.isClosProjected = function () {
```
normanjaeckel: Here the same.

openslides/core/config_variables.py
```diff
@@ -157,11 +157,28 @@ def get_config_variables():
...
+        weight=190,
         group='Projector')
+
+    yield ConfigVariable(
+        name='projector_broadcast',
+        default_value=0,
+        input_type='integer',
+        label='Projector which is broadcasted',
+        weight=200,
+        group='projector',
```
normanjaeckel: Typo.

openslides/core/config_variables.py
```diff
@@ -157,11 +157,28 @@ def get_config_variables():
...
+        hidden=True)
+
+    yield ConfigVariable(
+        name='projector_currentListOfSpeakers_reference',
+        default_value=1,
+        input_type='integer',
+        label='Projector reference for list of speakers',
+        weight=201,
+        group='projector',
```
normanjaeckel: Typo.

openslides/core/models.py
```diff
@@ -152,6 +163,25 @@ def need_full_update_for(self, collection_element):
         return True
 
 
+class ProjectionDefault(RESTModelMixin, models.Model):
```
normanjaeckel: What is die advantage of splitting up this into two models Projector and ProjectionDefault? Even if we have two models on server side (which may be ok also for readability), you can combine them in the REST API. Only Projector may be the root rest model. (Remove access_permissions here and use get_root_rest_element(self): return self.projector.)

openslides/core/models.py
```diff
@@ -243,3 +273,61 @@ class Meta:
     def get_session_store_class(cls):
         from .session_backend import SessionStore
         return SessionStore
+
+
+class Message(RESTModelMixin, models.Model):
```
normanjaeckel: Ah this looks like saving redundant data. At least the uuid exists twice: In the projector config field and in this database field. Why is this necessary? You can store the same message even in multiple projectors. If they should live there.

Or you also change the Message projector element and transform this to a normal slide like all others.

openslides/core/models.py
```diff
@@ -243,3 +273,61 @@ class Meta:
...
+
+    uuid = models.CharField(
+        max_length=255,
+        default='')
+
+    class Meta:
+        default_permissions = ()
+
+
+class Countdown(RESTModelMixin, models.Model):
```
normanjaeckel: Here the same.

openslides/core/projector.py
```diff
@@ -20,7 +23,7 @@ class Countdown(ProjectorElement):
     To start the countdown write into the config field:
 
         {
-            "status": "running",
+            "running": Frue,
```
normanjaeckel: ?

openslides/core/serializers.py
```diff
@@ -30,7 +39,16 @@ class ProjectorSerializer(ModelSerializer):
 
     class Meta:
         model = Projector
-        fields = ('id', 'config', 'elements', 'scale', 'scroll', 'width', 'height',)
+        fields = ('id', 'config', 'elements', 'scale', 'scroll', 'name', 'blank', 'width', 'height', )
```
normanjaeckel: Some of the fields are read only: scale, scroll and blank, I think.

openslides/core/serializers.py
```diff
@@ -50,3 +68,37 @@ class Meta:
...
         read_only_fields = ('user', )
+
+
+class MessageSerializer(ModelSerializer):
+    """
+    Serializer for core.models.Message objects.
+    """
+    class Meta:
+        model = Message
+        fields = ('id', 'message', 'projectors', 'visible', 'uuid')
```
normanjaeckel: uuid is read_only I think.

openslides/core/serializers.py
```diff
@@ -50,3 +68,37 @@ class Meta:
...
+        return super().create(data)
+
+
+class CountdownSerializer(ModelSerializer):
+    """
+    Serializer for core.models.Countdown objects.
+    """
+    class Meta:
+        model = Countdown
+        fields = ('id', 'projectors', 'countdown_time', 'default_time', 'description', 'running', 'visible', 'uuid')
```
normanjaeckel: Here the same.

openslides/core/signals.py
```diff
@@ -20,3 +22,56 @@ def delete_django_app_permissions(sender, **kwargs):
...
+     - assignments
+     - mediafiles
+     - motion
+     - users
+     - list_of_speakers
+     - current_list_of_speakers
+    """
+
+    # Check whether the pk's from 1 to 8 are free.
+    if ProjectionDefault.objects.filter(pk__in=range(1, 9)).exists():
```
normanjaeckel: ProjectionDefault.objectsProjectionDefault.objects.all(). It should do nothing if at least one element exists.

openslides/agenda/projector.py
```diff
@@ -59,3 +59,12 @@ def get_requirements(self, config_entry):
...
                     yield speaker.user
+
+
+class CurrentListOfSpeakersSlide(ProjectorElement):
+    """
+    Slide for the current list of speakers.
+
+    Nothing special to check.
+    """
+    name = 'agenda/current-list-of-speakers'
```
normanjaeckel: You may have to use the get_requirements method to get speakers via autoupdate.

openslides/core/models.py
```diff
@@ -152,6 +163,25 @@ def need_full_update_for(self, collection_element):
         return True
 
 
+class ProjectionDefault(RESTModelMixin, models.Model):
+    """
+    Model for the ProjectionDefaults like Motion, Agenda, List of speakers,...
+    """
+    access_permissions = ProjectionDefaultAccessPermissions()
+
+    name = models.CharField(max_length=256)
```
normanjaeckel: Please add some comments or better extend the docstring about what name and display name should be here. Which choices are possible?

openslides/core/projector.py
```diff
@@ -63,50 +66,72 @@ def validate_config(cls, config_data):
...
-                        value['countdown_time'] = value.get('default', config['projector_default_countdown'])
-                else:
-                    index += -1
-            projector_config[key] = value
-        if found:
-            projector_instance.config = projector_config
-            projector_instance.save()
+                raise ValueError("Action must be 'start', 'stop' or 'reset', not {}.".format(action))
+
+        if (Countdown_model.objects.all().exists()):
```
normanjaeckel: It looks like that this logic can be moved to a viewset if we decide to use models for countdowns.

openslides/agenda/static/js/agenda/base.js
FinnStutzenstein: technically the last projector, because this function iterates through all projectors and just saves the id if there is a match. But: An element could only be projected on one projector at the time. So there is no need to give back a list of projectors.
This is the difference to the messages and countdowns. One of these could be displayed on multiple projectors.

openslides/core/models.py
FinnStutzenstein: Now there is one advantage: If I want to have the default projector id of motions, i could search the ProjectionDefault with name motions and get the id. If the ProjectionDefaults are suborderd under Projector. I have to iterate through all projectors to find the Projector with the given ProjectionDefault. Is this the solution?

normanjaeckel: Where do you need the default projector id? On server side, you can still get it as you wrote if we use two models. On client side you can add a short projector method to do the thing easily.

openslides/agenda/static/js/agenda/base.js
normanjaeckel: I think we should keep the option that a projector control plugin provides buttons to display one element an more than one projector.

emanuelschuetze:
Some merge conflicts... @FinnStutzenstein

openslides/core/models.py
FinnStutzenstein: Only on client side. I will implement this.

openslides/agenda/static/js/agenda/site.js
FinnStutzenstein: Done.

openslides/agenda/static/js/agenda/site.js
FinnStutzenstein: Done.

openslides/core/config_variables.py

openslides/core/config_variables.py

openslides/core/models.py
FinnStutzenstein: Done.

openslides/core/models.py
FinnStutzenstein: I think the problem is that messages and countdowns are stable (they have to be stable!), but if one wants to remove a message, he has to give the uuid. My idea was to give one message always the same uuid (generated at the moment of creation), so if the message is projected on two projectors. it has the same uuid on both projectors. For writing in a projectors config and removing from there the uuid is necessary.

openslides/core/projector.py
FinnStutzenstein: Type, done.

openslides/core/serializers.py
FinnStutzenstein: yes, you are right. Also width and height. Done.

openslides/core/serializers.py
FinnStutzenstein: Done.

openslides/core/serializers.py
FinnStutzenstein: Done.

openslides/core/signals.py
FinnStutzenstein: Done with ProjectionDefaults.objects.all().exists().

openslides/agenda/projector.py
FinnStutzenstein: Done.

openslides/agenda/static/js/agenda/site.js
FinnStutzenstein: Done.

openslides/core/models.py
normanjaeckel: Yes and no. If I want to project the same motion on two projectors (which should be possible by server side) I also get different uuids. But I do not want to save uuids to every model we have. So we should use only one way I think.

openslides/agenda/static/js/agenda/base.js
FinnStutzenstein: Could the plugin override this method? This would be the simplest solution, because for the current purpose changing this function affect all written code for the projectorbuttons etc. And the projectorbuttons don't support multiple projectors, so this wouldn't make sense just to change this piece of code.

@emanuelschuetze mentioned some days ago, that this should be an extra feature: Project elements one multiple projectors simultaneously. Like now the messages and countdowns.

CHANGELOG
FinnStutzenstein: Done.

openslides/core/models.py
FinnStutzenstein: But with the motions, you can remove them with purge_elements. This is not possible for Messages and Countdowns, and with saving the uuid there is a direct access to these elements without searching for them.

openslides/agenda/static/js/agenda/base.js
normanjaeckel: Plugins can not override this code but they may provide a button which fires some http request that let slide XY appear on several projectors. It would be clean if this method returns an array of all projectors that show this slide, not only one (random) id.

openslides/core/models.py
FinnStutzenstein: @normanjaeckel Ping. please test this

openslides/core/models.py
normanjaeckel: You can not use `prune_elements` for a motion with the flag `stable: true`. Or if you have more unstable slides on a projector than a motion you can also not use the `prune_elements` viewpoint because you would disable other slides as well. So it is no special thing the you have to look for the uuid of the element that you want to disable. If we transform the message and the countdown to a semi model (saved in model but controlled by ProjectorViewSet) we have to care about a lot of additional code and about data redundancy (what if a countdown in model has a different uuid than in projector, what if one is missing, ...). I think it is eaysier to run a short lodash function in the client to find our uuid. (We do this already in current master, I think; not for multiple projectors of course.)

FinnStutzenstein: So your suggestion is to remove the models completely and just "save" the messages to the projector configs?

normanjaeckel: Yes. The other way would be to transform the messages and countdowns to real root rest models and remove the extra view points in the projector view set. But I like the current way a little bit more: Just save messages into projector as it is in master.

FinnStutzenstein: Ok, I will have a try to bring this to multiple projectors.

normanjaeckel: Thanks.

FinnStutzenstein: Done.

openslides/agenda/static/js/agenda/base.js
FinnStutzenstein: Can I create an Issue for this? I think this is some kind of feature for new plugins, and the possibility to project things multiple times should be implemented in the master. But not with this PR, because this was not specified at the beginning and not requested by the customer.

normanjaeckel: You can do so. But please update the comment that this function currently returns only the id of the last projector that shows this element (because we do not ensure that this is the only one).

FinnStutzenstein:
@emanuelschuetze should work now.

openslides/agenda/static/js/agenda/base.js
FinnStutzenstein: Done in core/base.js (this is just overritten)

emanuelschuetze:
@normanjaeckel Rebased. Ready for last review.

openslides/core/static/js/core/base.js
```diff
@@ -305,34 +313,93 @@ angular.module('OpenSlidesApp.core', [
...
-            return $http.post(
-                '/rest/core/projector/1/prune_elements/',
-                [{name: this.getResourceName(), id: this.id}]
-            );
+        BaseModel.prototype.project = function(projectorId) {
+            // if this object is already projected on projectorId, delete this element from this projector
+            var isProjectedId = this.isProjected();
+            if (isProjectedId > 0) {
+                $http.post(
+                    '/rest/core/projector/' + isProjectedId + '/prune_elements/', []
```

openslides/core/static/js/core/base.js
```diff
@@ -396,10 +463,74 @@ angular.module('OpenSlidesApp.core', [
...
+            },
+            methods: {
+                controlProjector: function(action, direction) {
+                    $http.post('/rest/core/projector/' + this.id + '/control_view/',
+                            {"action": action, "direction": direction}
+                    );
+                },
+                getStateForCurrentSlide: function () {
+                    var return_dict;
+                    $.each(this.elements, function(key, value) {
```
normanjaeckel: Use angular.forEach or lodash.

openslides/core/static/js/core/site.js
```diff
@@ -1155,209 +1158,437 @@ angular.module('OpenSlidesApp.core.site', [
...
+                $interval.cancel(countdown.interval);
+            });
+        };
+
+        // Get all message and countdown data from the defaultprojector (id=1)    
+        var rebuildAllElements = function () {
+            $scope.countdowns = [];
+            $scope.messages = [];
+
+            $.each(Projector.get(1).elements, function (uuid, element) {
```
normanjaeckel: Here the same.

openslides/core/static/js/core/site.js
```diff
@@ -1155,209 +1158,437 @@ angular.module('OpenSlidesApp.core.site', [
...
-                }
-            }
-
+            cancelIntervalTimers();
         });
 
-        // watch for changes in Config
-        var last_conf;
+        // watch for changes in projector_broadcast
+        var last_bc;
```
normanjaeckel: Please do not user abbrevation here.

openslides/core/static/js/core/site.js
```diff
@@ -1155,209 +1158,437 @@ angular.module('OpenSlidesApp.core.site', [
...
-                "description": countdown.description,
-                "default": parseInt(countdown.default)
-            };
-            if (countdown.status == "stop") {
-                data[countdown.uuid].countdown_time = parseInt(countdown.default);
-            }
-            $http.post('/rest/core/projector/1/update_elements/', data);
+            $scope.projectors.forEach(function (projector) {
+                var countdowns = [];
+                $.each(projector.elements, function (uuid, element) {
```
normanjaeckel: Here the same.

openslides/core/static/js/core/site.js
```diff
@@ -1155,209 +1158,437 @@ angular.module('OpenSlidesApp.core.site', [
...
-            var data = {};
-            // calculate end point of countdown (in seconds!)
-            var endTimestamp = Date.now() / 1000 - $scope.serverOffset + countdown.countdown_time;
-            data[countdown.uuid] = {
-                "status": "running",
-                "countdown_time": endTimestamp
-            };
-            $http.post('/rest/core/projector/1/update_elements/', data);
+            $scope.projectors.forEach(function (projector) {
+                $.each(projector.elements, function (uuid, element) {
```
normanjaeckel: Here the same.

openslides/core/static/js/core/site.js
```diff
@@ -1155,209 +1158,437 @@ angular.module('OpenSlidesApp.core.site', [
...
+    'Projector',
+    'ProjectionDefault',
+    'Config',
+    'gettextCatalog',
+    function ($scope, $http, $state, $timeout, Projector, ProjectionDefault, Config, gettextCatalog) {
+        ProjectionDefault.bindAll({}, $scope, 'projectiondefaults');
+
+        // watch for changes in projector_broadcast
+        // and projector_currentListOfSpeakers_reference
+        var last_bc, last_clos;
```
normanjaeckel: Here the same.

openslides/core/static/js/core/site.js
```diff
@@ -1155,209 +1158,437 @@ angular.module('OpenSlidesApp.core.site', [
...
+                    };
+                    $scope.edit[projector.id] = false;
                 }
             });
+            if ($scope.projectors.length) {
+                first_watch = false;
+            }
+        });
+
+        // Misc stuff
```
normanjaeckel: ?

openslides/core/views.py
```diff
@@ -214,7 +225,10 @@ def activate_elements(self, request, pk):
         for element in request.data:
             if element.get('name') is None:
                 raise ValidationError({'detail': 'Invalid projector element. Name is missing.'})
-            projector_config[uuid.uuid4().hex] = element
+            if element.get('uuid') is None:
```
normanjaeckel: This can be reverted, no?

openslides/core/views.py
```diff
@@ -447,6 +461,68 @@ def set_scroll(self, request, pk):
...
+        """
+        REST API operation to (un-)broadcast the given projector.
+        This method takes care, that all other projectors get the new requirements.
+
+        It expects a POST request to
+        /rest/core/projector/<pk>/broadcast/ without an argument
+        """
+        if config['projector_broadcast'] == 0:
+            config['projector_broadcast'] = pk
+            projector_broadcast_data(pk)
```
normanjaeckel: Ah. Here it is. :-)

Please use the need_full_update_for_this() logic and add the check in projector model. (If this config entry is still used.)

openslides/core/views.py
```diff
@@ -447,6 +461,68 @@ def set_scroll(self, request, pk):
...
+        """
+        if not isinstance(request.data, int):
+            raise ValidationError({'detail': 'Data must be an int.'})
+
+        try:
+            projectiondefault = ProjectionDefault.objects.get(pk=request.data)
+            projector_instance = self.get_object()
+            projectiondefault.projector = projector_instance
+            projectiondefault.save()
+        except ObjectDoesNotExist:
```
normanjaeckel: Please use try except else and put only the statement that may raise the exception into the try block.
Then use ProjectionDefault.DoesNotExist as exception.

openslides/agenda/static/js/agenda/base.js
```diff
@@ -112,61 +112,89 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
...
-                    return $http.post(
-                        '/rest/core/projector/1/prune_elements/',
-                        [{name: this.content_object.collection, id: this.content_object.id}]
-                    );
+                project: function (projectorId, tree) {
+                    var isProjectedId = this.isProjected(tree);
+                    if (isProjectedId > 0) {
+                        // Deactivate
+                        $http.post(
+                            '/rest/core/projector/' + isProjectedId + '/prune_elements/', []
```
FinnStutzenstein: Done.

openslides/agenda/static/js/agenda/base.js
```diff
@@ -112,61 +112,89 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
...
-                    return $http.post(
-                        '/rest/core/projector/1/prune_elements/',
-                        [{name: 'agenda/list-of-speakers', id: this.id}]
-                    );
+                projectListOfSpeakers: function(projectorId) {
+                    var isProjectedId = this.isListOfSpeakersProjected();
+                    if (isProjectedId > 0) {
+                        // Deactivate
+                        $http.post(
+                            '/rest/core/projector/' + isProjectedId + '/prune_elements/', []
```
FinnStutzenstein: Done.

openslides/agenda/static/js/agenda/base.js
```diff
@@ -112,61 +112,89 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
...
+                    if (isProjectedId != projectorId) {
+                        return $http.post(
+                            '/rest/core/projector/' + projectorId + '/prune_elements/',
+                            [{name: 'agenda/list-of-speakers', id: this.id}]
+                        );
+                    }
                 },
                 // check if list of speakers is projected
                 isListOfSpeakersProjected: function () {
                     // Returns true if there is a projector element with the
```
FinnStutzenstein: Done.

openslides/agenda/static/js/agenda/base.js
```diff
@@ -265,6 +293,54 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
...
+    function (Projector, Assignment, Topic, Motion, Agenda) {
+        return {
+            getItem: function (projectorId) { 
+                var elementPromise;
+                return Projector.find(projectorId).then(function (projector) {
+                    // scan all elements
+                    _.forEach(projector.elements, function(element) {
+                        switch(element.name) {
+                            case 'motions/motion':
+                                elementPromise = Motion.find(element.id).then(function(motion) {
```
FinnStutzenstein: Done.

openslides/agenda/static/js/agenda/site.js
```diff
@@ -110,6 +111,17 @@ angular.module('OpenSlidesApp.agenda.site', ['OpenSlidesApp.agenda'])
...
             }
         });
+        Projector.bindAll({}, $scope, 'projectors');
+        $scope.mainListTree = true;
+        $scope.$watch(function () {
+            return Projector.lastModified();
+        }, function () {
+            var projectiondefault =  ProjectionDefault.filter({name: 'agenda_all_items'})[0];
+            if (projectiondefault) {
+                $scope.defPrId_all_items = projectiondefault.projector_id;
```
FinnStutzenstein: Done.

openslides/core/static/js/core/base.js

openslides/core/static/js/core/base.js
FinnStutzenstein: Done. Now using _.forEach.

openslides/core/static/js/core/site.js
FinnStutzenstein: Done.

openslides/core/static/js/core/site.js
FinnStutzenstein: Done.

openslides/core/static/js/core/site.js
FinnStutzenstein: Done.

openslides/core/static/js/core/site.js
FinnStutzenstein: Done.

openslides/core/static/js/core/site.js
FinnStutzenstein: Reorganized code. Done.

openslides/core/views.py
FinnStutzenstein: Done.

openslides/core/views.py
FinnStutzenstein: Done. Was originally for the message and countdown models.

openslides/assignments/static/js/assignments/base.js
```diff
@@ -267,42 +267,47 @@ angular.module('OpenSlidesApp.assignments', [])
...
                 // override project function of jsDataModel factory
-                project: function (poll_id) {
-                    return $http.post(
-                        '/rest/core/projector/1/prune_elements/',
-                        [{name: 'assignments/assignment', id: this.id, poll: poll_id}]
-                    );
+                project: function (projectorId, pollId) {
+                    var isProjectedId = this.isProjected(pollId);
+                    if (isProjectedId > 0) {
+                        $http.post('/rest/core/projector/' + isProjectedId + '/prune_elements/', []);
```
FinnStutzenstein: Done.

openslides/core/projector.py
```diff
@@ -63,50 +62,43 @@ def validate_config(cls, config_data):
...
-
-        projector_instance = Projector.objects.get(pk=projector_id)
-        projector_config = {}
-        found = False
-        for key, value in projector_instance.config.items():
-            if value['name'] == cls.name:
-                if index == 0:
+                raise ValueError("Action must be 'start', 'stop' or 'reset', not {}.".format(action))
+
+        # Use the first countdown if there is one
```
FinnStutzenstein: It is used in agenda/models.py

It will always take the first countdown, not all. This was a wish from Emanuel. Currently if no countdown is found it does nothing. I will implement, that a countdown will be created.

FinnStutzenstein: Done.

openslides/utils/autoupdate.py
```diff
@@ -138,17 +138,48 @@ def send_data(message):
...
-        if send_all is None:
-            send_all = projector.need_full_update_for_this(collection_element)
-        if send_all:
-            output = get_projector_element_data(projector)
-        else:
-            output = []
-        output.append(collection_element.as_autoupdate_for_projector())
-        if output:
+    broadcast_id = config['projector_broadcast']
+    if broadcast_id > 0:
```
FinnStutzenstein: Done.

openslides/core/views.py
FinnStutzenstein: One question: Is the autoupdate triggered if there is a request to `/broadcast/`? Now I trigger the autoupdate with this method. I can implement this to `need_full_update_for()`, but does the server recognize, if the need_full_update status change?

tests/integration/agenda/test_viewsets.py
```diff
@@ -216,50 +216,28 @@ def test_end_speech_no_current_speaker(self):
 
     def test_begin_speech_with_countdown(self):
         config['agenda_couple_countdown_and_speakers'] = True
-        projector = Projector.objects.get(pk=1)
-        projector.config['03e87dea9c3f43c88b756c06a4c044fb'] = {
```
FinnStutzenstein: I've implemented this ;) Done.

tests/integration/core/test_views.py
```diff
@@ -25,9 +25,11 @@ def test_slide_on_default_projector(self):
         default_projector.save()
 
         response = self.client.get(reverse('projector-detail', args=['1']))
+        content = json.loads(response.content.decode())
+        del content['projectiondefaults']
```
FinnStutzenstein: So just assert for name, id, width, height,..?

openslides/agenda/projector.py
```diff
@@ -64,3 +64,32 @@ def need_full_update_for_this(self, collection_element):
...
+
+    def get_requirements(self, config_entry):
+        pk = config['projector_currentListOfSpeakers_reference']
```
FinnStutzenstein: New Approach: Every projector has a setting 'Current list of speakers reference' and this could be set per projector in the manage view. But also the global selector (like now) is available to set the reference from all projectors.

FinnStutzenstein: @emanuelschuetze mentioned that this should be a different feature. I think like together with the changes in `isProjected()` for supporting one slide on multiple projectors.

FinnStutzenstein:
Follow up issues:

- Feature request: Project one slide on multiple projectors
- remove find() from factory `CurrentListOfSpeakersItem` (agenda/base.js)

openslides/core/views.py
FinnStutzenstein: I just saw: If broadcast is active, `send_all` is true, so the `need_full_update_for_this()` is not triggered. So now I just have to trigger the autoupdate if there is a request to `/broadcast/`.
Now I do this with
```
projector_instance = self.get_object()
inform_changed_data(projector_instance)
```
instead of `broadcast_projector_data`. Is this ok?

normanjaeckel:
I did another review and noted a lot of comments offline. I propose that I discuss the important things with @emanuelschuetze and then push an extra commit on top of your branch. (So please do not push anything to this branch any more.) Some things are just stylistic. I would change them without remark. Some things can be moved out into a separate issue that can be fixed by @FinnStutzenstein or me in October.

PR #2401, opened 2016-09-18
Added management command to change config values. Fixed #2400.

### LINKED ISSUE #2400: Add management command to change config values
@boehlke Do you want to use such a command during install OpenSlides into your containers?

normanjaeckel:


emanuelschuetze:
Ok, but not tested.

PR #2404, opened 2016-09-18
Cache2

ostcar:
Adds a cache system for CollectionElement that caches all instances for autoupdate, rest-api etc.


emanuelschuetze:
Autoupdate successfully tested. Missing login message fixed in new commit.

openslides/utils/collection.py
```diff
@@ -31,8 +42,15 @@ def __init__(self, instance=None, is_deleted=False, collection_string=None, id=N
                 'Invalid state. Use CollectionElement.from_instance() or '
                 'CollectionElement.from_values() but not CollectionElement() '
                 'directly.')
-        self.instance = instance
-        self.deleted = is_deleted
+
+        # Delete the element from the cache, if is_deleted is True:
+        if is_deleted:
```
normanjaeckel: Proposal: Use self.is_deleted() here.

openslides/utils/collection.py
```diff
@@ -31,8 +42,15 @@ def __init__(self, instance=None, is_deleted=False, collection_string=None, id=N
...
+            self.delete_from_cache()
+
```
normanjaeckel: Remove blank line and move comment into the else clause.

openslides/utils/collection.py
```diff
@@ -117,14 +152,217 @@ def get_full_data(self):
...
+            # Use the cache version if self.instance is not set.
+            # After this line full_data can be None, if the element is not in the cache.
+            self.full_data = cache.get(self.get_cache_key())
+
+        if self.full_data is None:
+            self.full_data = self.get_access_permissions().get_full_data(self.get_instance())
+            self.save_to_cache()
+        return self.full_data
 
     def is_deleted(self):
```
normanjaeckel: Is there any reason why this is a method. We may just use an attribute instead.

openslides/utils/collection.py
```diff
@@ -102,8 +129,16 @@ def get_instance(self):
...
         if self.is_deleted():
             raise RuntimeError("The collection element is deleted.")
+
         if self.instance is None:
-            self.instance = self.get_model().objects.get(pk=self.id)
+            # The config instance has to be get from the config element, because
+            # some config values are not in the db.
+            from openslides.core.config import config
+            if (self.collection_string == config.get_collection_string() and
+                    isinstance(self.id, str)):
```
normanjaeckel: In which config case is self.id not a string?

openslides/utils/collection.py
```diff
@@ -117,14 +152,217 @@ def get_full_data(self):
...
+            self.save_to_cache()
+        return self.full_data
 
     def is_deleted(self):
         """
         Returns Ture if the item is marked as deleted.
         """
         return self.deleted
 
+    def get_cache_key(self):
```
normanjaeckel: Proposal: Move caching methods (and especially cache key and cache key prefix functions) to an extra class `Cache` for better code structure and readability.

openslides/utils/collection.py
```diff
@@ -1,25 +1,36 @@
...
+
+        This will also update the instance in the cache.
+
+        If is_deleted is set to True, the element is deleted from the cache.
         """
         return cls(instance=instance, is_deleted=is_deleted)
 
     @classmethod
-    def from_values(cls, collection_string, id, is_deleted=False):
+    def from_values(cls, collection_string, id, is_deleted=False, full_data=None):
```
normanjaeckel: Ok. Please add a short note in the docstring, e. g. `The argument full_data may already contain all data. They are fetched out of the cache.`

openslides/utils/collection.py
```diff
@@ -117,14 +152,217 @@ def get_full_data(self):
...
+                missing_ids.add(get_collection_id_from_cache_key(cache_key)[1])
+
+        # Generate collection elements that where is the cache.
+        for cache_key, cached_full_data in cached_full_data_dict.items():
+            yield CollectionElement.from_values(
+                *get_collection_id_from_cache_key(cache_key),
+                full_data=cached_full_data)
+
+        # Generate collection element that where not in the cache.
+        # If config elements are not in the cache, they have to be read from
```
normanjaeckel: I do not understand this comment. How can we read configs from cache if they are not in? Do you mean 'config object' at the end of the sentence?

openslides/utils/collection.py
```diff
@@ -117,14 +152,217 @@ def get_full_data(self):
...
+                yield CollectionElement.from_instance(instance)
+
+    def as_autoupdate_for_projector(self):
+        """
+        Returns a list of dictonaries to send them to the projector.
+        """
+        output = []
+        for collection_element in self.element_generator():
+            content = collection_element.as_autoupdate_for_projector()
+            if content is not None:
```
normanjaeckel: It is impossible that content is None, is it? See our last changes above. We send `deleted` if the user is not allowed to see it.

openslides/utils/collection.py
```diff
@@ -117,14 +152,217 @@ def get_full_data(self):
...
+
+    def as_list_for_user(self, user):
+        """
+        Returns a list of dictonaries to send them to a user, for example over
+        the rest api.
+        """
+        output = []
+        for collection_element in self.element_generator():
+            content = collection_element.as_dict_for_user(user)
+            if content is not None:
```
normanjaeckel: Here the same.

openslides/utils/rest_api.py
```diff
@@ -1,15 +1,13 @@
...
 from rest_framework.decorators import detail_route, list_route  # noqa
 from rest_framework.metadata import SimpleMetadata  # noqa
-from rest_framework.mixins import (  # noqa
-    DestroyModelMixin,
-    ListModelMixin,
-    RetrieveModelMixin,
-    UpdateModelMixin,
-)
-from rest_framework.response import Response  # noqa
+from rest_framework.mixins import ListModelMixin as _ListModelMixin  # noqa
```
normanjaeckel: Remove # noqa

openslides/utils/settings.py.tpl
```diff
@@ -76,6 +76,16 @@ DATABASES = {
 #     }
 # }
 
+# CACHES = {
```
normanjaeckel: Remove theses lines. They are already there.

openslides/core/access_permissions.py
```diff
@@ -94,9 +94,8 @@ def check_permissions(self, user):
         """
         from .config import config
 
-        # Every authenticated user can see the metadata and list or retrieve
-        # the config. Anonymous users can do so if they are enabled.
-        return user.is_authenticated() or config['general_system_enable_anonymous']
```
ostcar: done. Please do it yourself next time

openslides/core/access_permissions.py
```diff
@@ -106,4 +105,8 @@ def get_full_data(self, instance):
 
         # Attention: The format of this response has to be the same as in
         # the retrieve method of ConfigViewSet.
+        if isinstance(instance, dict):
```
ostcar: I don't see the difference. You code does exactly the same. If you like, you can change it.

openslides/core/views.py
```diff
@@ -523,10 +524,8 @@ def check_view_permissions(self):
         if self.action in ('list', 'retrieve'):
             result = self.get_access_permissions().check_permissions(self.request.user)
         elif self.action == 'metadata':
-            # Every authenticated user can see the metadata and list or
```
ostcar: Done. Please do it yourself next time.

openslides/core/views.py
```diff
@@ -535,22 +534,24 @@ def check_view_permissions(self):
...
         try:
-            value = config[key]
-        except ConfigNotFound:
+            content = collection_element.as_dict_for_user(request.user)
+        except ConfigStore.DoesNotExist:
             raise Http404
-        # Attention: The format of this response has to be the same as in
-        # the get_full_data method of ConfigAccessPermissions.
-        return Response({'key': key, 'value': value})
+        if content is None:
```
ostcar: See the next line: when the user has no permission to see the content. See also the comment in utils/access_permissions.py from get_restricted_data()

openslides/users/views.py
```diff
@@ -177,7 +164,7 @@ class UserLoginView(APIView):
     """
     Login the user.
     """
-    http_method_names = ['get', 'post']
+    http_method_names = ['get', post']
```
ostcar: Done. Please do it yourself next time.

openslides/utils/collection.py
ostcar: Done. You can change the docstring if you like.

openslides/utils/collection.py
ostcar: I like this style. I won't protest, if you change it to the style you like.

openslides/utils/collection.py
ostcar: When a config value is changed, the save method a ConfigStore-element is called and therefore the autoupdate system is called with this ConfigStore-element. In this case the id is the numeric id of this db-element.

openslides/utils/collection.py
ostcar: Done. Please do it your self next time.

openslides/utils/collection.py
ostcar: It may change in the future.

openslides/utils/collection.py
ostcar: To much work for to less gain. I am not against it.

openslides/utils/collection.py
ostcar: Yes. done

openslides/utils/collection.py
ostcar: Yes. Done

openslides/utils/collection.py
ostcar: No. Here it is different. as_dict_for_user returns access_permissions.restricted_data(user) which can return None.

openslides/utils/settings.py.tpl
ostcar: done

openslides/topics/models.py
```diff
@@ -17,19 +18,26 @@ class Topic(RESTModelMixin, models.Model):
...
     class Meta:
         default_permissions = ()
 
     def __str__(self):
         return self.title
 
+    @classmethod
+    def all_query(cls):
+        return cls.objects.prefetch_related('attachments', 'agenda_items')
+
```
ostcar: @normanjaeckel please have a look at this. I would like to add a method like this to any of our rest models. I would like to add it as a manager instead of a classmethod but therefore we would have to add a Manager to every model we have. What do you think?

I use the method to reduce qb queries. Currently there are 2 extra queries for each(!) topic on the agenda. So at the moment there are 201 queries for the agenda list view with 100 topics.

normanjaeckel: Does this refer to *every* model or to every agenda related model (topic, motion, assignment, but not user)? Is it possible subclass the manager only for the three agenda related models?

ostcar: I want to add it to every model which Serializer needs more then one model. For example if the user would add the name of every group in the get_full_data() method, then the user model would also need this method.

I have not tested yet which models need the method at the moment. Maybe it is only the agenda related models, maybe not.

ostcar: But I think we should add a Manager. It is cleaner then the classmethod

normanjaeckel: Hm. What about mixing in a custom manager. This manager looks for a attribute like `give_me_prefetch_related_stuff`. This is empty by default but may contain the names of the related models like items, mediafiles or groups in your example.

normanjaeckel: Or we use Django magic and loop over all model fields and prefetch all related fields. (reverse foreign key or m2m)

ostcar: The Django magic would be too mutch. For example we do not need the Group be the user. Or users on agenda_items (only the speaker object with the user_id). I try the manager that reads strings from the model

ostcar: @normanjaeckel Please see my new commit. I was not able to write add a manager via the RestModel. Maybe it is a django bug, but a Manager added via a Mixin does not work at all. If you succeed, then please let me know.

Do you think it is ok to rewrite the default queryset? I don't see any problem.

normanjaeckel: [This](https://docs.djangoproject.com/en/1.10/topics/db/managers/#custom-managers-and-model-inheritance) does not work? I would prefer not to override the default manager `objects` but to add a custom manager `autoupdate_objects` that uses the `get_queryset()` method. So we still have access to the default methods with default Django behavior where we do not want to prefetch some stuff (e. g. calling Model.objects.all() with only one query).

ostcar: No. It did not work. I also read the documentation. Therefore I think it is a django bug that Managers on Mixins are not supported.

Currently I override the objects manager but with a different method ```get_full_queryset()```.

Please have a look at my newest commit. I added tests for each RestModel viewset that counts the queries used for a list view. Currently it is a disaster :) I fixt some cases but there are still much more queries then needed. Maybe you can also have a look at it? This has nothing to do with caching. In this tests caching is disabled (as usual in tests).

tests/integration/assignments/test_viewset.py
```diff
@@ -4,9 +4,61 @@
...
+        * 1 request to get all related users,
+        * 1 request to get the agenda item,
+        * 1 request to get the polls,
+        * 1 request to get the tags,
+
+        * lots of permissions requests.
+
+        TODO: The last requests are a bug.
+        """
+        with self.assertNumQueries(57):
```
normanjaeckel: :-)

normanjaeckel:
Is this working in total? Which things are missing?

ostcar:
The core-things are working. What is not is the commit "no signal no cry". There I added a system to only update the needed elements on a projector change. The code works, but there are still todos on other parts that this feature works.

ostcar:
The core-things are working. What is not is the commit "no signal no cry". There I added a system to only update the needed elements on a projector change. The code works, but there are still todos on other parts that this feature works.

openslides/utils/collection.py
normanjaeckel: See #2463

openslides/utils/collection.py
```diff
@@ -117,14 +161,236 @@ def get_full_data(self):
...
+                for instance in query.filter(pk__in=missing_ids):
+                    yield CollectionElement.from_instance(instance)
+
+    def as_autoupdate_for_projector(self):
+        """
+        Returns a list of dictonaries to send them to the projector.
+        """
+        output = []
+        for collection_element in self.element_generator():
+            content = collection_element.as_autoupdate_for_projector()
```
normanjaeckel: Here content may be None, right? Then we to not want `None` to be appended to the output.

openslides/utils/models.py
```diff
@@ -54,3 +54,36 @@ def get_rest_pk(self):
...
+        """
+        Calls the django delete-method and afterwards hits the autoupdate system.
+
+        See the save method above.
+        """
+        # TODO: Fix circular imports
+        from .autoupdate import inform_changed_data
+        # Django sets the pk of the instance to None after deleting it. But
+        # we need the pk to tell the autoupdate system which element was deleted.
+        instance_pk = self.pk
```
normanjaeckel: Are you sure that we (or Django) do not expect anywhere that pk is None if a model instance is deleted?

ostcar: This can be a problem and we should fix it.

One solution would be, that we set self.pk to None after inform_change_data is called. Another solution would be, that inform_change_data does not need an instance but also accepts a collection_string and an id. Because when an element is deleted, then we do not need the full instance data.
What do you think?

normanjaeckel:
Broadcast does not work. I can do this.

PR #2414, opened 2016-09-22
Add Dockerfile

boehlke:


normanjaeckel:
@ostcar Please let us know if you should choose another structure here.

PR #2423, opened 2016-09-23
Implemented need_full_update check. Fixed #2396.

### LINKED ISSUE #2396: Implement Projector.need_full_update_for()
There is a new method that has to be implemented for all projector elements. At the moment it always returns True.

normanjaeckel:


openslides/assignments/projector.py
```diff
@@ -23,7 +23,16 @@ def get_requirements(self, config_entry):
             yield assignment
             yield assignment.agenda_item
             for user in assignment.related_users.all():
+                # Yield user instances of current candidates (i. d. future
```
emanuelschuetze: typo? do you mean "i. e."?

normanjaeckel: Corrected.

emanuelschuetze:
Ok, tested successfully with votecolletor plugin.

PR #2445, opened 2016-09-27
Using a datetimepicker for agenda start time

FinnStutzenstein:
A new dependency but some kind of better handling. Please test.


emanuelschuetze:
Works good. But some translations strings are still missing (e.g. for the buttons "Today", "Clear", "Time", "Close"). Is there a possibility to marks theses strings for our gettext transslations?

normanjaeckel:
May be you can override the `constant` [here](https://github.com/Gillardo/bootstrap-ui-datetime-picker/blob/master/dist/datetime-picker.js#L5).

FinnStutzenstein:
I've added a factory for the datetimepicker translation. It just generates the buttonBar object with the given translation (all translations are currently missing). This object can be passed through the html attribut button-bar. For testing I changed 'now' to 'Agenda' and in german I get in the config 'Tagesordnung'. So this should work.

PR #2461, opened 2016-09-29
some minor fixes

FinnStutzenstein:
1. Do not show the description header if there is no description of an election
2. The serverOffset has to be initialized with 0 assuming that the servers time is equal with the clients time. This fixes the projector clock bug, that when the projector is loaded the clock shows 01:00 until the first autoupdate comes in.
3. Many users of OpenSlides find the caption '100%' in the pdf control irritating. Most users don't get, that this is a button but an indicator for the zoom. If one zooms in, it stays at 100% and he thinks that this is a bug. For that reason I changed the '100%' to the full screen icon also used by images.

### DIFF
--- a/openslides/assignments/static/templates/assignments/assignment-detail.html
+++ b/openslides/assignments/static/templates/assignments/assignment-detail.html
@@ -78,8 +78,10 @@
 </div>
 
 <div class="details">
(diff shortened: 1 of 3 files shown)


emanuelschuetze:
Ok, successfully tested. Merge...

PR #2476, opened 2016-10-01
Removed old code needed be the server to serve pdf

ostcar:
I would remove this code immediately but you can wait until the client pdf is ready. If you are sure that the client will handle all pdfs in the 2.1, then there is not need to wait.


ostcar:
Ping
There are problem with the old code when trying to install openslides on jessie. It would be easier if we just removed it.

Now this PR has to be rebased. Please give me a response if I should rebase this pull request or remove it.

normanjaeckel:
I prefer removing Reportlab as soon as possible but I do not know how many PDF views are already ported to JS.

emanuelschuetze:
Currently, assignments pdf and users access data pdf is still missing as pdfmake PDF. After this is finished we will make a 2.1-beta1 release.

emanuelschuetze:
@ostcar Please rebase. The last server pdf file is the assignment catalog pdf - Sean is working on this. But if you like we can temporary deactivate this pdf to merge this pull request.

emanuelschuetze:
Rebased. Ready for last review.

normanjaeckel:
Very good. I will merge this now. :)

PR #2477, opened 2016-10-01
Support to send many objects through the autoupdate system

ostcar:


openslides/utils/autoupdate.py
```diff
@@ -133,50 +136,80 @@ def send_data(message):
...
+    for root_instance in root_instances:
+        collection_elements.append(
+            CollectionElement.from_instance(
+                root_instance,
+                information=information))
+    # If currently there is an open database transaction, then the
+    # send_autoupdate function is only called, when the transaction is
+    # commited. If there is currently no transaction, then the function
+    # is called immediately.
+    transaction.on_commit(lambda: send_autoupdate(collection_elements))
```
normanjaeckel: You can somehow skip this line if the collection_elements set is empty.

openslides/utils/autoupdate.py
```diff
@@ -133,50 +136,80 @@ def send_data(message):
...
+        # Make surce instance is an iterable
+        instances = (instances, )
+    for instance in instances:
+        try:
+            root_instances.add(instance.get_root_rest_element())
+        except AttributeError:
+            # Instance has no method get_root_rest_element. Just ignore it.
+            pass
+
+    # Generates an collection element list for each root_instance.
```
normanjaeckel: The comment does not match to the code. Do you want to create many collection element lists or only one collection element list that contains all root instances?

openslides/utils/autoupdate.py
```diff
@@ -133,50 +136,80 @@ def send_data(message):
...
+    inform_deleted_data('motions/motion', 1, 'assignments/assignment', 5)
+
+    The argument information is added to each collection element.
     """
-    collection_element = CollectionElement.from_values(
-        collection_string=collection_string,
-        id=id,
-        deleted=True,
-        information=information)
+    if len(args) % 2 or not args:
```
normanjaeckel: Hm. This might be `if not args or not len(args) % 2:` because you want to catch zero args or an odd number of args.

openslides/utils/autoupdate.py
```diff
@@ -133,50 +136,80 @@ def send_data(message):
...
+        ValueError("inform_deleted_data has to be called with the same number of "
+                   "collection strings and ids. It has to be at least one "
+                   "collection string and one id.")
+
+    # Go through each pair of collection_string and id and generate a collection
+    # element from it.
+    collection_elements = CollectionElementList()
+    for index in range(len(args) // 2):
```
normanjaeckel: This looks wrong: If I have two pairs then range(len(args) // 2) returns [0,1]. Then the first loop run is correct, but the second loop run uses collection_string=args[1] and id=args[2] but it should be collection_string=args[2] and id=args[3].

May be you can use the step argument of the range function: `range(0, len(args), 2`? (Not tested.)

openslides/utils/autoupdate.py
normanjaeckel: Ah this is done below. Is this slower that doing the check here because of the transaction layer? I do not know. (It is not necessary to put the check in the function below because in delete case we always want to send something out. The check is only useful for the changed case. But this is no problem of course.)

openslides/utils/autoupdate.py
ostcar: Only one list. Updated the comment.

openslides/utils/autoupdate.py
ostcar: I did not want to do the if-clause two times.

openslides/utils/autoupdate.py
ostcar: Yes, and therefore my code is right.

openslides/utils/autoupdate.py
ostcar: Yes you are right. (I think I need tests for the code...)

openslides/utils/autoupdate.py
normanjaeckel: Ok.

PR #2484, opened 2016-10-04
Does not use gulp 'production' flag for openslides.js.

emanuelschuetze:
Comment out uglify function for openslides.js.
Added TODO that all OpenSlides js files needed rework before uglified code works correctly.


normanjaeckel:
One typo in comment, then merge.

PR #2489, opened 2016-10-06
list-of-speakers overlay (issue #2211, Version 4)

matakuka:
After all the master changes, it was easier to start a new branch instead of repairing the old one.
Previous discussion see #2371.

### DIFF
--- a/openslides/agenda/projector.py
+++ b/openslides/agenda/projector.py
@@ -1,5 +1,6 @@
 from ..core.config import config
 from ..core.exceptions import ProjectorException
+from ..core.models import Projector
 from ..utils.collection import CollectionElement
 from ..utils.projector import ProjectorElement
 from .models import Item
@@ -32,7 +33,6 @@ class ItemListSlide(ProjectorElement):
 class ListOfSpeakersSlide(ProjectorElement):
     """
     Slide definitions for Item model.
-
     This is only for list of speakers slide. You have to set 'id'.
     """
     name = 'agenda/list-of-speakers'
@@ -69,31 +69,56 @@ class ListOfSpeakersSlide(ProjectorElement):
             output.extend(self.get_requirements_as_collection_elements(config_entry))
         return output
 
+    def update_data(self):
+        return {'agenda_item_id': self.config_entry.get('id')}
 
-class CurrentListOfSpeakersSlide(ProjectorElement):
+
+class CurrentListOfSpeakersMetaClass(ProjectorElement):
+    """
+    Main class for the list of speaker slides.
+
+    """
+    def get_requirements(self, config_entry):
+        items = self.get_agenda_items(config['projector_currentListOfSpeakers_reference'])
+        for item in items:
+            yield item
+            for speaker in item.speakers.filter(end_time=None):
+                yield speaker.user
+            query = (item.speakers.exclude(end_time=None)
+                     .order_by('-end_time')[:config['agenda_show_last_speakers']])
+            for speaker in query:
+                # Yield last speakers
+                yield speaker.user
+
+    def get_agenda_items(self, projector_id):
+        projector = Projector.objects.get(pk=projector_id)
+        for element in projector.elements.values():
+            agenda_item_id = element.get('agenda_item_id')
+            if agenda_item_id is not None:
+                yield Item.objects.get(pk=agenda_item_id)
+
+    def get_collection_elements_required_for_this(self, collection_element, config_entry):
+        output = super().get_collection_elements_required_for_this(collection_element, config_entry)
+        # Full update if agenda_item changes because then we may have new
+        # candidates and therefor need new users.
+        items = self.get_agenda_items(config['projector_currentListOfSpeakers_reference'])
+        for item in items:
+            if collection_element == CollectionElement.from_values(item.get_collection_string(), item.pk):
+                output.extend(self.get_requirements_as_collection_elements(config_entry))
+                break
+        return output
+
+
+class CurrentListOfSpeakersSlide(CurrentListOfSpeakersMetaClass):
     """
     Slide for the current list of speakers.
-
-    Nothing special to check.
     """
     name = 'agenda/current-list-of-speakers'
 
-    def get_requirements(self, config_entry):
-        pk = config['projector_currentListOfSpeakers_reference']
-        if pk is not None:
-            # List of speakers slide.
-            try:
-                item = Item.objects.get(pk=pk)
-            except Item.DoesNotExist:
-                # Item does not exist. Just do nothing.
-                pass
-            else:
-                yield item
-                for speaker in item.speakers.filter(end_time=None):
-                    # Yield current speaker and next speakers
-                    yield speaker.user
-                query = (item.speakers.exclude(end_time=None)
-                         .order_by('-end_time')[:config['agenda_show_last_speakers']])
-                for speaker in query:
-                    # Yield last speakers
-                    yield speaker.user
+
+class CurrentListOfSpeakersOverlaySlide(CurrentListOfSpeakersMetaClass):
+    """
+    List of speakers overlay.
+    Subclass of ListOfSpeakers
+    """
+    name = 'agenda/current-list-of-speakers-overlay'
--- a/openslides/agenda/static/js/agenda/base.js
+++ b/openslides/agenda/static/js/agenda/base.js
@@ -298,7 +298,7 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
     'Agenda',
     function (Projector, Assignment, Topic, Motion, Agenda) {
         return {
-            getItem: function (projectorId) { 
+            getItem: function (projectorId) {
                 var elementPromise;
                 return Projector.find(projectorId).then(function (projector) {
                     // scan all elements
@@ -337,6 +337,59 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
         };
     }
 ])
+
+.factory('ListOfSpeakersOverlay', [
+    '$http',
+    'Projector',
+    'gettextCatalog',
+    'gettext',
+    function($http, Projector, gettextCatalog, gettext) {
+        var name = 'agenda/current-list-of-speakers-overlay';
+        return {
+            name: name,
+            verboseName: gettext('List of speakers overlay'),
+            project: function (projectorId, overlay) {
+                var isProjectedId = this.isProjected(overlay);
+                if (isProjectedId > 0) {
+                    // Deactivate
+                    var projector = Projector.get(isProjectedId);
+                    var uuid;
+                    _.forEach(projector.elements, function (element) {
+                        if (element.name == 'agenda/current-list-of-speakers-overlay') {
+                            uuid = element.uuid;
+                        }
+                    });
+                    $http.post('/rest/core/projector/' + isProjectedId + '/deactivate_elements/',
+                               [uuid]);
+                }
+                // Activate, if the projector_id is a new projector.
+                if (isProjectedId != projectorId) {
+                    return $http.post(
+                        '/rest/core/projector/' + projectorId + '/activate_elements/',
+                         [{name: 'agenda/current-list-of-speakers-overlay',stable: true}]);
+                }
+            },
+            isProjected: function (additionalId) {
+                // Returns the id of the last projector with an agenda-item element. Else return 0.
+                // additionalId is not needed
+                var isProjected = 0;
+                var predicate = function (element) {
+                    var value;
+                    value = element.name == 'agenda/current-list-of-speakers-overlay';
+                    return value;
+                };
+                Projector.getAll().forEach(function (projector) {
+                    if (typeof _.findKey(projector.elements, predicate) === 'string') {
+                        isProjected = projector.id;
+                    }
+                });
+                return isProjected;
+            }
+        };
+    }
+])
+
+
 
 // Make sure that the Agenda resource is loaded.
 .run(['Agenda', function(Agenda) {}]);
--- a/openslides/agenda/static/js/agenda/projector.js
+++ b/openslides/agenda/static/js/agenda/projector.js
@@ -16,6 +16,9 @@ angular.module('OpenSlidesApp.agenda.projector', ['OpenSlidesApp.agenda'])
         slidesProvider.registerSlide('agenda/current-list-of-speakers', {
             template: 'static/templates/agenda/slide-current-list-of-speakers.html',
         });
+        slidesProvider.registerSlide('agenda/current-list-of-speakers-overlay', {
+            template: 'static/templates/agenda/slide-current-list-of-speakers-overlay.html',
+        });
     }
 ])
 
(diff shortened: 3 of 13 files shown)


matakuka:
(question copied from the old PR)
@normanjaeckel 
I am trying it  on a new branch. I am almost ready, but I have two problems left.


**(1)**  The task is to make the geting of agenda_item in agenda/base.js generic.

DS.get() in [agenda/base.js](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/agenda/static/js/agenda/base.js#L291-L339) DS.get returns 'undefined' when I tried to refactor it and using the factory from from [core/site.js](https://github.com/matakuka/OpenSlides/blob/issue2211-multiprojector/openslides/core/static/js/core/site.js#L1543).

Maybe it is an easy error on my part. Please take a look on the approach I think of.
```
.factory('CurrentListOfSpeakersItem', [
    'DS',
    'Projector',
    'Agenda',
    function (DS, Projector, Agenda) {
        return {
            getItem: function (projectorId) {
                _.forEach(Projector.get(projectorId).elements, function (element) {
                    if(element.hasOwnProperty("agenda_item_id")){
                        return Agenda.get(DS.get(element.name, element.id).agenda_item_id);
                    }
                })
            }
        }
    }
])
```
(the currently existing approach works, so this is no blocker, but I understand you wanted it generic)

**(2)**  Yielding the speakers in [agenda/projector.py](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/agenda/projector.py#L73-L99). 
Here the current approach in master seems not to work in class ListOfSpeakersSlide. 

Currently, as I understand it, it gets the *agenda_item(projector_id)*, which may only work in tests if both are 1. I think that there is a need to get the Projector, THEN the item on the projector.

However, I am unsure how to get the Projector(projector_currentListOfSpeakers_reference) in python.
I thought it was 
`ProjectorManager.get_full_queryset().filter(pk=id)`, but it does not work.

Is this correct? If yes, how to handle it?

normanjaeckel:
(1): At the moment: Slides which provide a list of speakers item (motions/motion e. g. in contrast to users/user) do not keep this info in the projector element attribute. So element.hasOwnProperty("agenda_item_id") looks wrong at the moment. May be you have to add some logic to the update_data() of the motion, topic and assignment ProjectorElement. Then you can get the item id into the projector element data and there you can fetch using your factory.

(2) Projector.objects.get(pk=pk) returns the projector instance you want to use.

normanjaeckel:
Additionally: The CurrentListOfSpeakersSlide and the CurrentListOfSpeakersSlideOverlay are almost the same stuff. One has stable true and another template, that's all. Please to not use different ProjectorElements if possbile. Only deal with the stable flag.
If this is not possible, use derivated classes so we do not have to repeat ourselves.

matakuka:
Partially testable: Functional. 
Remaining issues (I'll need help on this): 

- [ ] When the list of speakers is updated, new speakers show up without names. I guess there is some missing resolve, but I do not understand where.
- [ ] I don't understand the failing integration/core/test_view test. It probably fails because I added a field (agenda_item_id) with update_data().

openslides/agenda/static/js/agenda/projector.js
```diff
@@ -36,6 +39,7 @@ angular.module('OpenSlidesApp.agenda.projector', ['OpenSlidesApp.agenda'])
         $scope.$watch(function () {
             return Agenda.lastModified();
         }, function () {
+            console.log("Agenda update");
```
matakuka: note to self: get rid

FinnStutzenstein:
The projector must get the users which should be displayed via autoupdate. So in the projector.py in get_requirements some logic should be added (or is not working; didn't look at your code)

matakuka:
I think it is working now as expected - except for the failing test

openslides/agenda/projector.py
```diff
@@ -69,31 +69,63 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
             output.extend(self.get_requirements_as_collection_elements(config_entry))
         return output
 
+    def update_data(self):
```
normanjaeckel: You can just return config_entry.get('id'). This is the same as the pk of the item. I think you want to get the pk and not the human readable item_number.

openslides/agenda/projector.py
```diff
@@ -69,31 +69,63 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
...
-                yield item
-                for speaker in item.speakers.filter(end_time=None):
-                    # Yield current speaker and next speakers
-                    yield speaker.user
+        items = self.get_agenda_items(config['projector_currentListOfSpeakers_reference'])
+        for item in items:
+            yield item
+            for speaker in item.speakers.filter(end_time=None):
+                yield speaker.user
                 query = (item.speakers.exclude(end_time=None)
```
normanjaeckel: This block should be less indended. I does not have to run inside the speakers loop.

openslides/agenda/projector.py
```diff
@@ -69,31 +69,63 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
...
                          .order_by('-end_time')[:config['agenda_show_last_speakers']])
                 for speaker in query:
                     # Yield last speakers
                     yield speaker.user
+
+    def get_agenda_items(self, projector_id):
+        projector = Projector.objects.get(pk=projector_id)
+        for element_searched in projector.elements:
```
normanjaeckel:
```python
for element in projector.elements.values():
    agenda_item_id = element.get('agenda_item_id')
    if agenda_item_id is not None:
        yield Item.objects.get(pk=agenda_item_id)
```

openslides/agenda/projector.py
```diff
@@ -69,31 +69,63 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
...
+                agenda_item_id = projector.elements[element_searched]['agenda_item_id']
+                # TODO: check in which cases agenda_item_id exist, but is ''
+                if agenda_item_id != '':
+                    yield Item.objects.get(pk=agenda_item_id)
+
+    def get_collection_elements_required_for_this(self, collection_element, config_entry):
+        output = super().get_collection_elements_required_for_this(collection_element, config_entry)
+        # Full update if agenda_item changes because then we may have new
+        # candidates and therefor need new users.
+        items = self.get_agenda_items(config['projector_currentListOfSpeakers_reference'])
```
normanjaeckel:
```python
items = self.get_agenda_items(config['projector_currentListOfSpeakers_reference'])
for item in items:
    if collection_element == CollectionElement.from_values(item.get_collection_string(), item.pk):
        output.extend(self.get_requirements_as_collection_elements(config_entry))
        break
return output
```

openslides/agenda/projector.py
```diff
@@ -69,31 +69,63 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
...
+    name = 'agenda/current-list-of-speakers'
+
+
+class CurrentListOfSpeakersOverlaySlide(CurrentListOfSpeakersMetaClass):
+    """
+    List of speakers overlay.
+    Subclass of ListOfSpeakers
+    """
+    name = 'agenda/current-list-of-speakers-overlay'
+    stable = True
```
normanjaeckel: This is not necessary. The client has to write the flag `stable` into the projector config. Please remove this line.

openslides/agenda/static/js/agenda/base.js
```diff
@@ -288,8 +288,6 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
     }
 ])
 
-// TODO: Remove all find() calls from the projector logic. It is also used on the site so this has to be
```
normanjaeckel: Hey, please don't delete the TODO withou fullfilling it. If this factory is not used by the projector any more, move it to site.js.

openslides/assignments/projector.py
```diff
@@ -54,3 +54,10 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
         if collection_element == CollectionElement.from_values(Assignment.get_collection_string(), config_entry.get('id')):
             output.extend(self.get_requirements_as_collection_elements(config_entry))
         return output
+
+    def update_data(self):
```
normanjaeckel:
```python
def update_data(self):
    data = None
    try:
        assignment = Assignment.objects.get(pk=self.config_entry.get('id'))
    except Assignment.DoesNotExist:
        # Assignment does not exist, so just do nothing.
        pass
    else:
        data = {'agenda_item_id': assignment.agenda_item_id}
    return data
```

openslides/core/static/js/core/site.js
```diff
@@ -764,18 +766,26 @@ angular.module('OpenSlidesApp.core.site', [
...
             }
-
+            $scope.getDefaultProjector();
             // stop ALL interval timer
             cancelIntervalTimers();
 
             rebuildAllElements();
         });
+        $scope.getDefaultProjector = function () {
+            var projectiondefault = ProjectionDefault.filter({name: 'agenda_current_list_of_speakers'})[0];
```
normanjaeckel: This line is not easy to understand. What do you mean by defaultProjector in this general context? Why do we use the first element of the filter array. You may use ProjectionDefault.get() instead. At least add a comment what this function is doing.

openslides/core/static/js/core/site.js
```diff
@@ -1003,6 +1014,58 @@ angular.module('OpenSlidesApp.core.site', [
...
+                _.findKey(Projector.get(currentOverlayProjector).elements, function(element) {
+                    if (element.name == 'agenda/current-list-of-speakers-overlay') {
+                        $http.post('/rest/core/projector/' + currentOverlayProjector + '/deactivate_elements/',
+                                   [element.uuid]);
+                    }
+                });
+            }
+            if (projector.id && projector.id > 0 && projector.id != currentOverlayProjector){
+                $http.post('/rest/core/projector/' + projector.id + '/activate_elements/',
+                           [{name: 'agenda/current-list-of-speakers-overlay'}]);
```
normanjaeckel: Here you might want to add `stable: true`. See countdown and messages.

openslides/motions/projector.py
```diff
@@ -37,3 +37,10 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
         if collection_element == CollectionElement.from_values(Motion.get_collection_string(), config_entry.get('id')):
             output.extend(self.get_requirements_as_collection_elements(config_entry))
         return output
+
+    def update_data(self):
```
normanjaeckel:
```python
def update_data(self):
    data = None
    try:
        motion = Motion.objects.get(pk=self.config_entry.get('id'))
    except Motion.DoesNotExist:
        # Motion does not exist, so just do nothing.
        pass
    else:
        data = {'agenda_item_id': motion.agenda_item_id}
    return data
```

openslides/topics/projector.py
```diff
@@ -22,3 +22,10 @@ def get_requirements(self, config_entry):
         else:
             yield topic
             yield topic.agenda_item
+
+    def update_data(self):
```
normanjaeckel:
```python
def update_data(self):
    data = None
    try:
        topic = Topic.objects.get(pk=self.config_entry.get('id'))
    except Topic.DoesNotExist:
        # Topic does not exist, so just do nothing.
        pass
    else:
        data = {'agenda_item_id': topic.agenda_item_id}
    return data
```

openslides/agenda/static/js/agenda/projector.js
matakuka: done

openslides/core/static/js/core/site.js
```diff
@@ -1003,6 +1014,59 @@ angular.module('OpenSlidesApp.core.site', [
...
+                return true;
+            } else {
+                return false;
+            }
+        };
+        $scope.goToListOfSpeakers = function() {
+            // go to the list of speakers (management) of the currently
+            // displayed projector slide
+            CurrentListOfSpeakersItem.getItem($scope.currentListOfSpeakersReference).then(function (success) {
+                console.log(success);
```
matakuka: done

openslides/agenda/projector.py
matakuka: done

openslides/agenda/projector.py
matakuka: done

openslides/agenda/projector.py
matakuka: done

openslides/agenda/static/js/agenda/base.js
matakuka: done, sorry

openslides/core/static/js/core/site.js
matakuka: renamed to `getDefaultOverlayProjector`. As for the filter, I oriented myself on all the other occurrences of projectionDefault throughout the code (e.g. topics/static/js/topics/site.js:166). I think it would be better to do this in one sepearate issue.

openslides/core/static/js/core/site.js
matakuka: done

openslides/motions/projector.py
matakuka: done

openslides/topics/projector.py
matakuka: done

matakuka:
@normanjaeckel 
(Almost) all remarks done (except the projectionDefault). I am sorry for the probable comment/mail spam, but I did not find any method for closing the review's point upon completion.

openslides/agenda/projector.py
```diff
@@ -70,10 +70,7 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
         return output
 
     def update_data(self):
-        data = {}
-        if Item.objects.filter(pk=self.config_entry.get('id')).exists():
-            data['agenda_item_id'] = Item.objects.get(pk=self.config_entry.get('id')).item_number
-        return data
+        return self.config_entry.get('id')
```
normanjaeckel: `return {'agenda_item_id': self.config_entry.get('id')}`?

normanjaeckel:
@emanuelschuetze Please test this carefully.

openslides/agenda/static/templates/agenda/slide-current-list-of-speakers-overlay.html
```diff
@@ -10,8 +10,10 @@ <h3 translate>Current list of speakers</h3>
     class="currentSpeaker">
     <i class="fa fa-microphone fa-lg"></i> {{ speaker.user.get_full_name() }}
     <ol class="nextSpeakers">
-      <li ng-repeat="speaker in agendaItem.speakers | filter: {begin_time: null} | orderBy:'weight' | limitTo: 3">
+      <li ng-repeat="speaker in nextSpeakers = (agendaItem.speakers | filter: {begin_time: null}) | orderBy:'weight' | limitTo: 3">
         {{ speaker.user.get_full_name() }}
     </ol>
+      <p ng-if="nextSpeakers.length > 3" class="lastSpeakers">
+       <i> {{ nextSpeakers.length - 3 }} <translate> more speakers</translate> </i>
```
emanuelschuetze: Suggestion: Use just this short string without translations:
`<i>+ {{ nextSpeakers.length - 3 }}</i>`

emanuelschuetze:
One remark. Rest works good. Please fix this last remark and rebase. Then it's ready for final review.

PR #2503, opened 2016-10-12
remove run.sh from Dockerfile

boehlke:


emanuelschuetze:
@normanjaeckel I'll merge it directly because @boehlke need it for automated build in https://hub.docker.com/r/openslides/openslides/

Issue #2258, opened 2016-08-08
Majority indicators on polls


matakuka:
(I am currently working on:)

Creating an indicator for each election candiate to see if a certain mayority is reached:
- relative mayority (more votes than other candidates)
- absolute mayority (50% + 1 of the valid votes)
- two/third mayority (66% + 1 of the valid votes)
- ... (suggestions?)

I think it should be a button with a drop-down menu (to select the mayority calculation). Candidates that pass the calculation will be displayed differently, so a voting committee can easily see which candidates reached the threshold. On mouseover/click, an informative text is displayed that tells details:
- which of the mayorities is reached/not reached
- how many votes are missing/in excess for the respective step
- example: `John Doe got 50 of 100 votes. This is the relative mayority (10 votes). This is not the absolute mayority of 51 votes (1 vote missing). This is not a two-third mayority of 67 votes (17 votes missing)`

I guess there will be some questions on the more complicated voting methods and their mayority calculations, especially in connection with e.g. yes/no/abstain method.

normanjaeckel:
Do not forget to care about the yes-no-abstain election modus. Then the  meaning of the terms of majority changes ...

matakuka:
Indeed it gets complicated. My *feeling* says the percentages are is always on the base of the valid votes, but I did not dig into the deeps of democratic calculations yet. I will soon prepare a list/chart of all election modes and possible results.
**When this list is done, I will need advice from a more user side view**, being "which is the correct definition of majority `#n`".

What I'm doing beside this democrato-mathematic stuff:
- creating two (server-side) poll variables that sort the candidates after the poll has been commited: 
 - `position` in the list of candidates on account of `Yes` votes (if candidates share the amount of yes votes, lower `No` Votes are first). 
   - Server-side because else I would need to retrieve every candidates' votes before displaying a result. Also, this sorting could get useful elsewhere (sorting the result slide by position instead of name) 
 - `is_uniqe` Set to False if there are other candidates with the same amount of votes.
   - If there are more candidates than posts left, the resulting color/message should be a warning, allowing for a second election round. If not, it is just a question of ego who is displayed first, which I would leave to `orderBy name`.
- client side calculations:
 -  votes | yesvotes > than (`Yes * 3 / 2`| `Yes /2` |`Yes * (whatever)`?
   - TODO here: which is the base?
 - is `unique= false` relevant for the current position? (If there are less or equal candidates than posts available, we can return a "green") 
[...]

PR #2509, opened 2016-10-14
Majorities (issue #2258)

matakuka:
This is a reduced, first version, for motions only. 
It checks the inputs of a poll, and if the calculation is possible, it shows how far from the quorum the decision is.

TODO:
- [ ]  beautify/design
- [ ]  move logic to a more generic place
- [ ] add assignment methods


matakuka:
Some integration tests fail after inserting another config variable at motions, I am unsure which ones.

emanuelschuetze:
One commit added. All done. I'll mark the translation strings of the new config values in a later pull request.
@normanjaeckel Please review, rebase and merge.

PR #2510, opened 2016-10-16
Support https as websocket protocol (wss).

emanuelschuetze:


openslides/core/static/js/core/base.js
```diff
@@ -52,6 +52,13 @@ angular.module('OpenSlidesApp.core', [
         var recInterval = null;
         $rootScope.connected = false;
 
+        var websocketProtocol;
+        if (location.protocol == 'https:') {
+            websocketProtocol = 'wss://';
```
normanjaeckel: The double slash is not part of the protocol. Please remove it from this variable and re-add it in line 83.

emanuelschuetze:
I fixed the remark and rebased. So ready for merge.

PR #2514, opened 2016-10-17
Round corners for projectorbuttons (fixes #2494)

### LINKED ISSUE #2494: Sharp corners in projector button
The projector button has two round and two sharp corners if there is only one projector. Expected: 4 round corners. If looks like that the dropdown is simply cut off.

FinnStutzenstein:
Hide the ul-node so the button gets round corners.

### DIFF
--- a/openslides/agenda/static/templates/agenda/item-detail.html
+++ b/openslides/agenda/static/templates/agenda/item-detail.html
@@ -26,7 +26,7 @@
           uib-dropdown-toggle>
           <span class="caret"></span>
         </button>
-        <ul class="dropdown-menu">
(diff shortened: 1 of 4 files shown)


emanuelschuetze:
Not fixed for (custom) projector buttons of agenda items and mediafiles. Rest looks good.

FinnStutzenstein:
@emanuelschuetze Ping.

emanuelschuetze:
Restested: Looks good for agenda items and medafiles. But the projection button for countdown and messages has still the old (sharp) corners. But this will be fixed by using the (default) projection button also for countdown and messages (see #/2464). So ready for merge.

PR #2516, opened 2016-10-17
Remove chat history

FinnStutzenstein:
New permission "can_manage_chat", trashbutton shows up in chat. ToDo: server logic and request in `clearChatHistory` (core/base.js)

@normanjaeckel should this permission be added to group 3 staff via `RunPython`?


normanjaeckel:
It is not possible to add new permissions to a group during migrations because of Django's way to create the new permission after all migration are finished.

PR #2519, opened 2016-10-17
Updated translation resource. Updated German translations.

emanuelschuetze:
I'll merge it dircectly.


gulpfile.js
```diff
@@ -170,6 +170,7 @@ gulp.task('pot', function () {
     return gulp.src([
             'openslides/core/static/templates/*.html',
             'openslides/*/static/templates/*/*.html',
+            'openslides/*/static/templates/*/*/*.html',
```
normanjaeckel: We should try `'openslides/*/static/templates/**/*.html'`. This is possbily more generic.

emanuelschuetze: Fixed in #2996

PR #2524, opened 2016-10-24
Fixed migrations by skipping autoupdate.

normanjaeckel:
@emanuelschuetze Please test this.


emanuelschuetze:
@normanjaeckel see #2525 (There I added also skip autoupdate for group migrations)

normanjaeckel:
Groups are not Autoupdate models. So you don't have to skip something.

PR #2525, opened 2016-10-24
Skip autoupdates for all migration files.

emanuelschuetze:
@normanjaeckel I changed all save methods of all migration files. Correct? Please read it carfully.


openslides/users/migrations/0004_groups.py
```diff
@@ -31,9 +31,9 @@ def migrate_groups_and_user_permissions(apps, schema_editor):
         if group_default.name == 'Guests' and group_registered.name == 'Registered users':
             # Rename groups pk 1 and 2.
             group_default.name = 'Default'
-            group_default.save()
+            group_default.save(skip_autoupdate=True)
```
normanjaeckel: This is not necessary.

emanuelschuetze:
duplicate, see #2524

PR #2537, opened 2016-10-27
Caching results of insertLineNumbers and extractRangeByLineNumbers

CatoTH:
The first, simpler part of the performance improvements

### DIFF
--- a/openslides/motions/static/js/motions/diff.js
+++ b/openslides/motions/static/js/motions/diff.js
@@ -4,261 +4,229 @@
 
 angular.module('OpenSlidesApp.motions.diff', ['OpenSlidesApp.motions.lineNumbering'])
 
-.service('diffService', function (lineNumberingService) {
-    var ELEMENT_NODE = 1,
-        TEXT_NODE = 3,
-        DOCUMENT_FRAGMENT_NODE = 11;
+.service('diffService', [
+    'lineNumberingService',
+    '$cacheFactory',
+    function (lineNumberingService, $cacheFactory) {
+        var ELEMENT_NODE = 1,
+            TEXT_NODE = 3,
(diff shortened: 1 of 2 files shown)


openslides/motions/static/js/motions/diff.js
```diff
@@ -4,11 +4,13 @@
 
 angular.module('OpenSlidesApp.motions.diff', ['OpenSlidesApp.motions.lineNumbering'])
 
-.service('diffService', function (lineNumberingService) {
+.service('diffService', function (lineNumberingService, $cacheFactory) {
```
normanjaeckel: Ah I think it is necessary to use the array notation:
```javascript
.service('diffService', [
    'lineNumberingService',
    '$cacheFactory',
    function (lineNumberingService, $cacheFactory) {
        ...
    }
])
```
This is required if we later want to compress and uglify the code via gulp.

openslides/motions/static/js/motions/linenumbering.js
```diff
@@ -14,14 +14,25 @@ angular.module('OpenSlidesApp.motions.lineNumbering', [])
  * No constructs like <a...><div></div></a> are allowed. CSS-attributes like 'display: block' are ignored.
  */
 
-.service('lineNumberingService', function () {
+.service('lineNumberingService', function ($cacheFactory) {
```
normanjaeckel: Here the same.

thoessl:
@normanjaeckel Like in the second commit now? If it's ok like this, I'll squash those two commits.

normanjaeckel:
Looks good.

emanuelschuetze:
@CatoTH Please rebase. Then I'll merge.

PR #2587, opened 2016-11-04
Block deleting of the last group with user manage permissions

### LINKED ISSUE #2544: Don't allow to delete user groups "staff" and "delegates"
OS 2.1s new group policy allows to delete every user group. OS offers predefined groups like "delegates". In my opinion it is not a good feature that one can delete these groups. The purpose of OS is to provide a tool for conventions and there will always be delegates and a "staff".

So in my opinion these groups should not be deleteable.

But maybe there are good reasons to allow this?

FinnStutzenstein:
This closes #2544 .

### DIFF
--- a/openslides/users/static/templates/users/group-list.html
+++ b/openslides/users/static/templates/users/group-list.html
@@ -41,6 +41,7 @@
                 &nbsp;
             <!--delete-->
             <a href="" class="text-danger"
+              ng-if="group_danger.id != group.id"
               ng-bootbox-confirm="{{ 'Are you sure you want to delete this entry?' | translate }}<br>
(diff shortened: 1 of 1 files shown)


normanjaeckel:
Nice. What about the server side?

FinnStutzenstein:
There is (sadly) no protection at all. A user is not blocked to lock himself out (the "dangerous permissions" users.can_see and users.can_manage) as well as not prevented to delete the last group with those permissions.
All logic on the client site could be taken to make checks on the server side if a group is updated (the permissions) or deleted.

emanuelschuetze:
Ready for merge.

PR #2591, opened 2016-11-04
Improve importing and sorting of users

### LINKED ISSUE #2136: regression: users are always sorted with their first name
In the settings under users there is the option "sort users with first name".
This option is now useless because the sorting is always with the first name. This is a regression to older OpenSlides versions as one can see in this screenshot of the manual for OpenSlides 1.5.1:
https://docs.openslides.org/de/_images/Tutorial_3_03.png

emanuelschuetze:
- Add is_present field for csv import of users.
- Refactor JS functions get_full_name and get_short_name (Fixes #2136)


openslides/users/models.py
```diff
@@ -175,44 +175,7 @@ class Meta:
...
-            else:
-                name = ', '.join((last_name, first_name))
-
-        # The user has only a first_name or a last_name or no name
-        else:
-            name = first_name or last_name or self.username
-
-        # Return result
-        return name
+        return ' '.join((self.first_name, self.last_name))
```
normanjaeckel: This is too short. If first_name and last_name is empty we want to use the username. See old logic in get_short_name().

openslides/users/static/js/users/site.js
```diff
@@ -1374,10 +1385,8 @@ angular.module('OpenSlidesApp.users.site', [
 
         // config strings in users/config_variables.py
         gettext('[Place for your welcome and help text.]');
-        gettext('Sort users by first name');
-        gettext('Disable for sorting by last name');
+        gettext('Sort name of participants by');
         gettext('Participants');
```
normanjaeckel: Add translation strings for new choices here.

openslides/users/models.py
emanuelschuetze: I changed it to old logic (although first AND last name can not be empty because you have to enter at least one name part. And the python __str__ method is now used for debugging and printing only, isn't it?)

openslides/users/static/js/users/site.js
emanuelschuetze: Done

emanuelschuetze:
@normanjaeckel All done.

emanuelschuetze:
@normanjaeckel ready for last review.

openslides/users/pdf.py
```diff
@@ -24,10 +24,7 @@ def users_to_pdf(pdf):
     Create a list of all users as PDF.
     """
     data = [['#', _('Name'), _('Structure level'), _('Group')]]
-    if config['users_sort_users_by_first_name']:
-        sort = 'first_name'
-    else:
-        sort = 'last_name'
+    sort = 'first_name'
```
normanjaeckel: Do you have CamelCase here?

openslides/users/config_variables.py
```diff
@@ -11,7 +11,7 @@ def get_config_variables():
     # Sorting
     yield ConfigVariable(
         name='users_sort_by',
-        default_value='first_name',
+        default_value='firstName',
```
normanjaeckel: I think we did not use CamelCase on server side before, did we?

openslides/users/pdf.py
emanuelschuetze: No, order_by requires the db field 'first_name' or 'last_name'.

openslides/users/config_variables.py
emanuelschuetze: Yes, but now we use this config variable on client side only. There we use CamelCase. So I don't want to change it again back.

normanjaeckel:
Rebase and merge.

PR #2592, opened 2016-11-04
New table design for users

FinnStutzenstein:
@emanuelschuetze This is the new user table. Please test it, I haven't had enough time to do this proper, but (as always) it should be functional ;) Please give some feedback for the layout of the single user rows.

Also in this PR I am going to make the table layout more clean (html and css) so that the basic structure could be adapted for every table. But this is currently in a wip state.

### DIFF
--- a/openslides/core/static/js/core/base.js
+++ b/openslides/core/static/js/core/base.js
@@ -699,32 +699,6 @@ angular.module('OpenSlidesApp.core', [
     }
 ])
 
-/*
- * This filter filters all items in an array. If the filterArray is empty, the
- * array is passed. The filterArray contains numbers of the multiselect, e. g. [1, 3, 4].
- * Then, all items in the array are passed, if the item_id (get with id_function) matches
- * one of the ids in filterArray. id_function could also return a list of ids. Example:
- * Item 1 has two tags with ids [1, 4]. filterArray == [3, 4] --> match
- */
-.filter('SelectMultipleFilter', [
-    function () {
-        return function (array, filterArray, idFunction) {
-            if (filterArray.length === 0) {
-                return array;
-            }
-            return Array.prototype.filter.call(array, function (item) {
-                var id = idFunction(item);
-                if (!id) {
-                    return false;
-                } else if (typeof id === 'number') {
-                    id = [id];
-                }
-                return _.intersection(id, filterArray).length > 0;
-            });
-        };
-    }
-])
-
 // mark HTML as "trusted"
 .filter('trusted', [
     '$sce',
--- a/openslides/core/static/js/core/site.js
+++ b/openslides/core/static/js/core/site.js
@@ -342,6 +342,91 @@ angular.module('OpenSlidesApp.core.site', [
     }
 ])
 
+.factory('Multiselect', [
+    function () {
+        var instance = function () {
+            var areFiltersSet = function () {
+                var areFiltersSet = false;
+                _.forEach(self.filters, function (filterList, filter) {
+                    if (filterList.length > 0) {
+                        areFiltersSet = true;
+                    }
+                });
+                return areFiltersSet;
+            };
+            var resetFilters = function () {
+                _.forEach(self.filters, function (filterList, filter) {
+                    self.filters[filter] = [];
+                });
+            };
+            var getFilterString = function (obj) {
+                var newList = [];
+                _.forEach(self.propertyList, function (property) {
+                    newList.push(obj[property]);
+                });
+                _.forEach(self.propertyFunctionList, function (fn) {
+                    newList.push(fn(obj));
+                });
+                _.forEach(self.propertyDict, function (idFunction, property) {
+                    newList.push(_.map(obj[property], idFunction).join(' '));
+                });
+                return newList.join(' ');
+            };
+            var operate = function (filter, id) {
+                if (_.indexOf(self.filters[filter], id) > -1) {
+                    // remove id
+                    self.filters[filter] = _.filter(self.filters[filter], function (_id) {
+                        return _id != id;
+                    });
+                } else {
+                    // add id
+                    self.filters[filter].push(id);
+                }
+            };
+
+            var self = {
+                filters: {},
+                areFiltersSet: areFiltersSet,
+                resetFilters: resetFilters,
+                operate: operate,
+                getFilterString: getFilterString,
+            };
+            resetFilters(); //Initiate filters
+            return self;
+        };
+
+        return {
+            instance: instance
+        };
+    }
+])
+
+/*
+ * This filter filters all items in an array. If the filterArray is empty, the
+ * array is passed. The filterArray contains numbers of the multiselect, e. g. [1, 3, 4].
+ * Then, all items in the array are passed, if the item_id (get with id_function) matches
+ * one of the ids in filterArray. id_function could also return a list of ids. Example:
+ * Item 1 has two tags with ids [1, 4]. filterArray == [3, 4] --> match
+ */
+.filter('MultiselectFilter', [
+    function () {
+        return function (array, filterArray, idFunction) {
+            if (filterArray.length === 0) {
+                return array;
+            }
+            return Array.prototype.filter.call(array, function (item) {
+                var id = idFunction(item);
+                if (!id) {
+                    return false;
+                } else if (typeof id === 'number') {
+                    id = [id];
+                }
+                return _.intersection(id, filterArray).length > 0;
+            });
+        };
+    }
+])
+
 // Load the django url patterns
 .run([
     'runtimeStates',
--- a/openslides/motions/static/js/motions/site.js
+++ b/openslides/motions/static/js/motions/site.js
@@ -653,6 +653,33 @@ angular.module('OpenSlidesApp.motions.site', [
     }
 ])
 
+.factory('MotionCsvExport', [
+    function () {
+        return function (element, motions) {
+            var csvRows = [
+                ['identifier', 'title', 'text', 'reason', 'submitter', 'category', 'origin'],
+            ];
+            _.forEach(motions, function (motion) {
+                var row = [];
+                row.push('"' + motion.identifier + '"');
+                row.push('"' + motion.getTitle() + '"');
+                row.push('"' + motion.getText() + '"');
+                row.push('"' + motion.getReason() + '"');
+                row.push('"' + motion.submitters[0].get_full_name() + '"');
+                var category = motion.category ? motion.category.name : '';
+                row.push('"' + category + '"');
+                row.push('"' + motion.origin + '"');
+                csvRows.push(row);
+            });
+
+            var csvString = csvRows.join("%0A");
+            element.href = 'data:text/csv;charset=utf-8,' + csvString;
+            element.download = 'motions-export.csv';
+            element.target = '_blank';
+        };
+    }
+])
+
 // Cache for MotionPollDetailCtrl so that users choices are keeped during user actions (e. g. save poll form).
 .value('MotionPollDetailCtrlCache', {})
 
@@ -723,9 +750,11 @@ angular.module('OpenSlidesApp.motions.site', [
     'HTMLValidizer',
     'Projector',
     'ProjectionDefault',
+    'MotionCsvExport',
+    'Multiselect',
     function($scope, $state, $http, ngDialog, MotionForm, Motion, Category, Tag, Workflow, User, Agenda, MotionBlock,
                 MotionDocxExport, MotionContentProvider, MotionCatalogContentProvider, PdfMakeConverter, PdfMakeDocumentProvider,
-                gettextCatalog, HTMLValidizer, Projector, ProjectionDefault) {
+                gettextCatalog, HTMLValidizer, Projector, ProjectionDefault, MotionCsvExport, Multiselect) {
         Motion.bindAll({}, $scope, 'motions');
         Category.bindAll({}, $scope, 'categories');
         MotionBlock.bindAll({}, $scope, 'motionBlocks');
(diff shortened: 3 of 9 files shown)


FinnStutzenstein:
Todo: Filter for motions that have no state (new option in state,... filters)

openslides/users/static/templates/users/user-list.html
```diff
@@ -134,94 +105,280 @@ <h1 translate>Participants</h1>
...
+          <small>
+            <div ng-if="user.groups_id.length">
+              <i class="fa fa-users"></i>
+              <span ng-repeat="group in user.groups_id | limitTo:2">
+                {{ (groups | filter: {id: group})[0].name | translate }}<span ng-if="!$last">,</span></span><span ng-if="user.groups_id.length > 2">,
+                ... [+{{ user.groups_id.length - 2}}]</span>
+              <!-- sorry for merging them together, but otherwise there would be a whitespace because of the new line -->
+            </div>
+            <div ng-if="user.structure_level">
+              <i class="fa fa-flag-o"></i>
```
emanuelschuetze: use a black flag: `fa-flag`

openslides/users/static/templates/users/user-list.html
```diff
@@ -134,94 +105,280 @@ <h1 translate>Participants</h1>
...
+        <projector-button model="user", default-projector-id="defaultProjectorId">
+        </projector-button>
+      </div>
+      <!-- main content column -->
+      <div class="col-xs-6 content">
+        <div class="spacer-right"> <!-- horizontal block -->
+            <strong>
+              <i ng-style="{'visibility': user.is_active === false ? 'visible' : 'hidden'}" class="fa fa-ban"
+                title="{{ 'Is inactive' | translate }}"></i>
+              <i ng-style="{'visibility': user.is_committee ? 'visible' : 'hidden'}" class="fa fa-users"
```
emanuelschuetze: Suggestion: Use 'fa-university' icon for a better distinction of user group icon.

openslides/users/static/templates/users/user-list.html
```diff
@@ -50,37 +30,28 @@ <h1 translate>Participants</h1>
...
-        </div>
-        <button class="btn btn-default" ng-click="$parent.isFilterOpen = !$parent.isFilterOpen"
-            ng-class="$parent.isFilterOpen ? 'btn-primary' : 'btn-default'">
-          <i class="fa fa-filter"></i>
-          <translate>Filter ...</translate>
+      <div class="dropdown pull-right" uib-dropdown>
+        <button os-perms="users.can_manage" class="btn btn-default" uib-dropdown-toggle
+            id="user-pdf-dropdown">
+          <i class="fa fa-file-pdf-o fa-lg"></i>
+          <translate>PDF Export</translate>
```
emanuelschuetze: Change the button text if table is filtered - as in motion list view.

emanuelschuetze: And maybe you can easily add a nice csv export of filtered data - like in motions?

emanuelschuetze:
Looks good. Some remarks.

FinnStutzenstein:
@emanuelschuetze ping

emanuelschuetze:
@normanjaeckel Please review.

openslides/users/static/js/users/base.js
```diff
@@ -100,8 +100,12 @@ angular.module('OpenSlidesApp.users', [])
                     var title = _.trim(this.title),
                         firstName = _.trim(this.first_name),
                         lastName = _.trim(this.last_name),
-                        name = '';
-                    if (Config.get('users_sort_by').value == 'lastName') {
+                        name = '',
```
normanjaeckel: As @emanuelschuetze said this change might be not necessary any more.

openslides/users/static/js/users/site.js
```diff
@@ -445,10 +445,27 @@ angular.module('OpenSlidesApp.users.site', [
...
+            {name: 'is_committee',
+             display_name: 'Committee'},
+            {name: 'number',
+             display_name: 'Number'},
+            {name: 'structure_level',
+             display_name: 'Structure level'},
+            {name: 'comment',
+             display_name: 'Comment'},
+        ];
+        $scope.sortColumn = ($scope.config('users_sort_by') == 'firstName') ? 'first_name' : 'last_name';
```
normanjaeckel: It would be better if the config option uses underscore and not camel case, so we can better rely on the model attribute which is also underscore in JavaScript.

openslides/users/static/js/users/site.js
```diff
@@ -458,6 +475,62 @@ angular.module('OpenSlidesApp.users.site', [
             }
             $scope.sortColumn = column;
         };
+        // define custom search filter string
+        $scope.getFilterString = function (user) {
```
normanjaeckel: This function looks a little bit like the one in motion list controller. Can we have a generic solution here?

openslides/users/static/js/users/site.js
```diff
@@ -458,6 +475,62 @@ angular.module('OpenSlidesApp.users.site', [
...
+                    function (group) {
+                        return group.name;
+                    }
+                ).join(" "),
+            ].join(" ");
+        };
+        $scope.multiselectFilter = {
+            group: [],
+        };
+        $scope.getItemId = {
```
normanjaeckel: Here the same. If we have a new style filter table service, it can be configured with single and multi select filters and provide this code in a generic way.

openslides/users/static/js/users/site.js
```diff
@@ -458,6 +475,62 @@ angular.module('OpenSlidesApp.users.site', [
...
+                        return _id != id;
+                    });
+                } else {
+                    // add id
+                    $scope.multiselectFilter[filter].push(id);
+                }
+            }
+        };
+        // for reset-button
+        $scope.reset_filters = function () {
```
normanjaeckel: Please use camel case.

openslides/users/static/js/users/site.js
```diff
@@ -458,6 +475,62 @@ angular.module('OpenSlidesApp.users.site', [
...
+                group: [],
+            };
+            if ($scope.filter) {
+                $scope.filter.search = '';
+            }
+            $scope.isPresentFilter = undefined;
+            $scope.isActiveFilter = undefined;
+            $scope.isCommitteeFilter = undefined;
+        };
+        $scope.are_filters_set = function () {
```
normanjaeckel: Please use camel case.

openslides/users/static/js/users/site.js
```diff
@@ -559,20 +633,46 @@ angular.module('OpenSlidesApp.users.site', [
...
         $scope.makePDF_userAccessDataList = function () {
             var filename = gettextCatalog.getString("List of access data")+".pdf";
             var userAccessDataListContentProvider = UserAccessDataListContentProvider.createInstance(
-                $scope.users, $scope.groups, Config);
+                $scope.usersFiltered, $scope.groups, Config);
             var documentProvider = PdfMakeDocumentProvider.createInstance(userAccessDataListContentProvider);
             pdfMake.createPdf(documentProvider.getDocument()).download(filename);
         };
+        // Export as a csv file
+        $scope.csv_export = function () {
```
normanjaeckel: Please use camel case.

normanjaeckel: Please move export code to a separate angular module.

openslides/users/static/js/users/site.js
FinnStutzenstein: I think not, because not all properties of user should be in the filterString (e.g. default_passwort). So we have to decide for every model, what should be in the string.

openslides/users/static/js/users/base.js
FinnStutzenstein: Done.

openslides/users/static/js/users/site.js
FinnStutzenstein: Done.

openslides/users/static/js/users/site.js
FinnStutzenstein: Done.

openslides/users/static/js/users/site.js
FinnStutzenstein: Done.

FinnStutzenstein:

openslides/core/static/js/core/base.js
```diff
@@ -689,7 +689,7 @@ angular.module('OpenSlidesApp.core', [
  * one of the ids in filterArray. id_function could also return a list of ids. Example:
  * Item 1 has two tags with ids [1, 4]. filterArray == [3, 4] --> match
  */
-.filter('SelectMultipleFilter', [
+.filter('MultiselectFilter', [
```
emanuelschuetze: Why this filter is in base.js and not in site.js? Does we need this filter also on projector?

openslides/core/static/js/core/site.js
```diff
@@ -351,6 +351,64 @@ angular.module('OpenSlidesApp.core.site', [
...
+        this.$get = function () {
+            var instance = function () {
+                var filters = {};
+                var getFilters = function () { return filters; };
+                var itemId = {
+                    group: function (obj) {return obj.groups_id;},
+                    state: function (obj) {return obj.state_id;},
+                    category: function (obj) {return obj.category_id;},
+                    motionBlock: function (obj) {return obj.motion_block_id;},
+                    tag: function (obj) {return obj.tags_id;}
```
emanuelschuetze: mmh, is this good to use in core such specific keys from our motion and users app?

emanuelschuetze:
Retested: works successfully. Some minor code styling remarks.

emanuelschuetze:
@normanjaeckel Please review again!

openslides/core/static/js/core/site.js
normanjaeckel: I agree. We should not use motion specific stuff in core files.

openslides/core/static/js/core/base.js
normanjaeckel: Question is still open.

openslides/core/static/js/core/site.js
```diff
@@ -351,6 +351,64 @@ angular.module('OpenSlidesApp.core.site', [
     }
 ])
 
+.provider('MultiselectFilterProvider', [
```
normanjaeckel: Why do you use a provider and not a factory (or a service)? A provider is useful if we want to configure something in the pre-run phase of Angular. Also the provider should not called Provider, see [here] (https://docs.angularjs.org/guide/providers#provider-recipe).

openslides/users/static/js/users/site.js
normanjaeckel: Yes but you can handle a list of property names that you want to use in the filter and a generic function cares about creating the whitespace join. So we do not have to program the logic of joining strings with whitespace twice.

FinnStutzenstein:
@emanuelschuetze @normanjaeckel ping

With the next table (probably agenda), I will try to put multiselect and sorting together to make this whole filter/sort unit generic.

emanuelschuetze:
Successfully tested. Two last suggestions:
* Please add tooltips for the icons for structure level, groups and comment.
* It would be great, if you can change the groups of each user in the list view by drop down menu via a cog icon (same like tags/categories in motion list view).

openslides/core/static/js/core/site.js
```diff
@@ -351,6 +351,95 @@ angular.module('OpenSlidesApp.core.site', [
...
+                };
+            };
+            var getFilterString = function (obj) {
+                var newList = [];
+                _.forEach(self.propertyList, function (property) {
+                    newList.push(obj[property]);
+                });
+                _.forEach(self.propertyFunctionList, function (fn) {
+                    var tmp = fn(obj);
+                    console.log(tmp);
```
emanuelschuetze: please remove this line

openslides/motions/static/js/motions/site.js
```diff
@@ -782,7 +835,7 @@ angular.module('OpenSlidesApp.motions.site', [
             $scope.sortColumn = column;
         };
         // define custom search filter string
-        $scope.getFilterString = function (motion) {
+        /*$scope.getFilterString = function (motion) {
```
emanuelschuetze: can these lines removed?

openslides/users/config_variables.py
```diff
@@ -11,12 +11,12 @@ def get_config_variables():
...
         name='users_sort_by',
-        default_value='firstName',
+        default_value='first_name',
         input_type='choice',
         label='Sort name of participants by',
         choices=(
-            {'value': 'firstName', 'display_name': 'First name'},
-            {'value': 'lastName', 'display_name': 'Last name'}),
+            {'value': 'first_name', 'display_name': 'First name'},
+            {'value': 'last_name', 'display_name': 'Last name'}),
```
emanuelschuetze: If you change the config value you have to change also the get_short_name function in /users/base.js:
https://github.com/OpenSlides/OpenSlides/blob/30fd55409b1b5ab919a87afabddae126d867915f/openslides/users/static/js/users/base.js#L104

FinnStutzenstein: Done.

openslides/core/static/js/core/site.js
FinnStutzenstein: Ok, every table now defines their own needed filters. Done.

openslides/motions/static/js/motions/site.js
```diff
@@ -743,35 +772,65 @@ angular.module('OpenSlidesApp.motions.site', [
...
+            'tags' : function (tag) {
+                return tag.name;
+            },
+        };
         $scope.getItemId = {
-            state: function (motion) {return motion.state_id;},
-            category: function (motion) {return motion.category_id;},
-            motionBlock: function (motion) {return motion.motion_block_id;},
-            tag: function (motion) {return motion.tags_id;}
+                state: function (motion) {return motion.state_id;},
```
normanjaeckel: Indention is not correct.

FinnStutzenstein:
Ok, todos: the remakrs and one stylistic change.

emanuelschuetze:
@FinnStutzenstein This pr is important. I want to merge it asap. Maybe you can split out my open remarks in an other pr.

FinnStutzenstein:
Going to be finished tomorrow.

FinnStutzenstein:
OK

PR #2608, opened 2016-11-08
Load angular-gettext language files from plugin (Fixed #2339)

### LINKED ISSUE #2339: Load angular-gettext language files from plugins
Find a way to load angular-gettext json files from plugins

Currently [here](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/core/static/js/core/base.js#L128) the language file from OpenSlides core is loaded - depending on selected language: `gettextCatalog.loadRemote("static/i18n/" + lang + ".json");`

emanuelschuetze:
Add new OpenSlidesPlugins provider where plugins can be registered.

### DIFF
--- a/openslides/core/static/js/core/base.js
+++ b/openslides/core/static/js/core/base.js
@@ -108,7 +108,8 @@ angular.module('OpenSlidesApp.core', [
 .factory('Languages', [
     'gettext',
     'gettextCatalog',
-    function (gettext, gettextCatalog) {
+    'OpenSlidesPlugins',
+    function (gettext, gettextCatalog, OpenSlidesPlugins) {
         return {
             // get all available languages
             getLanguages: function () {
@@ -144,6 +145,7 @@ angular.module('OpenSlidesApp.core', [
             // set current language and return updated languages object array
             setCurrentLanguage: function (lang) {
                 var languages = this.getLanguages();
+                var plugins = OpenSlidesPlugins.getAll();
                 angular.forEach(languages, function (language) {
                     language.selected = false;
                     if (language.code == lang) {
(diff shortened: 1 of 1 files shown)


emanuelschuetze:
The plugin side can be found here https://github.com/OpenSlides/openslides-votecollector/pull/42 - please review too. Thanks

openslides/core/static/js/core/base.js
```diff
@@ -460,6 +470,26 @@ angular.module('OpenSlidesApp.core', [
     }
 ])
 
+
+.provider('OpenSlidesPlugins', [
```
normanjaeckel: Please add a docstring or comment about what this is doing and how to use it.

openslides/core/static/js/core/base.js
```diff
@@ -144,13 +145,22 @@ angular.module('OpenSlidesApp.core', [
...
                     language.selected = false;
                     if (language.code == lang) {
                         language.selected = true;
                         gettextCatalog.setCurrentLanguage(lang);
                         if (lang != 'en') {
                             gettextCatalog.loadRemote("static/i18n/" + lang + ".json");
+                            // load language files from plugins
+                            if (plugins.length > 0) {
+                                angular.forEach(plugins, function (plugin) {
+                                    if ($.inArray(lang, plugin.languages) != -1) {
```
normanjaeckel: Please use `plugins.indexOf(lang) != -1` instead of this jQuery method. (Or use a lodash method ...)

openslides/core/static/js/core/base.js
```diff
@@ -144,13 +145,22 @@ angular.module('OpenSlidesApp.core', [
...
+                var plugins = OpenSlidesPlugins.getAll();
                 angular.forEach(languages, function (language) {
                     language.selected = false;
                     if (language.code == lang) {
                         language.selected = true;
                         gettextCatalog.setCurrentLanguage(lang);
                         if (lang != 'en') {
                             gettextCatalog.loadRemote("static/i18n/" + lang + ".json");
+                            // load language files from plugins
+                            if (plugins.length > 0) {
```
normanjaeckel: I think this check is not necessary. If the length is 0 then the forEach will skip automaticly, right?

emanuelschuetze: Right. Fixed.

openslides/core/static/js/core/base.js

openslides/core/static/js/core/base.js

emanuelschuetze:
@normanjaeckel All done. Ready for last review.

PR #2611, opened 2016-11-09
Rearrange JavaScript files for mediafile app.

normanjaeckel:
This is a proof of concept. @emanuelschuetze and @ostcar and all other: How do you like the new file and module structure?


emanuelschuetze:
@ostcar Do you have any opinion? Otherwise we will decide (and merge) this.

FinnStutzenstein:
@emanuelschuetze asked about my opinion:

* I would split up the create and update controller. So for every app each controller lays in their own file to keep this straight through all apps.
* The separation of the state is fine, this makes in almost every app a huge part of the site.js.
* Splitting special things up from the site.js and projector.js is good to keep an module-like structure. The mainMenu is a good example for this. So each app do not have to import the whole core, but just the modules that the app really needs.
* I do not like the private filter in the list.js. If just the controller is in the file, the private filter should go to the site.js or get a separate file like filters.js. This would make sense, if we always put the filters in a filter-file, to keep them together, but do not merge them into some other code. Also this is something like a module. The mediafileapp provides a filter module with filters for mediafiles.
* renaming base.js to resources.js is ok, if there are only the factories for the models. Filters, directives, special configs (especially for core) should go in filters.js, directives.js, configs.js (exclude the stateconfig)

For the tl;dr people:
Conclusion of @normanjaeckel suggestion and my remarks:
- resource.js with the 'models'
- site.js and projector.js combine all necessary modules
- [module].js with special importent things like mainMenu.js or the inline editing service
- [view].js only with one controller like list.js, create.js, form.js, ...
- states.js with the stateconfig
- pdf.js; csv.js; docx.js; ... for each possibility to export

The following are not necessary to split up; they may be placed in site.js if it's only one not so important filter etc. (or keep the sturcture straight and always put in separate files?):
- config.js (except state config, if exists)
- filter.js (if exists)
- directives.js (if exists)

normanjaeckel:
Thanks for your feedback. I think we should not move filters to a filter.js and directives to a directive.js because this looks like kind of double labeling (the angular component has already a kind of a label). We can either move such components to the logic it belongs to (here the mediafile filter to list.js because it is designed to be used only there) or to a separate file if it might be used somewhere else (site.js or utils.js).

FinnStutzenstein:
Then (especially for directives) i prefer utils.js. Ok, if the filter (e.g. the private filter) is only used for the list controller, ok. But where to put the os-filter which is used in almost every list view?

normanjaeckel:
Commonly used filters and extra stuff can be in an extra module utils.js or listViewUtils.js or something like this.

normanjaeckel:
All done. @emanuelschuetze please review and merge.

Issue #2590, opened 2016-11-04
Move httpProvider to base.js


FinnStutzenstein:
To fix [this TODO](https://github.com/OpenSlides/OpenSlides/blob/f2fa8de7bed64d2cf8d2ae081d2e70506b0b27e9/openslides/motions/static/js/motions/projector.js#L55) the httpProvider config has to be moved from core/site.js to core/base.js.

For testing, issue #2547 has to be fixed first.

PR #2620, opened 2016-11-11
Highlighting motions for multiprojector and move httpProvider (fixes …

FinnStutzenstein:
…#2590)

Some fixes I've overlooked with the Multiprojector. Also the todo in motions/projector.js is solved.

### DIFF
--- a/openslides/motions/static/js/motions/site.js
+++ b/openslides/motions/static/js/motions/site.js
@@ -1066,15 +1066,17 @@ angular.module('OpenSlidesApp.motions.site', [
         $scope.linesForProjector = false;
         // Set 0 for disable highlighting on projector
         var setHighlightOnProjector = function (line) {
-            var elements = _.map(Projector.get(1).elements, function(element) { return element; });
-            elements.forEach(function (element) {
-                if (element.name == 'motions/motion') {
-                    var data = {};
-                    data[element.uuid] = {
-                        highlightAndScroll: line,
-                    };
-                    $http.post('/rest/core/projector/1/update_elements/', data);
-                }
+            _.forEach(Projector.getAll(), function (projector) {
+                var elements = _.map(projector.elements, function(element) { return element; });
+                elements.forEach(function (element) {
+                    if (element.name == 'motions/motion' && element.id == motion.id) {
+                        var data = {};
+                        data[element.uuid] = {
+                            highlightAndScroll: line,
+                        };
+                        $http.post('/rest/core/projector/' + projector.id + '/update_elements/', data);
+                    }
+                });
             });
         };
         $scope.scrollToAndHighlight = function (line) {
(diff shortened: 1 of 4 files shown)


normanjaeckel:
Hm. Am I right that this implementation means that the projector sends http requests to the server to scroll himself to a calculated value? What if I have different projectors. Do they make a scroll battle? It looks somehow really strange ...

FinnStutzenstein:
@normanjaeckel yes, the projector scrolls itself. The problem is that only the projector knows on which location (esp. y-coordinate in px) the specific line is. With this information, one could calculate where to scroll. This is the limitation of the current scroll system. Try to scroll the projector with your mouse; this will not work. So we can't scroll to an anchor etc. but with the y-position the projector itself can scroll.

normanjaeckel:
But what happens if I have two projectors with different resolution. To which value does the projector scroll?

FinnStutzenstein:
First, OS currently does not support displaying one motion on two projectors. But if this gets implemented and one wants to jump to line 30, every single slide checks the location of the line (because the motion_slide controller is loaded two times on both projectors) and scrolls just one projector:

"jump to 30" --> slide 1 (motion 4 on projector 1) checks y_1 --> scroll projector 1 to y_1
                    --> slide 2 (motion 4 on projector 2) checks y_2 --> scroll projector 2 to y_2

normanjaeckel:
Sorry but I think this is broken by design:
1. Projector motion with long text and focus on a line somewhere in the middle of the text.
2. Activate line jumping -> projector scrolls to the position
3. Open another tab with the same projector -> all ok
4. Just open another browser window with the same projector **as real projector** and decrease width of the browser window. -> The projector scrolls again and the correct projectors loose the focus.

I think the approach should be that the server has to calculate the scroll value. It has the projector resolution and it knows the measures of the template and the highlighted line number.

FinnStutzenstein:
One importent thing: what is the line height on the projector? If we can set a fixed line height the calculation would be easy - that was my problem. (also for zooming the projector) For this the html has to be changed.

openslides/motions/static/js/motions/site.js
```diff
@@ -1066,15 +1066,17 @@ angular.module('OpenSlidesApp.motions.site', [
...
-                    var data = {};
-                    data[element.uuid] = {
-                        highlightAndScroll: line,
-                    };
-                    $http.post('/rest/core/projector/1/update_elements/', data);
-                }
+            _.forEach(Projector.getAll(), function (projector) {
+                var elements = _.map(projector.elements, function(element) { return element; });
+                elements.forEach(function (element) {
+                    if (element.name == 'motions/motion') {
```
normanjaeckel: If we have more than one projector element with motions/motion we send multiple requests. I think we need a _.find or something similar that breaks after the first item found. forEach is not so good to use here.

PR #2631, opened 2016-11-13
Overhaul change recommendation statuses

CatoTH:
"Proposed" is removed as a status for change recommendations. When a change recommendation is created, it's in the "non-rejected" status "agreed". 
The "Changed version" shows all changes, including the rejected ones.
The "Resolution" shows all non-rejected changes.
In the change summary, the status of a change recommendation is only shown if it has been rejected.

We still have to decide on the design of the control to switch between the two states. Something like http://www.bootstrap-switch.org/examples.html would fit, but would require adding another CSS/JS-library (a small one, though).


openslides/motions/models.py
```diff
@@ -758,7 +758,7 @@ class MotionChangeRecommendation(RESTModelMixin, models.Model):
     """The motion version to which the change recommendation belongs."""
 
     status = models.PositiveIntegerField(default=0)
```
normanjaeckel: I thought we wanted to use a boolean field (rejected false/true). Status 0 should not means accepted because the change recomendation is not “accepted” by default.

openslides/motions/static/js/motions/base.js
```diff
@@ -255,9 +255,14 @@ angular.module('OpenSlidesApp.motions', [
...
-                    return this.getTextWithChangeRecommendations(versionId, function(status) {
-                        return (status == 1);
+                getTextWithAllChangeRecommendations: function (versionId) {
+                    return this._getTextWithChangeRecommendations(versionId, function(status) {
+                        return true;
+                    });
+                },
+                getTextWithoutRejectedChangeRecommendations: function (versionId) {
+                    return this._getTextWithChangeRecommendations(versionId, function(status) {
+                        return (status != 1);
```
normanjaeckel: If we use a boolean this can be simplified. (See also line 249.)

openslides/motions/static/templates/motions/motion-detail/view-diff.html
```diff
@@ -11,21 +11,15 @@
                 <div class="action-row" ng-if="motion.isAllowed('can_manage')">
                     <div class="btn-group" data-toggle="buttons">
                         <label class="btn btn-sm btn-default" ng-class="{active: (change.status == 0)}"
-                               title="{{ 'Suggested' | translate }}" ng-click="change.status = 0; change.saveStatus();">
+                               title="{{ 'Accepted' | translate }}" ng-click="change.status = 0; change.saveStatus();">
```
normanjaeckel: Here the same: I do not think that the default is accepted. The default has not specific decision and OpenSlides only provides to reject single recommendations explicitly. The rest is part of the motion and will be “accepted” if the motion in total is set to accepted.

CatoTH:
I've renamed the status field into "rejected".
It would be safe to test this feature again.

emanuelschuetze:
Successfully tested. 
@normanjaeckel Please review and merge.

PR #2632, opened 2016-11-13
Allow to explicitly set the type of a change recommendation

CatoTH:


openslides/motions/static/js/motions/site.js
```diff
@@ -285,7 +285,23 @@ angular.module('OpenSlidesApp.motions.site', [
     'gettextCatalog',
     'Editor',
     'Config',
-    function(gettextCatalog, Editor, Config) {
+    'formlyConfig',
+    function(gettextCatalog, Editor, Config, formlyConfig) {
+        formlyConfig.setType({
```
normanjaeckel: Is this something generic? It looks so, so I propose to move it to a reusable factory or other service.

CatoTH: I moved the type definition to the site-wide formly-configuration in core/site.js to the other custom formly types.

emanuelschuetze:

normanjaeckel:
I do not know whether @emanuelschuetze wants to test this again. If not please merge immediately.

emanuelschuetze:
Retested. Ok.

PR #2651, opened 2016-11-19
Editing change recommendations

CatoTH:
This allows editing the change recommendations. The line numbers are fixed, though.

### DIFF
--- a/openslides/motions/static/js/motions/site.js
+++ b/openslides/motions/static/js/motions/site.js
@@ -311,6 +311,20 @@ angular.module('OpenSlidesApp.motions.site', [
                     }
                 };
             },
+            getEditDialog: function(change) {
+                return {
+                    template: 'static/templates/motions/change-recommendation-form.html',
+                    controller: 'ChangeRecommendationUpdateCtrl',
+                    className: 'ngdialog-theme-default wide-form',
+                    closeByEscape: false,
+                    closeByDocument: false,
+                    resolve: {
+                        change: function() {
+                            return change;
+                        }
+                    }
+                };
+            },
             // angular-formly fields for motion form
             getFormFields: function (line_from, line_to) {
                 return [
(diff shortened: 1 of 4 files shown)


openslides/motions/static/js/motions/site.js
```diff
@@ -1231,6 +1245,41 @@ angular.module('OpenSlidesApp.motions.site', [
...
+            // inject the changed change recommendation (copy) object back into DS store
+            MotionChangeRecommendation.inject(change);
+            // save changed change recommendation object on server
+            MotionChangeRecommendation.save(change, { method: 'PATCH' }).then(
+                function(success) {
+                    $scope.closeThisDialog();
+                },
+                function (error) {
+                    // save error: revert all changes by restore
+                    // (refresh) original motion object from server
```
normanjaeckel: Comment is not up to date.

openslides/motions/static/js/motions/site.js
```diff
@@ -1231,6 +1245,41 @@ angular.module('OpenSlidesApp.motions.site', [
...
 ])
 
+.controller('ChangeRecommendationUpdateCtrl', [
+    '$scope',
+    'MotionChangeRecommendation',
+    'ChangeRecommendationForm',
+    'change',
+    function ($scope, MotionChangeRecommendation, ChangeRecommendationForm, change) {
+        $scope.alert = {};
+        $scope.model = change;
```
emanuelschuetze: You have to use `$scope.model = angular.copy(change);` - like in [MotionUpdateCtrl](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/motions/static/js/motions/site.js#L1364). Otherwise you can change and cancel a change recommendation and the changes are displayed... Then you need a reload to be in sync with server data. So please update this line.

openslides/motions/static/js/motions/site.js
emanuelschuetze: ping

emanuelschuetze:
Successfully tested.

Issue #2615, opened 2016-11-10
OS 2.1: double-click required to create a change recommendation


donovaly:
- go with the mouse between the line number of a motion text line and the first character of the line
- click on the appearing plus button

result: nothing happens

It only works if one double-clicks on the button. This might look as a minor problem but it cost me a lot of time to find out why i cannot create a change recommendation by clicking on the button.

emanuelschuetze:
This behaviour is correct:
First you have to click the start line, finally you have to click the end line of your change recommendation.
In your special case if you want to change one line only you have to click two times on the plus button (start/end line is here the same).

Any idea how to improve this?

FinnStutzenstein:
@emanuelschuetze I would prefer this behaviuor:

- the first click selects the clicked line --> you can change this line
- on a second click, the region is expanded (but the second click is not necessary)

I don't know if the region has to be fix to change the text, if not one could change the region while editing this.

donovaly:
I agree with the proposal of @FinnStutzenstein.

CatoTH:
The problem here is that changing the text is done in an overlay. So if the first click opens the overlay already, you won't be able to select the second line anymore.
I agree that this behavior is somewhat confusing an some clarification would be helpful, though...

donovaly:
I described the behavior now in the docs. I leave it up to you developers if you change it or not.

normanjaeckel:
What about open a small tooltip oder hint after the first click like “Move down or click on the same line to create a change recommendation ...”?

emanuelschuetze:
@normanjaeckel Yes, I also think about this solution with @CatoTH.
@CatoTH Please try to implement this.

CatoTH:
Here's a proposal:
![bildschirmfoto 2016-11-19 um 16 51 43](https://cloud.githubusercontent.com/assets/533440/20456552/0c088880-ae79-11e6-93a6-cac1ec37c199.png)

donovaly:
Looks good.

emanuelschuetze:
Done in current master.

PR #2655, opened 2016-11-19
A small tooltip after selecting the first line when creating a change recommendation

CatoTH:
This shows a little tooltip left of the first selected line number when starting creating a change recommendation. Hopefully this prevents some confusion if you don't know you have to select another line for anything to happen. This is an attempt to solve #2615.

### DIFF
--- a/openslides/motions/static/templates/motions/motion-detail/view-original.html
+++ b/openslides/motions/static/templates/motions/motion-detail/view-original.html
@@ -46,3 +46,8 @@
         </div>
     </li>
 </ul>
+
+<div class="tooltip top tt_change_recommendation_create_help" role="tooltip">
+    <div class="tooltip-arrow"></div>
+    <div class="tooltip-inner" translate>Now choose the last line to be changed</div>
+</div>
--- a/openslides/core/static/css/app.css
+++ b/openslides/core/static/css/app.css
@@ -487,6 +487,16 @@ img {
     background-color: #337ab7;
 }
 
+.tt_change_recommendation_create_help {
+    display: none;
+    max-width: 150px;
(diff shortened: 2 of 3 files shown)


emanuelschuetze:
@CatoTH ping

CatoTH:
It's to the right now, overlapping the text... I'm not really sure if this is better.

openslides/motions/static/templates/motions/motion-detail/view-original.html
```diff
@@ -46,3 +46,8 @@
         </div>
     </li>
 </ul>
+
+<div class="tooltip right tt_change_recommendation_create_help" role="tooltip">
```
emanuelschuetze: please use `uib-tooltip` from angular-bootstrap (see https://angular-ui.github.io/bootstrap/#/tooltip or pull request #2680).

The tooltip position to left and right are really not very usefule...Suggestion: Show tooltop to top so that the tooltop does not overlap the following text lines.

CatoTH:
I've moved it to the top. You're right, this seems like the most useful solution.

I don't see uib-tooltip here, though. As I see it, uib-tooltip needs to be defined as an attribute on a element that is parsed by angular. However, the create-buttons here are pseudo-elements / the line numbers inserted via ng-bind-html. From my understanding, we would need to include uib-tooltip in the line numbering algorithm and then run a separate compile step as in http://stackoverflow.com/questions/17417607/angular-ng-bind-html-and-directive-within-it instead of ng-bind-html. Aside from the work of refactoring it, this would negatively affect the performance (compiling and having n tooltips instead of only one), so I'm rather sceptical about this.

PR #2679, opened 2016-11-21
Added endpoint for mass import.

normanjaeckel:


emanuelschuetze:
@normanjaeckel Please add patch to #1533 that we can discuss it later.

normanjaeckel:
Done.

PR #2688, opened 2016-11-25
New assignment and mediafile tables and table generics

FinnStutzenstein:
New things:

- assignment and mediafile tables are in the new design.
- multiselect factory --> osTableFilter factory. It now contains the multiselect filters, boolean filters (is present, ...) and the filter string and filter querystring creation.
- new osTableSort factory: This covers all things handling the sorting.
- css is more generic as well as some parts of the tables.
- new sort option in multiselect: Option, to select all items without a property. Like all motions without a tag.
- globally renaming the `DeleteMode` to `SelectMode` if may some more options will be available in the future.
- Some general html improvements and cleanup.
- moved the `osFilter` from base.js to site.js.

### DIFF
--- a/openslides/core/static/js/core/site.js
+++ b/openslides/core/static/js/core/site.js
@@ -342,61 +342,107 @@ angular.module('OpenSlidesApp.core.site', [
     }
 ])
 
-.factory('Multiselect', [
+/* This factory handles the filtering of the OS-data-tables. It contains
+ * all logic needed for the table header filtering. Things to configure:
+ * - multiselectFilters: A dict associating the filter name to a list (empty per default). E.g.
+ *       { tag: [],
+ *         category: [], }
+ * - booleanFilters: A dict containing a dict for every filter. The value property is a must.
+ *   For displaying properties like displayName, choiceYes and choiceNo could be usefull. E.g.
+ *      { isPresent: {
+ *          value: undefined,
+ *          displayName: gettext('Is present'), } }
+ * - propertyList, propertyFunctionList, propertyDict: See function getObjectQueryString
+ */
+.factory('osTableFilter', [
     function () {
-        var instance = function () {
-            var areFiltersSet = function () {
-                var areFiltersSet = false;
-                _.forEach(self.filters, function (filterList, filter) {
-                    if (filterList.length > 0) {
-                        areFiltersSet = true;
+        var createInstance = function () {
+            var self = {
+                multiselectFilters: {},
+                booleanFilters: {},
+                filterString: '',
+            };
+            self.areFiltersSet = function () {
+                var areFiltersSet = _.find(self.multiselectFilters, function (filterList) {
+                    return filterList.length > 0;
+                });
+                areFiltersSet = areFiltersSet || _.find(self.booleanFilters, function (filterDict) {
+                    return filterDict.value !== undefined;
+                });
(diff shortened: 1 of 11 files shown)


FinnStutzenstein:
@emanuelschuetze failure with negative filter for tags could not be recreated; rest is now fixed.

emanuelschuetze:
@FinnStutzenstein Two remarks
@normanjaeckel Please start your review.

openslides/core/static/js/core/site.js
```diff
@@ -342,61 +342,114 @@ angular.module('OpenSlidesApp.core.site', [
...
+            };
+            self.areFiltersSet = function () {
                 var areFiltersSet = false;
-                _.forEach(self.filters, function (filterList, filter) {
+                _.forEach(self.multiselectFilters, function (filterList, filter) {
                     if (filterList.length > 0) {
                         areFiltersSet = true;
                     }
                 });
+                _.forEach(self.booleanFilters, function (filterDict, filter) {
```
normanjaeckel: Here you can use a `_.find` or something similar that breaks the loop if one truthy value was found.

openslides/core/static/js/core/site.js
```diff
@@ -342,61 +342,114 @@ angular.module('OpenSlidesApp.core.site', [
...
+                    self.booleanFilters[filter].value = undefined;
                 });
+                self.filterString = '';
+            };
+            self.operateMultiselectFilter = function (filter, id, danger) {
+                if (!danger) {
+                    if (_.indexOf(self.multiselectFilters[filter], id) > -1) {
+                        // remove id
+                        self.multiselectFilters[filter] = _.filter(self.multiselectFilters[filter], function (_id) {
```
normanjaeckel: To remove an element of an array you may use use something like `array.splice(index, 1)`. this returns the remove element (which can be ignored) and the array is changed then.

openslides/core/static/js/core/site.js
```diff
@@ -407,26 +460,45 @@ angular.module('OpenSlidesApp.core.site', [
...
             if (filterArray.length === 0) {
                 return array;
             }
+            var itemsWithoutProperty = _.indexOf(filterArray, -1) > -1;
             return Array.prototype.filter.call(array, function (item) {
                 var id = idFunction(item);
-                if (!id) {
-                    return false;
-                } else if (typeof id === 'number') {
+                console.log(item);
```
normanjaeckel: Remove console.log lines.

openslides/motions/static/templates/motions/motion-list.html
```diff
@@ -354,8 +395,8 @@ <h1 translate>Motions</h1>
       </div>
       <!-- additional content column -->
       <style>
-        #motion-table .row .col-xs-4 {
-          width: calc(50% - {{ isDeleteMode ? '120' : '70' }}px);
+        .os-table .row .col-xs-4 {
```
normanjaeckel: Is it possible to move this stylesheets to our css file?

FinnStutzenstein:

emanuelschuetze:
@FinnStutzenstein Last remark: Please show only number of candidates in the badge - not the number of posts, e.g. [0] not [0 / 1]. The tooltip shows the number of posts with something like ... 1 needed.

FinnStutzenstein:
@emanuelschuetze ok, now I understand you previous comment. Badge is done and all rebased.

PR #2691, opened 2016-11-27
Improved forms in ng-dialogs.

emanuelschuetze:
- users: use form field in grid layout and add more option
  to reduce scrolling of long dialog.
- agenda/motions/assignment: show parent agenda item field for create
  dialog only (to prevent corrupt agenda by action 'set my child item as my new
  parent item')
- refactor ng dialog and formly code.


emanuelschuetze:
Ok, I don't want to do a generic solution now. Maybe we can think about later...

PR #2698, opened 2016-11-29
Translate default projector (closes #2531)

### LINKED ISSUE #2531: OS 2.1: missing translation in agenda projection submenu
- In OS 2.1 dev create a subentry of an agenda point
- click on the arrow button besides the projection button to show the submenu

result: there are 2 subentries while there should only be one and the second one is in English instead of in German, see the screenshot:

bugxxx

FinnStutzenstein:
Hope, that all places with 'Default projector' are covered.


FinnStutzenstein:
@emanuelschuetze merge conflict fixed.

emanuelschuetze:
> What happens if I call my projector "Agenda". Is this translated then?

Yes. Like with all other stings which are marked for translation (e.g. group names, config values). I'll merge it now. Make an issue if you like.

PR #2713, opened 2016-12-01
Added settings for client to basic JavaScript code on client startup.…

### LINKED ISSUE #2697: Add hook to get settings values in client
We want to set some custom settings values like special labels or disabling some buttons. These values should be available in the client.

normanjaeckel:
… Closed #2697.

@emanuelschuetze You can test this by adding a new settings value (e. g. `ASSIGNMENT_POSTS_LABEL = 'Mandate'`) to your personal settings.py and a respective value (e. g. 'posts') to utils/settings.py.tpl. Then you have to add the settings key to core/views line 145. See my comment. Then you can access this value in every angular service and controller via dependency injection: Add `OpenSlidesSettings` which is a  [Angular constant](https://docs.angularjs.org/api/ng/type/angular.Module#constant) to dependencies of e. g. AssignmentDetailCtrl and get the value via `OpenSlidesSettings.ASSIGNMENT_POSTS_LABEL`.

### DIFF
--- a/openslides/core/views.py
+++ b/openslides/core/views.py
@@ -141,6 +141,18 @@ class WebclientJavaScriptView(utils_views.View):
                 else:
                     js_files.extend(app_js_files)
 
+        client_settings_keys = [
+            # Add new settings to personal settings.py, utils/settings.py.tpl and to this list. Remove this comment later.
+        ]
+        client_settings = {}
+        for key in client_settings_keys:
+            try:
+                client_settings[key] = getattr(settings, key)
+            except AttributeError:
+                # Settings key does not exist. Do nothing. The client will
+                # treat this as undefined.
+                pass
+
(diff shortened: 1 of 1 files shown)


normanjaeckel:
Next to the code where we add config to rootscope we may also add this new constant for the settings to rootscope so we can easily access this in the template.

openslides/core/views.py
```diff
@@ -141,6 +141,18 @@ def get(self, *args, **kwargs):
                 else:
                     js_files.extend(app_js_files)
 
+        client_settings_keys = [
+            # Add new settings to personal settings.py, utils/settings.py.tpl and to this list. Remove this comment later.
```
normanjaeckel: @emanuelschuetze Only add new settings keys (as strings) to this list. Of course add key and value to settings.py and settings.py.tpl.

PR #1972, opened 2016-02-14
Use tinymce instead of CKEditor.

emanuelschuetze:
The main reason of switching to tinymce:
The data loss problem with MS Word is still unfixed in CKEditor, see https://dev.ckeditor.com/ticket/13174. Also the integration into angular-formly is easier with tinymce. And ui-tinymce looks more active than angular-ckeditor (e.g. Edge support with angular-ckeditor seems to be still open).

What I changed:
- better integration of tinymce in bower and gulp
- Improve support for html tags in reportlab's motion pdf.
- Now paste from word works without problems.
- The editor is now used for customslides (text), motions (text,
  reason) and users (about).

PR #2720, opened 2016-12-02
Going back from tinymce to ckeditor

matakuka:
This is an attempt to revert to CKEditor. The purpose is first evaluation, I'll keep working on the issues.

**Reason**: *copy+ pasting from Microsoft Word to OpenSlides is always difficult. Currently, after some research, I think CKEditor handles it better. The older issues we had with CKEditor (e.g. text disappearing) have been fixed in the latest release.*

What works better:
- better at keeping numbers of sections (tinyMCE messed up my test document numbering)
- better adherence to the original style
- tables

What does not (yet) work fine (in both editors):
- line spacing/ empty paragraphs
- more exotic numbered lists (e.g. latin letters, beginning with C)

TODO:
- [ ]  get bower/gulp to really only pull/concatenate the needed files to remove excess size.
- [x] fix inline editing
- [x] fix paragraph spacing/empty paragraphs
- [x] fix line numbering
- [x] numbered list other than standard `<li>`
- [x] <s>table handling (set visible borders, don't handle each cell as "line")</s> moved to #2784
- [x] beautification of UI, removal/adding of plugins according to needs.
- [ ] refactoring my ugly hacks (e.g. defining a path in index.html)
- [x] make the editor language dependant on openslides setting, not on browser setting

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -178,6 +178,7 @@ OpenSlides uses the following projects or parts of them:
   * `angular-bootstrap <http://angular-ui.github.io/bootstrap>`_, License: MIT
   * `angular-bootstrap-colorpicker <https://github.com/buberdds/angular-bootstrap-colorpicker>`_, License: MIT
   * `angular-chosen-localytics <http://github.com/leocaseiro/angular-chosen>`_, License: MIT
+  * `angular-ckeditor <https://github.com/lemonde/angular-ckeditor/>`_, License: MIT
   * `angular-csv-import <https://github.com/bahaaldine/angular-csv-import>`_, License: MIT
   * `angular-formly <http://formly-js.github.io/angular-formly/>`_, License: MIT
   * `angular-formly-templates-bootstrap <https://github.com/formly-js/angular-formly-templates-bootstrap>`_, License: MIT
@@ -189,13 +190,13 @@ OpenSlides uses the following projects or parts of them:
   * `angular-sanitize <http://angularjs.org>`_, License: MIT
   * `angular-scroll-glue <https://github.com/Luegg/angularjs-scroll-glue>`_, License: MIT
   * `angular-ui-router <http://angular-ui.github.io/ui-router/>`_, License: MIT
-  * `angular-ui-tinymce <http://angular-ui.github.com>`_, License: MIT
   * `angular-ui-tree <https://github.com/angular-ui-tree/angular-ui-tree>`_, License: MIT
   * `angular-xeditable <https://github.com/vitalets/angular-xeditable>`_, License: MIT
   * `api-check <https://github.com/kentcdodds/api-check>`_, License: MIT
   * `bootstrap <http://getbootstrap.com>`_, License: MIT
   * `bootstrap-ui-datetime-picker <https://github.com/Gillardo/bootstrap-ui-datetime-picker>`_, License: MIT
   * `chosen <http://harvesthq.github.io/chosen/>`_, License: MIT
+  * `ckeditor <http://ckeditor.com>`_,  License: For licensing, see LICENSE.md or http://ckeditor.com/license.
   * `font-awesome-bower <https://github.com/tdg5/font-awesome-bower>`_, License: MIT
   * `jquery <https://jquery.com>`_, License: MIT
   * `jquery.cookie <https://plugins.jquery.com/cookie>`_, License: MIT
--- a/README.rst
+++ b/README.rst
@@ -196,7 +196,7 @@ OpenSlides uses the following projects or parts of them:
   * `bootstrap <http://getbootstrap.com>`_, License: MIT
   * `bootstrap-ui-datetime-picker <https://github.com/Gillardo/bootstrap-ui-datetime-picker>`_, License: MIT
   * `chosen <http://harvesthq.github.io/chosen/>`_, License: MIT
-  * `ckeditor <http://ckeditor.com>`_,  License: For licensing, see LICENSE.md or http://ckeditor.com/license.
+  * `ckeditor <http://ckeditor.com>`_,  License: GPL 2+, LGPL 2.1+ or MPL 1.1.
   * `font-awesome-bower <https://github.com/tdg5/font-awesome-bower>`_, License: MIT
   * `jquery <https://jquery.com>`_, License: MIT
   * `jquery.cookie <https://plugins.jquery.com/cookie>`_, License: MIT
@@ -211,8 +211,6 @@ OpenSlides uses the following projects or parts of them:
   * `open-sans-fontface <https://github.com/FontFaceKit/open-sans>`_, License: Apache License version 2.0
   * `pdfjs-dist <http://mozilla.github.io/pdf.js/>`_, License: Apache-2.0
   * `roboto-condensed <https://github.com/davidcunningham/roboto-condensed>`_, License: Apache 2.0
-  * `tinymce <http://www.tinymce.com>`_, License: LGPL-2.1
-  * `tinymce-i18n <https://github.com/OpenSlides/tinymce-i18n>`_, License: LGPL-2.1
 
 
 License and authors
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -108,37 +108,12 @@ gulp.task('angular-chosen-img', function () {
         .pipe(gulp.dest(path.join(output_directory, 'css')));
 });
 
-// Catches all skins files for TinyMCE editor.
-gulp.task('tinymce-skins', function () {
-    return gulp.src(path.join('bower_components', 'tinymce', 'skins', '**'))
-        .pipe(gulp.dest(path.join(output_directory, 'tinymce', 'skins')));
+// Extra task only for CKEditor
+gulp.task('ckeditor', function () {
+    return gulp.src(path.join('bower_components', 'ckeditor', '**'))
+        .pipe(gulp.dest(path.join(output_directory, 'ckeditor')));
 });
 
-// Catches all required i18n files for TinyMCE editor.
-gulp.task('tinymce-i18n', function () {
-    return gulp.src([
-            'bower_components/tinymce-i18n/langs/en_GB.js',
-            'bower_components/tinymce-i18n/langs/cs.js',
-            'bower_components/tinymce-i18n/langs/de.js',
-            'bower_components/tinymce-i18n/langs/es.js',
-            'bower_components/tinymce-i18n/langs/fr_FR.js',
-            'bower_components/tinymce-i18n/langs/pt_PT.js',
-            ])
-        .pipe(rename(function (path) {
-            if (path.basename === 'en_GB') {
-                path.basename = 'en';
-            } else if (path.basename === 'fr_FR') {
-                path.basename = 'fr';
-            } else if (path.basename === 'pt_PT') {
-                path.basename = 'pt';
-            }
-        }))
-        .pipe(gulpif(argv.production, uglify()))
-        .pipe(gulp.dest(path.join(output_directory, 'tinymce', 'i18n')));
-});
-
-// Combines all TinyMCE related tasks.
-gulp.task('tinymce', ['tinymce-skins', 'tinymce-i18n'], function () {});
 
 // Compiles translation files (*.po) to *.json and saves them in the directory
 // openslides/static/i18n/.
@@ -151,7 +126,7 @@ gulp.task('translations', function () {
 });
 
 // Gulp default task. Runs all other tasks before.
-gulp.task('default', ['js', 'js-libs', 'templates', 'css-libs', 'fonts-libs', 'tinymce', 'angular-chosen-img', 'translations'], function () {});
+gulp.task('default', ['js', 'js-libs', 'templates', 'css-libs', 'fonts-libs',  'ckeditor', 'angular-chosen-img', 'translations'], function () {});
 
 
 /**
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -108,12 +108,59 @@ gulp.task('angular-chosen-img', function () {
         .pipe(gulp.dest(path.join(output_directory, 'css')));
 });
 
-// Extra task only for CKEditor
-gulp.task('ckeditor', function () {
-    return gulp.src(path.join('bower_components', 'ckeditor', '**'))
+// CKEditor defaults
+gulp.task('ckeditor-defaults', function () {
+    return gulp.src([
+            path.join('bower_components', 'ckeditor', 'styles.js'),
+            path.join('bower_components', 'ckeditor', 'contents.css'),
+        ])
         .pipe(gulp.dest(path.join(output_directory, 'ckeditor')));
 });
 
+// CKEditor skins
+gulp.task('ckeditor-skins', function () {
+    return gulp.src([
+            path.join('bower_components', 'ckeditor', 'skins', 'moono-lisa', '**', '*'),
+        ],{ base: path.join('bower_components', 'ckeditor', 'skins') })
+        .pipe(gulp.dest(path.join(output_directory, 'ckeditor', 'skins')));
+});
+
+// CKEditor plugins
+gulp.task('ckeditor-plugins', function () {
+    return gulp.src([
+            path.join('bower_components', 'ckeditor', 'plugins', 'clipboard', '**', '*'),
+            path.join('bower_components', 'ckeditor', 'plugins', 'colorbutton', '**', '*'),
+            path.join('bower_components', 'ckeditor', 'plugins', 'dialog', '**', '*'),
+            path.join('bower_components', 'ckeditor', 'plugins', 'find', '**', '*'),
+            path.join('bower_components', 'ckeditor', 'plugins', 'image', '**', '*'),
+            path.join('bower_components', 'ckeditor', 'plugins', 'justify', '**', '*'),
+            path.join('bower_components', 'ckeditor', 'plugins', 'liststyle', '**', '*'),
+            path.join('bower_components', 'ckeditor', 'plugins', 'magicline', '**', '*'),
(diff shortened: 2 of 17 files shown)


matakuka:
@emanuelschuetze : I found a **security issue** currently present in ckeditor. Please do not test with confidential data. The plugin "scayt" (spellcheck as you type) sends data to a third party unless disabled. It seems to not be enabled by default, but as soon as spell check is turned on, it starts sending data. I failed so far in disabling this "service", as my configuration done in core/base.js seems to be partially ignored.

matakuka:
@emanuelschuetze. It is not a problem anymore. Seems more of a cache issue on my side. The browser disables the offending *scayt* now. Still an opportunity to remind oneself to always check for such "services".

emanuelschuetze:
@matakuka Thanks for your pull request. I rebased your 3 commits and resolved the conflict in README. I added also a commit to define ckeditor toolbar. Normal editing works. Some tasks are still open before we can merge this. But for first tests with MS word it is enough...

To finish this pullrequest you can also check the old pr #1972 where I replaced ckeditor by tinymce.

matakuka:
note to self (and other potential testers): Check whether support for Edge is acceptable, as this was an issue in #1972.

matakuka:
@emanuelschuetze I think I got the "remove empty paragraphs on import" working. My code seems to me like unnecessarily complicated, but I think I got a working first draft at least (we could wrap it into a config if someone doesn't want this behaviour):
- On import the html is parsed through the ckeditor filter first. If there are `<p>` tags remaining that have no content but `<br>` or other `<p>`, these are discarded. 

I am not 100% sure this is done right, but I think it is testable. Also, subscript and superscript are included, which I did not test, but I assume they work out-of-the-box.

matakuka:
I currently have a problem with line numbering, as ckeditor's output is a bit different from tinymce's output:

**tags create extra "lines":**
`<p>` create one extra line, I think it is due to the combination of a generated `<p><span>` combination.
`<li>` creates three to four "lines". I see `<li><p><span>`.

I think the effect is created by each tag passing through `_insertLineNumbersToBlockNode` and somehow getting a line number there. 

The desired effect is to not get the own number there, but just get the child node's line numbers (`_moveLeadingLineBreaksToOuterNode`?). 
Setting `this._prependLineNumberToFirstText = false` seemed not to have an effect.

emanuelschuetze:
> I currently have a problem with line numbering, as ckeditor's output is a bit different from tinymce's output:

@CatoTH Please look at this issue in this branch as discussed. Thanks.

CatoTH:
@emanuelschuetze Ok, seems like a rather simple bug to fix. I'll try to fix it today evening

CatoTH:
@matakuka @emanuelschuetze This should solve the line numbering problem: https://github.com/OpenSlides/OpenSlides/pull/2768/commits/d1e9d29c6224858d4c8f2b15e6358f992acab0b8

matakuka:
@CatoTH thank you a lot.

@emanuelschuetze: inline editing is mostly done, and testable. There are some issues remaining with inline editing:
- the angularjs - popup "The content has changed- save" behaves very sluggish, although the watched variables (active and changed) fire as they should. I'm not sure on the cause of this.
- the css-classes "`motion-text motion-text-original line-numbers-{{ lineNumberMode }}`" are not included in ckeditor. Thus, the line numbers are there in the source text, but not shown.
- Currently, the ckeditor language is set on page load, ignoring later changes.

matakuka:
second point (inline numbers not showing) is solved; I did not allow the content "data-line-number" in ckeditor, which the css uses to display numbers.

emanuelschuetze:
> second point (inline numbers not showing) is solved;

I can't confirm that. If I activate the inline editing the line numbers are not displayed any more.
And there is a strange behavior if I press "backspace" key at the beginning of the first sentence of a paragraph (the line moved to vertical at the left border)... So the inline editing mode needs more work ...

openslides/motions/static/js/motions/site.js
```diff
@@ -1305,10 +1305,14 @@ angular.module('OpenSlidesApp.motions.site', [
             return Boolean(isAllowed);
         };
 
+
         // Inline editing functions
         $scope.inlineEditing = MotionInlineEditing;
         $scope.inlineEditing.init($scope, motion);
 
+        // TODO: properties of inlineEditing are currently not watched continuosly,
```
matakuka: The inlineEditing.editor.onChange event modifies inlineEditing.changed, but this is only registered by the 'ng-if' triggers in the template when the object is rerendered (by the tab losing focus). 

I think it is something in angularjs. The result is the "save button" not appearing on time.

matakuka:
Another thing: Currently I included colors (text-color, background-color) in ckeditor. Do we support this?

 (Reason for asking: Inline editing after some text is highlighted highlighted text currently confuses ckeditor, the highlight becomes part of the "source" html in the process).

emanuelschuetze:
@matakuka First tests looks good!

openslides/motions/static/js/motions/site.js
matakuka: a dirty approach to solve this would be to always have a "save" button and giving a message if nothing has changed at all. But I'd prefer these ng-ifs to work

emanuelschuetze:
I pushed a commit which updates the toolbar.

matakuka:
sorry, my attention slipped for a minute :). I hope I merged it correctly.

matakuka:
The language selection is a bit difficult. Currently we select the editor language the moment the ckeditor options in the factory (in core/base.js) are parsed; which is mostly on the first template loading. For inline editing I could change it to the moment the editor starts.

But I think this has lower priority, as most people won't change the language mid-session.

emanuelschuetze:
Tested successfully. Looks good! I'll do the last open task for gulp/bower/index.html.
@matakuka Please rebase
@normanjaeckel Please make a first review. Should go into 2.1b1 today.

matakuka:
@normanjaeckel rebase is done; 

I moved the remaining "table" issues to a new issue #2784 .

README.rst
```diff
@@ -189,13 +190,13 @@ OpenSlides uses the following projects or parts of them:
...
   * `angular-scroll-glue <https://github.com/Luegg/angularjs-scroll-glue>`_, License: MIT
   * `angular-ui-router <http://angular-ui.github.io/ui-router/>`_, License: MIT
-  * `angular-ui-tinymce <http://angular-ui.github.com>`_, License: MIT
   * `angular-ui-tree <https://github.com/angular-ui-tree/angular-ui-tree>`_, License: MIT
   * `angular-xeditable <https://github.com/vitalets/angular-xeditable>`_, License: MIT
   * `api-check <https://github.com/kentcdodds/api-check>`_, License: MIT
   * `bootstrap <http://getbootstrap.com>`_, License: MIT
   * `bootstrap-ui-datetime-picker <https://github.com/Gillardo/bootstrap-ui-datetime-picker>`_, License: MIT
   * `chosen <http://harvesthq.github.io/chosen/>`_, License: MIT
+  * `ckeditor <http://ckeditor.com>`_,  License: For licensing, see LICENSE.md or http://ckeditor.com/license.
```
normanjaeckel: Here we should mention LGPL and MPL in correct version number.

normanjaeckel: Below the lines for tinymce have to be removed.

gulpfile.js
```diff
@@ -108,37 +108,12 @@ gulp.task('angular-chosen-img', function () {
         .pipe(gulp.dest(path.join(output_directory, 'css')));
 });
 
-// Catches all skins files for TinyMCE editor.
-gulp.task('tinymce-skins', function () {
-    return gulp.src(path.join('bower_components', 'tinymce', 'skins', '**'))
-        .pipe(gulp.dest(path.join(output_directory, 'tinymce', 'skins')));
+// Extra task only for CKEditor
```
normanjaeckel: But what does this task do? Why do we have to copy these files to the output directory *and* catch all these files in the bower.json main section?

openslides/core/static/js/core/base.js
```diff
@@ -530,33 +530,96 @@ angular.module('OpenSlidesApp.core', [
     }
 ])
 
-// Options for TinyMCE editor used in various create and edit views.
-// Required in core/base.js because MotionComment factory which used this
-// factory has to placed in motions/base.js.
+// Options for CKEditor used in various create and edit views.
```
normanjaeckel: Please readd the comment about the place of this factory.

openslides/core/static/js/core/base.js
```diff
@@ -530,33 +530,96 @@ angular.module('OpenSlidesApp.core', [
...
 .factory('Editor', [
     'gettextCatalog',
-    function (gettextCatalog) {
+    function(gettextCatalog) {
```
normanjaeckel: Please readd the whitespace. ;-)

openslides/core/static/js/core/base.js
```diff
@@ -530,33 +530,96 @@ angular.module('OpenSlidesApp.core', [
...
-                    plugins: [
-                      'lists link autolink charmap preview searchreplace code fullscreen',
-                      'paste textcolor colorpicker image imagetools wordcount'
-                    ],
-                    menubar: '',
-                    toolbar: 'undo redo searchreplace | styleselect | bold italic underline strikethrough ' +
-                        'forecolor backcolor removeformat | bullist numlist | outdent indent | ' +
-                        'link image charmap table | code preview fullscreen'
+                    on: {
+                        instanceReady: function() {
```
normanjaeckel: Is this code copied from somewhere? Please extend comment or docstring about what this is doing. (We have two combined loops (children and children.children). This can be explained a little bit more.)

openslides/core/static/js/core/base.js
```diff
@@ -530,33 +530,96 @@ angular.module('OpenSlidesApp.core', [
...
+                            // TODO: check all kind of clipboard html content if "isEmpty" is a reliable property
+                            this.on('paste', function(evt) {
+                                if (evt.data.type == 'html') {
+                                    var fragment = CKEDITOR.htmlParser.fragment.fromHtml(evt.data.dataValue);
+                                    var writer = new CKEDITOR.htmlParser.basicWriter();
+                                    // html content will now be in a dom-like structure inside 'fragment'.
+                                    this.filter.applyTo(fragment);
+                                    if (fragment.children) {
+                                        var new_content_children = [];
+                                        for (var i = 0; i < fragment.children.length; i++) {
```
normanjaeckel: Here we can use a lodash function.

openslides/core/static/js/core/base.js
```diff
@@ -530,33 +530,96 @@ angular.module('OpenSlidesApp.core', [
...
+                                            var empty = true;
+                                            if (fragment.children[i].children){
+                                                for (var j = 0; j < fragment.children[i].children.length; j++) {
```

openslides/core/static/templates/index.html
```diff
@@ -10,10 +10,19 @@
 <link rel="stylesheet" href="static/css/openslides-libs.css">
 <link rel="stylesheet" href="static/css/app.css">
 <link rel="icon" href="/static/img/favicon.png">
+<!-- TODO: there is probably a better place for it:-->
```
normanjaeckel: Yes. What about a `run` call in angular?

openslides/core/static/templates/index.html
```diff
@@ -10,10 +10,19 @@
...
+<script>
+  window.CKEDITOR_BASEPATH = '/static/ckeditor/';
+</script>
+<!-- -->
 <script src="static/js/openslides-libs.js"></script>
 <script src="static/js/openslides.js"></script>
 <script src="static/js/openslides-templates.js"></script>
-
+<!-- TODO: move inside openslides-libs (?):-->
```
normanjaeckel: Here the same.

emanuelschuetze:
I'll try to resolve the remarks now...

README.rst
emanuelschuetze: I wrote `License: LGPL and MPL` (why not GPL?). Which correct version number do you mean?.
Tinymce lines are removed.

openslides/core/static/js/core/base.js
```diff
@@ -530,33 +530,96 @@ angular.module('OpenSlidesApp.core', [
...
+                            // additionally strips out all empty <p> paragraphs
+                            // TODO: check all kind of clipboard html content if "isEmpty" is a reliable property
+                            this.on('paste', function(evt) {
+                                if (evt.data.type == 'html') {
+                                    var fragment = CKEDITOR.htmlParser.fragment.fromHtml(evt.data.dataValue);
+                                    var writer = new CKEDITOR.htmlParser.basicWriter();
+                                    // html content will now be in a dom-like structure inside 'fragment'.
+                                    this.filter.applyTo(fragment);
+                                    if (fragment.children) {
+                                        var new_content_children = [];
```
matakuka: if this fragment is DOM-like, it may contain nested properties (being html nodes). Traverse the children and check if it is a child only containing empty `<br>` or `<p>`.
`new_content_children` will finally contain all nodes that are not empty.

openslides/core/static/js/core/base.js
```diff
@@ -530,33 +530,96 @@ angular.module('OpenSlidesApp.core', [
...
+                                                if (fragment.children[i].name != 'p' && fragment.children[i].name != 'br' &&
+                                                    fragment.children[i].isEmpty !== true){
+                                                    new_content_children.push(fragment.children[i]);
+                                                }
+                                            }
+                                        }
+                                        fragment.children = new_content_children;
+                                    }
+                                    fragment.writeHtml(writer);
+                                    evt.data.dataValue = writer.getHtml();
```
matakuka: return the re-created fragment without the empty `<p>` and `<br>` into the editor import processing (same as at the begin of the function: by ckeditor)

gulpfile.js
emanuelschuetze: Done.

README.rst
emanuelschuetze: Done.

openslides/core/static/js/core/base.js

openslides/core/static/js/core/base.js

openslides/core/static/js/core/base.js

openslides/core/static/templates/index.html
emanuelschuetze: Done. Moved to .config() in base.js

openslides/core/static/templates/index.html
emanuelschuetze: Done: Issue #2787 created to move it later into openslides-libs.js

gulpfile.js
```diff
@@ -108,37 +108,59 @@ gulp.task('angular-chosen-img', function () {
...
 });
 
-// Catches all skins files for TinyMCE editor.
-gulp.task('tinymce-skins', function () {
-    return gulp.src(path.join('bower_components', 'tinymce', 'skins', '**'))
-        .pipe(gulp.dest(path.join(output_directory, 'tinymce', 'skins')));
+// CKEditor defaults
+gulp.task('ckeditor-defaults', function () {
+    return gulp.src([
+            'bower_components/ckeditor/styles.js',
```
normanjaeckel: Please use path.join('bower_components', 'ckeditor', ...) for Windows support.

openslides/core/static/js/core/base.js
```diff
@@ -43,6 +43,12 @@ angular.module('OpenSlidesApp.core', [
     }
 ])
 
+.config(
```
normanjaeckel: Please add docstring or comment. Maybe move this down just above the Editor factory.

openslides/core/static/js/core/base.js
```diff
@@ -530,33 +530,96 @@ angular.module('OpenSlidesApp.core', [
...
-                    menubar: '',
-                    toolbar: 'undo redo searchreplace | styleselect | bold italic underline strikethrough ' +
-                        'forecolor backcolor removeformat | bullist numlist | outdent indent | ' +
-                        'link image charmap table | code preview fullscreen'
+                    on: {
+                        instanceReady: function() {
+                            // add a listener to ckeditor that parses the clipboard content and, after the regular filter,
+                            // additionally strips out all empty <p> paragraphs
+                            // TODO: check all kind of clipboard html content if "isEmpty" is a reliable property
+                            this.on('paste', function(evt) {
```
emanuelschuetze: Replaced this comment.

openslides/core/static/templates/index.html
```diff
@@ -10,10 +10,15 @@
...
+<!-- TODO (#2787): move into openslides-libs-->
+<script>
+  window.CKEDITOR_BASEPATH = '/static/ckeditor/';
+</script>
+<!-- -->
 <script src="static/js/openslides-libs.js"></script>
 <script src="static/js/openslides.js"></script>
 <script src="static/js/openslides-templates.js"></script>
-
+<!-- -->
```
normanjaeckel: Remove this line?

gulpfile.js
emanuelschuetze: Done

openslides/core/static/js/core/base.js

openslides/core/static/templates/index.html
emanuelschuetze: Done

openslides/core/static/js/core/base.js
```diff
@@ -530,33 +530,111 @@ angular.module('OpenSlidesApp.core', [
...
+                                                if (child.name != 'p' && child.name != 'br' &&
+                                                    child.isEmpty !== true){
+                                                    new_content_children.push(child);
+                                                }
+                                            }
+                                        });
+                                        fragment.children = new_content_children;
+                                    }
+                                    fragment.writeHtml(writer);
+                                    evt.data.dataValue = writer.getHtml();
```
normanjaeckel: Here is the last comment missing:
`Return the re-created fragment without the empty <p> and <br> into the editor import processing (same as at the begin of the function: by ckeditor)`

PR #2734, opened 2016-12-04
Hot fix for cache update while user import (Fixes #2727).

### LINKED ISSUE #2727: Cache not correctly updated on user import.
When importing some users the group memberships are not shown in the list view. Looking at the networktraffic, the `groups_id` is correctly send to the server and the server has the `groups_id` in the response, so the server has saved the group memberships.
Looking at the rest API, no groups are set. If you restart the server (or wait some more time) the cache gets updated and now the user have groups set in the rest API and in the list view.

@normanjaeckel @ostcar Please have a look at this, i think this is a cache bug.

emanuelschuetze:


openslides/users/serializers.py
```diff
@@ -3,6 +3,8 @@
 from django.utils.translation import ugettext as _
 from django.utils.translation import ugettext_lazy
 
+from openslides.utils.autoupdate import inform_changed_data
```
normanjaeckel: Please use dot notation: `from ..utils.autoupdate import inform_changed_data`.

openslides/users/serializers.py
```diff
@@ -79,7 +81,9 @@ def create(self, validated_data):
             validated_data['default_password'] = User.objects.generate_password()
         validated_data['password'] = make_password(validated_data['default_password'], '', 'md5')
         # Perform creation in the database and return new user.
-        return super().create(validated_data)
+        user = super().create(validated_data)
+        inform_changed_data(user)
```
normanjaeckel: Please add Comment here that this autoupdate call is redundant. Please add issue that caching for new objects with many 2 many fields has to be refactored (or just fixed). Then you can add the new issue number here.

openslides/users/serializers.py
emanuelschuetze: Done

openslides/users/serializers.py
emanuelschuetze: Done. See also #2736.

emanuelschuetze:
All done. Rebased.

PR #2737, opened 2016-12-05
Use session cookies and store filter settings in sessionStorage

FinnStutzenstein:
Tested with Chrome and FF. Please check [this](https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-CSRF_COOKIE_AGE) whether the described effects with the bookmark are such a negative thing. I can not confirm this. (Info: if the csrf cookie is wrong, you will just get a 403 - forbidden; clearing the cookie helps).


openslides/core/static/js/core/site.js
```diff
@@ -343,6 +343,13 @@ angular.module('OpenSlidesApp.core.site', [
     }
 ])
 
+.config([
+    '$sessionStorageProvider',
+    function ($sessionStorageProvider) {
+        $sessionStorageProvider.setKeyPrefix('OS-');
```
normanjaeckel: Can we use a more explicit prefix here like `OpenSlides-`?

FinnStutzenstein: Ok, should the `csrftoken` also be renamed to `OpenSlidesCsrfToken` to keep every cookie and storage with the same prefix?

normanjaeckel: Ok, why not.

emanuelschuetze:
@FinnStutzenstein Do you still want to rename the csrftoken cookie?

FinnStutzenstein:
@emanuelschuetze yes, I'll do this tomorrow morning, so this pr can be merged tomorrow.

FinnStutzenstein:
@normanjaeckel Ok.

PR #2739, opened 2016-12-05
Attempt to mitigate the list / line number problem

CatoTH:
See #2561


emanuelschuetze:
Tested: Works good. But please add more vertical space before ul/ol list in pdf output with outside numbering (as in pdf with inline numbering).

Example:
```
[..] text
* a
* b
```
should be
```
[..] text

* a
* b
```
Ok?

CatoTH:
@emanuelschuetze ok, done.

PR #2740, opened 2016-12-06
changing csv import regex for motions

FinnStutzenstein:
Should this be changed for all imports?

Difference: [\s\S] matches everything (includes line breaks; the dot does not match line breaks) and the `m` flag for multiline.

### DIFF
--- a/openslides/motions/static/js/motions/site.js
+++ b/openslides/motions/static/js/motions/site.js
@@ -1580,7 +1580,7 @@ angular.module('OpenSlidesApp.motions.site', [
         // detect if csv file is loaded
         $scope.$watch('csv.result', function () {
             $scope.motions = [];
-            var quotionRe = /^"(.*)"$/;
+            var quotionRe = /^"([\s\S]*)"$/m;
             angular.forEach($scope.csv.result, function (motion) {
                 if (motion.identifier) {
                     motion.identifier = motion.identifier.replace(quotionRe, '$1');


FinnStutzenstein:
@emanuelschuetze and others: Could you explain the different versions of angular-csv-import?
this

This is after deleting the `openslides/static` and `bower_components` folder and running `npm install`.

In our bower.json is version 0.0.36 explicit given. The change from `split('\n')` to `split(new RegExp('\n(?![^"]*"(?:(?:[^"]*"){2})*[^"]*$)'))` (see screenshot) is done with [this commit](https://github.com/bahaaldine/angular-csv-import/commit/aa81a880cef19a9899b2c30a1cffe439380688eb) tagged with 0.0.37; so why does my code do have this change?

normanjaeckel:
I used angular-csv-import in another project but it is a little bit buggy. May be we should change to something like [PapaParse](https://github.com/mholt/PapaParse).

emanuelschuetze:
> @emanuelschuetze and others: Could you explain the different versions of angular-csv-import?

Discussed and debuged together with @FinnStutzenstein : angular-csv-import 0.0.36 contains the right regexp - with wrong version number in file header.

FinnStutzenstein:
So, the module does work if there is an even number of quotes. This has to be verified (maybe an example), because this was a quick result of a little research. If this is confirmed, we have to discuss about a solution. I have something in mind...

emanuelschuetze:
> So, the module does work if there is an even number of quotes. This has to be verified (maybe an example)

Yes, verified. The following example shows an _empty_ csv preview because the forbidden " in `test2` line:
```
identifier;title;text;reason;submitter;category
;test1;"foo bar";;Max;
;test2;"foo "bar";;Max;
```

FinnStutzenstein:
Ok, in reason of this problem together with Issue #2741 we need to think about a plan for handling csv files. I'll do some research, and i we can talk about this on friday.

normanjaeckel:
I once again suggest to use PapaParse instead of angular-csv-import. The angular implementation is easy to build.

emanuelschuetze:
@FinnStutzenstein Please test PapaParse first before your work more on angular-csv-import.

emanuelschuetze:
Closed this because we will use PapaParse (see #2783)

PR #2742, opened 2016-12-06
Mark missing translation strings in table headers.

emanuelschuetze:


openslides/assignments/static/js/assignments/site.js
```diff
@@ -356,13 +356,13 @@ angular.module('OpenSlidesApp.assignments.site', [
...
+             display_name: gettext('Item')},
             {name: 'title',
-             display_name: 'Title'},
-            {name: 'open_posts',
-             display_name: 'Open posts'},
+             display_name: gettext('Title')},
             {name: 'phase',
-             display_name: 'Phase'},
+             display_name: gettext('Phase')},
+            {name: 'assignment.assignment_related_users.length',
```
emanuelschuetze: @FinnStutzenstein I changed the sort field `open_posts` to number of candidates which is more useful. But with this name (`assignment.assignment_related_users.length`) the sorting by number is not correct. Do you know whats wrong?

donovaly:
Please note that in the agenda table there are also other strings to be translated:

- "Set duration..."
- "Set comment..."

In the election table there is the string

- "needed" (in the popup when pointing with the mouse over the number of positions to be elected)

emanuelschuetze:
> "Set duration..."
> "Set comment..."
> "needed" 

Stings added.

openslides/assignments/static/js/assignments/site.js
FinnStutzenstein: @emanuelschuetze Please make an issue.

emanuelschuetze: see #2750

emanuelschuetze:
@normanjaeckel Ready for review and merge.

openslides/users/static/js/users/pdf.js
```diff
@@ -213,11 +213,11 @@ angular.module('OpenSlidesApp.users.pdf', ['OpenSlidesApp.core.pdf'])
             var createWelcomeText = function() {
                 return [
                     {
-                        text:   getConfigValue('users_pdf_welcometitle'),
+                        text:   gettextCatalog.getString(getConfigValue('users_pdf_welcometitle')),
                         style: 'userDataHeading'
                     },
                     {
-                        text:   getConfigValue('users_pdf_welcometext'),
+                        text:   gettextCatalog.getString(getConfigValue('users_pdf_welcometext')),
```
normanjaeckel: Here the same.

openslides/motions/static/js/motions/site.js
```diff
@@ -804,16 +806,15 @@ angular.module('OpenSlidesApp.motions.site', [
...
     'PdfMakeDocumentProvider',
-    'gettextCatalog',
     'HTMLValidizer',
     'Projector',
     'ProjectionDefault',
     'MotionCsvExport',
     'osTableFilter',
     'osTableSort',
-    function($scope, $state, $http, ngDialog, MotionForm, Motion, Category, Tag, Workflow, User, Agenda, MotionBlock,
+    function($scope, $state, $http, gettext, gettextCatalog, ngDialog, MotionForm, Motion, Category, Tag, Workflow, User, Agenda, MotionBlock,
```
emanuelschuetze: Done

openslides/users/static/js/users/pdf.js
```diff
@@ -213,11 +213,11 @@ angular.module('OpenSlidesApp.users.pdf', ['OpenSlidesApp.core.pdf'])
             var createWelcomeText = function() {
                 return [
                     {
-                        text:   getConfigValue('users_pdf_welcometitle'),
+                        text:   gettextCatalog.getString(getConfigValue('users_pdf_welcometitle')),
```
emanuelschuetze: Do you want to remove the getConfigValue function? Otherwise I would change this line to:
`Config.translate(getConfigValue('users_pdf_welcometitle'))`

normanjaeckel: I do not think we need a generic methofd `getConfigValue`.

emanuelschuetze: if you want to rename this function in pdf please suggest a new name. otherwise please approve

normanjaeckel: Hm. This looks messy. It translates "-" in some circumstances. Better would be to do the translation in the getConfigValue method which can also be renamed to getConfigOrPlaceholer.

Or you just use `Config.translate(Config.get(...).value)` || '-'`

emanuelschuetze: Done.

openslides/users/static/js/users/pdf.js
emanuelschuetze: Done

normanjaeckel:
Ok.

PR #2746, opened 2016-12-07
Use Roboto font in all templates.

emanuelschuetze:
- Only install needed font files for Roboto and Roboto Condensed
  which strongly reduced the directory size of 'static/font'.
- Make it easier to adjust custom font by using global css font definition
  in app.css and projector.css.
- Remove Open Sans font.

@normanjaeckel @FinnStutzenstein I want to use the Roboto font in all templates and reduce the handling by using one font source only. Do you like this new look? I am not 100% sure. So I need your opinion. Please test this (run `npm install` first).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -35,8 +35,7 @@
     "ngstorage": "~0.3.11",
     "ngBootbox": "~0.1.3",
     "pdfmake-dist": "~0.1.27",
-    "open-sans-fontface": "https://github.com/OpenSlides/open-sans.git#1.4.2.post1",
-    "roboto-condensed": "~0.3.0",
+    "roboto-fontface": "~0.6.0",
     "tinymce": "~4.4.3",
     "tinymce-i18n": "OpenSlides/tinymce-i18n#a186ad61e0aa30fdf657e88f405f966d790f0805"
   },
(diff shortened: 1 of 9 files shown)


FinnStutzenstein:
I first thought that this font is a little bit blurred, but after some minutes I can read it well. This font is a little bit more playful and not as straight als the current font. Also the letters are a little bit higher, for instance the labels are higher, too, which I thing is good.

I do like the font, but I'm not a designer, so you should get some other opinions too.

normanjaeckel:
Hm. I can hardly see any difference. The new font is a little bit too narrow but if you like it ...

emanuelschuetze:
I improved it now that "Roboto Medium" is used for strong/bold tags, e.g. motion titles in motion list view. Looks good now for me. Ready to merge.
@FinnStutzenstein Please verify.

FinnStutzenstein:
roboto_alignment
The alignment with the two fonts does not match

FinnStutzenstein:
And the font size could be 120% but all in all I like the new font

emanuelschuetze:
@FinnStutzenstein You can make a new pull request with your (minor) style changes. I'll merge it now...

PR #2749, opened 2016-12-09
changing user import (closes #2666)

### LINKED ISSUE #2666: wrong participant group number in csv example file and description
When importing participants, OS states "Default groups: Delegates 2, Staff 3 Committees 4". But in the offered example file the group number 5 is used for the committee.

This is all wrong: currently the group 6 is the one for delegates and 4 for staff. At least on my system.
When using the offered example file I only get garbage since e.g. the committee has group 5 but there is no such group in OS. I expected that there is a group for committees as described and if it does not exist it should be created during the import.

Maybe I completely misunderstood the feature but it is nevertheless quite unusable as it is.

FinnStutzenstein:
Now, group names are given in the csv file. You can take the default (englisch) name or the translated name for the default groups. E.g. for me in Germany I can take `Delegates` or `Delegierte`. Groups, that are new become created (compare to motions import, categories gets created, too)

### DIFF
--- a/openslides/users/static/js/users/csv.js
+++ b/openslides/users/static/js/users/csv.js
@@ -5,30 +5,62 @@
 angular.module('OpenSlidesApp.users.csv', [])
 
 .factory('UserCsvExport', [
-    function () {
-        return function (element, users) {
-            var csvRows = [
-                ['title', 'first_name', 'last_name', 'structure_level', 'number', 'groups', 'comment', 'is_active', 'is_present', 'is_committee'],
-            ];
-            _.forEach(users, function (user) {
-                var row = [];
-                row.push('"' + user.title + '"');
-                row.push('"' + user.first_name + '"');
-                row.push('"' + user.last_name + '"');
-                row.push('"' + user.structure_level + '"');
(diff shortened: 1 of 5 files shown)


openslides/users/static/js/users/csv.js
```diff
@@ -5,30 +5,63 @@
...
+                    // example entries
+                    ['Dr.', 'Max', 'Mustermann', 'Berlin','1234567890', csvGroups, 'xyz', '1', '1', ''],
+                    ['', 'John', 'Doe', 'Washington','75/99/8-2', csvGroup, 'abc', '1', '1', ''],
+                    ['', 'Fred', 'Bloggs', 'London', '', '', '', '', '', ''],
+                    ['', '', 'Executive Board', '', '', '', '', '', '', '1'],
+
+                ];
+                var csvString = csvRows.join("%0A");
+                element.href = 'data:text/csv;charset=utf-8,' + csvString;
+                element.download = gettextCatalog.getString('users') + '-' +
```
FinnStutzenstein: @emanuelschuetze Opinion?

FinnStutzenstein:
rebased.

Issue #2747, opened 2016-12-07
Regression: Countdown time change is no longer taken into account


donovaly:
- Assure that you don't have any countdown in your OS
- now create a new countdown
- now edit this new countdown because you want to change its time from 1:00 to 5:00
- do  this change and press the blue button with the checkmark to save this

result: the time is still 1:00 not the 5:00 you set

- edit the countdown again and you can see that the 5:00 is there

This is a regression bug.

emanuelschuetze:
> result: the time is still 1:00 not the 5:00 you set

Confirmed. Now you have to click the "Stop" button. The time is reset to 5:00.

Yes, we should reset it automatically after pressing blue save button.
@FinnStutzenstein Please fix it if you have time...

FinnStutzenstein:
closed with PR #2751 .

donovaly:
Confirmed fixed, Many thanks Finn for all your great work in fixing!

Issue #2660, opened 2016-11-19
ugly linebreak in projector preview


donovaly:
- project the agenda
- scroll two times and zoom in one step
result:
![localhost 8000 2016-11-19 19-30-31](https://cloud.githubusercontent.com/assets/1828501/20457582/ac6121ce-ae8e-11e6-9c23-35ba70da2ff1.jpg)

- now zoom out twice
result:
![localhost 8000 2016-11-19 19-27-52](https://cloud.githubusercontent.com/assets/1828501/20457574/8c5f150c-ae8e-11e6-805f-62b620e213fa.jpg)

This looks ugly and is unnecessary since there is enough space to increase the width of the projector preview a bit so tat two negative values can be displayed in one line.

emanuelschuetze:
I can confirm it in Chrome for values "-1" and "-1" - but in projector mangage view only (not in side bar).
@FinnStutzenstein Please look at this (minor) issue.

FinnStutzenstein:
@emanuelschuetze  Closed with PR #2751

donovaly:
Confirmed fixed.

PR #2751, opened 2016-12-09
Some stylistic changes and new motion filter option

FinnStutzenstein:
* fixing issue #2747 
* fixing issue #2660 
* filter motions for recommendations
* adding recommendation status in motion table
* fixing ugly linebreaks in motion table

### DIFF
--- a/openslides/motions/static/js/motions/site.js
+++ b/openslides/motions/static/js/motions/site.js
@@ -806,6 +806,25 @@ angular.module('OpenSlidesApp.motions.site', [
         });
         $scope.alert = {};
 
+        // collect all states and all recommendations of all workflows
+        $scope.states = [];
+        $scope.recommendations = [];
+        var workflows = Workflow.getAll();
+        _.forEach(workflows, function (workflow) {
+            var workflowHeader = {
+                headername: workflow.name,
+                workflowHeader: true,
+            };
+            $scope.states.push(workflowHeader);
+            $scope.recommendations.push(workflowHeader);
+            _.forEach(workflow.states, function (state) {
+                $scope.states.push(state);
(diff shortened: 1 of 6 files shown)


openslides/motions/static/js/motions/site.js
```diff
@@ -806,6 +806,26 @@ angular.module('OpenSlidesApp.motions.site', [
         });
         $scope.alert = {};
 
+        // collect all states and all recommendations of all workflows
+        $scope.states = [];
+        $scope.recommendations = [];
+        var workflows = Workflow.getAll();
+        _.forEach(workflows, function (workflow) {
+            if (workflows.length > 1) {
```
normanjaeckel: This check is always true because if workflows is an empty array the loop won't run.

openslides/motions/static/js/motions/site.js
```diff
@@ -806,6 +806,26 @@ angular.module('OpenSlidesApp.motions.site', [
...
+                var wf = {};
+                wf.headername = workflow.name;
```
normanjaeckel: Here you can simplify:
```
var wf = {
    headername: workflow.name,
    workflowHeader: true
}
```

openslides/motions/static/js/motions/site.js
FinnStutzenstein: done

.travis.yml
```diff
@@ -9,7 +9,7 @@ python:
   - "3.4"
   - "3.5"
 env:
-  - TRAVIS_NODE_VERSION="4"
+  - TRAVIS_NODE_VERSION="6"
```
normanjaeckel: Do we really want to drop support of Node 4.x?

PR #2760, opened 2016-12-12
Added global config for motion preamble.

normanjaeckel:
@emanuelschuetze Any idea how to add the preamble to the motion form? May we use an extra (uneditable) form field?

(For CSV I did nothing because I think we do not want to export the preamble.)


openslides/motions/config_variables.py
```diff
@@ -49,7 +49,7 @@ def get_config_variables():
 
     yield ConfigVariable(
         name='motions_preamble',
-        default_value='The assembly may decide,',
+        default_value='The assembly may decide:',
```
normanjaeckel: Do you have to change a (translation string) line in motion/site.js too?

emanuelschuetze: Done now

PR #2765, opened 2016-12-12
rename mediafiles private flag to hidden

### LINKED ISSUE #2451: Protection for mediafiles
Now, all mediafiles are reachable over the url `media/file/{filename}` for all registered and unregistered users. Also with the private flag the files are not protected; just hidden in the list view. But the name "private" suggests that this file is real private. So the user thinks that no one could access it, but this is not true.

I think that data protection is an important thing to do. So imho all files should just be accessible if the user is logged in and has the permission to view the file. And two options for the private flag:
1. Rename it to "hidden" with a hint, that the file is just hidden in the list view, but _could_ be accessed by everyone with permissions to see every file. So this flag is named for the purpose what is does: Hide the file.
2. Implement a protection that just the user who uploaded this file can access it. But: why should he upload such a file, because this kind of privacy prevents the user to use the file in motions, elections, ...

I think this topic need some discussion.

FinnStutzenstein:
closes #2451


normanjaeckel:
Migration pull request updated.

PR #2768, opened 2016-12-13
Bugfix for line numbering with CKeditor

CatoTH:


emanuelschuetze:
@CatoTH Thanks. Your fix works successfully. I have cherry-picked your commit to #2720. So I closed this pr here.

PR #2782, opened 2016-12-16
Sort candidates in assignments

FinnStutzenstein:
@emanuelschuetze @normanjaeckel Please check this


FinnStutzenstein:
@normanjaeckel Should I merge the assignment migrations 0003 and 0004 into 0002 or 0001?

openslides/assignments/migrations/0003_assignmentrelateduser_weight.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
normanjaeckel: Please use only one new migration file name give it a explicit name like `0003_candidate_weight.py`.

openslides/assignments/models.py
```diff
@@ -28,14 +28,31 @@ class AssignmentRelatedUser(RESTModelMixin, models.Model):
...
+    """
+    ForeinKey to the user who is related to the assignment.
+    """
+
     elected = models.BooleanField(default=False)
+    """
+    Saves the election state of each user
+    """
+
+    weight = models.IntegerField(null=True)
```
normanjaeckel: Please do not use null=True but default=0.

openslides/assignments/models.py
```diff
@@ -28,14 +28,31 @@ class AssignmentRelatedUser(RESTModelMixin, models.Model):
...
+    """
+    The sort order of the list of speakers. None, if he has already spoken.
```
normanjaeckel: Docstring is wrong. Copy and pasted ... :-)

openslides/assignments/models.py
```diff
@@ -249,7 +271,14 @@ def create_poll(self):
         poll = self.polls.create(
             description=self.poll_description_default,
             pollmethod=pollmethod)
-        poll.set_options({'candidate': user} for user in candidates)
+        options = []
+        for user in candidates:
```
normanjaeckel: Candidates is already a db evaluation so we should not evaluate the db again and again inside the loop. Try to solve this with only one query for all users / related users. Maybe use prefetch_related (https://docs.djangoproject.com/en/1.10/ref/models/querysets/#prefetch-related).

openslides/assignments/models.py
```diff
@@ -361,6 +390,7 @@ class AssignmentOption(RESTModelMixin, BaseOption):
         settings.AUTH_USER_MODEL,
         on_delete=models.CASCADE)
     vote_class = AssignmentVote
+    weight = models.IntegerField(null=True)
```
normanjaeckel: Please do not use null=True but default=0.

openslides/assignments/static/js/assignments/site.js
```diff
@@ -470,6 +471,12 @@ angular.module('OpenSlidesApp.assignments.site', [
                 $scope.defaultProjectorId = projectiondefault.projector_id;
             }
         });
+        $scope.$watch(function () {
```
normanjaeckel: If we use such a watcher here we do not need the Assignment.bindOne in line 464. You have to extend the watch callback. See http://www.js-data.io/docs/js-data-angular#dsbindone

openslides/assignments/views.py
```diff
@@ -186,6 +188,50 @@ def create_poll(self, request, pk=None):
...
+            if not isinstance(user_id, int) or users.get(user_id) is None:
+                raise ValidationError(
+                    {'detail': _('Invalid data.')})
+            valid_users.append(users[user_id])
+
+        weight = 1
+        with transaction.atomic():
+            for valid_user in valid_users:
+                query = Q(user__id=valid_user.id) & Q(assignment__id=assignment.id)
+                assignment_related_user = AssignmentRelatedUser.objects.get(query)
```
normanjaeckel: Here the same. Try not to use extra queries inside the loop.

openslides/assignments/models.py
FinnStutzenstein: Ok, now I fetch the related users directly from `AssignmentRelatedUser`. This is now only one query.

openslides/assignments/static/js/assignments/site.js
FinnStutzenstein: Done.

FinnStutzenstein:
@normanjaeckel ready for review.

openslides/assignments/models.py
```diff
@@ -361,6 +389,7 @@ class AssignmentOption(RESTModelMixin, BaseOption):
         settings.AUTH_USER_MODEL,
         on_delete=models.CASCADE)
     vote_class = AssignmentVote
+    weight = models.IntegerField(default=0)
```
normanjaeckel: Please move this model field one line above. `vote_class` is not a model field (it should have been be separated by one line, but this has not be done before).

FinnStutzenstein:
@normanjaeckel ready for merge.

PR #2797, opened 2016-12-19
projector language (closes #2653)

### LINKED ISSUE #2653: missing translation for agenda title in the projector
- use English as language for OS
- display the agenda

result: one gets in the projector "Tagesordnung" instead of "Agenda":

localhost 8000 2016-11-19 15-07-12

I thought I already reported it and the reason is that OS uses the browser language instead of its own language settings. But I cannot find this bug report.

FinnStutzenstein:


openslides/core/config_variables.py
```diff
@@ -108,6 +108,22 @@ def get_config_variables():
     # Projector
 
     yield ConfigVariable(
+        name='projector_language',
+        default_value='en',
```
emanuelschuetze: the default should be 'browser', please. (Like current behavior.)

openslides/core/config_variables.py
```diff
@@ -108,6 +108,22 @@ def get_config_variables():
     # Projector
 
     yield ConfigVariable(
+        name='projector_language',
+        default_value='browser',
+        input_type='choice',
+        label='Projector language',
```
emanuelschuetze: Please add this string for translation into core/site.js.

openslides/core/config_variables.py
```diff
@@ -108,6 +108,22 @@ def get_config_variables():
...
+        choices=(
+            {'value': 'en', 'display_name': 'English'},
+            {'value': 'de', 'display_name': 'Deutsch'},
+            {'value': 'fr', 'display_name': 'Français'},
+            {'value': 'es', 'display_name': 'Español'},
+            {'value': 'pt', 'display_name': 'Português'},
+            {'value': 'cs', 'display_name': 'Čeština'},
+            {'value': 'browser', 'display_name': 'Current browser language'}),
```
emanuelschuetze: Add this browser string to site.js. (The language can be used in native language)

emanuelschuetze:
@normanjaeckel Please start review.

openslides/core/config_variables.py
normanjaeckel: Please move this line to the top of the choices set.

openslides/core/static/js/core/projector.js
```diff
@@ -56,6 +56,30 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
     }
 ])
 
+.controller('LanguageCtrl', [
+    '$scope',
+    'gettextCatalog',
```
normanjaeckel: This dep is not used.

openslides/core/static/js/core/projector.js
```diff
@@ -56,6 +56,30 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+    'Languages',
+    'Config',
+    'ProjectorID',
+    function ($scope, gettextCatalog, Languages, Config, ProjectorID) {
+        // for the dynamic title
+        $scope.projectorId = ProjectorID();
```
normanjaeckel: You can remove `$scope.projectorId` in ProjectorContainerCtrl now.

FinnStutzenstein: No, see Line 99. The container needs to know the projector id for receiving the resolution.

openslides/core/config_variables.py

openslides/core/config_variables.py

openslides/core/static/js/core/projector.js
FinnStutzenstein: Done.

FinnStutzenstein:
@normanjaeckel ok.

openslides/core/static/js/core/projector.js
normanjaeckel: I think the ProjectorContainerCtrl is a child of the LanguageCtrl because it is attached inside the html tag. So the scope of the ProjectorContainerCtrl inherits the scope of the LanguageCtrl. So our scope var has not to be applied again.

FinnStutzenstein:
@emanuelschuetze @normanjaeckel mergeconflict resolved.

normanjaeckel:
My comment refering to `$scope.projectorId`is still not solved.

FinnStutzenstein:
@normanjaeckel Done.

Issue #2654, opened 2016-11-19
OS does not notify when a login attempt failed


donovaly:
- disable cookies in your browser and afterwards delete all cookies in your browser
- start the OS server and try to login as admin
- OS now correctly warns you that cookies are required, proceed anyway to login as admin

result: your are correctly not logged in but without any notice that the login failed. You see the content of OS as if you were a guest. That is in principle correct. but users might still wonder why they cannot do anything. Therefore OS must issue an error message If a login attempt failed. if it failed because of disabled cookies, this should displayed. with this info.

The error message should be displayed like a message with red text to assure that the users really get this/cannot overlook it.

emanuelschuetze:
@FinnStutzenstein If you find a quick solution for 2.1 please implement. Otherwise we move it to 2.2.

FinnStutzenstein:
@donovaly @emanuelschuetze hiw do you disable cookies? In you browser settings or with a plugin (which plugin)?

donovaly:
> how do you disable cookies?

In the browser settings. In some organizations they are by default disabled. I don't know how this is technically done but I have seen this. Personally I always browse with disabled cookies unless I need them (like now to use GitHub). To achieve this one can explicitly disable them or use the safe browse mode. Also on my smart phone there is a privacy browse mode that denies cookies by default.

FinnStutzenstein:
@donovaly please retest, should be fixed with the PR.

emanuelschuetze:
@donovaly please reopen if this is not fixed for you.

donovaly:
> @donovaly please retest, should be fixed with the PR.

This is fixed in beta3. many thanks!

PR #2798, opened 2016-12-19
Cookie handling on login

FinnStutzenstein:
Just a suggestion: The client sends `cookies: false` in login request, if cookies are disabled. So the server could just give an error and the user stays at the login form (if guests are enabled). This prevents the user of thinking he is logged in (See issue #2654 ).

@normanjaeckel (and others): Whats you opinion?

### DIFF
--- a/openslides/users/views.py
+++ b/openslides/users/views.py
@@ -164,6 +164,9 @@ class UserLoginView(APIView):
     http_method_names = ['get', 'post']
 
     def post(self, *args, **kwargs):
+        # If the client tells that cookies are disabled, do not continue as guest (if enabled)
+        if not self.request.data.get('cookies', True):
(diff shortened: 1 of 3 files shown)


normanjaeckel:
> and the user stays at the login form (if guests are enabled).and the user stays at the login form (if guests are enabled).

Do you mean `if guests are disabled`?

openslides/users/views.py
```diff
@@ -164,6 +164,9 @@ class UserLoginView(APIView):
     http_method_names = ['get', 'post']
 
     def post(self, *args, **kwargs):
+        # If the client tells that cookies are disabled, do not continue as guest (if enabled)
```
normanjaeckel: Ok, I understand now.

PR #2800, opened 2016-12-20
Consistent time handling and generic time converter code

### LINKED ISSUE #2748: Incosistent time handling in agenda import/export
OS offers an example CSV file to see how agenda entries are added to a CSV import file. This file uses the format "hh:mm" for the duration. So let's test it:

- download the offered CSV example file
- use this file for an import
- now export the resulting agenda

result: in the exported CSV the duration has the format "mm". This is inconsistent and confusing although the exported file can be used for another import.

To be consistent either the format "hh:mm" or "mm" should be used for import and export.

FinnStutzenstein:
Closes #2748 and removed some more code repetitions. Also this factory could be extended if more parse options are needed.

### DIFF
--- a/openslides/agenda/static/js/agenda/csv.js
+++ b/openslides/agenda/static/js/agenda/csv.js
@@ -5,16 +5,20 @@
 angular.module('OpenSlidesApp.agenda.csv', [])
 
 .factory('AgendaCsvExport', [
-    function () {
+    'HumanTimeConverter',
+    function (HumanTimeConverter) {
         return function (element, agenda) {
             var csvRows = [
                 ['title', 'text', 'duration', 'comment', 'is_hidden'],
             ];
             _.forEach(agenda, function (item) {
                 var row = [];
+                var duration = item.duration ? HumanTimeConverter.secondsToHumanTime(item.duration*60,
+                        { seconds: 'disabled',
+                            hours: 'enabled' }) : '';
                 row.push('"' + (item.title || '') + '"');
                 row.push('"' + (item.text || '') + '"');
-                row.push('"' + (item.duration || '') + '"');
+                row.push('"' + duration + '"');
                 row.push('"' + (item.comment || '') + '"');
                 row.push('"' + (item.is_hidden ? '1' : '')  + '"');
                 csvRows.push(row);
--- a/openslides/core/static/js/core/base.js
+++ b/openslides/core/static/js/core/base.js
@@ -852,48 +852,105 @@ angular.module('OpenSlidesApp.core', [
     }
 ])
 
+/* Two functions to convert between time duration in seconds <-> human readable time span.
+ * E.g. 90 sec <-> 1:30 (min), 3661 sec <-> 1:01:01 (h)
+ *
+ * secondsToHumanTime: Expects seconds and give [h*:]mm[:ss]. The minutes part is always given, the hours
+ *      and minutes could be controlled. The default are forced seconds and hours just if it is not 0.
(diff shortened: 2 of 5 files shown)


openslides/core/static/js/core/base.js
```diff
@@ -852,48 +852,94 @@ angular.module('OpenSlidesApp.core', [
     }
 ])
 
+/* Two functions to convert between time duration in seconds <-> human readable time span.
+ * E.g. 90 sec <-> 1:30 (min), 3661 sec <-> 1:01:01 (h)
+ *
+ * secondsToHumanTime: Expects seconds and give [h*:]mm[:ss]. The minutes part is always given, the hours
+ *      and minutes could be controlled. The default are forced seconds and hours just if it is not 0.
+ *      - forceSeconds (default true): show seconds even if 00
```
normanjaeckel: If I understand this correctly we have 3 cases. Always seconds, only seconds if they are greater than 0, no seconds.

Am I right? If yes please implement this using only one option flag (accepting strings like 'enable', 'auto', 'disable' or something like this) and not two bool flag.

openslides/core/static/js/core/base.js
```diff
@@ -852,48 +852,94 @@ angular.module('OpenSlidesApp.core', [
...
+ *      - ignoreSeconds (default false): Do not show seconds, even is greater 0. This overrides forceSeconds.
+ *      - forceHours: (default false): Show hours even is hours=0. Just one zero is appended.
```

openslides/core/static/js/core/base.js
```diff
@@ -852,48 +852,94 @@ angular.module('OpenSlidesApp.core', [
...
+                    time = h + ':' + time;
+                }
+                if (seconds < 0) {
+                    time = '-'+time;
+                }
+                return time;
+            },
+            humanTimeToSeconds: function (data, params) {
+                if (!params)
+                    params = {seconds: false, hours: false};
```
normanjaeckel: Missing `{ }`

openslides/core/static/js/core/base.js
```diff
@@ -852,48 +852,94 @@ angular.module('OpenSlidesApp.core', [
...
+                var minLength = 1;
+                if (params.seconds) minLength++;
```
normanjaeckel: Missing `{ }`

openslides/core/static/js/core/base.js
```diff
@@ -852,48 +852,94 @@ angular.module('OpenSlidesApp.core', [
...
+                if (params.hours) minLength++;
```
normanjaeckel: Missing `{ }`

openslides/core/static/js/core/base.js
FinnStutzenstein: Yes, good suggestion..

FinnStutzenstein:
@normanjaeckel ok, all done and rebased.

openslides/agenda/static/js/agenda/csv.js
```diff
@@ -5,16 +5,20 @@
...
             ];
             _.forEach(agenda, function (item) {
                 var row = [];
+                var duration = item.duration ? HumanTimeConverter.secondsToHumanTime(item.duration*60,
+                        { seconds: 'disabled',
+                            hours: 'enabled' }) : '';
                 row.push('"' + (item.title || '') + '"');
                 row.push('"' + (item.text || '') + '"');
-                row.push('"' + (item.duration || '') + '"');
+                row.push('"' + (duration || '') + '"');
```
normanjaeckel: duration is already an empty string (if item.duration was falsy) so we do not have to use the or clause again.

FinnStutzenstein:
@normanjaeckel Done.

PR #2801, opened 2016-12-20
More select mode options for users

FinnStutzenstein:


openslides/users/static/js/users/site.js
```diff
@@ -609,60 +613,58 @@ angular.module('OpenSlidesApp.users.site', [
...
+            });
         };
         // add group for selected users
         $scope.addGroupMultiple = function (group) {
-            angular.forEach($scope.users, function (user) {
-                if (user.selected) {
-                    user.groups_id.push(group);
-                    User.save(user);
-                }
+            if (!group) return;
```
normanjaeckel: Please add `{ }` and please use positive condition check: if group then call selectModeAction ...

openslides/users/static/js/users/site.js
```diff
@@ -609,60 +613,58 @@ angular.module('OpenSlidesApp.users.site', [
...
         // remove group for selected users
         $scope.removeGroupMultiple = function (group) {
-            angular.forEach($scope.users, function (user) {
-                if (user.selected) {
-                    var groupIndex = user.groups_id.indexOf(parseInt(group));
-                    if (groupIndex > -1) {
-                        user.groups_id.splice(groupIndex, 1);
-                        User.save(user);
-                    }
+            if (!group) return;
```
normanjaeckel: Here the same.

normanjaeckel:
Also add CHANGELOG entry if reasonable.

FinnStutzenstein:
@normanjaeckel Done. Added general note for more multiselect actions (includes also generate password etc.)

openslides/users/static/js/users/site.js
```diff
@@ -609,60 +613,60 @@ angular.module('OpenSlidesApp.users.site', [
...
+        $scope.deleteMultiple = function () {
+            selectModeAction(function (user) {
+                $scope.delete(user);
+            });
         };
         // add group for selected users
         $scope.addGroupMultiple = function (group) {
-            angular.forEach($scope.users, function (user) {
-                if (user.selected) {
+            if (!group) {
```
normanjaeckel: Are you sure that this is `!group` instead of `group`?

FinnStutzenstein: Oh, I think I didn't run gulp, so I didn't noticed..; Now its fixed

PR #2803, opened 2016-12-21
Added "No." for participant number in get_full_name().

emanuelschuetze:
Forgot to add a clear prefix for participant number in #2802.

### DIFF
--- a/openslides/users/static/js/users/base.js
+++ b/openslides/users/static/js/users/base.js
@@ -69,8 +69,9 @@ angular.module('OpenSlidesApp.users', [])
     'Group',
     'jsDataModel',
     'gettext',
+    'gettextCatalog',
     'Config',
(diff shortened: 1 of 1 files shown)


openslides/users/static/js/users/base.js
```diff
@@ -133,7 +134,7 @@ angular.module('OpenSlidesApp.users', [])
                         addition.push(structure_level);
                     }
                     if (number) {
-                        addition.push(number);
+                        addition.push(gettextCatalog.getString('No.') + ' ' + number);
```
normanjaeckel: Here we need a context description for the translaters to show that this is an abbrevation of number and not the opposite of yes.

emanuelschuetze: Done

PR #2804, opened 2016-12-25
Disconnect from the update_last_login signal so we skip the autoupdate after login

ostcar:
At the moment, when a user logs in, his field "last_login" is updated. Therefore save() is called and therefore inform_changed_data. But at the moment last_login is not part of the user serializer and therefore no change is send to the client.

But to update the last_login field is a problem, when a lot of users login at the same time.

### DIFF
--- a/openslides/users/apps.py
+++ b/openslides/users/apps.py
@@ -13,6 +13,8 @@ class UsersAppConfig(AppConfig):
         from . import projector  # noqa
 
         # Import all required stuff.
+        from django.contrib.auth.signals import user_logged_in
+        from django.contrib.auth.models import update_last_login
         from ..core.config import config
         from ..core.signals import post_permission_creation
         from ..utils.rest_api import router
@@ -31,3 +33,6 @@ class UsersAppConfig(AppConfig):
         # Register viewsets.
         router.register(self.get_model('User').get_collection_string(), UserViewSet)
         router.register('users/group', GroupViewSet)
+
+        # Disconnect the update_last_login signal. We don't use it
+        user_logged_in.disconnect(update_last_login)


normanjaeckel:
Ok.

@emanuelschuetze Be sure that you do not want to deploy the information when a user logs in to the clients. If you want to trigger this (may be in an extended user table that the admin can see when someone logs in) we have to look for another solution.

emanuelschuetze:
@normanjaeckel Good point. What I want to implement later are the following two features:
a) Which users are _currently_ logged in (to observe the load of concurrent users).
b) The number of users/accounts who has logged in (just for statistic that you can calculate the percentage of participants who uses OpenSlides)

I don't need the information WHEN a user has logged in last time. Is (a) and (b) possible after merging this pull request?

ostcar:
> a) Which users are currently logged in (to observe the load of concurrent users).
Currently this can be done with this function: https://github.com/OpenSlides/OpenSlides/blob/master/openslides/utils/autoupdate.py#L18

But this function will be removed with #2794. After the merge of this pull request, it is still possible. For example you can use ```utils.cache.websocket_user_cache.get_all().keys()``` to get a list of all user ids, that have opened a websocket connection. (There can be false-positive in that list, when a websocket connection has be closed without notice.)

> b) The number of users/accounts who has logged in (just for statistic that you can calculate the percentage of participants who uses OpenSlides)
This will not work any more. If we want to have this feature, I think the best way would be, if we copied the signal, that I disabled in this pull request, to our sourcecode, so we can call ```user.save(informed_changed_data=False)```. At least for the case, that the user had been logged in before.

normanjaeckel:
I propose the following way: We do not disconnect the signal here but override the receiver and set user.save(informed_changed_data=False) without any difference. The advantage is that the last login timestamp is in the database and can be extracted via REST API if someone wants this to do. If @emanuelschuetze later wants to add the described features he can add some logic to send the data in some cases.

Other possibility: We disconnect the signal if DEBUG is True or better if a special env var oder settings var is set so @ostcar can use this var for his tests to disconnect the signal if he needs to do so.

ostcar:
I don't like the second possibility. I don't want to test a fake version.

I see two options. Either merge this right now (this will not break nothing) before you code your new feature, or keep this unneeded signal receiver until you build you feature. In the second case: please do it fast.

normanjaeckel:
We do it via #2831 as discussed.

PR #2818, opened 2017-01-06
Add new personal settings MOTION_IDENTIFIER_WITHOUT_BLANKS.

emanuelschuetze:
Allow to remove all whitespaces from motion identifier if settings
value is True. Default: False.
Example: "B 1 - Ä 1" -> "B1-Ä1"


emanuelschuetze:
Added changelog entry. Rebased.

Issue #2667, opened 2016-11-20
ugly display of import participant table


donovaly:
Use german menus and try to import participants via a CSV table. Result:

![localhost 8000 2016-11-20 18-39-55](https://cloud.githubusercontent.com/assets/1828501/20464879/4868328e-af51-11e6-8902-ee87f3a1aa28.png)

This could easily be fixed by introducing line breaks in the table header line. "Teilnehmendennummer" is a very long word and having it in 2 lines would help a lot. The same for "Gliederungsebene	".

Btw. why "Teilnehmendennummer"? Googling shows only 51 hits while "Teilnehmernummer" gives 136.000 hits. As a native German I also have never heard or read Teilnehmendennummer.

emanuelschuetze:
@FinnStutzenstein Can you improve the preview table. Maybe with adding a second line below the name like in users list table? And use "No." if number is set.

emanuelschuetze:
Done with #2820.

PR #2820, opened 2017-01-06
Change user import table

FinnStutzenstein:
@emanuelschuetze Do you want to have an extra column (the first) with the checkboxes? And do we need a checkbox (de-)select all (like in all other tables)?

### DIFF
--- a/openslides/core/static/css/app.css
+++ b/openslides/core/static/css/app.css
@@ -1249,6 +1249,10 @@ img {
     padding-right: 0 !important;
 }
 
+.scroll-x-container {
+    overflow-x: auto;
+}
+
 /* user details */
 .user_details fieldset {
     margin-bottom: 10px;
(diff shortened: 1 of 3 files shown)


emanuelschuetze:
> Do you want to have an extra column (the first) with the checkboxes? 

No, I want to click on the black check-circle-icon to select/unselect the row. Then the icon (and the counter) should change.

> And do we need a checkbox (de-)select all (like in all other tables)?

No, not necessary.

emanuelschuetze:
Overflow scrolling successfully tested. Still open: select/deselect rows.

FinnStutzenstein:
@emanuelschuetze Ready to test.

emanuelschuetze:
> What is the difference to the old behaviour?

If fixes #2667 and makes some minor improvements for preview table.

> Do we need a CHANGLOG entry?

No. So ready for merge...

PR #2821, opened 2017-01-06
Add new personal settings MOTIONS_ALLOW_AMENDMENTS_OF_AMENDMENTS.

emanuelschuetze:
Set settings value to 0 to disable new amendment button in motion detail
template. Default: 1.


emanuelschuetze:
Added changelog entry. Rebased.

> Is it possible to use a boolean value here instead of 0 and 1?

No, it tested it.  A python boolean "True/False" in settings.py is invalid (unknown value) in JavaScript (which requires "true/false"). So, 0 and 1 is just the easiest way.

normanjaeckel:
Ah I see.
Please test a fix in line 187 of core/views.py: The variable client_settings contains a dict but we should transform it to JSON (i. e. a string) and render this string into the template. Then `True` should be exposed to `true`.

emanuelschuetze:
> we should transform it to JSON (i. e. a string) and render this string into the template

How can I do this? Can you give me an example that I can try it?

normanjaeckel:
Head
`import json`

Line 187
```
""".format(realm=realm, angular_modules=angular_modules, settings=json.dumps(client_settings), js_files=js_files) +
```

emanuelschuetze:
@normanjaeckel I added a new commit. Please review.

PR #2836, opened 2017-01-11
Added groups_id to createopenslides user command.

emanuelschuetze:


openslides/users/management/commands/createopenslidesuser.py
```diff
@@ -26,10 +26,16 @@ def add_arguments(self, parser):
...
+        )
 
     def handle(self, *args, **options):
         user_data = {
             'first_name': options['first_name'],
             'last_name': options['last_name'],
         }
-        User.objects.create_user(options['username'], options['password'], **user_data)
+        user = User.objects.create_user(options['username'], options['password'], **user_data)
+        user.save()
```
normanjaeckel: This save is not necessary.

openslides/users/management/commands/createopenslidesuser.py
```diff
@@ -26,10 +26,16 @@ def add_arguments(self, parser):
...
+        user.groups.add(Group.objects.get(id=options['groups_id']))
```
normanjaeckel: It is probably possbile to use only the id and not to fetch the group from the db.

normanjaeckel: Also you need to validate the input. Only a string that can be transformed to an integer is allowed.

openslides/users/management/commands/createopenslidesuser.py
```diff
@@ -26,10 +26,16 @@ def add_arguments(self, parser):
             'password',
             help='The password of the new user.'
         )
+        parser.add_argument(
+            'groups_id',
+            help='The group ids of the new user.'
```
normanjaeckel: It is only one id.

emanuelschuetze: done

openslides/users/management/commands/createopenslidesuser.py
emanuelschuetze: done

openslides/users/management/commands/createopenslidesuser.py
emanuelschuetze: done (2x)

emanuelschuetze:
Done

PR #2837, opened 2017-01-12
Use gulp production mode in docker images.

emanuelschuetze:


emanuelschuetze:
@normanjaeckel I'll merge it directly because I need today an updated docker image.

normanjaeckel:
Ok.

PR #2875, opened 2017-01-15
delete list of speaker error messages (closes #2237)

### LINKED ISSUE #2237: Delete multiple speakers – How to handle error
If you want to delete multiple speakers, you can send an array with ids. How to handle an error here  (e. g. an invalid speaker id)?

a) Delete nobody? Then we have to add `with transaction.atomic():` to the viewset method.
b) Delete all valid speakers? Then we have to add `break` to the viewset method.
c) Something else ...?

matakuka:
If there are more than 4 removed, there will be an error message with numbers, 
else all removed users will be listed in the message

### DIFF
--- a/tests/integration/agenda/test_viewset.py
+++ b/tests/integration/agenda/test_viewset.py
@@ -1,5 +1,6 @@
 from django.contrib.auth import get_user_model
 from django.core.urlresolvers import reverse
+from django.utils.translation import ugettext
 from rest_framework import status
 from rest_framework.test import APIClient
 
@@ -203,13 +204,17 @@ class ManageSpeaker(TestCase):
         response = self.client.delete(
             reverse('item-manage-speaker', args=[self.item.pk]),
             {'speaker': '1'})
-        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data.get('detail'),
+                         ugettext('No speakers have been removed from the list of speakers.'))
 
     def test_remove_someone_else_invalid_data(self):
         response = self.client.delete(
             reverse('item-manage-speaker', args=[self.item.pk]),
             {'speaker': 'invalid'})
-        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data.get('detail'),
+                         ugettext('No speakers have been removed from the list of speakers.'))
 
     def test_remove_someone_else_non_admin(self):
         admin = get_user_model().objects.get(username='admin')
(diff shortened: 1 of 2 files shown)


matakuka:
Change as discussed: Now the name will only be displayed if it is 1 user, else only the number will be shown.
rebased and ready for review @normanjaeckel

tests/integration/agenda/test_viewset.py
```diff
@@ -195,13 +195,13 @@ def test_remove_someone_else_not_on_list(self):
         response = self.client.delete(
             reverse('item-manage-speaker', args=[self.item.pk]),
             {'speaker': '1'})
-        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.status_code, 200)
```
normanjaeckel: Please add a test line that the correct message is retrieved.

tests/integration/agenda/test_viewset.py
```diff
@@ -195,13 +195,13 @@ def test_remove_someone_else_not_on_list(self):
...
 
     def test_remove_someone_else_invalid_data(self):
         response = self.client.delete(
             reverse('item-manage-speaker', args=[self.item.pk]),
             {'speaker': 'invalid'})
-        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.status_code, 200)
```
normanjaeckel: Please add a test line that the correct message is retrieved.

tests/integration/agenda/test_viewset.py
```diff
@@ -195,13 +196,17 @@ def test_remove_someone_else_not_on_list(self):
         response = self.client.delete(
             reverse('item-manage-speaker', args=[self.item.pk]),
             {'speaker': '1'})
-        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.content.decode('UTF-8'), '{"detail":"' +
+                         ugettext('No speakers have been removed from the list of speakers.' + '"}'))
```
normanjaeckel: This ugettext looks strange. The `}` is inside the function call.

tests/integration/agenda/test_viewset.py
```diff
@@ -195,13 +196,17 @@ def test_remove_someone_else_not_on_list(self):
...
 
     def test_remove_someone_else_invalid_data(self):
         response = self.client.delete(
             reverse('item-manage-speaker', args=[self.item.pk]),
             {'speaker': 'invalid'})
-        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.content.decode('UTF-8'), '{"detail":"' +
+                         ugettext('No speakers have been removed from the list of speakers.' + '"}'))
```
normanjaeckel: Here the same.

emanuelschuetze:

PR #2881, opened 2017-01-15
Changed the function has_perm to support an user id or None as the fi…

ostcar:
…rst argument.

Also fixt the case, where an anonymous user connects to the websocket connection.

Also added a function anonymous_is_enabled() to see, if the anyonmous user is enabled.

### DIFF
--- a/openslides/utils/autoupdate.py
+++ b/openslides/utils/autoupdate.py
@@ -9,7 +9,7 @@ from django.db import transaction
 
 from ..core.config import config
 from ..core.models import Projector
-from .auth import AnonymousUser
+from .auth import AnonymousUser, anonymous_is_enabled
 from .cache import websocket_user_cache
 from .collection import Collection, CollectionElement, CollectionElementList
 
@@ -40,7 +40,7 @@ def ws_add_site(message):
             # Skip apps that do not implement get_startup_elements
             continue
         for collection in get_startup_elements():
-            output.extend(collection.as_autoupdate_for_user(message.user))
+            output.extend(collection.as_autoupdate_for_user(message.user.id))
 
     # Send all data. If there is no data, then only accept the connection
     if output:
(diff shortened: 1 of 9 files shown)


openslides/utils/autoupdate.py
```diff
@@ -40,7 +40,7 @@ def ws_add_site(message):
             # Skip apps that do not implement get_startup_elements
             continue
         for collection in get_startup_elements():
-            output.extend(collection.as_autoupdate_for_user(message.user))
+            output.extend(collection.as_autoupdate_for_user(message.user.id))
```
normanjaeckel: Why is it not possible to use `message.user` instead of `message.user.id`? If we have to use the id then we should change the docstrings and argument names of all as_..._for_user(user.id) methods in utils/collection.py for better readablitity.

ostcar: The docstrings are not wrong. They only don't tell about a feature, which they already have. But I have no problem, if you like to change the docstrings.

message.user is the django anonymous user. Of cause you can add some code to check for it and if anonymous user is enabled and user is the django anonymous user, change it to our anonymous user. But I like it, that the user id can be used. So in the future, it would be possible not to get the hole user object from the session, but only the user id. It should be possible to get read of one db query, when only the id is needed

normanjaeckel: Ok. It was not self explaining that `user` as namend in the method wokrs the same as `user.id` as given by this line. So I would appreciate you to add a comment or docstring extension or thelike (and if you are in good mood a unit test for the as_autoupdate_for_user method called with an id and None). :-)

emanuelschuetze:
@ostcar one remark from @normanjaeckel is still open

emanuelschuetze:
@ostcar ping

ostcar:
I added the docstrings

normanjaeckel:
Travis failed.

ostcar:
This was a temporary error. Fixed it by restarting travis

Issue #2779, opened 2016-12-15
Improve relation to agenda item.


normanjaeckel:
At the moment we often send two single autoupdate messages if a agenda related instance is created or changed. These blobs should be bundle together to improve performance for the websocket handler (daphne).

normanjaeckel:
At the moment I do not have any idea how to solve this without much effort.

normanjaeckel:
Maybe we use a django middleware here. Or something like [this](https://github.com/django/channels/blob/69c59ee8b4924cd346e0a2b1724979375ad9924c/channels/message.py#L66).

FinnStutzenstein:
More from PR #2886 : If one creates a new object the client waits until it's created and then modified the coupled agenda item. This could be done on server side.

emanuelschuetze:
@normanjaeckel I think we can close this issue because #3341 is merged, right?

PR #2886, opened 2017-01-17
Removing unnecessary resolvers (closes #2869)

### LINKED ISSUE #2869: Remove resolver
after merging of #2863 remove now all other findAll() resolve statements from site.js

FinnStutzenstein:
*All* state have to be checked if all resolves are removed and every injection of Ids etc. works.

### DIFF
--- a/openslides/agenda/static/js/agenda/base.js
+++ b/openslides/agenda/static/js/agenda/base.js
@@ -23,21 +23,26 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
 
 .factory('AgendaUpdate',[
     'Agenda',
-    function(Agenda) {
+    'operator',
+    function(Agenda, operator) {
         return {
             saveChanges: function (item_id, changes) {
-                Agenda.find(item_id).then(function(item) {
-                    var something = false;
-                    _.each(changes, function(change) {
-                        if (change.value !== item[change.key]) {
-                            item[change.key] = change.value;
-                            something = true;
+                // change agenda item only if user has the permission to do that
+                if (operator.hasPerms('agenda.can_manage agenda.can_see_hidden_items')) {
+                    Agenda.find(item_id).then(function (item) {
+                        console.log(item);
+                        var something = false;
+                        _.each(changes, function(change) {
+                            if (change.value !== item[change.key]) {
+                                item[change.key] = change.value;
+                                something = true;
+                            }
+                        });
+                        if (something === true) {
+                            Agenda.save(item);
                         }
                     });
-                    if (something === true) {
-                        Agenda.save(item);
-                    }
-                });
+                }
             }
         };
     }
--- a/openslides/agenda/static/templates/agenda/item-list.html
+++ b/openslides/agenda/static/templates/agenda/item-list.html
@@ -8,11 +8,14 @@
         <translate>New</translate>
       </a>
       <!-- import -->
-      <a ui-sref="topics.topic.import" os-perms="agenda.can_manage"
+      <span os-perms="agenda.can_manage">
+        <a ui-sref="topics.topic.import"
+          os-perms="agenda.can_see_hidden_items"
           class="btn btn-default btn-sm">
-        <i class="fa fa-download fa-lg"></i>
-        <translate>Import</translate>
-      </a>
+          <i class="fa fa-download fa-lg"></i>
+          <translate>Import</translate>
+        </a>
+      </span>
       <!-- current list of speakers -->
       <a ui-sref="agenda.current-list-of-speakers" os-perms="users.can_see_name"
           class="btn btn-default btn-sm">
--- a/openslides/assignments/static/js/assignments/site.js
+++ b/openslides/assignments/static/js/assignments/site.js
@@ -42,48 +42,20 @@ angular.module('OpenSlidesApp.assignments.site', [
             })
             .state('assignments.assignment.list', {
                 resolve: {
-                    assignments: function(Assignment) {
-                        return Assignment.findAll();
-                    },
-                    tags: function(Tag) {
-                        return Tag.findAll();
-                    },
-                    items: function(Agenda) {
-                        return Agenda.findAll().catch(
-                            function() {
-                                return null;
-                            }
-                        );
-                    },
-                    phases: function(Assignment) {
+                    phases: ['Assignment', function (Assignment) {
                         return Assignment.getPhases();
-                    }
+                    }],
                 }
             })
             .state('assignments.assignment.detail', {
                 controller: 'AssignmentDetailCtrl',
                 resolve: {
-                    assignment: function(Assignment, $stateParams) {
-                        return Assignment.find($stateParams.id).then(function(assignment) {
-                            return assignment;
-                        });
-                    },
-                    users: function(User) {
-                        return User.findAll();
-                    },
-                    tags: function(Tag) {
-                        return Tag.findAll();
-                    },
-                    items: function(Agenda) {
-                        return Agenda.findAll().catch(
-                            function() {
-                                return null;
-                            }
-                        );
-                    },
-                    phases: function(Assignment) {
+                    assignmentId: ['$stateParams', function($stateParams) {
+                        return $stateParams.id;
+                    }],
+                    phases: ['Assignment', function (Assignment) {
                         return Assignment.getPhases();
-                    }
+                    }],
                 }
             })
             // redirects to assignment detail and opens assignment edit form dialog, uses edit url,
@@ -91,8 +63,8 @@ angular.module('OpenSlidesApp.assignments.site', [
             // (from assignment controller use AssignmentForm factory instead to open dialog in front
             // of current view without redirect)
             .state('assignments.assignment.detail.update', {
-                onEnter: ['$stateParams', '$state', 'ngDialog', 'Assignment',
-                    function($stateParams, $state, ngDialog, Assignment) {
+                onEnter: ['$stateParams', '$state', 'ngDialog',
+                    function($stateParams, $state, ngDialog) {
                         ngDialog.open({
                             template: 'static/templates/assignments/assignment-form.html',
                             controller: 'AssignmentUpdateCtrl',
@@ -100,16 +72,8 @@ angular.module('OpenSlidesApp.assignments.site', [
                             closeByEscape: false,
                             closeByDocument: false,
                             resolve: {
-                                assignment: function() {
-                                    return Assignment.find($stateParams.id).then(function(assignment) {
-                                        return Assignment.loadRelations(assignment, 'agenda_item');
-                                    });
-                                },
-                                items: function(Agenda) {
-                                    return Agenda.findAll().catch(
-                                        function() {
-                                            return null;
-                                        });
+                                assignmentId: function() {
+                                    return $stateParams.id;
                                 },
                             },
                             preCloseCallback: function() {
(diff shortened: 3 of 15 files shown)


openslides/assignments/static/js/assignments/site.js
```diff
@@ -459,8 +413,9 @@ angular.module('OpenSlidesApp.assignments.site', [
     'PdfMakeBallotPaperProvider',
     'gettextCatalog',
     function($scope, $http, $filter, filterFilter, gettext, ngDialog, AssignmentForm, operator, Assignment,
-        User, assignment, phases, Projector, ProjectionDefault, AssignmentContentProvider, BallotContentProvider,
+        User, assignmentId, phases, Projector, ProjectionDefault, AssignmentContentProvider, BallotContentProvider,
         PdfMakeDocumentProvider, PdfMakeBallotPaperProvider, gettextCatalog) {
+        var assignment = Assignment.get(assignmentId);
```
FinnStutzenstein: Question: Quite some times I changed the injection in the way that not the object but the id is injected. In this (and more) cases the controller needs the object and does not only bind the object. So I `get` the object in the controller. Is this acceptable or should we inject the object:
```
resolve: {
    assignment: ['$statePrams', 'Assignment', function ($stateParams, Assignement) {
        return Assignment.get($stateParams.id);
    }]
}
```
so the commented line could be removed?

openslides/assignments/static/js/assignments/site.js
```diff
@@ -419,7 +373,7 @@ angular.module('OpenSlidesApp.assignments.site', [
         };
         // create the PDF List
         $scope.makePDF_assignmentList = function () {
-            User.findAll().then( function(users) {
+            User.getAll().then( function(users) {
```
normanjaeckel: User.getAll should not return a promise so you should not use it with `then()`.

openslides/assignments/static/js/assignments/site.js
normanjaeckel: I think it is better to use get inside the controller. This is more readable.

openslides/assignments/static/js/assignments/site.js
```diff
@@ -703,8 +653,10 @@ angular.module('OpenSlidesApp.assignments.site', [
     'AssignmentForm',
     'Agenda',
     'AgendaUpdate',
-    'assignment',
-    function($scope, Assignment, AssignmentForm, Agenda, AgendaUpdate, assignment) {
+    'assignmentId',
+    function($scope, Assignment, AssignmentForm, Agenda, AgendaUpdate, assignmentId) {
+        var assignment = Assignment.get(assignmentId);
+        Assignment.loadRelations(assignment, 'agenda_item');
```
normanjaeckel: This line can be removed. All agenda_items should be present in DS store.

openslides/assignments/static/js/assignments/site.js
```diff
@@ -753,11 +705,12 @@ angular.module('OpenSlidesApp.assignments.site', [
...
     'gettextCatalog',
     'AssignmentPoll',
-    'assignmentpoll',
+    'assignmentpollId',
     'ballot',
-    function($scope, $filter, gettextCatalog, AssignmentPoll, assignmentpoll, ballot) {
+    function($scope, $filter, gettextCatalog, AssignmentPoll, assignmentpollId, ballot) {
         // set initial values for form model by create deep copy of assignmentpoll object
         // so detail view is not updated while editing poll
+        var assignmentpoll = AssignmentPoll.get(assignmentpollId);
```
normanjaeckel: Is it necessary to use angular.copy here? Cf. other update controller in other apps.

openslides/motions/static/js/motions/site.js
```diff
@@ -803,6 +663,11 @@ angular.module('OpenSlidesApp.motions.site', [
         });
         $scope.alert = {};
 
+        // is this necessary?
+        _.forEach(Motion.getAll(), function(motion) {
+            Motion.loadRelations(motion, 'agenda_item');
```
normanjaeckel: I think loadRelations is not necessary.

openslides/motions/static/js/motions/site.js
```diff
@@ -1091,21 +956,22 @@ angular.module('OpenSlidesApp.motions.site', [
...
+             Motion, MotionComment, Category, Mediafile, Tag, User, Workflow, Config, motionId, MotionInlineEditing,
              MotionCommentsInlineEditing, Projector, ProjectionDefault) {
+        var motion = Motion.get(motionId);
         Category.bindAll({}, $scope, 'categories');
         Mediafile.bindAll({}, $scope, 'mediafiles');
         Tag.bindAll({}, $scope, 'tags');
         User.bindAll({}, $scope, 'users');
         Workflow.bindAll({}, $scope, 'workflows');
-        Motion.loadRelations(motion, 'agenda_item');
+        Motion.loadRelations(Motion.get(motionId), 'agenda_item');
```
normanjaeckel: Here the same.

openslides/motions/static/js/motions/site.js
```diff
@@ -1592,7 +1459,7 @@ angular.module('OpenSlidesApp.motions.site', [
             // save change motion object on server
             Motion.save(motion, { method: 'PATCH' }).then(
                 function(success) {
-                    Agenda.find(success.agenda_item_id).then(function(item) {
+                    Agenda.get(success.agenda_item_id).then(function(item) {
```
normanjaeckel: Don't use `then()` here.

openslides/topics/static/js/topics/site.js
```diff
@@ -182,7 +165,7 @@ angular.module('OpenSlidesApp.topics.site', ['OpenSlidesApp.topics', 'OpenSlides
                 $scope.defaultProjectorId = projectiondefault.projector_id;
             }
         });
-        Topic.loadRelations(topic, 'agenda_item');
+        Topic.loadRelations(Topic.get(topicId), 'agenda_item');
```
normanjaeckel: Here the same.

openslides/assignments/static/js/assignments/site.js
FinnStutzenstein: Done.

openslides/assignments/static/js/assignments/site.js
FinnStutzenstein: Done.

openslides/motions/static/js/motions/site.js
FinnStutzenstein: Done.

openslides/motions/static/js/motions/site.js
FinnStutzenstein: Done.

openslides/motions/static/js/motions/site.js
FinnStutzenstein: Removed the Agenda.get completely. item is never used.

openslides/topics/static/js/topics/site.js
FinnStutzenstein: Done (two times in this file).

openslides/assignments/static/js/assignments/site.js
FinnStutzenstein: Yes, if you do not copy the object the original object will be changed in your browser. So when editing a user you see the user change in the background (if detail view open).

openslides/agenda/static/templates/agenda/item-list.html
```diff
@@ -8,11 +8,14 @@
         <translate>New</translate>
       </a>
       <!-- import -->
-      <a ui-sref="topics.topic.import" os-perms="agenda.can_manage"
+      <span os-perms="agenda.can_manage">
```
FinnStutzenstein: @normanjaeckel @emanuelschuetze You can only import topics correctly if you have both permissions. Without 'can_manage' you cannot set duration, weight,.. and without 'can_see_hidden_items' you cannot set the type.

Same changes in AgendaUpdate factory.

Is it useful that a person with 'can_manage' does not have 'can_see_hidden_items'?

openslides/core/static/js/core/projector.js
```diff
@@ -171,7 +171,7 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
         var STEPS = 5;
         $scope.scroll = 0;
         var setScroll = function (scroll) {
-            scroll = 80 * scroll;
+            scroll = -80 * scroll;
```
FinnStutzenstein: @emanuelschuetze Does you projector scrolls also in the wrong direction in current master? (Should be fixed with this)

openslides/agenda/static/js/agenda/base.js
```diff
@@ -23,21 +23,26 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
...
             saveChanges: function (item_id, changes) {
-                Agenda.find(item_id).then(function(item) {
-                    var something = false;
-                    _.each(changes, function(change) {
-                        if (change.value !== item[change.key]) {
-                            item[change.key] = change.value;
-                            something = true;
+                // change agenda item only if user has the permission to do that
+                if (operator.hasPerms('agenda.can_manage agenda.can_see_hidden_items')) {
+                    Agenda.find(item_id).then(function (item) {
```
FinnStutzenstein: FYI: We cannot get rid of this find: This function is always called if a motion, topic, assignment is created to set properties of the also created agenda item. But the autoupdate is not fast enough to deliver the new agenda item, so a `Agenda.get()` always return undefined. So we have to wait for this item.

FinnStutzenstein:
@emanuelschuetze @normanjaeckel Please review.

openslides/agenda/static/templates/agenda/item-list.html
emanuelschuetze: > Is it useful that a person with 'can_manage' does not have 'can_see_hidden_items'?

No, it's not useful. But we need the option to set 'can_see_hidden_items' permission for normal users who should see more than official agenda items. So I can accept it that staff users have to consider that they have also this extra permission for internal items. No changes required for me.

openslides/core/static/js/core/projector.js
emanuelschuetze: Right, currently we use "natural scrolling". But this is not very intuitive with our arrow buttons. So,  it's very good that you fix this here!

openslides/agenda/static/js/agenda/base.js
normanjaeckel: Ah this is bad. The problem is that we wait for the response of the (e. g. motion) create request and then want to fire a new agenda update request but have to wait for the server to create the agenda item.

Suggestion: We refactor the create requests and let the server do the changes. Then we remove this code here completely.

Please open an issue, if you think, this is good.

Hint: At the moment we have code also to update agenda items on (e. g. motion) update requests but we do not use it because we removed the missleading form fields (flag parent item) or we should remove this form fields (hidden item flag).

openslides/assignments/static/js/assignments/site.js
normanjaeckel: Then please insert a `copy()` here. :-)

openslides/agenda/static/js/agenda/base.js
FinnStutzenstein: Yeah, this were my thoughts, too. I think there is issue #2779 that is coupled with this issue, so when creating an object, the server should handle the object and the coupled agenda item at the same time as well as send one request. I'll link this PR with this issue.

openslides/assignments/static/js/assignments/site.js
FinnStutzenstein: Oh, didn't redone the test changes ;)

openslides/agenda/static/templates/agenda/item-list.html
FinnStutzenstein: Ok.

FinnStutzenstein:
@normanjaeckel Done and rebased.

FinnStutzenstein:
@emanuelschuetze please test new commit

emanuelschuetze:
user profile and change password view  looks good now.

Issue #2480, opened 2016-10-03
Generic get_collection_string() does not support camel case.


normanjaeckel:
E. g. a model `MotionBlock` gets `motions/motionblock` as collection string but the client expects `motions/motion-block`.

normanjaeckel:
This should be fixed now.

FinnStutzenstein:
@normanjaeckel I tried to revert the changes in PR #2622 but this didn't work.

PR #2891, opened 2017-01-18
Added support for CamelCase to RESTModelMixin. See #2480.

normanjaeckel:


openslides/utils/utils.py
```diff
@@ -1,6 +1,23 @@
+import re
 import roman
 
 
+first_re = re.compile('(.)([A-Z][a-z]+)')
+second_re = re.compile('([a-z0-9])([A-Z])')
```
ostcar: Maybe you could rename this two variables. Keep in mind, that we could add more regular expressions for other function in this file in the future.

openslides/utils/utils.py
```diff
@@ -1,6 +1,23 @@
...
+
+
+def convert_camel_case(text):
```
ostcar: The name is unclear. When I read the code, I though a string was converted to CamelCase. For example a function
```python
covert_lower_case(my_string)
```
should make my_string lower case and not expect a lower_case_string.

Proposal: Rename the function to camel_to_snake_case.

openslides/utils/utils.py
```diff
@@ -1,6 +1,23 @@
...
+    """
+    Converts ThisText to this-text.
```
ostcar: What about "ThatText"? :)

openslides/utils/utils.py
normanjaeckel: Changed.

openslides/utils/utils.py
normanjaeckel: Changed.

openslides/utils/utils.py
normanjaeckel: NoProblemWithThisText.

openslides/utils/utils.py
```diff
@@ -1,6 +1,23 @@
...
+
+    Credits: epost.
+    See http://stackoverflow.com/questions/1175208
```
normanjaeckel: Done.

normanjaeckel:
All done now. Please test this by checking out master and adding a projector message. Then check this branch and just make sure everything runs successfully (projector message, chat, projection defaults, motion blocks, motion change recommendation).

PR #2895, opened 2017-01-20
Possibility to send extra startup data (closes #2885)

### LINKED ISSUE #2885: Remove OPTION request for assignment phases.
same as #2862 . See [this function](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/assignments/static/js/assignments/base.js#L319). This is called within some assignment resolvers.

FinnStutzenstein:


FinnStutzenstein:
@normanjaeckel coverage failed. Do we have tests for autoupdate or the methods in the apps.py?

openslides/core/static/js/core/base.js
```diff
@@ -294,6 +297,29 @@ angular.module('OpenSlidesApp.core', [
     }
 ])
 
+.factory('StartupExtraData', [
+    'Assignment',
```
normanjaeckel: Please do not make 'Assignment' to be a dependecy of our core module. So solve this extraData thing we need another approach. Maybe something like we do for custom django settings. See https://github.com/OpenSlides/OpenSlides/blob/master/openslides/core/views.py#L181

FinnStutzenstein:
@normanjaeckel Please look at last commit. Now the phases are passed as an angular constant (in the wrong module ;) ). Now we have the dependency in the core/views.py accessing the assignment model to get the phases. Whats about this concept? (If its ok I'll clean up the code etc.)

normanjaeckel:
No, I think also that the python module assignments should not be a dependency of core. So you can use a generic solution that loops over all apps (see [Django app registry](https://docs.djangoproject.com/en/1.10/ref/applications/#application-registry)) greps extra Angular constants (like we do for site and projector modules) and render this stuff into the response of the startup viewpoint.

FinnStutzenstein:
@normanjaeckel Please review.

normanjaeckel:
Maybe just add a test the checks if the custom variables and settings are part of the response of the api view call.

FinnStutzenstein:
@normanjaeckel Added test.

FinnStutzenstein:
@emanuelschuetze could you restart travis?

emanuelschuetze:
Something seems to be wrong with travis connection here. I'll try to close/reopen...

emanuelschuetze:
Travis done now.
@normanjaeckel Please review test.

PR #2896, opened 2017-01-20
Prevent XSS-attacks (fixes #2871)

### LINKED ISSUE #2871: Validate Strings against XSS attacks
The server accepts strings with something like `<script>alert('XSS')</script>`. Some of them are pasted trusted in the client DOM (just search for 'trusted' to find those places) and this is a security leak.

The simplest solution is to check every string and replace `<script` with `&lt;script`. Maybe we want to use a library to handle the checking and and XSS protection.

FinnStutzenstein:
@normanjaeckel Your opinion for this method? Or do you want to use a dependency to make the checks?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -41,6 +41,7 @@ Core:
 - Replaced angular-csv-import through Papa Parse for csv parsing.
 - Added smooth projector scroll.
 - Added watching permissions in client and change the view immediately on changes.
+- Validate HTML strings from CKEditor against XSS attacks.
 
 Motions:
 - Added adjustable line numbering mode (outside, inside, none) for each
--- a/openslides/core/serializers.py
+++ b/openslides/core/serializers.py
@@ -1,4 +1,5 @@
 from openslides.utils.rest_api import Field, ModelSerializer, ValidationError
+from openslides.utils.validate import validate_html
 
 from .models import (
     ChatMessage,
@@ -77,6 +78,10 @@ class ProjectorMessageSerializer(ModelSerializer):
     class Meta:
         model = ProjectorMessage
         fields = ('id', 'message', )
+
+    def validate(self, data):
+        data['message'] = validate_html(data.get('message', ''))
+        return data
 
 
 class CountdownSerializer(ModelSerializer):
--- a/openslides/motions/serializers.py
+++ b/openslides/motions/serializers.py
@@ -12,6 +12,7 @@ from openslides.utils.rest_api import (
     SerializerMethodField,
     ValidationError,
 )
+from openslides.utils.validate import validate_html
 
 from .models import (
     Category,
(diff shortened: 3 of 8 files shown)


emanuelschuetze:
@FinnStutzenstein travis failed

tests/unit/utils/test_rest_api.py
```diff
@@ -0,0 +1,17 @@
+from unittest import TestCase
+
+from openslides.utils.rest_api import ModelSerializer
+
+
+class ValidatorTest(TestCase):
+    def test_XSS_protection(self):
+        data = {
+            'String alaskjdlk': 'slwq<a><p>test<script></p>relel</a>slsl',
```
normanjaeckel: The test strings are not consistent: Please use a strict test szenario (e. g. always 10 random chars):
```python
'String 1': 'tuveegi2Ho<a><p>tuveegi2Ho<script></p>Boovai7esu</a>ee4Yaiw0ei',
'List 1': [
    'sAa3efae1ji<script><p>iek9in6ahV</p>',
    {
        'String 1': 'oiKuaGie8v',
        'String 2': 'ohJie1Usha<script iek9in6ahV'
    }
]
```

normanjaeckel:
@emanuelschuetze What do you think about whitelisting and solving the problem on client side (this is where it occurs at the moment because we use bindHTML or something like this).

FinnStutzenstein:
@normanjaeckel The CKEditor blocks the script tag already. But if I create a custom XHS Request, I can put everything in the database I want. Do not trust the user and validate this on the server.

normanjaeckel:
With client side solution I do not mean to block the user that wants to send malicious data but to block the execution of this by the recipient client. AngularJS supports tag escaping by default but we use bindHTML or something like this to enable HTML tags there.

normanjaeckel:
(Do not forget a CHANGELOG entry.)

emanuelschuetze:
> With client side solution I do not mean to block the user that wants to send malicious data

why not? Firstly, we should prevent that a user try to insert malicious data into database (e.g. by entering `<cript> ..</script>` into ckeditor - which is already ignored by ckeditor OR by creating customm request, which has to parsed on server side). After this I see no way to execute malicious code in template by recipient client because no malicious code is saved.

tests/unit/utils/test_rest_api.py
FinnStutzenstein: Done.

openslides/utils/rest_api.py
```diff
@@ -163,6 +163,23 @@ def get_fields(self):
             fields[field_name] = field
         return fields
 
+    # simple global check for XSS attacs
```
FinnStutzenstein: Done.

normanjaeckel:
Ok then we should define a whitelist of allowed tags for several model fields that may contain HTML. We should use bleech as we did in OpenSlides 1.6 or 1.7

FinnStutzenstein:
@normanjaeckel Now using bleach. The allowed tags, attributes and styles should be everything the CKEditor produces as output at the moment.

emanuelschuetze:
@normanjaeckel ping

normanjaeckel:
Please add CHANGELOG entry. Then rebase and merge.

emanuelschuetze:
I added CHANGELOG entry.

emanuelschuetze:
Ready for merge...

PR #2910, opened 2017-01-26
Use custom redis prefix for asgi and cache entries.

emanuelschuetze:
Note: These prefix have to be unique for each OpenSlides instance!

@ostcar Do you accept the new defaults for geiss?


emanuelschuetze:
As discussed with @ostcar I changed the channels prefix to common default "asgi:".

PR #2931, opened 2017-02-03
Reorganize the current list of speakers. Fixes #2669 and find()-calls

### LINKED ISSUE #2669: List of speakers offers inexisting features
Below the projector preview there is a new section "List of speakers". This is a great feature but it only makes sense if the currently projected thing has a speakers list.

Take for example the case that you project a user. Cklick in the list of speakers on the projection symbol and/or press the microphone icon.
Result: nothing happens because a user has technically  no speakers list.

The fix would simply be to disable the speakers list buttons if the currently projected thing provides no speakers list.

This issue is in principle the same as issue #2661.

FinnStutzenstein:
- Now just one slide with the option to have an overlay or not.
- Configurable in the projector element (overlay attribute).
- Fixed errors with projecting the slide.
- Fixed goto manage current list of speakers button (closes #2669).
- Fixed [many Todos](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/agenda/static/js/agenda/base.js#L292).
- This creates the base for supporting multiple current lists of speakers.


FinnStutzenstein:
One scenario is not covered:

When changing the current list of speakers reference (the config value) no autoupdate is send for the clos-slides. The projectors recognize the change and try to watch on the new reference. This will fail, because they didn't get the new reference projector.

@normanjaeckel any idea?

normanjaeckel:
Maybe we have to send a full projector update if the config value changes as we do for broadcast config value. See core/models.py:179.

openslides/agenda/projector.py
```diff
@@ -73,13 +73,19 @@ def update_data(self):
...
     """
-    Main class for the list of speaker slides.
-
+    Slide for the current list of speakers.
     """
+    name = 'agenda/current-list-of-speakers'
+
     def get_requirements(self, config_entry):
-        items = self.get_agenda_items(config['projector_currentListOfSpeakers_reference'])
+        # The query mechanism on client need the referenced projector.
```
normanjaeckel: Typo `needs`

openslides/agenda/projector.py
```diff
@@ -90,35 +96,27 @@ def get_requirements(self, config_entry):
...
-        items = self.get_agenda_items(config['projector_currentListOfSpeakers_reference'])
-        for item in items:
-            if collection_element == CollectionElement.from_values(item.get_collection_string(), item.pk):
-                output.extend(self.get_requirements_as_collection_elements(config_entry))
-                break
+        # Full update if agenda_item or referenced projector changes because
+        # then we may have new candidates and therefor need new users.
+        reference_projector = Projector.objects.get(
+            pk=config['projector_currentListOfSpeakers_reference'])
+
```
normanjaeckel: Please remove this blank line.

FinnStutzenstein:
@normanjaeckel @emanuelschuetze Fixed autoupdate.

openslides/agenda/projector.py
FinnStutzenstein: Done.

openslides/agenda/projector.py
FinnStutzenstein: Done.

FinnStutzenstein:
@normanjaeckel Ping. You can review now.

openslides/agenda/projector.py
```diff
@@ -90,35 +96,30 @@ def get_requirements(self, config_entry):
...
-                break
+        # Full update if agenda_item or referenced projector changes because
+        # then we may have new candidates and therefor need new users.
+        reference_projector = Projector.objects.get(
+            pk=config['projector_currentListOfSpeakers_reference'])
+        is_reference_projector = collection_element == CollectionElement.from_values(
+                reference_projector.get_collection_string(),
+                reference_projector.pk)
+        is_config = collection_element.collection_string == 'core/config' and \
+            collection_element.get_instance()['key'] == 'projector_currentListOfSpeakers_reference'
```
normanjaeckel: I think we should not get the instance from the collection element but use something like this:
`collection_element.information.get('changed_config') == 'projector_currentListOfSpeakers_reference'`

And please do not use `\` but braces.

FinnStutzenstein:
@normanjaeckel Done with the remark. Rebased.

PR #2937, opened 2017-02-06
Inline edit the ballot description (closes #2560)

### LINKED ISSUE #2560: Can not change the poll description field for a specific ballot
* Create a new election and enter something into the poll description default field.
* Add a candidate and create first ballot.
* create ballot paper -> The default poll description is shown correctly in pdf.
* Now, try to edit the poll description for this specific ballot only. The form field in poll edit form does not save the value - and it's also the wrong place, because you have to enter poll results which are not yet available. So we need a possibility to change the poll description independent from poll edit form.

Workaround: Change poll description default field in assignment form. After that create a new ballot with new (chaned) description. Note: it has to changed _before_ new ballot is created.

FinnStutzenstein:


openslides/assignments/static/templates/assignments/assignment-detail.html
```diff
@@ -133,7 +133,13 @@ <h3 translate>Election result</h3>
         index="$index" heading="{{ 'Ballot' | translate }} {{ assignment.polls.length - $index }}">
       <div ng-controller="AssignmentPollDetailCtrl">
         <!-- action buttons -->
-        <div class="pull-right">
+        <div class="pull-right" os-perms="assignments.can_manage">
+          <span class="spacer-right" editable-text="poll.description" onaftersave="saveDescriptionChange(poll)">
+            <translate>Comment on the ballot paper</translate>:
```
emanuelschuetze: Suggestion: Set this label as tooltip. Otherwise it's very long line with label and inline editing value...
Use 'hint' instead of 'comment' (it's not an internal comment).

openslides/assignments/static/templates/assignments/assignment-detail.html
```diff
@@ -133,7 +133,13 @@ <h3 translate>Election result</h3>
...
+            <span ng-if="!poll.description" translate>Set description ...</span>
```
emanuelschuetze: 'description' is not a good word. Suggestion: 'Set hint for ballot paper ...'

openslides/assignments/static/templates/assignments/assignment-detail.html
FinnStutzenstein: Done.

openslides/assignments/static/templates/assignments/assignment-detail.html
FinnStutzenstein: Done.

PR #2940, opened 2017-02-08
Updated requirements.

emanuelschuetze:
Channels uses now latest twisted 16.6.0 which requires MS Visual C++
2015 build tools on Windows.


emanuelschuetze:
Added notice for windows section in DEVELOPMENT.rst. I think, the CHANGELOG is not the right place for news about requirement updates or build system -> so not changed.

PR #2943, opened 2017-02-10
New full text search on client-side (Fixed #2865).

### LINKED ISSUE #2865: Server side search with whoosh fails with many clients.

emanuelschuetze:
Removed server-side search by whoosh.


openslides/agenda/static/js/agenda/base.js
```diff
@@ -87,6 +87,26 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
...
+                // return true if a specific relation matches for given searchquery
+                // e.g. speakers
+                hasSearchResult: function (results) {
+                    var item = this;
+                    var foundSomething;
+                    // search for speakers (check if any user.id from already found users matches)
+                    _.forEach(results, function(result) {
+                        switch(result.getResourceName()) {
+                            case "users/user":
+                                if (_.find(item.speakers, {'user_id': result.id})) {
```
normanjaeckel: You can also use _.some here but I do not know whether it is better than _.find

openslides/core/static/js/core/site.js
```diff
@@ -952,76 +973,59 @@ angular.module('OpenSlidesApp.core.site', [
...
+                result.urlDetailState = searchModule.urlDetailState;
+                result.weight = searchModule.weight;
+                result.checked = true;
+                result.elements = $filter('filter')(DS.getAll(searchModule.collectionName), $scope.searchquery);
+                $scope.results.push(result);
+                _.forEach(result.elements, function(element) {
+                    foundObjects.push(element);
+                });
+            });
+            // search additional in specific releations of all defined searchModule
```
normanjaeckel: missing s at the end of the comment

openslides/core/static/js/core/site.js
```diff
@@ -952,76 +973,59 @@ angular.module('OpenSlidesApp.core.site', [
...
+                        if (object.hasSearchResult(foundObjects, $scope.searchquery)) {
+                            // releation found, check if object is not yet in search results
+                            _.forEach($scope.results, function(result) {
+                                if ((object.getResourceName() == result.collectionName) &&
+                                        _.findIndex(result.elements, {'id': object.id}) === -1) {
+                                    result.elements.push(object);
+                                }
+                            });
+                        }
+                    } catch(err) {
```
normanjaeckel: What error do you want to catch here? If it is whether a ressource supports relational search the please use something like `if (object.hasSearchResult === !void 0){...}` or even better `if (_.isFunction(object.hasSearchResult)) {...}`

openslides/core/static/js/core/site.js
```diff
@@ -952,76 +973,59 @@ angular.module('OpenSlidesApp.core.site', [
...
+                });
+            });
+            // search additional in specific releations of all defined searchModule
+            _.forEach(searchModules, function(searchModule) {
+                _.forEach(DS.getAll(searchModule.collectionName), function(object) {
+                    try {
+                        if (object.hasSearchResult(foundObjects, $scope.searchquery)) {
+                            // releation found, check if object is not yet in search results
+                            _.forEach($scope.results, function(result) {
+                                if ((object.getResourceName() == result.collectionName) &&
```
normanjaeckel: Please use `===`.

openslides/motions/static/js/motions/base.js
```diff
@@ -385,9 +386,27 @@ angular.module('OpenSlidesApp.motions', [
...
+                            case "users/user":
+                                if (_.find(motion.submitters, {'id': result.id})) {
+                                    foundSomething = true;
+                                } else if (_.find(motion.supporters, { 'id': result.id })) {
+                                    foundSomething = true;
+                                }
+                        }
+                    });
+                    // search for category
+                    if (motion.category && motion.category.name.match(new RegExp(searchquery, 'i'))) {
```
normanjaeckel: Is this faster than the switch-case block above? It is more complex to check users in one way and categories in another.

emanuelschuetze: The resource `motions/category` is not a registed search module (because only base apps should be registed and add the app name as search filter). So a new switch/case does not work. - The easiest way is to search the categories manually. A more generic solution is not really required because it is only important for motion category.

openslides/core/static/js/core/site.js
emanuelschuetze: Done

openslides/core/static/js/core/site.js
emanuelschuetze: Done

openslides/core/static/js/core/site.js
emanuelschuetze: Done

openslides/agenda/static/js/agenda/base.js
emanuelschuetze: Done (for agenda, assignments, motions)

openslides/agenda/static/js/agenda/base.js
```diff
@@ -87,6 +87,26 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
...
+                getSearchResultName: function () {
+                    return this.getAgendaTitle();
+                },
+                // return true if a specific relation matches for given searchquery
+                // e.g. speakers
+                hasSearchResult: function (results) {
+                    var item = this;
+                    var foundSomething;
+                    // search for speakers (check if any user.id from already found users matches)
+                    _.forEach(results, function(result) {
```
emanuelschuetze: Done (for agenda, assignments, motions)

emanuelschuetze:
@normanjaeckel All done. Ready for last review and merge.

emanuelschuetze:
@normanjaeckel I already rebased and I'll merge it now. But, as discussed, I pushed a new backup branch that you can review my [changes](https://github.com/emanuelschuetze/OpenSlides/commit/1896200ce281c04919114dd509421c7a93adda3d) later (after merge). If you have more remarks I'll change it after merge.

openslides/motions/static/js/motions/base.js
normanjaeckel: Ok.

openslides/core/static/js/core/site.js
```diff
@@ -952,76 +973,59 @@ angular.module('OpenSlidesApp.core.site', [
...
+                            // releation found, check if object is not yet in search results
+                            _.forEach($scope.results, function(result) {
+                                if ((object.getResourceName() === result.collectionName) &&
+                                        _.findIndex(result.elements, {'id': object.id}) === -1) {
+                                    result.elements.push(object);
+                                }
+                            });
+                        }
+                    } else {
+                        return false;
```
normanjaeckel: I think the else clause and the return false is not necessary in this forEach loop.

openslides/motions/static/js/motions/base.js
```diff
@@ -192,6 +192,7 @@ angular.module('OpenSlidesApp.motions', [
 .factory('Motion', [
     'DS',
     '$http',
+    '$filter',
```
normanjaeckel: The $filter injection is not used.

emanuelschuetze: Fixed in #2996

openslides/core/static/js/core/site.js
emanuelschuetze: Tested without else-return. Result: It is required. - Otherwise you have to iterate over all users although no users (the users factory) provide the `hasSearchResult`. So the return cancel the iteration already on first run and `if (_.isFunction(object.hasSearchResult))` is false.

normanjaeckel: Ok.

PR #2953, opened 2017-02-13
set ckeditor to fixed Version (fixes #2921)

### LINKED ISSUE #2921: CKEditor in config: Cannot read property 'isInline' of null
If you have an editor open in Config and change into another app, you will get this error.

Changing state from a motion with active inline editing to another state works; @matakuka Do you know, if the editor become disabled if you leave a motion? Did someone implements this already?

matakuka:
A bug was introduced in ckeditor 4.6.2, see #2921. This "works" around by not pulling the version affected, which does not has any new features we need.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -22,7 +22,7 @@
     "angular-xeditable": "~0.5.0",
     "bootstrap-css-only": "~3.3.6",
     "bootstrap-ui-datetime-picker": "~2.4.0",
-    "ckeditor": "~4.6.1",
+    "ckeditor": "4.6.1",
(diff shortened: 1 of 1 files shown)


normanjaeckel:
Do you think (or know) whether 4.6.3 will have a fix? If yes I would prefere an exclude of 4.6.2 in the version range instead of a fix version to 4.6.1.

matakuka:
The bug report target is set to the next major version (4.7), but I am not sure of their (real) timings. But as the fix seems to be easy (already provided in the bug reports), this will probably get fixed.

PR #2959, opened 2017-02-14
Added missing topic text to agenda csv export (Fixed #2954).

### LINKED ISSUE #2954: regression: agenda eport looses "Text"
OS offers an example CSV file to see how agenda entries are added to a CSV import file. So let's test it:

1. download the offered CSV example file
1. use this file for an import
1. now export the resulting agenda

result: in the exported CSV the text of the "text" property is lost.

The example file of OS 2.1 b3 is this one:
[agenda-example.txt](https://github.com/OpenSlides/OpenSlides/files/772717/agenda-example.txt)
and this is what one gets after the export:
[agenda-export-2.txt](https://github.com/OpenSlides/OpenSlides/files/772718/agenda-export-2.txt)

expected result: the exported file is identic to the imported file because nothing has been changed in the meantime. And of course the Text data must not be lost.

The dataloss is a regression to Os 2.0 and to OS 2.1 b1.

emanuelschuetze:


openslides/agenda/static/js/agenda/base.js
```diff
@@ -87,6 +87,16 @@ angular.module('OpenSlidesApp.agenda', ['OpenSlidesApp.users'])
                 getAgendaTitle: function () {
                     return this.title;
                 },
+                getTopicText: function () {
```
normanjaeckel: This is a bad hack but I can live with it.

Proposal: Rename this method to `getCSVExportText`. Then every app or plugin may decide whether to provide extra text or not and it is not semi-hard coded to the topic app.

emanuelschuetze: Done (I use 'getCSVExportText' now).

emanuelschuetze:
Rebased. Ready for merge...

PR #2969, opened 2017-02-15
Solving two bugs in the diff - for #2961

CatoTH:
This solves:
- Double Line-numbering in Diff View
- In Diff view, inserted paragraphs were not marked as such (only the first paragraph), as could be seen in the test database ("Der Wolf hatte danach richtig schlechte laune, trank eine Flasche Rum,
machte eine Weltreise und kam danach wieder um die Ziegen zu fressen." wasn't green before)

### DIFF
--- a/openslides/core/static/css/app.css
+++ b/openslides/core/static/css/app.css
@@ -609,11 +609,11 @@ img {
     padding-top: 5px;
 }
 .change-recommendation-overview {
-    margin-bottom: 50px;
-    padding: 10px;
+    margin-bottom: 30px;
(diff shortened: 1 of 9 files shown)


CatoTH:
@emanuelschuetze if you're referring to the negative margin of the diff box, this should be fixed now.

emanuelschuetze:
@CatoTH Thanks for your layot fix. I added a new commit to this pull request which fixed an other bug in slide template (wrong order of text passages) + more layout fixes. Now it looks very good on motion detail and projector.

Issue #1871, opened 2016-01-14
Fix serializers for PUT requests


normanjaeckel:
At the moment we have two bad hacks: In motion update view and user update view a non staff client can not send a PUT request because on the one hand he is not allowed to change some of the fields, on the other hand the serializer expects all these fields. It has to send a PATCH request and we strip out any field the are not on the whitelist.

(Another bad hack might be also to transform PUT requests to PATCH requests internally.)

I propose the following solution: We split up the serializers (maybe using class inheritance) so that non staff clients get serializers where the fields of the "blacklist" are set to read only fields. Maybe we can find help [here](http://www.django-rest-framework.org/api-guide/serializers/#dynamically-modifying-fields). Then the client can send PUT or PATCH as it like and also the API is not broken.

ostcar:
So for read requests (get one or get all) it would be equal which serializer would be used? This is important to add caching in the future.

normanjaeckel:
If we use the serializer info also to generate a correct OPTIONS view than the serializer has to know which fields would be read only for a specific client and which not.

ostcar:
OPTIONS requests are not problem for the caching

normanjaeckel:
I the proposed solution we have different serializers for different users.

We can do this for 2.1

normanjaeckel:
Look also for PATCH requests in mediafiles and agenda.

normanjaeckel:
Note to myself: Agenda is no problem. Mediafile is solved. Only motion is still open.

PR #2973, opened 2017-02-17
Dialogs for some views

### LINKED ISSUE #2952: User cannot change password without 'user.can_see' permission

### LINKED ISSUE #2965: Clearing chat messages produces console error
Add some chat messages and then clear them. We get a console error and an unclear behavior in other clients concerning the bagde with the number of currently unread messages. Proposal: The badge should be set to zero and disappear.

### LINKED ISSUE #2970: User profile form is not saved
A user cannot changed the own user profile. Fields are not saved.

### LINKED ISSUE #2971: User is logged out after changing password
* Click "Change password" in user menu
* Save new password

-> Password is changed. User is redirected to users list view. The user is logged out, but It seems that the session is already active (user name is shown in user menu).

Please prevent a logout after changing password!

FinnStutzenstein:
- fixed #2971 , user stays logged in after password change
- fixed #2970 (with new form)
- fixed #2965 (check, if there are chatmessages)
- fixed #2952 (with new form)
- Profile and password view are now dialogs.
- Tag and category forms are now dialogs.
- Clicking on 'edit current slide' does not change the state; Dialog opens right away.
- Same for clicking on edit in the agenda. No state transition.
- removed tag and category detail view. Currently you couldn't get to them, because all links to the detail views are removed. And also, why have a detail view for a tag or category when they have just one or two properties?

There are two things left:

- The mediafile form is not handled like all other forms. Much old code. I'll do this, but not in this PR because in this PR there are already so many things..
- @normanjaeckel I check in the users/views.py if the operator changes itself (via profile form). But I want now to limit the data. Please see the comment where I explain the problem.


openslides/users/views.py
```diff
@@ -68,6 +71,7 @@ def update(self, request, *args, **kwargs):
             # Check permissions only to update yourself.
             if str(request.user.pk) != self.kwargs['pk']:
                 self.permission_denied(request)
+            #del request.data['groups_id']
```
FinnStutzenstein: Now the user can update himself/herself. But in the request data are things like `groups_id`, `is_committee`, `username`, `is_present`... A user without `users.can_manage` is not allowed to change these entries. The form forbids to change them, but you can manually create a request and change you groups etc.

In the previous lines is checked witch permissions the user has. If he has not the right permissions, the special data should be removed. When commenting in `del request.data['groups_id']` I get an error, that `groups_id` is required. Is there a possibility just to partially update the user or strip out the forbidden fields?

emanuelschuetze:
Tested, found only 2 issues:
* Category edit view doesn't work
* MotionBlock edit view (from detail view) does not work
Rest looks and works very good!

FinnStutzenstein:
@emanuelschuetze Should work now.

openslides/agenda/static/js/agenda/site.js
```diff
@@ -268,13 +269,19 @@ angular.module('OpenSlidesApp.agenda.site', [
...
+        $scope.getDetailStatePrefix = function (item) {
             var prefix = item.content_object.collection.replace('/','.');
             // Hotfix for Issue 2566.
             // The changes could be reverted if Issue 2480 is closed.
             prefix = prefix.replace('motion-block', 'motionBlock');
             return prefix;
         };
+        $scope.edit = function (item) {
+            var formName = item.content_object.collection.split('/')[1];
+            formName = formName.charAt(0).toUpperCase() + formName.slice(1) + 'Form';
```
normanjaeckel: Hm. This does not work if I have a collection `foo/bar-baz`. It would be transformed to `Bar-bazForm` instead of `BarBazForm`.

openslides/agenda/static/js/agenda/site.js
```diff
@@ -268,13 +269,19 @@ angular.module('OpenSlidesApp.agenda.site', [
                     return false;
             }
         };
-        $scope.getUpdateStatePrefix = function (item) {
+        $scope.getDetailStatePrefix = function (item) {
             var prefix = item.content_object.collection.replace('/','.');
             // Hotfix for Issue 2566.
             // The changes could be reverted if Issue 2480 is closed.
```
normanjaeckel: Issue 2480 is closed. A pull request was merged yesterday or the day before. Please check if this hotfix is still necessary.

openslides/core/static/js/core/base.js
```diff
@@ -701,24 +702,26 @@ angular.module('OpenSlidesApp.core', [
...
                 },
-                getStateForCurrentSlide: function () {
+                getFormForCurrentSlide: function () {
                     var return_dict;
                     angular.forEach(this.elements, function(value, key) {
                         if (value.name == 'agenda/list-of-speakers') {
                             return_dict = {
-                                'state': 'agenda.item.detail',
-                                'param': {id: value.id}
+                                form: $injector.get('TopicForm'),
```
normanjaeckel: This looks wrong. If I have a list of speakers slide, it has not necessarily something to do with topics. Also motion items have list of speakers. And an ItemForm does not exist.

openslides/core/static/js/core/base.js
```diff
@@ -701,24 +702,26 @@ angular.module('OpenSlidesApp.core', [
...
                             value.name != 'core/clock' &&
                             value.name != 'core/countdown' &&
                             value.name != 'core/projector-message' &&
                             value.name != 'agenda/current-list-of-speakers' ) {
-                            return_dict = {
-                                'state': value.name.replace('/', '.')+'.detail.update',
-                                'param': {id: value.id}
-                            };
+                                var formName = value.name.split('/')[1];
+                                formName = formName.charAt(0).toUpperCase() + formName.slice(1) + 'Form';
```
normanjaeckel: Here the same. Maybe we need a factory that transforms collection names in CamelCase angular stuff.

openslides/agenda/static/js/agenda/site.js
```diff
@@ -268,13 +269,22 @@ angular.module('OpenSlidesApp.agenda.site', [
...
             prefix = prefix.replace('motion-block', 'motionBlock');
             return prefix;
         };
+        $scope.edit = function (item) {
+            var formName = item.content_object.collection.split('/')[1];
+            // Hotfix for Issue 2566.
+            // The changes could be reverted if Issue 2480 is closed.
+            formName = formName.replace('motion-block', 'motionBlock');
+            formName = formName.charAt(0).toUpperCase() + formName.slice(1) + 'Form';
```
normanjaeckel: Hm. This does not work if I have a collection foo/bar-baz. It would be transformed to Bar-bazForm instead of BarBazForm.

openslides/core/static/js/core/base.js
```diff
@@ -701,24 +702,29 @@ angular.module('OpenSlidesApp.core', [
...
                             value.name != 'agenda/current-list-of-speakers' ) {
-                            return_dict = {
-                                'state': value.name.replace('/', '.')+'.detail.update',
-                                'param': {id: value.id}
-                            };
+                                var formName = value.name.split('/')[1];
+                                // Hotfix for Issue 2566.
+                                // The changes could be reverted if Issue 2480 is closed.
+                                formName = formName.replace('motion-block', 'motionBlock');
+                                formName = formName.charAt(0).toUpperCase() + formName.slice(1) + 'Form';
```
normanjaeckel: Here the same. Maybe we need a factory for this stuff.

openslides/users/views.py
```diff
@@ -37,10 +38,12 @@ def check_view_permissions(self):
         """
         Returns True if the user has required permissions.
         """
+        is_operator = (str(self.request.user.pk) == self.kwargs.get('pk'))
```
normanjaeckel: Let us please use a consistent solution:
* 'list', 'retrieve' use only check_permission. There we set the result always to True. The permission check is then done in get_restricted_data which is already implemented.
* 'metadata' is only for users with can_see_name
* 'update' and 'partial_update' is for all users that are logged in (see e. g. some lines in core/views.py as example); we check the permissions all in the update() method below.

openslides/users/views.py
```diff
@@ -68,6 +71,7 @@ def update(self, request, *args, **kwargs):
             # Check permissions only to update yourself.
             if str(request.user.pk) != self.kwargs['pk']:
                 self.permission_denied(request)
+            # del request.data['groups_id']
```
normanjaeckel: Here we need a solution that works together with the problem described here: https://github.com/OpenSlides/OpenSlides/issues/1871

I solved some of the problems but motion also also still open.

openslides/agenda/static/js/agenda/site.js
FinnStutzenstein: Yes, it is. The collection from a motion block is `motions/motion-block`. For the states and forms we need `motions/motionBlock`.

openslides/agenda/static/js/agenda/site.js
FinnStutzenstein: Yes, it's the same problem like the `getDetailStatePrefix`.

openslides/users/views.py
FinnStutzenstein: I would prefer splitting this up.

openslides/core/static/js/core/base.js

openslides/agenda/static/js/agenda/site.js
FinnStutzenstein: I thought this is fixed with the CamelCase-PR?

openslides/users/views.py
FinnStutzenstein: Done.

FinnStutzenstein:
@normanjaeckel So whats about the `fuu-bar` --> `fuuBar` issue? Should the server send this, or should this be changed in the `onMessage` method when the client recieves data? Everything else should be done.

openslides/agenda/static/js/agenda/site.js
normanjaeckel: Yes it is. The server transforms the python class name MotionBlock in the app motions to the collection name `motions/motion-block`. This name is send to the client. The resource definition in the client is `motions/motion-block`. Now you have to create a nice generic method that transforms `motions/motion-block` to `MotionBlock` + `Form` for this generic form dialog thing here. Can you please do it in #2983?

openslides/core/static/js/core/base.js
normanjaeckel: Here the same. See #2983.

openslides/users/views.py
```diff
@@ -39,8 +40,10 @@ def check_view_permissions(self):
         """
         if self.action in ('list', 'retrieve'):
             result = self.get_access_permissions().check_permissions(self.request.user)
-        elif self.action in ('metadata', 'update', 'partial_update'):
+        elif self.action == 'metadata':
             result = has_perm(self.request.user, 'users.can_see_name')
+        elif self.action in ('update', 'partial_update'):
+            result = self.request.user.is_authenticated()
```
normanjaeckel: Ok but now we need to check the has_perm(self.request.user, 'users.can_see_name') inside the update method.

normanjaeckel:
@emanuelschuetze We have two issues which can be done later (#2983 and #2984). But one thing should to be done in this pull request: The update method in views.py must check if the admin user that wants to change other users can see users.

emanuelschuetze:
> But one thing should to be done in this pull request: The update method in views.py must check if the admin user that wants to change other users can see users.

@normanjaeckel As discussed, I create a new issue for you, see #2986.

> Also we can add a CHANGELOG notice about the new dialogs.

Done in new commit.

Ready for merge now...

PR #2981, opened 2017-02-17
Send also motion.state.workflow to motion block slide.

emanuelschuetze:
to show recommendation names in slide template

### DIFF
--- a/openslides/motions/projector.py
+++ b/openslides/motions/projector.py
@@ -1,7 +1,7 @@
 from ..core.exceptions import ProjectorException
 from ..utils.collection import CollectionElement
 from ..utils.projector import ProjectorElement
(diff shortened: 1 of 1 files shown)


openslides/motions/projector.py
```diff
@@ -70,7 +70,9 @@ def get_requirements(self, config_entry):
         else:
             yield motion_block
             yield motion_block.agenda_item
-            yield from motion_block.motion_set.all()
+            for motion in motion_block.motion_set.all():
+                yield motion
+                yield motion.state.workflow
```
normanjaeckel: We do not need to send the workflows multiple times (once per motion) but only once at all.

emanuelschuetze: I use now `yield from Workflow.objects.all()`. Ok?

Issue #2972, opened 2017-02-17
RecursionError after sorting agenda


emanuelschuetze:
@normanjaeckel As discussed, please catch the corner case on server side that it comes to a RecursionError while sorting agenda tree. It is not reproduceable for me but after heavy drag&drop of agenda items one item was a parent of itself. That resulted in a corrupted and unusable database (RecursionError):

```
[..]
  File "/app/openslides/agenda/models.py", line 317, in is_hidden
    (self.parent is not None and self.parent.is_hidden()))
[..]
RecursionError: maximum recursion depth exceeded in comparison
```

emanuelschuetze:
RecursionError is catched on server side now (with #2972). Later we will discuss if we need a better client side sorting ui which can better handles auto updates.

PR #2985, opened 2017-02-17
Added check for hierarchical loops in agenda sort view. See #2972.

normanjaeckel:


emanuelschuetze:
I added an alert error message to template.

PR #2989, opened 2017-02-20
Allowed to export/import initial user password.

emanuelschuetze:


openslides/users/static/templates/users/user-import.html
```diff
@@ -82,7 +83,8 @@ <h3 translate>Preview</h3>
             <th translate>Comment
             <th translate>Is active
             <th translate>Is present
-            <th translate>Is committee</th>
+            <th translate>Is committee
+            <th translate>Initiales Password</th>
```
normanjaeckel: Wrong language

emanuelschuetze: Fixed.

normanjaeckel:
Please add CHANGELOG notice. Then rebase and merge.

PR #2993, opened 2017-02-21
Added motion log entry for creating polls.

emanuelschuetze:
@normanjaeckel How can I add a motion log for poll update in MotionPollViewSet?

### DIFF
--- a/openslides/motions/views.py
+++ b/openslides/motions/views.py
@@ -338,6 +338,7 @@ class MotionViewSet(ModelViewSet):
                 motion.create_poll()
         except WorkflowError as e:
             raise ValidationError({'detail': e})
+        motion.write_log([ugettext_noop('Vote created')], request.user)
         return Response({'detail': _('Vote created successfully.')})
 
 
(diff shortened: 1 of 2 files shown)


normanjaeckel:
You must extend a method like
```python
def update(self, *args, **kwargs):
    result = super().update(*args, **kwargs)
    # do something like poll = self.get_object(); poll.motion.write_log(...)
    return result
```

openslides/motions/views.py
```diff
@@ -357,6 +358,23 @@ def check_view_permissions(self):
...
+        poll.motion.write_log([ugettext_noop('Vote updated')], self.request.user)
+        return result
+
+    def destroy(self, *args, **kwargs):
+        """
+        Customized view endpoint to delete a motion poll.
+        """
+        poll = self.get_object()
+        poll.motion.write_log([ugettext_noop('Vote deleted')], self.request.user)
+        return super(MotionPollViewSet, self).destroy(*args, **kwargs)
```
normanjaeckel: Please use short form `super().destroy(*args, **kwargs)`

openslides/motions/views.py
```diff
@@ -357,6 +358,23 @@ def check_view_permissions(self):
...
+        poll = self.get_object()
+        poll.motion.write_log([ugettext_noop('Vote updated')], self.request.user)
+        return result
+
+    def destroy(self, *args, **kwargs):
+        """
+        Customized view endpoint to delete a motion poll.
+        """
+        poll = self.get_object()
+        poll.motion.write_log([ugettext_noop('Vote deleted')], self.request.user)
```
normanjaeckel: Please do not write log first because the destroy call may fail for whatever reason. You can save motion = poll.motion first, then run super to destroy the poll and then run motion.write_log().

emanuelschuetze:
@normanjaeckel I added timezone support for motion log time and fixed your remark. Please review again.

openslides/motions/views.py
normanjaeckel: Still open !?

emanuelschuetze: Done now, also for update.

emanuelschuetze:
Please review again.

normanjaeckel:
Rebase an merge.

emanuelschuetze:
Rebased.

PR #2995, opened 2017-02-21
Autoupdate on permission change

FinnStutzenstein:
@normanjaeckel Please have a look.

This just covers the case 'A group gets/loses permissions'. In the view the changes are detected, then a signal is send. Currently the user and motion app are connected to this signal. If it's the right permission, all necessary data is returned. The view now sends the data to all affected users.

Is this approach ok (it's currently a bit messy..)? If so, I'll work on.

### DIFF
--- a/openslides/agenda/signals.py
+++ b/openslides/agenda/signals.py
@@ -1,3 +1,4 @@
+from django.apps import apps
 from django.contrib.contenttypes.models import ContentType
 
 from openslides.utils.autoupdate import inform_changed_data
@@ -36,3 +37,17 @@ def listen_to_related_object_post_delete(sender, instance, **kwargs):
         except Item.DoesNotExist:
             # Item does not exist so we do not have to delete it.
             pass
+
+
+def get_permission_change_data(sender, permissions, **kwargs):
+    """
+    Yields all necessary collections if 'agenda.can_see' or
+    'agenda.can_see_hidden_items' permissions changes.
+    """
+    agenda_app = apps.get_app_config(app_label='agenda')
+    for permission in permissions:
+        # There could be only one 'agenda.can_see' and then we want to return data.
+        if (permission.content_type.app_label == agenda_app.label
+                and permission.codename in ('can_see', 'can_see_hidden_items')):
+            yield from agenda_app.get_startup_elements()
+            break
--- a/openslides/motions/signals.py
+++ b/openslides/motions/signals.py
@@ -1,3 +1,4 @@
+from django.apps import apps
 from django.utils.translation import ugettext_noop
 
 from .models import State, Workflow
@@ -102,3 +103,15 @@ def create_builtin_workflows(sender, **kwargs):
     state_2_2.next_states.add(state_2_3, state_2_4, state_2_5, state_2_6, state_2_7, state_2_8, state_2_9)
     workflow_2.first_state = state_2_1
     workflow_2.save()
+
+
+def get_permission_change_data(sender, permissions=None, **kwargs):
+    """
+    Returns all necessary collections if a 'can_see' permission changes.
+    """
+    motion_app = apps.get_app_config(app_label='motions')
+    for permission in permissions:
+        # There could be only one 'motion.can_see' and then we want to return data.
+        if permission.content_type.app_label == motion_app.label and permission.codename == 'can_see':
+            return motion_app.get_startup_elements()
+    return None
--- a/openslides/motions/signals.py
+++ b/openslides/motions/signals.py
@@ -105,13 +105,12 @@ def create_builtin_workflows(sender, **kwargs):
     workflow_2.save()
 
 
-def get_permission_change_data(sender, permissions=None, **kwargs):
+def get_permission_change_data(sender, permissions, **kwargs):
     """
-    Returns all necessary collections if a 'can_see' permission changes.
+    Yields all necessary collections if 'motions.can_see' permission changes.
     """
-    motion_app = apps.get_app_config(app_label='motions')
+    motions_app = apps.get_app_config(app_label='motions')
     for permission in permissions:
-        # There could be only one 'motion.can_see' and then we want to return data.
-        if permission.content_type.app_label == motion_app.label and permission.codename == 'can_see':
-            return motion_app.get_startup_elements()
-    return None
+        # There could be only one 'motions.can_see' and then we want to return data.
+        if permission.content_type.app_label == motions_app.label and permission.codename == 'can_see':
+            yield from motions_app.get_startup_elements()
--- a/openslides/users/views.py
+++ b/openslides/users/views.py
@@ -6,7 +6,9 @@ from django.utils.encoding import force_text
 from django.utils.translation import ugettext as _
 
 from ..core.config import config
+from ..core.signals import permission_change
 from ..utils.auth import anonymous_is_enabled, has_perm
+from ..utils.autoupdate import send_collections_to_users, inform_changed_data
 from ..utils.collection import CollectionElement
 from ..utils.rest_api import (
     ModelViewSet,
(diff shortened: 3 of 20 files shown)


openslides/users/views.py
```diff
@@ -162,6 +164,55 @@ def destroy(self, request, *args, **kwargs):
         self.perform_destroy(instance)
         return Response(status=status.HTTP_204_NO_CONTENT)
 
+    def update(self, request, *args, **kwargs):
```
normanjaeckel: This is too much here. It should be done in a more structured way. Proposal: Send out the signal if any group changes. Do not send just the perms one by one but all new perms at once through the signal. Then use inform_changed_data to send out all received objects. Of course this is a little bit more load but I think groups are not changed so often.

For the case of removed perms we have to discuss with @emanuelschuetze: Do we really want that the client suddenly removes stuff from the view someone is heavily reading at the moment? Think about someone reads a motion detail view with a great poem. Now the admin removes the can_see perm. Boom. I don't like this.

emanuelschuetze: As discussed with @normanjaeckel 
First, please fix the 'add' case only (without complex diff code).
Let us move the 'remove' case to 2.2.

FinnStutzenstein:
@normanjaeckel Please check again. I'm not a fan of just sending out all permissions to the apps. This overload isn't necessary and causes a lot of traffic especially with huge motions. So the diff should not be removed. See this example:
Group 'test' has 'agenda.can_see', 'motion.can_see'.
The request sends 'test' with permissions 'agenda.can_see', 'motion.can_see', 'users.can_see_names'.
With diff-code: Just 'users.can_see_names' is send out; without diff-code every permission (too much data).

With the diff-code the server also can determinate what permissions are added or removed. This is important for handling the remove-case where every permission has to be checked carefully.

Why I should use `inform_changed_data` here? There is no change in data (except the group, but this is handled by the `super().update()`.) So I introduced `inform_changed_permissions` that handles the signal handling.



The remove-case could be done later.

openslides/core/static/js/core/base.js
```diff
@@ -262,6 +262,9 @@ angular.module('OpenSlidesApp.core', [
                 _.forEach(list, function(data) {
                     // Uncomment this line for debugging to log all autoupdates:
                     // console.log("Received object: " + data.collection + ", " + data.id);
+                    if (data.collection !== 'core/config') {
+                        console.log(data.action + " object: " + data.collection + ", " + data.id, data);
```
normanjaeckel: Don't forget to remove this. :-)

openslides/motions/signals.py
```diff
@@ -102,3 +103,12 @@ def create_builtin_workflows(sender, **kwargs):
     state_2_2.next_states.add(state_2_3, state_2_4, state_2_5, state_2_6, state_2_7, state_2_8, state_2_9)
     workflow_2.first_state = state_2_1
     workflow_2.save()
+
+
+def get_permission_change_data(sender, permissions=None, **kwargs):
+    motion_app = apps.get_app_config(app_label='motions')
```
normanjaeckel: A docstring here would be nice.

openslides/users/views.py
```diff
@@ -169,6 +170,20 @@ def destroy(self, request, *args, **kwargs):
...
+        # Collect old and new permissions to get the difference.
+        permission_names = request.data['permissions']
+        if isinstance(permission_names, str):
+            permission_names = [permission_names]
+
+        permissionSerializer = PermissionRelatedField(read_only=True)
+        given_permissions = [
+            permissionSerializer.to_internal_value(data=perm) for perm in permission_names]
+        group = Group.objects.get(pk=int(kwargs['pk']))
+        inform_changed_permissions(group, given_permissions)
```
normanjaeckel: I think the information should be send after the update was successful. So first run super(). Then run the code to check for objects to send.

openslides/utils/autoupdate.py
```diff
@@ -187,6 +188,21 @@ def send_data(message):
                     {'text': json.dumps(output)})
 
 
+def send_collections_to_users(collections, users_ids):
```
normanjaeckel: Maybe we can combine this code with the one used in send_data() so that we do not have duplicates?

openslides/utils/autoupdate.py
```diff
@@ -253,6 +269,44 @@ def inform_deleted_data(*args, information=None):
...
+        This function informs the autoupdate system about changed permissions
+        and manage the necessary data to be send to affected clients.
+
+        given_permissions is a list of permissions retrieved from the client. First
+        the difference between these permissions and the old group permissions is
+        determinated to get the new (added) permissions.
+        Second all apps are queried to give the new required collections. These
+        are send out to all members of the group.
+        """
+        old_permissions = group.permissions.all()
```
normanjaeckel: I would like to have permission related code inside user app, not here. The code in the user app can do the signal stuff and then call a short helper here.

I once more suggest to use inform_changed_data because we gain the support of the extra channel way (internal channel 'autoupdate.send_data') and can reuse all code without repeating ourselves).

The diff is ok. Maybe you can describe the diff function a little bit more because `first-second` is not much intuitive. (Ah, I think the problem is that you mean `diff = first – second`. I didn't realize this. Also if first is 0 and second is 2, we do not get a result like –2, do we?)

Then please check not to use too much database queries. Use CollectionElement.from_values() instead of Model.objects.get() and thelike.

openslides/users/views.py
```diff
@@ -183,8 +184,26 @@ def update(self, request, *args, **kwargs):
...
+
+            # some permissions are added
+            if(len(new_permissions) > 0):
+                signal_results = permission_change.send(None, action='added', permissions=new_permissions)
+                collections_to_send = []
+                for (receiver, signal_collections) in signal_results:
+                    if signal_collections:
+                        collections_to_send.extend(signal_collections)
+                #send_collections_to_users(collections_to_send, [user.id for user in group.user_set.all()])
+                # inform_changed_data(what?)
```
FinnStutzenstein: @normanjaeckel I need your help here: in `collections_to_send` are all collections all users in `group.user_set.all()` need. How I have to do this with `inform_changed_data`, especially only for the affected users?

normanjaeckel: Ok, if you want to send the objects only to some users then you can't use inform_changed_data. I thought it would be easier just to send some objects to all users.

FinnStutzenstein: Hmm, this is ok for me; whats the best way to send all collections via inform changed data?

openslides/utils/autoupdate.py
FinnStutzenstein: Todo: this can be removed if using inform_changed_data

FinnStutzenstein:
Todo:

 - use inform_change_data
 - remove send_collections_to_users
 - make Issue for remove-case after merge.

emanuelschuetze:
@normanjaeckel Can you please fix the last autoupdate todos mentioned by @FinnStutzenstein. Thanks.

normanjaeckel:
@emanuelschuetze Ready for testing.

normanjaeckel:
@FinnStutzenstein May check this if he likes?!

FinnStutzenstein:
@normanjaeckel I do like :+1: 
I added the missed agenda app (also sending topics on agenda.can_see now) and fixed the user can_see_name permision.

openslides/agenda/signals.py
```diff
@@ -36,3 +37,16 @@ def listen_to_related_object_post_delete(sender, instance, **kwargs):
...
+    """
+    Yields all necessary collections if 'agenda.can_see' or
+    'agenda.can_see_hidden_items' permissions changes.
+    """
+    agenda_app = apps.get_app_config(app_label='agenda')
+    for permission in permissions:
+        # There could be only one 'agenda.can_see' and then we want to return data.
+        if (permission.content_type.app_label == agenda_app.label
+                and permission.codename in ('can_see', 'can_see_hidden_items')):
+            yield from agenda_app.get_startup_elements()
```
normanjaeckel: Please add break after this line because we do not want to send the same elements twice if both permissions are added at once.

FinnStutzenstein: Done.

FinnStutzenstein:
@emanuelschuetze @normanjaeckel all Done.

Issue #2984, opened 2017-02-17
Do not allow users to update some fields of themselves.


normanjaeckel:
At the moment any user can update himself and then change more than just his name. We have to restrict this. See #2973.

FinnStutzenstein:
Coupled with issue #1871 . Is this possible to fix for 2.1?

emanuelschuetze:
As discussed with @normanjaeckel 
Normal user should be allowed to update username and about me only.
@normanjaeckel Please add your commit to #3001.

emanuelschuetze:
Done with #3001

PR #3001, opened 2017-02-22
Reduced user profile form

emanuelschuetze:
A normal user should no longer edit the fields 'title, first/last name, structure level and number'. 
Reason: A user can changed and misused his/her name and structure level (e.g. use fake name) - and these fields are published on list of speakers, submitter, candidate etc.

@normanjaeckel If you accept this we should discuss if we prevent to change these fields also on server side.


normanjaeckel:
Yes, this has to be done also on server side. See #2984.

normanjaeckel:
Please test with different permissions.

Issue #3004, opened 2017-02-23
Normal user cannot save amendment


emanuelschuetze:
To reproduce:

* Create new (empty) group with permissions "motions.can_see" and "motions.can_create"
* Create new user and add this new group to user.
* Create new motion with admin as submitter.
* Activate amendments in config.
* Login as user and go to motion detail view.
* Click "new amendment" button.
* Save amendment form.
=> 403 (Forbidden) -> "Sie sind nicht berechtigt, diese Aktion durchzuführen."

emanuelschuetze:
Fixed with #3005

PR #3005, opened 2017-02-24
Fixing create view for amendments (fixing #3004)

FinnStutzenstein:


openslides/motions/views.py
```diff
@@ -87,12 +87,25 @@ def create(self, request, *args, **kwargs):
...
+                    whitelist.extend([
+                        'parent_id',
+                        'category_id',      # This will be set to the matching
+                        'motion_block_id',  # values from parent_motion.
+                    ])
+                    request.data['category_id'] = parent_motion.category.id
+                    request.data['motion_block_id'] = parent_motion.motion_block.id
             for key in request.data.keys():
                 if key not in whitelist:
                     # Non-staff users are allowed to send only some data.
```
FinnStutzenstein: @emanuelschuetze can you put here a `print("{} not in whitelist".format(key))`? This may give us the wrong key.

FinnStutzenstein:
@emanuelschuetze Please retest.

emanuelschuetze:
> Please add CHANGELOG notice.

Sorry, but bugfixing is not a entry in Changelog... we already have a generic entry "Reimplemented amendments"... So won't fix. I'll merge it.

normanjaeckel:
Ok. (I thought about this special behavior when a new motion is created (that it is set to the same category and block)).

PR #3017, opened 2017-02-27
Fixed use of PATCH and PUT. Fixed #1871.

### LINKED ISSUE #1871: Fix serializers for PUT requests

normanjaeckel:
@emanuelschuetze Please test motion update and change recommendation update for staff and for non staff users.

Still not solved: Non stuff users are not really able to change motion comments, are they?

### DIFF
--- a/openslides/assignments/views.py
+++ b/openslides/assignments/views.py
@@ -235,7 +235,7 @@ class AssignmentPollViewSet(UpdateModelMixin, DestroyModelMixin, GenericViewSet)
     """
     API endpoint for assignment polls.
 
-    There are the following views: update and destroy.
+    There are the following views: update, partial_update and destroy.
     """
     queryset = AssignmentPoll.objects.all()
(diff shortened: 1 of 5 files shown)


emanuelschuetze:
> Still not solved: Non stuff users are not really able to change motion comments, are they?

Right, see #3036

PR #3026, opened 2017-03-01
Several bug fixes in the diff

CatoTH:
Fixes the following problems:
- An empty line with line number was created at the end of a motion in the Diff View, if the last line was modified
- The line number is now inserted before INS, if INS is the first element of a line.
- Some messy situations when a whole paragraph was deleted without any replacement are resolved.

### DIFF
--- a/openslides/motions/static/js/motions/base.js
+++ b/openslides/motions/static/js/motions/base.js
@@ -278,11 +278,16 @@ angular.module('OpenSlidesApp.motions', [
                         html = lineNumberingService.insertLineNumbers(this.getVersion(versionId).text, lineLength);
 
                     var data = diffService.extractRangeByLineNumbers(html, maxLine, null);
-                    html = data.outerContextStart + data.innerContextStart +
-                        data.html +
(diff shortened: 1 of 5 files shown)


openslides/motions/static/js/motions/base.js
```diff
@@ -278,6 +278,12 @@ angular.module('OpenSlidesApp.motions', [
                         html = lineNumberingService.insertLineNumbers(this.getVersion(versionId).text, lineLength);
 
                     var data = diffService.extractRangeByLineNumbers(html, maxLine, null);
+
+                    if (data.html === "") {
+                        // Prevents empty lines at the end of the motion
+                        return "";
```
normanjaeckel: For better readability is it possible to write as following?
```javascript
if (data.html !== '') {
    html = ...
} else {
   html = '';
}
return html;
```

emanuelschuetze: Done and rebased

emanuelschuetze:
Remark is fixed and rebased.

PR #3051, opened 2017-03-04
Excluded already elected persons for new ballot (Fixed #2142)

### LINKED ISSUE #2142: Wrong number of eligible persons
- create an election and say that there have 3 persons to be elected
- add 5 candidates to the candidate list
- in the first voting round mark only 1 persion as elected
- start the second voting round
- print the ballot paper for this voting

result: OpenSlides adds this "_2. Wahlgang, 4 Kandidaten/innen, 3 verfügbare Posten_".
So the number of candidates is correct but since one person was already elected, there are now only 3 available position not 3

emanuelschuetze:
- Reorder assignment ballot tabs in detail template.
- Removed logic to set phase automatically to 'finished' if
  elected candidates reached, because autoupdate makes it unreliable.

### DIFF
--- a/openslides/assignments/static/js/assignments/site.js
+++ b/openslides/assignments/static/js/assignments/site.js
@@ -405,6 +405,7 @@ angular.module('OpenSlidesApp.assignments.site', [
     '$scope',
     '$http',
     '$filter',
+    '$timeout',
     'filterFilter',
     'gettext',
     'ngDialog',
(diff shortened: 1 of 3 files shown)


openslides/assignments/static/js/assignments/site.js
```diff
@@ -527,20 +528,27 @@ angular.module('OpenSlidesApp.assignments.site', [
...
         // create new ballot
         $scope.createBallot = function () {
-            $scope.activeTab = 0;
             $http.post('/rest/assignments/assignment/' + assignment.id + '/create_poll/').then(
                 function (success) {
                     $scope.alert.show = false;
                     if (assignment.phase === 0) {
                         $scope.updatePhase(1);
                     }
+                    $timeout(function() {
```
FinnStutzenstein: Is the timeout here really necessary? This should be in the angular context. If so, check line 549, if a timeout is required there, too.

emanuelschuetze: Yes, we need a timeout to select a recently created tab, see https://github.com/angular-ui/bootstrap/issues/5656. Otherwise the last tab is not selected. 
For delete tab the timeout is not required (I tested it).

normanjaeckel: Can you please add a short not with a link to this issue? Maybe they will fix it later (or maybe not if ui-bootstrap is a dead project (in favor of Angular 2).

emanuelschuetze: Done.

PR #3064, opened 2017-03-07
Changes required for Voting plugin

jwinzer:
These changes are necessary for my Voting plugin it work. I believe they are also required by the VoteCollector plugin


openslides/core/static/js/core/base.js
```diff
@@ -808,6 +808,7 @@ angular.module('OpenSlidesApp.core', [
                                 id: value.id,
                             };
                         } else if (
+                            value.name != 'voting/prompt' &&
```
normanjaeckel: This is a non-core feature. So I think we need a new interface for plugins to use this method properly. Please create a new issue for this and add a short comment here about the problem and refer to the new issue.

openslides/motions/apps.py
```diff
@@ -39,7 +39,8 @@ def ready(self):
         router.register(self.get_model('Workflow').get_collection_string(), WorkflowViewSet)
         router.register(self.get_model('MotionChangeRecommendation').get_collection_string(),
                         MotionChangeRecommendationViewSet)
-        router.register('motions/motionpoll', MotionPollViewSet)
+        #router.register('motions/motion-poll', MotionPollViewSet)
```
normanjaeckel: Please remove the uncommented line.

openslides/motions/views.py
```diff
@@ -395,9 +395,9 @@ def destroy(self, *args, **kwargs):
         """
         Customized view endpoint to delete a motion poll.
         """
-        result = super().destroy(*args, **kwargs)
         poll = self.get_object()
         poll.motion.write_log([ugettext_noop('Vote deleted')], self.request.user)
+        result = super().destroy(*args, **kwargs)
```
normanjaeckel: Please do not write the log before the poll was deleted. Maybe this line has to be one line earlier.

normanjaeckel:
My remarks are still open and I think @emanuelschuetze would not like the changes in the translation files.

openslides/motions/static/templates/motions/slide_motion.html
```diff
@@ -69,7 +69,7 @@ <h3 ng-if="motion.polls.length > 0 && motion.polls[0].has_votes" translate>Votin
   <!-- Title -->
   <div id="title">
     <h1>{{  motion.agenda_item.getTitle() || motion.getTitle() }}</h1>
-    <h2>
+    <h2 ng-if="config('motions_slide_has_motion_id')">
```
normanjaeckel: Where is this config value defined?

openslides/core/static/js/core/base.js
jwinzer: See #3112

openslides/motions/apps.py
jwinzer: Right!

openslides/motions/views.py
jwinzer: You are right.

openslides/motions/static/templates/motions/slide_motion.html
jwinzer: Value is not defined yet. I just needed a way to exclude the motion subheader in the slide view. I'm not sure if @emanuelschuetze will approve this change request

openslides/core/static/js/core/base.js
FinnStutzenstein: Whats about changing this from a blacklist to a white list allowing just the named forms (like `motion/motions`) to return an `EditForm`?
But we need a hook for plugins here:
1) We have a blacklist: Plugins need to set them on there, because if this is not possible, you will get the error you described. (!)
2) We have a whitelist: A Plugin may have EditForms, so they want to hook those in here.
@emanuelschuetze ?

normanjaeckel: I prefer a whitelist. Later we can add also a check against the plugin api (something where plugins can register their collection strings for this function here).

openslides/motions/views.py
```diff
@@ -395,8 +395,8 @@ def destroy(self, *args, **kwargs):
         """
         Customized view endpoint to delete a motion poll.
         """
-        result = super().destroy(*args, **kwargs)
         poll = self.get_object()
+        result = super().destroy(*args, **kwargs)
```
emanuelschuetze: @jwinzer @normanjaeckel Why we have to change the order here? Is it really required?

openslides/core/static/js/core/base.js
emanuelschuetze: I'll do a whiteliste in a new pull request

openslides/motions/views.py
normanjaeckel: I do not know. The change does not look required.

emanuelschuetze:
New pullrequest created, see #3127

openslides/motions/views.py
jwinzer: If you call destroy first self.get_object() will fail since the object was destroyed.

emanuelschuetze: Ok, fixed in #3127.

PR #3072, opened 2017-03-08
Fixes broken diff test cases

CatoTH:
All test cases work again. One failure was an actual bug, which did not occur in the real environment, though, as it was only an issue if plain text without a wrapping HTML-element was used for the diff.

Now it should be possible to add the Karma-based tests to Travis.

### DIFF
--- a/openslides/motions/static/js/motions/diff.js
+++ b/openslides/motions/static/js/motions/diff.js
@@ -760,17 +760,17 @@ angular.module('OpenSlidesApp.motions.diff', ['OpenSlidesApp.motions.lineNumberi
         };
 
         this._tokenizeHtml = function (str) {
-            var splitArrayEntries = function (arr, by, prepend) {
+            var splitArrayEntriesEmbedSeparator = function (arr, by, prepend) {
                 var newArr = [];
                 for (var i = 0; i < arr.length; i++) {
-                    if (arr[i][0] == '<' && (by == " " || by == "\n")) {
+                    if (arr[i][0] === '<' && (by === " " || by === "\n")) {
                         // Don't split HTML tags
                         newArr.push(arr[i]);
                         continue;
                     }
 
                     var parts = arr[i].split(by);
-                    if (parts.length == 1) {
+                    if (parts.length === 1) {
                         newArr.push(arr[i]);
                     } else {
                         var j;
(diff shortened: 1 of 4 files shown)


emanuelschuetze:
@CatoTH Travis failed with error:
`JSHint failed for: openslides/motions/static/js/motions/diff.js`

CatoTH:
@emanuelschuetze Travis-Error is fixed.

Now this PR also fixes the JS-Exception:
TypeError: Failed to execute 'insertBefore' on
'Node'

emanuelschuetze:
Retested. Now the new words are shown but the whole line is marked as "added" (green), see example:
change-recommendation1

CatoTH:
I don't really see this as a regression. Inserted paragraphs were subject to line numbering before as well. The inserted line break happens because the added words exceeds the maximum line length.
I can disable line breaks in inserted paragraphs, of cause - but I'm not sure this really improves the readability of the text... or did I miss the problem?

I can't reproduce the bug of the word vanishing in the PDF. Can you maybe send me your database with the bug?

CatoTH:
To illustrate the problem with cases like that:
We have two tokenized strings:
Before:
``['<P>', 'rief ', 'sie ', 'alle ', 'sieben ', 'herbei ', 'und ', 'sprach ', ''liebe ', 'Kinder, ', 'ich ', 'will ', 'hinaus ', 'in ', 'den ', 'Wald, ', 'seid', '</P>']``
After:
``['<P>', 'rief ', 'sie ', 'alle ', 'sieben ', 'herbei ', 'und ', 'sprach ', ''liebe ', 'Kinder, ', 'ich ', 'will ', 'hinaus ', 'in ', 'den ', 'Wald, ', 'seid ', 'Noch', '</P>', '<P>', 'Test ', '123', '</P>']``

It seems somewhat impossible to me to teach a diff algorithm to leave the middle ``</P>`` untouched, while inserting the words before and the ones after.

The possible solution would be:
- Split the "Before" and "After" Strings into arrays of block elements, each. In the example above, that would be one block element for "Before", and two for "After".
- Try to match the block elements by similarity, so we can guess which one has only by slightly altered and which ones have been inserted / deleted.
- Perform the inline diff only on those lines that are slightly altered.

emanuelschuetze:
As discussed with @CatoTH this behavior is ok (no inline diff for paragraphs).

CatoTH:
@emanuelschuetze not for today :)

emanuelschuetze:
Just for not: Sill open is the regression which you already fixed in #3044. Any idea why this is broken again?

emanuelschuetze:
@normanjaeckel Please merge.

CatoTH:
@emanuelschuetze how can I see the regression regarding STRIKE? The test case is still passing...

emanuelschuetze:
@CatoTH 
> how can I see the regression regarding STRIKE? The test case is still passing...

show a removed paragraph in pdf (the line numbers are also removed - this was fixed with #3044), see example:
change-recommendation-pdf

PR #3118, opened 2017-03-21
Allow style attributes for about field in user detail template (Fixed #3081)

### LINKED ISSUE #3081: regression: images size is ignored
- click on your user name and chosse then to edit your profile
- add some text and an image to the "About me" field
- for example use this image: https://upload.wikimedia.org/wikipedia/commons/0/09/Kubach_crystal_cave1.JPG
- since this image is quite large, set in the editor of OS the image size to 200 x 157 pixels:
localhost 8000 2017-03-12 23-25-41

result: after saving the image appears with its original size:
localhost 8000 2017-03-12 23-27-58

Therefore the image feature is de-facto unusable to be used! Al text edit fields of OS are affected.

emanuelschuetze:


openslides/users/static/templates/users/user-detail.html
```diff
@@ -33,7 +33,7 @@ <h2 translate>Participant</h2>
       <label translate>Participant number</label>
         {{ user.number }}
       <label translate>About me</label>
-        <div ng-bind-html="user.about_me"></div>
+        <div ng-bind-html="user.about_me | trusted"></div>
```
normanjaeckel: Uuuh. Do you really want users to add whatever they want? Everybody could get an exploit here!?

emanuelschuetze: We use bleach for validate html code, see [utils/validate.py](https://github.com/OpenSlides/OpenSlides/blob/aade91cead8d92db1e64619a6914c5ba4203b68a/openslides/utils/validate.py). Script tags are not possible, e.g..

normanjaeckel: Ok.

PR #3119, opened 2017-03-21
Minor fixes

emanuelschuetze:
- Added motion change recommendations to full text search.
- Show csv/docx export entries in list views for manager only.

### DIFF
--- a/openslides/agenda/static/templates/agenda/item-list.html
+++ b/openslides/agenda/static/templates/agenda/item-list.html
@@ -104,7 +104,7 @@
             </a>
           </li>
           <!-- CSV export -->
-          <li>
+          <li os-perms="agenda.can_manage">
             <a href="" id="downloadLinkCSV"
               ng-click="csvExport()">
               <i class="fa fa-file-text-o"></i>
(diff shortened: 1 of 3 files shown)


openslides/agenda/static/templates/agenda/item-list.html
```diff
@@ -104,7 +104,7 @@ <h1 translate>Agenda</h1>
             </a>
           </li>
           <!-- CSV export -->
-          <li>
+          <li os-perms="motions.can_manage">
```
normanjaeckel: We should not couple the agenda app to the motion app. Agenda is a dependency of motion, not vice versa.

openslides/motions/static/js/motions/base.js
```diff
@@ -408,6 +408,18 @@ angular.module('OpenSlidesApp.motions', [
                     if (!foundSomething && motion.category && motion.category.name.match(new RegExp(searchquery, 'i'))) {
                         foundSomething = true;
                     }
+
+                    // search for change recommendation
+                    if (!foundSomething) {
+                        var recommendations = MotionChangeRecommendation.filter({
+                            where: {motion_version_id: this.active_version}
+                        });
+                        _.forEach(recommendations, function(recommendation) {
```
normanjaeckel: Please use lodash `_.some` so that the loop aborts on first match.

openslides/agenda/static/templates/agenda/item-list.html
emanuelschuetze: Sorry, copy/paste error. Fixed now.

openslides/motions/static/js/motions/base.js
emanuelschuetze: Done.

PR #3122, opened 2017-03-21
Example data

normanjaeckel:
@emanuelschuetze First test for you. See new README about how it works. I will think about speed improvement later.

### DIFF
--- a/tests/example_data_generator/management/commands/create-example-data.py
+++ b/tests/example_data_generator/management/commands/create-example-data.py
@@ -1,0 +1,226 @@
+from textwrap import dedent
+
+from django.contrib.auth.hashers import make_password
+from django.core.management.base import BaseCommand, CommandError
+from django.utils.crypto import get_random_string
+
+from openslides.agenda.models import Item
+from openslides.assignments.models import Assignment
+from openslides.motions.models import Motion
+from openslides.topics.models import Topic
+from openslides.users.models import Group, User
+
+MOTION_NUMBER_OF_PARAGRAPHS = 4
+
+LOREM_IPSUM = [
+    """\
+    <p>Lorem ipsum dolor sit amet, consectetur adipisici elit, sed eiusmod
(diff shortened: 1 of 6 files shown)


tests/example_data_generator/management/commands/create-example-data.py
```diff
@@ -0,0 +1,165 @@
...
+    help = 'Create example data for OpenSlides.'
+
+    chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '-t', '--topics',
+            default=100,
+            type=int,
+            help='Number of topics to be created.',
```
emanuelschuetze: It would be helpful to get the default number in the help output. Also for other arguments.

normanjaeckel:
You can set the flags to zero, e. g. `-t 0 -m 200 -a 0 -u 0 0`. Another possibility is that I set all default values to zero and you have to set each flag manually.

FinnStutzenstein:
I would prefer a --only or --zero that sets all not given values to 0.

emanuelschuetze:
> I would prefer a --only or --zero that sets all not given values to 0.

+1

normanjaeckel:
Only flag added. Topics, assignments and default users are created much faster. Ready for next review. I do not have any idea how to make motions and staff users less slow. @emanuelschuetze Can you live with the speed?

emanuelschuetze:
Successfully tetested: 
Speed is ok. Only flag works as expected.

General question: 
Why I have to restart the OpenSlides server if I run the example data script in an other console? (I use sqlite). Browser reload does not get the new created data. Only after restart server the new data are visible in browser.

BTW: travis failed (in `tests.unit.users.test_models.UserManagerTest`)

Please rebase and merge.

normanjaeckel:
Server restart is required because the cache has to be updated. For Redis solutions it should be sufficient to clear the Redis cache via command line (maybe this is required though).

PR #3125, opened 2017-03-22
Beginning: passing line numbers through the diff

CatoTH:


openslides/motions/static/js/motions/diff.js
```diff
@@ -842,13 +849,11 @@ angular.module('OpenSlidesApp.motions.diff', ['OpenSlidesApp.motions.lineNumberi
 
             if (out.n.length === 0) {
                 for (i = 0; i < out.o.length; i++) {
-                    //this._outputcharcode('del', out.o[i]);
```
emanuelschuetze: Can you remove this line? You could add it yourself for debugging.

normanjaeckel: This line is already removed, isn't it?

openslides/motions/static/js/motions/diff.js
```diff
@@ -816,15 +816,22 @@ angular.module('OpenSlidesApp.motions.diff', ['OpenSlidesApp.motions.lineNumberi
...
         };
 
-        this._outputcharcode = function (pre, str) {
+        this._outputcharcode = function (str) {
             var arr = [];
             for (var i = 0; i < str.length; i++) {
                 arr.push(str.charCodeAt(i));
             }
-            console.log(str, pre, arr);
+            console.log(arr.join(", "));
```
normanjaeckel: But this line should be removed. :-)

openslides/motions/static/js/motions/diff.js
CatoTH: My change removes this line. If I remove this changing line, this will result in this line to remain in the file :)

openslides/motions/static/js/motions/diff.js
CatoTH: This line was the whole point of this function, as this was just a helper function to debug different kind of whitespaces and other non-obvious-to-spot differences between two strings. But ok, I removed the whole function.

openslides/motions/static/js/motions/diff.js
emanuelschuetze: Ok, right. Ok for me now.
@normanjaeckel ping

PR #3128, opened 2017-03-22
Fixed error with multiple workers on motion create. Fixed #3062.

### LINKED ISSUE #3062: https://github.com/OpenSlides/OpenSlides/blob/6b74bf635d88ce7693bb5384dff83acaf824b05f/openslides/motions/models.py#L195
I'm testing OpenSlides in big mode (runserver with 4 workers, postgres, daphne). When importing some motions about every 4th motion is imported. For the others I get:
```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "motions_motion_identifier_key"
DETAIL:  Key (identifier)=(122) already exists.
```
It seems like [here](https://github.com/OpenSlides/OpenSlides/blob/6b74bf635d88ce7693bb5384dff83acaf824b05f/openslides/motions/models.py#L195) the transaction is very importent.

When using Geiss this works, but for the reason that with `manage.py start --use-geiss` only one worker is started. An option like `--threads 4` will be usefull for big assemblies.

normanjaeckel:


normanjaeckel:
Ready to merge (if it works).

emanuelschuetze:
I added a new commit for motion update (can be tested by permitting multiple motions without identifier in complex workflow).

emanuelschuetze:
I tested the import of 50 motions in two variants:
1. with 4 workers*, postgres, daphne WITHOUT redis**: 100% successfully imported
2. with 4 workers*, postgres, daphne WITH redis**: import failed for some motions!!

@ostcar For (2) we need your help! Do you have an idea why our fix does not work with redis?

*) I patched the `openslides/__main__.pyL241`: nothreading=True => `manage.py start` runs 4 workers
**) Set use_redis to false/true in settings.py.

normanjaeckel:
We need another approach. I will open a new pull request tomorrow.

Issue #3140, opened 2017-03-24
security issue: user without permission needs to set title, text, reason for motion


FinnStutzenstein:
Scenario: We need a user with just motions.can_see and motions.can_see_and_manage_comments. And at least on comment in the config as well as at least a motion.

The user loggs in and edit the comment. On save the motion text (so the reason, why not the title??) is deleted.

This is a general issue in design: The client needs to send these infos on every save. See this [special save function](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/motions/static/js/motions/site.js#L849) in the list controller. This also needs to be send, if a user want to save comments. [See here](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/motions/static/js/motions/motion-services.js#L189). With PR #3061 I forbid users who are not submitter or have motions.can_manage to set the title/text/reason. This results in setting the text and reason to empty string. --> The server somehow needs the text/reason, but this could not always be delivered by the client.

We need a server side solution here: if no title/text/reason is given (may because the user isn't allowed to set those) the server cannot clear this. It should get the title/text/reason from the latest version.
So we can just do a `Motion.save(motion)` on client side without always passing the title/text/reason.

normanjaeckel:
I like git blame because we can blame each other for introducing bugs ... ;-).

The problem is [this line](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/motions/serializers.py#L314) and the following one. Here we validate some HTML and set the validated_data for text and reason to an empty string if no data are given. Later these empty strings are used to update the version [here](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/motions/serializers.py#L373).

It was https://github.com/OpenSlides/OpenSlides/commit/d55d3742110e04075075d955d2a8f24a8b2a889c that introduced the bug.

So the solution is not to set `data['text']` to an empty string if `data['text']` is given. The same for reason.

I do not know whether this problem also occurs for ProjectorMessage, MotionChangeRecommendation or Topic. @FinnStutzenstein Can you have a short look at this?

FinnStutzenstein:
I could do this Thuesday; if you are faster please do it.

Ok, thats right, we have to change this validations. But what makes me stumble is the following: The `save` method in the js controller exists a long time longer than we have the html validation. There must be a reason that (at least 8 month) we have this save method to fix something like deleting the text.
If you make a PR test this carefully.

normanjaeckel:
[...]

PR #3141, opened 2017-03-24
Hotfix for issue #3140

FinnStutzenstein:


normanjaeckel:
This is not the solution and not even a fix because if the client does not send any text and reason, text and reason are reset to an empty string. See my comment in #3140.

PR #3156, opened 2017-03-27
Disable edit button and hide remove button if election is finished.

### LINKED ISSUE #3099: Possible to start new election round for finished elections
- go to an election that already took place. Its phase is therefore "Finished"
- click on "new ballot"

result: a new ballot is started. This must be impossible because after an election has been finished be the assembly leaders it cannot be changed. For political parties this is even a restriction in the German laws for political parties.

It is OK that one can reset the state from "Voting" to "Searching for candidates" because participants are allowed to enter a motion to change the voting process during the voting phase.
So to fix this bug, only the state "Finished" must not be resettable. If something went wrong one is forced to make a new election. Maybe other users of OS don't have such a strict rule. If so, there should be an option to block resetting the "Finished" phase.

emanuelschuetze:
(Fixed #3099)

### DIFF
--- a/openslides/assignments/static/templates/assignments/assignment-detail.html
+++ b/openslides/assignments/static/templates/assignments/assignment-detail.html
@@ -80,7 +80,7 @@
     <div class="white-space-pre-line">{{ assignment.description }}</div>
   </div>
 
-  <div ng-if="assignment.phase != 2">
+  <div ng-if="assignment.phase !== 2">
     <h3 translate>Candidates</h3>
     <div ui-tree="treeOptions" ng-if="assignment.assignment_related_users.length"
         data-empty-placeholder-enabled="false">
@@ -123,7 +123,7 @@
   </div>
 
   <h3 translate>Election result</h3>
-  <button os-perms="assignments.can_manage" ng-show="assignment.phase != 2" ng-click="createBallot()"
+  <button os-perms="assignments.can_manage" ng-show="assignment.phase !== 2" ng-click="createBallot()"
       class="btn btn-default btn-sm">
     <i class="fa fa-bar-chart fa-lg"></i>
     <translate>New ballot</translate>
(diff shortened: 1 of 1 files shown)


openslides/assignments/static/templates/assignments/assignment-detail.html
```diff
@@ -133,7 +133,7 @@ <h3 translate>Election result</h3>
         index="$index" heading="{{ 'Ballot' | translate }} {{ $index + 1 }}">
       <div ng-controller="AssignmentPollDetailCtrl">
         <!-- action buttons -->
-        <div class="pull-right" os-perms="assignments.can_manage">
+        <div class="pull-right" os-perms="assignments.can_manage" ng-show="assignment.phase != 2">
```
normanjaeckel: You can use $scope.phases I think: `ng-show="assignment.phase != _.find(phases, {'display_name': 'Finished'}).value`

So we do not have a hard coded value here.

openslides/assignments/static/templates/assignments/assignment-detail.html
```diff
@@ -158,6 +158,7 @@ <h3 translate>Election result</h3>
           </a>
           <!-- Edit -->
           <button ng-click="editPollDialog(poll, $index+1)"
+              ng-class="{ 'disabled': assignment.phase == 2 }"
```
normanjaeckel: Here the same.

openslides/assignments/static/templates/assignments/assignment-detail.html
emanuelschuetze: I tested it: lodash function in template does not work. And I don't like the very long form (via 'display_name'). We already use the values  (0,1,2) in all code parts of this detail template an in pdf and projector. So: won't fix.

FinnStutzenstein: But: use `===` and `!==`.

emanuelschuetze: Done

openslides/assignments/static/templates/assignments/assignment-detail.html
emanuelschuetze: Done

emanuelschuetze:
Done and rebased.

PR #3159, opened 2017-03-28
fixing validate_html and motion saving (fixes #3140)

### LINKED ISSUE #3140: security issue: user without permission needs to set title, text, reason for motion

FinnStutzenstein:


emanuelschuetze:
This failure already exists in master. So perhaps no regression of this pull request.
@FinnStutzenstein @normanjaeckel Any idea? Should I merge this?
@FinnStutzenstein Can you rebase before merge please?

normanjaeckel:
The problem is that the client sends extra fields (comments as array and extra single comment fields) when comment fields are present and the user has this perm. Then the server sends 403 [here](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/motions/views.py#L120).

emanuelschuetze:
Rebased.

> This failure already exists in master. So perhaps no regression of this pull request.

see #3165

PR #3165, opened 2017-03-28
Fixed creating motion for users with comments permissions.

normanjaeckel:


emanuelschuetze:
travis failed:

```
$ flake8 openslides tests
tests/integration/motions/test_viewset.py:297:14: E131 continuation line unaligned for hanging indent
```

normanjaeckel:
Travis fixed. Ready to merge.

The problem with the agenda is only a client problem: The server creates the agenda item but the client tries to change it (from hidden to non-hidden). This request (PUT) fails. It is not about a POST request to create the agenda item. Only the server is able to create agenda items.

PR #3184, opened 2017-04-06
Support logos

FinnStutzenstein:


openslides/core/config.py
```diff
@@ -104,6 +105,29 @@ def __setitem__(self, key, value):
...
+                'path',
+                'display_name',
+            )
+            for required_entry in whitelist:
+                if required_entry not in value:
+                    raise ConfigError(_('{} has to be given.'.format(required_entry)))
+                if not isinstance(value[required_entry], str):
+                    raise ConfigError(_('{} has to be a string.'.format(required_entry)))
+            keys = list(value.keys())
+            # TODO: If this is uncommented, I get a 404 not found on the client.
```
FinnStutzenstein: @normanjaeckel Do you have an idea?

openslides/core/config.py
```diff
@@ -104,6 +105,29 @@ def __setitem__(self, key, value):
...
+            whitelist = (
+                'path',
+                'display_name',
+            )
+            for required_entry in whitelist:
+                if required_entry not in value:
+                    raise ConfigError(_('{} has to be given.'.format(required_entry)))
+                if not isinstance(value[required_entry], str):
+                    raise ConfigError(_('{} has to be a string.'.format(required_entry)))
+            # TODO: If this is uncommented, I get a 404 not found on the client.
```
normanjaeckel: Please remove this TODO and the code. I don't see a problem of someone saves additional unused properties into the config value.

openslides/mediafiles/static/js/mediafiles/list.js
```diff
@@ -41,7 +42,7 @@ angular.module('OpenSlidesApp.mediafiles.list', [
             }
         });
 
-        function updatePresentedMediafiles () {
+        function updatePresentedMediafiles() {
```
normanjaeckel: Please readd whitespace. :)

openslides/mediafiles/static/js/mediafiles/resources.js
```diff
@@ -93,6 +103,49 @@ angular.module('OpenSlidesApp.mediafiles.resources', [
     }
 ])
 
+.factory('Logos', [
```
normanjaeckel: This factory is only loosly coupled to the Logo ressource so I would prefere moving this code to another module, maybe `logo.js`.

openslides/core/config.py
FinnStutzenstein: Hm, Ok. But why does I get this 404 here? (just for learning ;) ). If I stop the script with the pdb and modify the dict by hand (also deleting `key`) it works..

openslides/mediafiles/static/js/mediafiles/resources.js
FinnStutzenstein: Done.

normanjaeckel:

FinnStutzenstein:
@normanjaeckel Done.

CHANGELOG
```diff
@@ -7,6 +7,8 @@ https://openslides.org/
 Version 2.1.2 (unreleased)
 ==========================
 
+Mediafiles:
```
normanjaeckel: Sorry but this is for OpenSlides 2.2

openslides/core/config.py
normanjaeckel: I do not know.

CHANGELOG
FinnStutzenstein: Done.

FinnStutzenstein:
@normanjaeckel Ping.

emanuelschuetze:
Rebased and resolved changelog conflic.

openslides/core/static/js/core/projector.js
```diff
@@ -150,7 +150,8 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
     'slides',
     'Config',
     'ProjectorID',
-    function($scope, $location, $timeout, Projector, slides, Config, ProjectorID) {
+    'Logos',
```
normanjaeckel: I am sorry but now I see that this is a problem. We make mediafile a dependency of core. Please let us think about a way not to add such a dep path.

openslides/mediafiles/static/js/mediafiles/list.js
```diff
@@ -22,8 +22,9 @@ angular.module('OpenSlidesApp.mediafiles.list', [
     'User',
     'Mediafile',
     'MediafileForm',
+    'Logos',
```
normanjaeckel: You should add `'OpenSlides.mediafiles.logo'` also to the module import list because we directly import `Logos` here. This is – of course – for readability only.

openslides/mediafiles/models.py
```diff
@@ -44,7 +44,8 @@ class Meta:
             ('can_see', 'Can see the list of files'),
             ('can_see_hidden', 'Can see hidden files'),
             ('can_upload', 'Can upload files'),
-            ('can_manage', 'Can manage files'))
+            ('can_manage', 'Can manage files'),
+            ('can_manage_logos', 'Can manage logos'))
```
normanjaeckel: This perm isn't used on server side, is it? Someone with 'can_manage_logos' but without 'can_manage_config' get errors?!

openslides/core/static/js/core/projector.js
FinnStutzenstein: @normanjaeckel The Logos basically use the config. I would prefer moving this `Logo` factory in the core/base.js. Only the changing happens in the mediafiles, so mediafiles is a dependency of core and not the other way around.

openslides/mediafiles/models.py
FinnStutzenstein: Ok yes. I will change the `check_view_permissions` for update and partial update if the config value to change is of type `logo`. If so, only users with `can_manage_logos` are allowed to change these config values.
I'll move the permission `can_manage_logos` also to the core. I thing we use this feature not only in the mediafile (only the logos are configured there) but this has effect on several places in OpenSlides. Also the mediafile won't get a dep of the core.

openslides/mediafiles/static/js/mediafiles/list.js
FinnStutzenstein: Not necessary anymore.

FinnStutzenstein:
@normanjaeckel Please check again.

emanuelschuetze:
Successfully retested.

PR #3188, opened 2017-04-07
Switched from npm to Yarn. Closed #3012.

### LINKED ISSUE #3012: Use Yarn instead of npm
We should evaluate switching to [Yarn](https://yarnpkg.com/lang/en/) and stop using npm. See [this announcement](https://code.facebook.com/posts/1840075619545360) to understand some issues.

@FinnStutzenstein Are you interested in looking at this (if you have some free time after 2.1 release)?

normanjaeckel:
For a clean switch remove `node_modules` (and possibly `bower_components`), install Yarn as mentioned in `DEVELOPMENT.rst` and run `yarn`. The first run is slower because Yarn has to fetch all packages. Later runs are much faster.

Later it is also possible that we work with an offline package repo.


DEVELOPMENT.rst
```diff
@@ -16,14 +16,18 @@ a. Check requirements
 '''''''''''''''''''''
 
 Make sure that you have installed `Python (>= 3.4) <https://www.python.org/>`_,
-`Node.js (>=4.x) <https://nodejs.org/>`_, `npm - Node Package Manager (>=3.x)
-<https://npmjs.org/>`_ and `Git <http://git-scm.com/>`_ on your system. You also
-need build-essential packages and header files and a static library for Python.
+`Node.js (>=4.x) <https://nodejs.org/>`_, `Yarn <https://yarnpkg.com/>`_ and
```
FinnStutzenstein: Is it necessary that Node/npm is installed parallel to yarn? Did we need node anymore? I didn't test this but yarn should be a drop in replacement for npm. Also in the yarn installation guide it does not tell to install node.

normanjaeckel: We need Node.js to run gulp and bower so we cannot go without it. Yarn is just a replacement for npm, nt for Node.js at all.

FinnStutzenstein: Ah yes, they are still Nodejs applications; or rather we need node to execute js on the server. You are right.

emanuelschuetze:
I tested it successfully and I added a new commit which updates the Dockerfile. Now the branch builds successfully on docker hub.

PR #3185, opened 2017-04-06
Export dialog for motions

FinnStutzenstein:

PR #3189, opened 2017-04-07
Pdf zip

FinnStutzenstein:
Info: This PR is based on the export dialog PR #3185 . So the export dialog has to be merged before this one.

@CatoTH Here are many changes, but just a few are important for this PR:
- In the motion/pdf.js is the new `MotionPdfExport` factory. There I've split up the generation of the `documentProvider`, which is passed through to the js-workers, and the actual download.
- The `getDocumentProvider` returns a promise that passes the provider if it is resolved.
- Also there is the `exportZip`. My goal if to have a dict assoziating filenames to the base64 content of the pdf. These will be put in a zip archive. For this, I iterate over all motions and generate the documentProvider for each motion.
- Then the `getDocument()` is called, so the document can get passed to the workers.

But I have found an issue there with the origin in the linenumberservice:
```
TypeError: Cannot read property 'length' of undefined
    at Object.djb2hash (openslides.js:10763)
    at Object.insertLineNumbers (openslides.js:11207)
    at MotionsmotionChangeRecommendation.getDiff (openslides.js:9005)
    at Motionsmotion.getTextByMode (openslides.js:8586)
    at motionText (openslides.js:12416)
    at Object.getContent (openslides.js:12484)
    at Object.getDocument (openslides.js:4700)
    at openslides.js:12861
    at processQueue (openslides-libs.js:26201)
    at openslides-libs.js:26217
```
If I try to export one motion, this works. If I want to export n motions, the first works but for all n-1 motions I get this error. May the problem be that the document generation has to be executed motion by motion??

To reproduce: Create 2-3 motions. Go to list view, click on export, select 'zip'. To download just one motion, use the filters.

Please say, if something is not clear here.

Thanks!!!

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -17,6 +17,7 @@ Motions:
 - Fixed empty motion comment field in motion update form [#3194].
 - Removed server side image to base64 transformation and
   added local transformation [#2705]
+- Added support for export motions in a zip archive [#3189].
 
 Core:
 - No reload on logoff. OpenSlides is now a full single page
--- a/bower.json
+++ b/bower.json
@@ -28,6 +28,7 @@
     "jquery.cookie": "~1.4.1",
     "js-data": "~2.9.0",
     "js-data-angular": "~3.2.1",
+    "jszip": "~3.1.3",
     "lodash": "~4.16.0",
     "ng-dialog": "~0.6.4",
     "ng-file-upload": "~11.2.3",
--- a/openslides/core/static/js/core/pdf.js
+++ b/openslides/core/static/js/core/pdf.js
@@ -859,10 +859,11 @@ angular.module('OpenSlidesApp.core.pdf', [])
 
 .factory('PdfCreate', [
     '$timeout',
+    '$q',
     'gettextCatalog',
     'FileSaver',
     'Messaging',
-    function ($timeout, gettextCatalog, FileSaver, Messaging) {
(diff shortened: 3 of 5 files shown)


emanuelschuetze:
@CatoTH As discussed please fix the linenumberservice to create motions parallel. You can reproduce the failure with this pull request. Please create a new pull request for your fix. Thanks!

CatoTH:
@FinnStutzenstein I think this error occurs because the functions are called with an invalid motionVersion, which cannot be resolved to a text. That's why the string passed to insertLineNumbers is null and the function fails. If you replace the motionVersion by "null" (the most recent motion version) in the getTextByMode-Call in pdf.js, the error message does not occur anymore.
Or are you sure the passed motionVersion-variable is correct?

FinnStutzenstein:
I'll check if the motionVersion is given.

FinnStutzenstein:
@CatoTH Yes you are right; I pass the `params` object to every call, but on the first call the motion version is set and this object is passed via reference. So for the other calls the version is wrong; a simple `_.clone()` helps. Thanks!

FinnStutzenstein:
@emanuelschuetze Ready to test!

FinnStutzenstein:
@emanuelschuetze Ready to test.

emanuelschuetze:
Successfully retested.
@normanjaeckel ping

Issue #3002, opened 2017-02-23
User without users.can_see permission cannot see names of submitter/speakers/candidates


emanuelschuetze:
To reproduce:
* Create new (empty) group with motions.can_see permission
* Create new user and add this new group to user.
* Create new motion with admin as submitter.
* Login as user and look at the motion
=> Submitter is not visible!

* Add permission 'users.can_see' for the new user group.
* Reload as user.
=> Now, submitter ("Administroator")  is visible!

Same problem also for speakers and candidates. This is very important for 2.1 because many organizations removed the 'users.can_see' permissions for normal users!

@normanjaeckel @FinnStutzenstein Any idea how to fix it?

FinnStutzenstein:
Technically the `users.can_see` permission does what should do. But: A submitter is strongly coupled with a motion. If a user is allowed to see this motion, he should be allowed to see the submitter. There are some levels of restrictions:

1. The server sends all users that are coupled with motions/assignments/topics/mediafiles/... They are now also visible in the list view. The user can visit the detail page.
2. The server sends all coupled users but cuts off the most data. The client just gets a mapping `id <-> Tttle, first name, last name`. The user list view does not show these users. The users are not clickable in the motions/...
3. We do not fix this and leave `users.can_see` very restrictive. If so, we have to modify some template, so that the submitters/... are not shown.

If we choose 1. or 2. the server has some more work to do to collect all needed users; so check which item the user is allowed to see and yield the users from these items.

normanjaeckel:
Discussed with @emanuelschuetze. The user app may inform all other apps via a Django signal about every specific user and ask whether the user has to be available for the client at least with LITTLE_DATA (see users/access_permissions:get_restricted_data()). The apps my answer Yes (True) if e. g. in motion app the user appears as submitter or supporter.

PR #3191, opened 2017-04-10
User without permission to see users can now see some of them.

normanjaeckel:
@emanuelschuetze Please review and test some cases.

- [x] motion submitters
- [x] motion supporters
- [x] candidates (assignment related users)
- [x] assignment option candidates (inside a poll)
- [x] item speakers
- [x] chat message
- [x] mediafile uploader

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -19,6 +19,12 @@ Core:
 - No reload on logoff. OpenSlides is now a full single page
   application [#3172].
 
+Users:
+- User without permission to see users can now see agenda item speakers,
+  motion submitters and supporters, assignment candidates, mediafile
+  uploader and chat message users if they have the respective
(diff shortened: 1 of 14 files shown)


emanuelschuetze:
Please update the commit message:
 * Use a short summary in the first line (the current message is very long without blank line after first line)
 * Add the fixed issue #3002.

normanjaeckel:
Commit message changed.

The problem with the newly created user is that we do not have an autoupdate on users if an assignment changes so the user (which is required now but wasn't required at creation time). This a design problem and is only loosely coupled with this pull request. For the projector we fixed this with complex server side logic. This should be discussed. See #3195.

emanuelschuetze:
Added pullrequest number to changlog entry.

PR #3194, opened 2017-04-11
Fixed empty motion comment field in motion update form.

emanuelschuetze:
Motion comment values were not extracted. Concerns edit button in
motion list view only.


normanjaeckel:

PR #3196, opened 2017-04-12
Fixed regression that opening of motion create form failed.

emanuelschuetze:
Introduced with commit 3de4a2af1c09b325c0cb9c464fca587cc6b03810.


normanjaeckel:
Ok.

PR #3201, opened 2017-04-18
Fixed compare of duplicated users while csv user import.

emanuelschuetze:
Now the joined full name (of title, first name, last name and structure level) is compared - not only e.g. the particular first name with an existing first name etc.


normanjaeckel:

PR #3202, opened 2017-04-18
Bugfix: changing line length did not invalidate cache

### LINKED ISSUE #3107: missing autoupdate for motion text length config
- Set in the OS Settings for motions a line length of 70.
- create now a new motion
- paste e.g. this text:
"In political parties it is common that a commission has to have a look at hundreds of motions. Their job is to deny invalid motions and to give recommendations. Therefore it would be an improvement in the workflow if one would be able to import the motions with the recommendation. At the moment one would have to set the recommendation for every motion which could take a lot of time."
- save the motion and project it

result: still up to 80 chars fill a line. This is a regression to prior OS 2.1 versions where this worked.

Moreover, assume the case that you have e.g. 20 motions using 80 chars line length. Now you see in the rules that only 65 chars are allowed in one line. Therefore you change the line length to this value. result. nothing happens. I expect all motions to be changed automatically. Otherwise it would be a nightmare in real life.

CatoTH:
The cache was based only on the html string, not the line length. Therefore, when changing the line length, the result of the function did not change until the page was reloaded.

This should fix #3107.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,6 +18,7 @@ Motions:
 - Removed server side image to base64 transformation and
   added local transformation [#2705]
 - Added support for export motions in a zip archive [#3189].
+- Bugfix: changing motion line length did not invalidate cache [#3202]
 
 Core:
 - No reload on logoff. OpenSlides is now a full single page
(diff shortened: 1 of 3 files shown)


normanjaeckel:

emanuelschuetze:
CHANGELOG entry added and rebased.

PR #3207, opened 2017-04-19
Configurable projector header

FinnStutzenstein:


openslides/core/config_variables.py
```diff
@@ -139,8 +139,8 @@ def get_config_variables():
         default_value=True,
         input_type='boolean',
         label='Show logo on projector',
-        help_text='You can replace the logo. Just copy a file to '
-                  '"static/img/logo-projector.png" in your OpenSlides data path.',
+        help_text='You can replace the logo by uploading an image and set it as '
+                  'the "main projector logo" in the mediafiles.',
```
emanuelschuetze: We use "files" instead of "mediafiles" in UI. 
Suggestion: ' .. in "files".'

FinnStutzenstein:
@emanuelschuetze Ping.

FinnStutzenstein:
Rebased.

FinnStutzenstein:
@normanjaeckel Whats about moving this option into the projector manage view, so you can enable/disable the clock for each projector on its own?

emanuelschuetze:
@normanjaeckel I think the new clock config fits well in the projector config because there we manage also other projector appearance (e.g. show logo/text, color etc). All configs has effect to _all_ projectors. So I am not sure if this is a good idea to move only the clock to the manage projector view.

+1  for merging this pull request without any changes and think about later about more ui improvements.
Ok?

PR #3209, opened 2017-04-19
Massimport of users (WIP)

FinnStutzenstein:
@emanuelschuetze @normanjaeckel Whats your opinion.

There is one trap: The username is generated on creation. So you can import user 'max' twice, because 'max' gets created, and for the second 'max' the username will be 'max 2'.
Now, all users are first validated, so there is no correction of the username, and then bulk_created.
So the drawback is, that we cannot import unique users twice (unique by username) on the same import.

If you want to have two 'max', you have to import him twice with two clicks on 'import'.

The 'importTrackId' is used for checking, which users gets imported. If 4 of 5 get imported, the client wants to know which users are imported. You can pass an `importTrackId` to every user and get an array of imported ids in the response. Helpful for the client.


emanuelschuetze:
I tested a bit. Importing 1000 users via csv in ca. 20sec! Wow! :-) Only one request. Also for importing via text box. Looks good.

Duplicated users (with same username) is an rare case. Then I would stop importing the second one and print out which users are not imported. Reimport of these failed users would be nice (but not really required).

Let me know when I should more retest.

FinnStutzenstein:
This sould already work with csv. Just one of the duplicates should get a green sign.
Just the button 'remove all imported' is missing.

FinnStutzenstein:
@emanuelschuetze Please test especially duplicates.

FinnStutzenstein:
The progressbar is some kind of a problem: Because we have a single request, we just get a single response. So I do not see any possibilities to transfer the status.
May with the notify system, the server could send status updates; but this is not fully implemeted yet.

I've tried to import 5000 users --> No problem. Takes about 30 sec. No Redis, Postgresql, just basic OS. Can you reproduce this?

FinnStutzenstein:
Tested: 5000 users with one group --> 503.
But if I wait a bit, the server execute all group assigning.
To speed up this group assigning, we need to have an explicit m2m table between users and groups. (In Django's `PermissionMixin` just a 'normal' m2m relation is established.) If we have this model (e.g. `UserGroupRelation`) we can just do a `bulk_create` for all these relations. For this we have to do:
 - Make a wrapper class around Django's `PermissionMixin`
 - Make a migration to copy all existing relations to the new model
 - (Maybe we have to change group/user code in the views, because of this new model)

@normanjaeckel @ostcar Your opinions?

normanjaeckel:
Hm. I am not sure whether we should spend much time on this old permission modeling. @ostcar has always been against the Django permission system and we can think about removing it completly and building an easier replacment ...

emanuelschuetze:
@ostcar ping

ostcar:
I have nothing to say here. I still think, that the current permission system is not good, but I don't have the time to change it.

emanuelschuetze:
@FinnStutzenstein Do you need more help/information?

FinnStutzenstein:
@emanuelschuetze I'll work on next week

FinnStutzenstein:
@normanjaeckel Please look at the new commit.

1. The migrationpath is broken; if we use this, we need to copy much data (how to do this?)
2. I don't like this way of changing the group relation ship.....

FinnStutzenstein:
Closed in favor of new PR.

PR #3212, opened 2017-04-19
Added notify system.

normanjaeckel:
@FinnStutzenstein Please play around a little bit. Server side (echo) should work for now.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,6 +21,8 @@ Motions:
 - Added support for export motions in a zip archive [#3189].
 - Bugfix: changing motion line length did not invalidate cache [#3202]
 - Bugfix: Added more distance in motion PDF for DEL-tags in new lines [#3211].
+- Added warning message if an edit dialog was already opened by another
+  client [#3212].
 
 Users:
 - User without permission to see users can now see agenda item speakers,
@@ -36,6 +38,7 @@ Core:
   application [#3172].
 - Adding support for choosing image files as logos [#3184, #3207].
 - Fixing error when clearing empty chat [#3199].
+- Added notify system [#3212].
 
 General:
 - Several bugfixes and minor improvements.
--- a/openslides/core/static/css/app.css
+++ b/openslides/core/static/css/app.css
@@ -1196,7 +1196,7 @@ img {
     border-color: #d6e9c6;
 }
 #messaging .warning {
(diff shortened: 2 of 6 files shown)


FinnStutzenstein:
@normanjaeckel I cannot push into your branch, so please look at my branch.

You can test my first idea of the notifying users, when others edit dialogs are open (currently just the array with names, the ui follows if everything works). With logged in users, this works fine.

I have a problem with anonymous users. With every notify, I pass a `senderId`, so the receiver can answer to this user directly. First, I would like to put [this line](https://github.com/FinnStutzenstein/OpenSlides/commit/5b7335737e4695b816f303de13759383a239a97a#diff-e97f2fd30e66e33689576c55af91d05eR381) to the server, so a user cannot corrupt this and second, whats about anonymous users?
I think, because they do not have an id, we should pass the channel name, so we do have a direct replay channel. @normanjaeckel Opinions?

normanjaeckel:
We can add the decorator `@channel_session_user` to the ws_receive_site function to get the user but I think the reply channel name is already part of the message object. Then we just need some server logic here ...

FinnStutzenstein:
Yes, this would be great. So the server includes the reply channel. If the clients notify includes `... replyChannel: 'channelname', ...` this message is only send to this channel.

This also fixes another issue: If user A has two tabs open and user B wants to reply to a specific message from user A, currently both tabs get this notify. With giving the replychannel, just the one tab, which has send the request, gets an answer.

We also should include always the senderId. If I want to notify a user, all running instances of OS by this user should get the message.

TL;DR:
The server should add:
 - replychannel (always),
 - userId (None, if anonymous)

normanjaeckel:
@FinnStutzenstein Done and rebased. Please rebase your own branch and test again (maybe read an correct my docstring in autoupdate.py).

FinnStutzenstein:
@normanjaeckel This works fine! Now, I just have to do the ui and how to display the warning.

normanjaeckel:
Ok, nice. We can merge it but then please open an issue about the following: The new code in MotionUpdateCtrl is very much. We should encapsulate it in an extra factory or service. Also some docstrings would be helpful to understand what this is doing (question query, response query, close signal ...).

emanuelschuetze:
>  We can merge it but then please open an issue about the following

Done: #3219

Issue #3204, opened 2017-04-19
django channels version broken


ostcar:
On the current master, you can not install the requirements. There are versions of channels, asgiref, daphne etc. installed, which are not compatible.

emanuelschuetze:
I can't reproduce it with a fresh python3.5 virtualenv. Every requirement was installed successfully:

```
pip install -r requirements_production.txt
[..]
Successfully installed Automat-0.5.0 Django-1.10.7 PyPDF2-1.26.0 appdirs-1.4.3 asgiref-1.1.1
attrs-16.3.0 autobahn-0.18.2 bleach-1.5.0 channels-1.0.3 constantly-15.1.0 daphne-1.2.0
djangorestframework-3.4.7 html5lib-0.9999999 incremental-16.10.1 jsonfield-1.0.3
packaging-16.8 pyparsing-2.2.0 roman-2.0.0 setuptools-34.4.1 six-1.10.0 twisted-17.1.0
txaio-2.7.0 zope.interface-4.3.3


pip install -r requirements.txt
[..]
Successfully installed coverage-4.3.4 flake8-3.3.0 isort-4.2.5 mccabe-0.6.1 pycodestyle-2.3.1
pyflakes-1.5.0

pip install -r requirements_big_mode.txt
[..]
Successfully installed asgi-redis-1.0.0 django-redis-4.7.0 django-redis-sessions-0.5.6
msgpack-python-0.4.8 psycopg2-2.6.2 redis-2.10.5

```

What's your problem?

ostcar:
The problem is not the install. The problem is, than python manage.py runserver does not work afterwards. At least, there are some errors.

```
2017-04-19 10:31:55,281 - ERROR - server - Error trying to receive messages: name 'txredisapi' is not defined
2017-04-19 10:32:00,284 - ERROR - server - Error trying to receive messages: name 'txredisapi' is not defined
2017-04-19 10:32:05,287 - ERROR - server - Error trying to receive messages: name 'txredisapi' is not defined
```

ostcar:
The error happens only when using redis as asgi backend.

normanjaeckel:
Confirmed.

emanuelschuetze:
Fixed with #3213.

PR #3213, opened 2017-04-20
Updated requirements (channels and asgi-redis).

normanjaeckel:
@ostcar may check this ...

### DIFF
--- a/requirements_big_mode.txt
+++ b/requirements_big_mode.txt
@@ -2,7 +2,7 @@
 -r requirements_production.txt
 
 # Requirements for Redis and PostgreSQL support
-asgi-redis>=1.0.0,<1.1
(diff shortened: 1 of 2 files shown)


ostcar:
You should merge this. The problem is, that geiss want work with it, but I try to update geiss in the next month. If not, you should nevertheless use a current version of channels.

One proposal: Add all channel-packages explicit to our requirements-file. So problems like #3204 wont happen again. The additional packages we need is asgiref and daphne

ostcar:
You should merge this. The problem is, that geiss want work with it, but I try to update geiss in the next month. If not, you should nevertheless use a current version of channels.

One proposal: Add all channel-packages explicit to our requirements-file. So problems like #3204 wont happen again. The additional packages we need is asgiref and daphne

PR #3223, opened 2017-04-27
Speed up WS connection

ostcar:
With this pull request the connection time is only 20% as before.

Please change docstrings, variable names etc as you like


ostcar:
Please fix travis, then merge

PR #3233, opened 2017-04-28
Fixes autoupdate bug for a user without user.can_see_name permission …

normanjaeckel:
…if a speaker is added.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -46,7 +46,8 @@ Core:
 - Fixing error when clearing empty chat [#3199].
 - Added notify system [#3212].
 - Enhanced performance esp. for server restart and first connection of all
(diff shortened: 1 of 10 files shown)


normanjaeckel:
This would be the "more code" solution for the problem that a user without user.can_see_name permission is able to see correct autoupdate of speakers. We need such code also for motions, assignments, mediafiles and core.

But maybe someone has a better idea?

normanjaeckel:
@emanuelschuetze Fix is now ready for review, test and merge.

emanuelschuetze:
travis failed (flake8 error)

normanjaeckel:
All done.

emanuelschuetze:
Changelog entry added and rebased.

PR #3240, opened 2017-05-05
Fixing wrong err call

FinnStutzenstein:
@emanuelschuetze @normanjaeckel Please merge this. console does not have a `err` function, so the generation of pdfs fails in some browsers. I changed this to `log` to can see, which browsers uses which color formats.


FinnStutzenstein:
@normanjaeckel Ok.

Issue #3242, opened 2017-05-07
Remove default permissions for personal note


emanuelschuetze:
For the new personal note feature there are introduced the django default permissions which are listed in the permission matrix for users:
```
  Can delete personal note 				
  Can change personal note 				
  Can add personal note 
```
This permissions should be removed from UI - as in all other apps.

PR #3244, opened 2017-05-09
Removed default permissions for personal note. Updated Yarn lock. Fix…

normanjaeckel:
…ed #3242.


normanjaeckel:
Hm. I do not like to provide migration files for every master snapshot. And I also do not want to add extra code to remove the in between created permissions. The migration path from 2.1 to 2.2 should work.

For a master snapshot instance it should be fine to remove the permissions manually from the db. It should not be necessary (and no harm) to run `migrate`.

emanuelschuetze:
As discussed with @normanjaeckel we will merge this pullrequest now. To migrate an old DB you can run this python code in a django shell:

```
$ python manage.py shell

>>> from django.contrib.auth.models import Permission
>>> from django.contrib.contenttypes.models import ContentType

>>> ct = ContentType.objects.filter(app_label='users', model='personalnote')
>>> Permission.objects.filter(content_type=ct)
<QuerySet [<Permission: users | personal note | Can add personal note>, 
  <Permission: users | personal note | Can change personal note>, 
  <Permission: users | personal note | Can delete personal note>]>

>>> Permission.objects.filter(content_type=ct).delete()
(3, {'auth.Permission': 3})

>>> Permission.objects.filter(content_type=ct)
<QuerySet []>
```

PR #3251, opened 2017-05-11
Improve zip creation and fixing page numbers

FinnStutzenstein:
When creating the motion zip archive, for every motion a worker is instantiated. Result with 100 motions: Browser freeze and no reaction at all.
Now: Just instantiate one worker, that does all converting.

Also fixed broken page counting and cleaned up the worker.


emanuelschuetze:
@FinnStutzenstein A short CHANGELOG entry would be nice.

emanuelschuetze:
Thanks!

PR #3256, opened 2017-05-17
extending the color parser (fixes #3243)

### LINKED ISSUE #3243: Could not parse color "red"/"green"
If you create a pdf in diff mode there are two errors in JS console:
```
Could not parse color "red"
Could not parse color "green"
```
This seems to be caused by pull request #3240.

Additional:
If you have some special text color values (e.g. `<span style="color:#000000">...</span>`) the diff lines are black (expected: added/removed lines should be override all custom text colors by using red/green color).
-> This is very typically for users who copy/paste from MS word.

FinnStutzenstein:
Just pass named colors; the parser can handle these.


normanjaeckel:
You can merge it but you may care about the case that someone uses invalid color names like `hamster` instead of `yellow`.

FinnStutzenstein:
Yeah I know. But do you know a way to check valid colors? I didn't find anything reasonable..

normanjaeckel:
No I don't. No problem if you want to merge now.

emanuelschuetze:
I'll merge it now.

Issue #3262, opened 2017-05-19
Move personal note to separate database model


emanuelschuetze:
Move information for personal notes and favorit of each motion to a separate database model which is based on user ids. Advantage: send all motion full data to client without split out personal part for each user. Should optimize performance.

PR #3267, opened 2017-05-23
Rework on personal notes.

normanjaeckel:
@FinnStutzenstein Ready for client rework. See #3262.


FinnStutzenstein:
@normanjaeckel I get an error when I try to POST this data: 
```
{"notes":{"motions":{"1":{"note":"note for the motion with id 1","star":true}}}}
```
Error: notes: Not a valid string.

normanjaeckel:
I think the problem is that the serializer field (users/serializers) expect a string instead of a JSONable object. So we have to set the serializer field explicitly. I can do this ...

FinnStutzenstein:
@normanjaeckel This would be nice!

normanjaeckel:
Please test again.

FinnStutzenstein:
@normanjaeckel Also rebased.

normanjaeckel:
@emanuelschuetze Ready for test and merge,

PR #3271, opened 2017-05-25
Marks split list items and hides the bullet points - fixes #3269

### LINKED ISSUE #3269: Wrong change view, when deleting parts in lists
When deleting parts in a list item, extra list bullets are shown in the change view:

Normal view:
deletion_prev

Change view:
deletion_changeview

CatoTH:


emanuelschuetze:
@CatoTH Can you give me a step by step instruction how to see that your fix works?

CatoTH:
@emanuelschuetze this class gets only added if not the whole list item is removed, but only parts of an item are effected. If that's not the case for you, can you maybe send me a dump with the change?

emanuelschuetze:
I can't reproduce your fix. Here an example what I tested:

Create a motion with some ul list items and remove some parts of two items in change recommendations (see original version, line 3 and line 8-11):
issue3269-a

Switch to diff version - each removed list parts get a new bullet point (line 3 and 8). And also the last part of item 1 gets a new bullet point (line 4) - which is not desired:

issue3269-b

@CatoTH Please try to test it again.

CatoTH:
@emanuelschuetze The extended pull request should fix your problem. The problem was triggered by the <p>-Tags inside of the split <li>s.
However, one other issue remains in the Diff view in that specific case: the split <p> causes a margin below that cannot be canceled out by the current classes. That would need further refactoring, and probably another Issue.

BTW: splitting the markup is not only necessary for editing the change recommendation as mentioned before, but also to display the change recommendation with the grey background and the action buttons to the right for the admins.

CatoTH:
The new PR should address the issue I mentioned in my last comment.

PR #3274, opened 2017-05-31
Do not reload all projector elements on every autoupdate (fixes #3259).

### LINKED ISSUE #3259: Black screen on page changing with a projected pdf
Project a pdf file and change the page. A black screen appears while changing.

Problem: The display properties (zoom, page number, visible, ...) is saved into the projector config. When changing the page, the projector config is updated and so the projector reloads the elements. The `ng-repeat` over all elements rebuild the DOM. In fact, the `ng-pdf` with the `canvas` is removed and immediately placed by angular. So the canvas reloads --> this causes the black screen.

Solution: Do not save the display information into the projector config. I would like to have a JSON-field in the mediafile model (name like 'display_information'), so updating e.g. the page number do not reload the projector elements. Also this helps syncing if we want to display one pdf on multiple projectors, so all projectors display the pdf equally.

@normanjaeckel can you change the mediafile model?

FinnStutzenstein:
@emanuelschuetze Zooming and page changing are working. To fix the little loading delay, I think we have to write our own viewer to load next pages in the background. The current viewer does not do this.

This has to be tested more carefully: I do not exactly know, if any slide needs the reloading when the element changes. Maybe a watcher is missing, but until now on every change the element gets reloaded, so this watcher was not necessary. This is changed with this PR!

### DIFF
--- a/openslides/core/static/js/core/projector.js
+++ b/openslides/core/static/js/core/projector.js
@@ -157,15 +157,48 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
         $scope.broadcast = 0;
 
         var setElements = function (projector) {
-            $scope.elements = [];
+            // Get all elements, that should be projected.
+            var newElements = [];
             _.forEach(slides.getElements(projector), function (element) {
                 if (!element.error) {
                     // Exclude the clock if it should be disabled.
                     if (Config.get('projector_enable_clock').value || element.name !== 'core/clock') {
-                        $scope.elements.push(element);
+                        newElements.push(element);
                     }
                 } else {
                     console.error("Error for slide " + element.name + ": " + element.error);
+                }
+            });
+
+            // Now we have to align $scope.elements to newElements:
+            // We cannot just assign them, because the ng-repeat would reload every
+            // element. This should be prevented (see #3259). To change $scope.elements:
+            // 1) remove all elements from scope, that are not in newElements (compared by the uuid)
+            // 2) Every new element in newElements, that is not in $scope.elements, get inserted there.
+            // 3) If there is the same element in newElements and $scope.elements every changed property
+            //    is copied from the new element to the scope element.
+
+            $scope.elements = _.filter($scope.elements, function (element) {
+                return _.some(newElements, function (newElement) {
+                    return element.uuid === newElement.uuid;
+                });
+            });
+
+            _.forEach(newElements, function (newElement) {
+                var matchingElement = _.find($scope.elements, function (element) {
+                    return element.uuid === newElement.uuid;
+                });
+                if (matchingElement) {
+                    // copy all changed properties.
+                    _.forEach(newElement, function (value, key) {
+                        if (newElement.hasOwnProperty(key) && !key.startsWith('$')) {
+                            if (typeof matchingElement[key] === 'undefined' || matchingElement[key] !== value) {
+                                matchingElement[key] = value;
+                            }
+                        }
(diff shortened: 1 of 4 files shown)


openslides/core/static/js/core/projector.js
```diff
@@ -157,17 +157,57 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+            // We cannot just assign them, because the ng-repeat would reload every
+            // element. This should be prevented (see #3259). To change $scope.elements:
+            // 1) remove all elements from scope, that are not in newElements (compared by the uuid)
+            // 2) Every new element in newElements, that is not in $scope.elements, get inserted there.
+            // 3) If there is the same element in newElements and $scope.elements every changed property
+            //    is copied from the new element to the scope element.
+
+            $scope.elements = _.filter($scope.elements, function (element) {
+                var found = false;
+                _.forEach(newElements, function (newElement) {
```
normanjaeckel: Here you may use a shorter lodash method like `_.some`.

openslides/core/static/js/core/projector.js
```diff
@@ -157,17 +157,57 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+                    if (element.uuid === newElement.uuid) {
+                        found = true;
+                    }
+                });
+                return found;
+            });
+
+            _.forEach(newElements, function (newElement) {
+                var found = false;
+                _.forEach($scope.elements, function (element) {
```
normanjaeckel: Here the same. Use `_.some` or `_.findIndex`

openslides/core/static/js/core/projector.js
```diff
@@ -157,17 +157,57 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
...
+                    if (element.uuid === newElement.uuid) {
+                        found = true;
+
+                        // copy all changed properties.
+                        _.forEach(newElement, function (value, key) {
```
normanjaeckel: What about using `_.assign` or any similar method?

openslides/mediafiles/static/js/mediafiles/projector.js
```diff
@@ -18,11 +18,36 @@ angular.module('OpenSlidesApp.mediafiles.projector', [
...
+            }
+        });
+
+        // Watch for scale changes. If the scale is changed, reload the pdf
+        // viewer by just disable and re-enable it.
+        $scope.$watch(function () {
+            return $scope.element.scale;
+        }, function () {
+            $scope.showPdf = false;
+            $timeout(function () {
```
normanjaeckel: This looks a little bit hacky but if we don't have a better solution we can merge it ...

openslides/core/static/js/core/projector.js
FinnStutzenstein: This will not work, because we do not want to copy angular stuff (e.g. the $hashCode changes every time).

openslides/core/static/js/core/projector.js
FinnStutzenstein: Done with `_.find`, because I need the element afterwards.

openslides/core/static/js/core/projector.js
FinnStutzenstein: Ok

openslides/core/static/js/core/projector.js
normanjaeckel: ok

PR #3286, opened 2017-06-02
Fixes some PDF issues

FinnStutzenstein:
- Fixes too low line numbers
- Fixes line numbers on page breaks also for lists
- Respect br tags in inserted texts.


FinnStutzenstein:
@normanjaeckel Done. Is it ok to just collect all PRs in one list item, that contains pdf fixes (not added/removed features)?

normanjaeckel:
Of course.

PR #3287, opened 2017-06-03
Fix a bug where a single moved word was not highlighted in the diff

CatoTH:


tests/karma/motions/diff.service.test.js
```diff
@@ -479,6 +479,14 @@ describe('linenumbering', function () {
 
       expect(diff).toBe('<P class="delete">...so frißt er Euch alle mit Haut und Haar.</P><P class="insert">...so frißt er <SPAN style="font-size: 2em; opacity: 0.5">Euch alle</SPAN> mit Haut und Haar.</P>');
     });
+
+    it('marks a single moved word as deleted and inserted again', function () {
+      var before = '<p>' + noMarkup(1) + 'Die notwendigen Ressourcen für erfolgreiche Jugendarbeit, sowohl personell als auch ' + brMarkup(2) + 'finanziell, müssen in den Regionen zur Verfügung gestellt werden. Insbesondere muss in </p>',
```
emanuelschuetze: Please use a dummy text - not a real motion text of customer!

CatoTH:
Ok, test case should be more abstract now.

PR #3288, opened 2017-06-05
Issue3269 issue3232 split elements2

CatoTH:
This is a more generalized version of #3271, which worked around a problem occurring when splitting lists items in change recommendations.

The problem of getting the spacing between paragraphs right when a part of it has been extracted and is being displayed inside a Diff-Box basically had the same reason: to display it correctly, information from the line extraction algorithm is needed to decide if there should be space between two paragraphs (if they have been paragraphs in the original version) or not (if they were the same paragraph originally), or if a bullet point should be displayed or not.

This is done by adding the two classes "os-split-before" and "os-split-after" (instead of only "os-split-li" in the previous PR). This also made necessary some extra normalization for the diff.

### DIFF
--- a/openslides/motions/static/js/motions/diff.js
+++ b/openslides/motions/static/js/motions/diff.js
@@ -309,6 +309,25 @@ angular.module('OpenSlidesApp.motions.diff', ['OpenSlidesApp.motions.lineNumberi
         };
 
         /**
+         * When a <li> with a os-split-before-class (set by extractRangeByLineNumbers) is edited when creating a
+         * change recommendation and is split again in CKEditor, the second list items also gets that class.
+         * This is not correct however, as the second one actually is a new list item. So we need to remove it again.
+         *
+         * @param {string} html
+         * @returns {string}
+         */
+        this.removeDuplicateClassesInsertedByCkeditor = function(html) {
+            var fragment = this.htmlToFragment(html);
+            var items = fragment.querySelectorAll('li.os-split-before');
+            for (var i = 0; i < items.length; i++) {
+                if (!this._isFirstNonemptyChild(items[i].parentNode, items[i])) {
+                    this.removeCSSClass(items[i], 'os-split-before');
+                }
+            }
+            return this._serializeDom(fragment, false);
+        };
+
+        /**
          * Returns the HTML snippet between two given line numbers.
          *
          * Hint:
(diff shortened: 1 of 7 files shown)


emanuelschuetze:
There is still a problem if you want to split a list item into 2 bullet points. I added a Linebreak in the change recommendation (which brings a new bullet point in changed view, ok.) But in diff version there is no new bullet point (before "TEST")... See example:


Original version:
pr-3288-0

Changed version:
pr-3288-1

Diff version:
pr-3288-2

CatoTH:
CKEditor is nasty here... it automatically adds the internal class to newly created <li> elements, and I don't see how to prevent it from doing that. I added a work around that tries to clean up the html before saving it to the change recommendation. If you re-create the change recommendation (or edit it and save it again), it should be displayed correctly.

emanuelschuetze:
@FinnStutzenstein Can you help test this pull request with several use cases? Thanks.

FinnStutzenstein:
@CatoTH I see your problem here: I select a part of a list item: Expected: no bullet point in ck editor here, but the editor creates a list item of this part.

Tests:

1. Select a part of a list item and just press save to create a change recommendation. CK editor shows bullet point, but the diff view is correct --> OK
2. Select a part of a list item and remove the bullet in the editor (a user may do this, because he thinks, that he would create a new list item). There is no point in the list view, but the text is not aligned to this list anymore (The ck editor removes the `ul`/`ol` --> new paragraph) --> not OK
test2

3. Select a part of a list item and just hit enter two times in the ck editor to split up the text into three parts. The ck editor shows three list items; click save. The first bullet point is not shown (OK) but the other two --> not OK
test3

4. Same as 3. but remove the list items in the ck editor as in 2. Same result as in 2. --> not OK

Whats about not to pass everything to the ck-editor and leave everything surrounded out, so just pass the text to the editor, so it doesn't know about any list things. When a user click on save, just the text from the ck-editor has to be inserted in every tags surround.

CatoTH:
@FinnStutzenstein I'm not sure if I really understood the problems in 2) and 3).
For 2), I understood that the UL/LI is changed into a regular paragraph. In that case, it would be normal that the regular paragraph does not have a bullet point and is not intended as a list item. This sounds more like a mistake made by the user, and I don't think, fixing this on the server side (or when displaying it afterwards) would be the correct way of dealing with it.

For 3) If you hit enter inside of a list item, you split it into two. So the second one (or the third one) would get a new bullet point. That sounds correct. Did I misunderstand the problem?

FinnStutzenstein:
@CatoTH As a user I would expect, when editing a part of an item, this part does not get a bullet in the ck-editor. A user would be irritated, whether he creates a new list point or not, so he may remove the bullet point in the ck-editor. This leads to 2) and a paragraph is inserted into the list item. Maybe @emanuelschuetze has an opinion, how OS should behave in this situation

For 3) I thought about making a paragraph inside list items, but it seems, that we want to create new items in this case. OK.

CatoTH:
We could remove the bullet point from CKEditor, however then it would be impossible to split an item into two or add a new one, not sure if that's really better.

FinnStutzenstein:
In my opinion the main problem is solved. We should move this disucssion in a new issue (if needed, @emanuelschuetze ?).

openslides/motions/static/js/motions/diff.js
```diff
@@ -546,12 +599,15 @@ angular.module('OpenSlidesApp.motions.diff', ['OpenSlidesApp.motions.lineNumberi
                 for (i = 0; i < children.length; i++) {
                     newNode.appendChild(children[i]);
                 }
+
```
normanjaeckel: Please remove some lines here.

CatoTH:
@normanjaeckel done.

normanjaeckel:
@emanuelschuetze Please approve and merge.

emanuelschuetze:
I resolved the conflicts in CHANGELOG.
Retested. Looks ok. Some minor regressions in pdf formatting (e.g. margin between change recommendation lines) can be fixed later in a new pr. I'll merge this pr.

PR #3290, opened 2017-06-07
Massimport for users

FinnStutzenstein:
Now it's very fast (also with groups) using atomic transactions to the db.

### DIFF
--- a/openslides/users/static/js/users/site.js
+++ b/openslides/users/static/js/users/site.js
@@ -755,13 +755,12 @@ angular.module('OpenSlidesApp.users.site', [
 .controller('UserUpdateCtrl', [
     '$scope',
     '$state',
-    '$http',
     'User',
     'UserForm',
     'Group',
     'userId',
     'ErrorMessage',
-    function($scope, $state, $http, User, UserForm, Group, userId, ErrorMessage) {
+    function($scope, $state, User, UserForm, Group, userId, ErrorMessage) {
         Group.bindAll({where: {id: {'>': 2}}}, $scope, 'groups');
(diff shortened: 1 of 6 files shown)


openslides/users/static/js/users/site.js
```diff
@@ -898,34 +897,46 @@ angular.module('OpenSlidesApp.users.site', [
...
-                User.create(user).then(
-                    function(success) {
-                        $scope.importcounter++;
-                    }
-                );
+            });
+            $http.post('/rest/users/user/mass_import/', {
+                users: users
+            }).then(function (success) {
+                console.log(success);
```
normanjaeckel: Please remove this line.

openslides/users/views.py
```diff
@@ -114,6 +119,45 @@ def reset_password(self, request, pk=None):
...
+                groups = data['groups_id']
+                del data['groups_id']
+
+                db_user = User(**data)
+                db_user.save(skip_autoupdate=True)
+                db_user.groups.add(*groups)
+                created_users.append(db_user)
+                if 'importTrackId' in user:
+                    imported_track_ids.append(user['importTrackId'])
+            except ValidationError:
```
normanjaeckel: Please catch this exception only at the line that may raise it (serializer.is_valid(raise_exception=True) I think).

openslides/users/static/js/users/site.js
FinnStutzenstein: Done.

openslides/users/views.py
FinnStutzenstein: Done.

PR #3293, opened 2017-06-10
Block anonymous from websocket, if they are deactivated

ostcar:


ostcar:
You should also fix the client.
1. Start openslides and login as admin
2. make sure anonymous user is activated (if not, activate it and start with step1)
3. deactivate anonymous users
4. stop daphne (the client starts to try to reconnect)
5. flush redis (and therefore all user sessions)
6. start dahpne (the client gets a 403 as response because the session is "expired" and anonymous is deactivated).

When the client tries to connect via websocket and gets a 403 from the server, it call /whoami/ once more to see, if anonymous user is activated or not. If not, then the websocket-reconnection should be stopped and the user should be redirected to the login page.

normanjaeckel:
@emanuelschuetze Can you please check @ostcar's last comment and open an issue (e. g. for @FinnStutzenstein)? Then merge this.

emanuelschuetze:
>  open an issue

done, see #3295

PR #3297, opened 2017-06-14
improved reconnect handling if the server was flushed (fixed #3295)

### LINKED ISSUE #3295: Block anonymous from websocket, if they are deactivated (client)
Splitted up from #3293:

You should also fix the client.
1. Start openslides and login as admin
2. make sure anonymous user is activated (if not, activate it and start with step1)
3. deactivate anonymous users
4. stop daphne (the client starts to try to reconnect)
5. flush redis (and therefore all user sessions)
6. start dahpne (the client gets a 403 as response because the session is "expired" and anonymous is deactivated).

When the client tries to connect via websocket and gets a 403 from the server, it call /whoami/ once more to see, if anonymous user is activated or not. If not, then the websocket-reconnection should be stopped and the user should be redirected to the login page.

FinnStutzenstein:
If you flush the cache and restart daphne, the user is now forced to relogin and the client does not try to reconnect with websockets.

### DIFF
--- a/openslides/core/static/js/core/base.js
+++ b/openslides/core/static/js/core/base.js
@@ -48,10 +48,11 @@ angular.module('OpenSlidesApp.core', [
     'REALM',
     'ProjectorID',
     '$q',
+    '$timeout',
     'ErrorMessage',
-    function (DS, REALM, ProjectorID, $q, ErrorMessage) {
+    function (DS, REALM, ProjectorID, $q, $timeout, ErrorMessage) {
         var socket = null;
-        var recInterval = null;
+        var retryConnectCallbacks = [];
 
         var websocketProtocol;
         if (location.protocol == 'https:') {
@@ -69,6 +70,21 @@ angular.module('OpenSlidesApp.core', [
           console.error('The constant REALM is not set properly.');
         }
 
+        /* The callbacks are invoked if the ws connection closed and this factory tries to
+         * reconnect after 1 second. The callbacks should return a promise. If the promise
+         * resolves, the retry-process is stopped, so the callback can indicate whether it
+         * has managed the reconnecting different.*/
+        var runRetryConnectCallbacks = function () {
+            var callbackPromises = _.map(retryConnectCallbacks, function (callback) {
+                return callback();
+            });
+            $q.all(callbackPromises).then(function (success) {
+                ErrorMessage.clearConnectionError();
+            }, function (error) {
+                $timeout(runRetryConnectCallbacks, 1000);
+            });
+        };
+
         var Autoupdate = {};
         Autoupdate.messageReceivers = [];
         // We use later a promise to defer the first message of the established ws connection.
@@ -78,15 +94,14 @@ angular.module('OpenSlidesApp.core', [
         };
         Autoupdate.newConnect = function () {
             socket = new WebSocket(websocketProtocol + '//' + location.host + websocketPath);
-            clearInterval(recInterval);
+            // Make shure the servers state hasn't changed: Send a whoami request. If no users is logged and
+            // anonymous are deactivated, reboot the client in fact that the server has lost all login information.
             socket.onclose = function (event) {
                 socket = null;
-                recInterval = setInterval(function () {
-                    Autoupdate.newConnect();
-                }, 1000);
                 if (event.code !== 1000) { // 1000 is a normal close, like the close on logout
                     ErrorMessage.setConnectionError();
                 }
+                $timeout(runRetryConnectCallbacks, 1000);
             };
             socket.onmessage = function (event) {
                 var dataList = [];
@@ -115,6 +130,9 @@ angular.module('OpenSlidesApp.core', [
                 socket.close();
             }
             Autoupdate.firstMessageDeferred = $q.defer();
+        };
+        Autoupdate.registerRetryConnectCallback = function (callback) {
+            retryConnectCallbacks.push(callback);
         };
         return Autoupdate;
     }
--- a/openslides/core/static/js/core/start.js
+++ b/openslides/core/static/js/core/start.js
@@ -8,13 +8,14 @@ angular.module('OpenSlidesApp.core.start', [])
     '$http',
     '$rootScope',
     '$state',
+    '$q',
     'DS',
     'autoupdate',
     'operator',
     'Group',
     'mainMenu',
-    function($http, $rootScope, $state, DS, autoupdate, operator, Group, mainMenu) {
-        return {
+    function($http, $rootScope, $state, $q, DS, autoupdate, operator, Group, mainMenu) {
+        var OpenSlides = {
             bootup: function () {
                 $rootScope.openslidesBootstrapDone = false;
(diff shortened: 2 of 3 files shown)


openslides/core/static/js/core/start.js
```diff
@@ -46,6 +47,27 @@ angular.module('OpenSlidesApp.core.start', [])
                 this.bootup();
             },
         };
+
+        // We need to 'ping' the server with a get request to whoami, because then we can decide,
+        // if the server is down or respond with a 403 (this cannot be differentiated with websockets)
+        autoupdate.registerRetryConnectCallback(function () {
+            return $q(function (resolve) {
```
normanjaeckel: I think this extra promise envelop is not necessary. $http.get().then() returns a promise. Just just have to return true or false in the success and error callback.

openslides/core/static/js/core/start.js
```diff
@@ -46,6 +47,27 @@ angular.module('OpenSlidesApp.core.start', [])
...
+                $http.get('/users/whoami').then(function (success) {
+                    if (success.data.user_id === null && !success.data.guest_enabled) {
+                        OpenSlides.shutdown();
```
normanjaeckel: Why not just OpenSlides.reboot()? Maybe we should rewrite bootup, shutdown and reboot so that they return each a promise.

openslides/core/static/js/core/base.js
```diff
@@ -48,10 +48,12 @@ angular.module('OpenSlidesApp.core', [
     'REALM',
     'ProjectorID',
     '$q',
+    '$timeout',
+    'operator',
```
normanjaeckel: operator is not used here

openslides/core/static/js/core/base.js
```diff
@@ -69,6 +71,28 @@ angular.module('OpenSlidesApp.core', [
           console.error('The constant REALM is not set properly.');
         }
 
+        /* The callbacks re invoked if the ws connection closes and this factory tries to
+         * reconnect after 1s. The callbacks should return a promise resolveing to true/false.
+         * If resolving to true, the retry-process is stopped, so the callback can indicate
+         * whether it has done the reconnecting different.*/
+        var runRetryConnectCallbacks = function () {
```
normanjaeckel: I do not understand the docstring.

But you can simplfy this:

```javascript
var runRetryConnectCallbacks = function () {
    var callbackPromises = _.map(retryConnectCallbacks, function (callback) {
        return callback();
    });
    $q.all(callbackPromises).then(
        function (data) {
            // All promises resolved successfully. Now try to reconnect.
            Autoupdate.newConnect();
        },
        function (error) {
            // At least one promise resolved with a rejection. Stop everything ... ???
            ErrorMessage.clearConnectionError();
        }
    );
}
```

Now you have to change the callbacks that they return a promise that resolve or reject correctly.

openslides/core/static/js/core/base.js
```diff
@@ -48,10 +48,12 @@ angular.module('OpenSlidesApp.core', [
...
     'ProjectorID',
     '$q',
+    '$timeout',
+    'operator',
     'ErrorMessage',
-    function (DS, REALM, ProjectorID, $q, ErrorMessage) {
+    function (DS, REALM, ProjectorID, $q, $timeout, operator, ErrorMessage) {
         var socket = null;
-        var recInterval = null;
+        var retryConnectCallbacks = [];
```
normanjaeckel: Do we want to support multiple retry callbacks? I do not see any usecase for this.

openslides/core/static/js/core/start.js
FinnStutzenstein: Because we can save an extra whoami request --> the `$state.go` is just copied from the bootup, because the query around it is the same.

openslides/core/static/js/core/base.js
FinnStutzenstein: But now we have the option ;). I need to hook into this from the `OpenSlides` factory, because I cannot use this factory as a dependency in `autoupdate`. This leads to circular dependencies.
And if I make the option to register one ore more callbacks - this is not the point I think.

openslides/core/static/js/core/base.js
FinnStutzenstein: Good idea, use resolve/reject instead of passing true/false.

openslides/core/static/js/core/start.js
FinnStutzenstein: But in this case I sometimes want to reject in the success callback of that request. I tested returning true/false, but this doesn't work for me..

normanjaeckel: If you return any value inside the successCallback or errorCallback of `then()`, the new promise will be resolved. If you throw an error, the new promise will be rejected.

See https://docs.angularjs.org/api/ng/service/$q#the-promise-api and https://developers.google.com/web/fundamentals/getting-started/primers/promises#chaining

FinnStutzenstein: Yes, but imho throwing errors should not be used for flow control your code. So either returning true/false or use the $q wrapper to manually resolve/reject the promise.

normanjaeckel: I understand this approach but I think the current solution is not consistent: Rejecting a promise is the async way of throwing an error.

You can simplify your code if you consider the check here not as check of connection established and session flush but as check if OpenSlides has to be shut down (i. e. the retry process has to be stopped).

This pseudo code is only to explain how we can consider this case.
```javascript
return $http().then(function (success) {
    if !(......) {
        // Error because we expected the session to be gone but whoami returned a valid user id.
        // So throw this "error". This is the same as the case that the server is down.
        throw ...
    }
    // Now do something for the expected case: OpenSlides has to be shut down, the retry process will stop, redirect to login page ...
}); // no error callback so that the promise is rejected automaticly in this case (is this true, I am not sure)
```

Now I see a new problem: We send the whoami request and if the server is down we immediatelly try to connect to websocket. This fails. Then we wait 1000 ms for the next try. I think this is a strange flow ...

FinnStutzenstein: Ok for the first part.

Whats your idea? I think we can just ping the server with the whoami request, and if this works, we can reconnect the websocket?

FinnStutzenstein:
@normanjaeckel Please look at the new commit. Now the client pings the server just via whoami request. Three options then:

1. Connections fails --> promise is rejecting --> try again in 1000 ms
2. Connection ok, user is logged in --> a new websocket connection is established
3. Connection ok, but user is not logged in anymore (server is flushed) --> shutdown os and force new login.

ostcar:
No. 3 has to possibilities:
No. 3a: Connection ok, but user is not logged in anymore and anonymous is disabled
No. 3b: Connection ok, but user is not logged in anymore and anonymous is enabled.

At least when you start openslides, this two cases are handled differently. I have no opinion, if they should also be handled differently in your scenario. I only want to point out, that you could handle them differently.

emanuelschuetze:
Successfully retested. CHANGELOG entry added. Rebased. Ready for merge.

PR #3300, opened 2017-06-16
Updated js pot file and German translations.

emanuelschuetze:


emanuelschuetze:
I'll merge it directly.

PR #3301, opened 2017-06-16
Highlight objects in list view, if related objects are projected

FinnStutzenstein:
@emanuelschuetze Color highlighting is done. Whats about a hint for the user (like 'the list of speakers is projected')? Where to put that and how to fit this nice into the template?

Currently highlighted:
 - projecting a list of speakers for motions/assignments (highlight in agenda list view and motion/assignment list view)
 - projecting any ballot of an assignment.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -50,6 +50,8 @@ Core:
   clients by refactoring autoupdate, Collection and AccessPermission [#3223].
 - Fixes autoupdate bug for a user without user.can_see_name permission [#3233].
 - Improved reconnect handling if the server was flushed [#3297].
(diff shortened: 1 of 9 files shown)


emanuelschuetze:
> Color highlighting is done.

Looks good! Not really self-explaining in the first moment but after some exercises an admin should get the different between dark blue (main slide is projected) and light blue (related slide, e.g. list of speakers, is projected). Then this new light blue is helpful for orientation in agenda.

> Whats about a hint for the user (like 'the list of speakers is projected')? 

I think we don't need a hint. I see no place where we can show it. So let us try this with out any hint.

PR #3304, opened 2017-06-18
Prevent colliding change recommendations - fixes #3298

### LINKED ISSUE #3298: Block double change recommendations
If I double-click on the save button when creating a change recommenadtion. thwo of them are created and I get massive js errors:
wrong_change_reco

@CatoTH Do you have time to fix this on server side? Maybe make the combination (motion_id, line_from, line_to) unique?

CatoTH:


CatoTH:
Please have a look at the code, as I'm not that confident in my Django skills yet :)

normanjaeckel:
I have some code styling remarks but I can update your code myself if you like.

But my main question is whether you want to accept one line overlapping or not: Example:

A changerecommendation ranges from line 10 to 12. A new one cames if with range 8 to10. Do you expect the server to accept it? (Same for 10 to 12 and 12 to 20).

CatoTH:
@normanjaeckel One line overlap is correct. The numbers are referring to the line number marker in the HTML, not the whole line. So if you select line 1 to 2 in the user interface, it's stored as 1 to 3 internally, with 3 (as the line number 3) marking the end of line two.

emanuelschuetze:
@CatoTH Travis failed.
@normanjaeckel please review again. Is something to do for you/@CatoTH?
@FinnStutzenstein Please test (and review if you like).

openslides/motions/models.py
```diff
@@ -771,6 +771,25 @@ class MotionChangeRecommendation(RESTModelMixin, models.Model):
     creation_time = models.DateTimeField(auto_now=True)
     """Time when the change recommendation was saved."""
 
+    def collides_with_other_recommendation(self):
+        recommendations = MotionChangeRecommendation.objects\
```
FinnStutzenstein: You can use brackets around the if condition (e.g. [here](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/motions/views.py#L163)) to avoid the backslash; the same three lines below.

CatoTH: Ok, I updated the branch to use brackets instead of backslashes.

CatoTH:
@FinnStutzenstein WorkflowError always results in a 500, doesn't it? I agree that 400 would be more correct, however the same applies for pretty much every other situation in which WorkflowError is raised in models.py ("You can not create a poll in state %s." etc.). So maybe it would make more send to generally catch this error and return a 400? Not only in this specific case.

@emanuelschuetze the failing test seems to be due to test coverage. If you can point me to any other test case that tests models, I can try to write a test case for it - however I only see test cases for views right now...

openslides/motions/models.py
```diff
@@ -771,6 +771,25 @@ class MotionChangeRecommendation(RESTModelMixin, models.Model):
...
+        for recommendation in recommendations:
+            if not ((self.line_from < recommendation.line_from and self.line_to <= recommendation.line_from) and
+                    not (self.line_from >= recommendation.line_to and self.line_to > recommendation.line_to)):
+                return True
+
+        return False
+
+    def save(self, *args, **kwargs):
+        if self.collides_with_other_recommendation():
+            raise WorkflowError('The recommendation collides with an existing one (line %s - %s).' %
```
normanjaeckel: Please do not use `WorkflowError` because this means an error concerning motion workflows (see models Workflow and State) and should not have anything to to with change recommendations.

You can add a new exception in motions/exceptions.py and optionally catch this exception somewhere in the change recommendation viewset. Or even better use a ValidationError. This should return a 400 code in the viewset.

openslides/motions/models.py
```diff
@@ -771,6 +771,25 @@ class MotionChangeRecommendation(RESTModelMixin, models.Model):
...
+                                (self.line_from, self.line_to))
+
+        super(MotionChangeRecommendation, self).save(*args, **kwargs)
```
normanjaeckel: Please remove the content of the brackets and use return: `return super().save(*args, **kwargs)`

I know that Django's save does not return anything at the moment but this may change in future and this way we show that we only want to step short before the exact save function. (No big point of course. :-)

normanjaeckel:
To fix Travis it looks like that we need a short unit test.

CatoTH:
Ok, added unit tests, integration tests and a 400 Bad Request

openslides/motions/views.py
```diff
@@ -449,6 +450,15 @@ def check_view_permissions(self):
             result = False
         return result
 
+    def create(self, request, *args, **kwargs):
+        """
+        Creating a Change Recommendation, custom excaption handling
```
emanuelschuetze: typo: "exception"

CatoTH:
@emanuelschuetze I fixed the typo. But I can't reproduce the problem right now, I tried it both with a single and with multiple lines - every time I got 400 Bad Requests for subsequent calls...
Did you start the server anew after changing to the PR-branch?

emanuelschuetze:
@FinnStutzenstein Can you approve this PR if everything is ok for you? Then I'll merge it.

normanjaeckel:
Maybe someone wants to add a CHANGELOG entry?

FinnStutzenstein:
@emanuelschuetze Works for me, even without throttling..

CatoTH:
Just added a changelog entry, hope that's ok

CHANGELOG
```diff
@@ -28,6 +28,8 @@ Motions:
   client [#3212].
 - Reworked DOCX export parser and added comments to DOCX [#3258].
 - New PDF export for personal note and comments [#3239].
+- Bugfix: Creating colliding change recommendation is now prevented
+  on server side [#3298]
```
emanuelschuetze: Please use the PR number #3304 in changelog - not the issue number.

CatoTH:
Ah ok, done.

emanuelschuetze:
Thanks!

Issue #2417, opened 2016-09-22
Projector container should not listen on the same channel as real projector.


normanjaeckel:
At the moment we send mass of data to the projector container but it uses only some small projector and config data. We should not send all other stuff (motions and thelike) to it.

@ostcar How can we do this?

ostcar:
At the moment the projector container uses the loadGlobalData function. This is not necessary, because it is called manually. You could call Config.findAll() in it instead.

FinnStutzenstein:
With multiprojector the resolution does not come from config, but from the projector which is shown in the iframe. So this should be changed to Projector.find(id). The projector container should only get the displayed projector as data and autoupdates for this, if the resolution has changed.

normanjaeckel:
Yeah but what about the autoupdate data? The projector container still connects to ws/projector I think.

ostcar:
Could we save the resolution right now to the projector element and not to the config, so we do not have to change the concept in the future?

FinnStutzenstein:
Can we make the projectorcontainer connect to site and specify in get_requirements the required projector. Technical the projector container isn't a projector. It is some kind of site which shows the projector, so this would make sense?

emanuelschuetze:
As discussed with @normanjaeckel and @FinnStutzenstein 
We need a 3rd channel for projector container which sends the required data (projector, some configs) only. Not really important for 2.1.

normanjaeckel:
This is high for more than three months. Maybe it is not high ... ?

jsaalfeld:
Since the client will be completely rewritten, this shouldn't be an issue anymore. Please reopen if I'm wrong.

PR #3305, opened 2017-06-21
fix projector reconnecting if websocket is closed

FinnStutzenstein:
See line 22-24: This is what we do on the gui side of OpenSlides. But for the projector we do not have `$state`, so what should we do if the user is logged off (and the default user - if activated - has no permission to see the projector) ?

### DIFF
--- a/openslides/core/static/js/core/projector.js
+++ b/openslides/core/static/js/core/projector.js
@@ -9,9 +9,24 @@ angular.module('OpenSlidesApp.core.projector', ['OpenSlidesApp.core'])
 .constant('REALM', 'projector')
 
 .run([
+    '$http',
     'autoupdate',
-    function (autoupdate) {
+    'DS',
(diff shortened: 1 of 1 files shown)


emanuelschuetze:
@normanjaeckel @FinnStutzenstein Can we finish and merge this pullrequest today? This is a very critical issue for production use. I want to create a new version for customer this evening.

FinnStutzenstein:
@emanuelschuetze This can be merged now. If the server is flushed, the data shore will be cleared. This means that the message will be projected. The users has to logon again and reload the projector. But this should be a rare case (just a critical situation).

@normanjaeckel If you want to handle this different, you can modify this as you want or create a follow up issue/pr.

FinnStutzenstein:
Remark: This has to be redone, if we separate the projector and the projector container (see issue #2417 )

PR #3306, opened 2017-06-21
remove OPTION request to query permissions (fixes #2862)

### LINKED ISSUE #2862: Remove OPTIONS request for group permissions
See `getPermissions`. We can parse all existing permissions from all groups instead.

FinnStutzenstein:


emanuelschuetze:
travis failed

FinnStutzenstein:
@normanjaeckel How do I test these methods?

normanjaeckel:
You may add one or more tests for the `WebclientJavaScriptView`. See tests/unit/agenda/test_views.py or tests/unit/motions/test_views.py how to write unit tests for our views.

FinnStutzenstein:
In the [GroupListCtrl](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/users/static/js/users/site.js#L1255) the permissions are used.

FinnStutzenstein:
@normanjaeckel For testing: I think to test this view, we have to patch too much; `get_angular_constants` may called in every app; If we add more constants, we have to path more apps. I don't think this is very usefull..

normanjaeckel:
Hm. Isn't it possible to test this view only partly (e. g. inside the user app): Test that the result contains the permissions content (but do not test all response content of this view).

FinnStutzenstein:
I can especially mock the function in the users app, but never the less I cannot test the whole JavaScriptView. So what we can do is to test the app itself?

normanjaeckel:
Why not add in users app a test like this?
```python
def text_...(self):
    request = MagicMock()
    view_instance = WebclientJavaScriptView()
    view_instance.request = request
    response = view_instance.get(request)
    self.assertTrue(".constant('permissions', ...)" in str(response.content))
```

tests/unit/core/test_views.py
```diff
@@ -164,3 +164,14 @@ def test_clear_elements_with_stable(self, mock_object):
...
+
+
+class WebclientJavaScriptView(TestCase):
+    def setUp(self):
+        self.request = MagicMock()
+        self.view_instance = views.WebclientJavaScriptView()
+        self.view_instance.request = self.request
+
+    def test_app_configs(self):
+        response = self.view_instance.get()
```
FinnStutzenstein: If I do this, django queries all permissions (e.g. see travis traceback): `for permission in Permission.objects.all()`.
This causes a `settings.DATABASES is improperly configured`-error. To execute the view, I have to mock everything where a DB query could occur, but this view is highly dynamic (finding plugins and special method in the app configs), that mocking all this is not good..

normanjaeckel: Hm. Ok. What about mocking the Permissions.objects.all() call only? Then the test is only about that the key permissions occurs in the WebClientJavascriptView as AngularJS constant.

PR #3308, opened 2017-06-21
call the own project method from a content object of an agenda item

FinnStutzenstein:
We may rewrite the `isProjected` to call this method direct on the content object. I'll make an issue, if this is merged.


emanuelschuetze:
@FinnStutzenstein Just as reminder:

> We may rewrite the isProjected to call this method direct on the content object. I'll make an issue, 
> if this is merged.

PR #3322, opened 2017-07-26
adding DEBUG flag (WIP)

FinnStutzenstein:
@normanjaeckel Does Django recognize, if I change the settings.DEBUG at this place? My tests worked so far. Is there anything to rethink of?

@normanjaeckel @emanuelschuetze If you have ideas for important debug output, please leave them here..

### DIFF
--- a/openslides/__main__.py
+++ b/openslides/__main__.py
@@ -129,6 +129,10 @@ def get_parser():
         '--use-geiss',
         action='store_true',
         help='Use Geiss instead of Daphne as ASGI protocol server.')
+    subcommand_start.add_argument(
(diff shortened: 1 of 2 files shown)


openslides/__main__.py
```diff
@@ -129,6 +129,10 @@ def get_parser():
         '--use-geiss',
         action='store_true',
         help='Use Geiss instead of Daphne as ASGI protocol server.')
+    subcommand_start.add_argument(
```
normanjaeckel: I do not like this extra command line option. I think the settings value itself is sufficient.

FinnStutzenstein:
Closed to do this later.

PR #3324, opened 2017-07-26
Configvalue for sequential number

FinnStutzenstein:


FinnStutzenstein:
@normanjaeckel I do not think so.

openslides/motions/config_variables.py
```diff
@@ -253,3 +253,12 @@ def get_config_variables():
         weight=375,
         group='Motions',
         subgroup='Export')
+
+    yield ConfigVariable(
+        name='motions_export_sequential_number',
+        default_value=True,
+        input_type='boolean',
+        label='Include the sequential number in the PDF and DOCX',
```
emanuelschuetze: Please mark this string for translation in site.js (and change to " ... number in PDF and DOCX" - without "the")

PR #3327, opened 2017-07-26
Configvalue for pagenumber alignment in PDF

FinnStutzenstein:


FinnStutzenstein:
Done.

openslides/motions/config_variables.py
```diff
@@ -253,3 +253,16 @@ def get_config_variables():
         weight=375,
         group='Motions',
         subgroup='Export')
+
+    yield ConfigVariable(
+        name='motions_export_pdf_pagenumber_alignment',
```
emanuelschuetze: I think about to move this config in the general config group because it's valid for all pdf generation - not only for motions. What do you think?
Remark: Please add label and value strings to site.js that we can translate all strings later.

FinnStutzenstein:
@emanuelschuetze Please look again.

FinnStutzenstein:
@emanuelschuetze Done.

PR #3330, opened 2017-07-27
Improvements for plugin integration

FinnStutzenstein:


openslides/core/static/js/core/base.js
```diff
@@ -493,22 +499,46 @@ angular.module('OpenSlidesApp.core', [
...
+                    var promises = _.map(hooks, function (hook) {
+                        // Either a template (html given as string) or a templateUrl has
+                        // to be given. If a scope is provided, the schope of this templateHook
+                        // is populated with the given functions/values.
+                        if (hook.template) {
+                            return $q(function (resolve) {
+                                resolve('<div>' + hook.template + '</div>');
+                            });
+                        } else {
+                            return $q(function (resolve) {
```
normanjaeckel: This wrapper around $http is not necessary I think. You can just return the $http.get().then(). Inside the then() function you may just return success.data.

openslides/core/static/js/core/base.js
```diff
@@ -476,6 +476,12 @@ angular.module('OpenSlidesApp.core', [
 ])
 
 // Template hooks
+// 2 possible uses:
+// - { Id: 'myHookId', template: '<button>click me</button>' }
+// - { Id: 'myHookId', templateUrl: '/static/templates/plugin_name/my-hook.html' }
```
normanjaeckel: I wonder whether there exists an AngularJS intern lib function to do what we want to do here: lazy loading of templates. If you don't have any idea just use this impelementation. It is ok for me.

openslides/core/static/js/core/base.js
```diff
@@ -493,22 +499,46 @@ angular.module('OpenSlidesApp.core', [
...
+                        } else {
+                            return $q(function (resolve) {
+                                $http.get(hook.templateUrl).then(function (success) {
+                                    resolve(success.data);
+                                });
+                            });
+                        }
+                    });
+                    // Populate scope
+                    _.forEach(hooks, function (hook) {
```
normanjaeckel: I wonder whether it is useful to do the scope stuff inside the first "loop" with the templates.

openslides/core/static/js/core/base.js
```diff
@@ -476,6 +476,12 @@ angular.module('OpenSlidesApp.core', [
 ])
 
 // Template hooks
+// 2 possible uses:
+// - { Id: 'myHookId', template: '<button>click me</button>' }
+// - { Id: 'myHookId', templateUrl: '/static/templates/plugin_name/my-hook.html' }
+// It is possible to provide a scope, that is merged into the scope of the templateHook.
```
normanjaeckel: Can you please extend the comment with an example?

openslides/core/static/js/core/base.js
FinnStutzenstein: I searched a bit but I didn't find any good solution. This implementation is very small and on point whats going on, so I would rather keep it.

openslides/core/static/js/core/base.js

openslides/core/static/js/core/base.js

openslides/core/static/js/core/base.js

FinnStutzenstein:

openslides/core/static/js/core/base.js
normanjaeckel: Ok, now I Know: All of our templates are cached via gulp in the template cache and I would expect any plugin also to to so. See gulpfile line 89: Any plugin should transform all templates to JS and use the hooks inside app.py to add these JS-files to the client. Then you do not need to do this hot reloading bay hand here. AngularJS should just know the template under its path.

normanjaeckel:
Please also add CHANGELOG entry.

emanuelschuetze:
Changelog entry added. Rebased.

PR #3332, opened 2017-07-27
Added config option for standard font size in PDF.

emanuelschuetze:


emanuelschuetze:
@normanjaeckel I have to merge it today. You can still review/remark after pr is merged.

openslides/core/config_variables.py
```diff
@@ -130,6 +130,19 @@ def get_config_variables():
         group='General',
         subgroup='Export')
 
+    yield ConfigVariable(
+        name='general_export_pdf_fontsize',
+        default_value='10',
+        input_type='choice',
+        label='Standard font size in PDF',
+        choices=(
+            {'value': '10', 'display_name': '10'},
```
normanjaeckel: Why don't you want to enable more font sizes by enabling a big range instead of this small choice. Minimum 6. Maximum 64 or even no maximum.

emanuelschuetze: It's just the _standard_ font size for text which sould be between 10 and 12pt. More is not useful because font size of h3 is still 12pt. And smaller than 10 is not readable for text; also because line numbers always 2 pt smaller (10pt text -> 8pt linenumbers). 
So, won't fix.

normanjaeckel: Ok, but this is not obvious. You mean font size for standard text. Standard font size sounds like a basic fontsize from which all other font sizes are derived. Maybe change label.

PR #3333, opened 2017-07-27
Updated German translations.

emanuelschuetze:
Fix also the order of  general config values.

### DIFF
--- a/openslides/core/config_variables.py
+++ b/openslides/core/config_variables.py
@@ -105,7 +105,7 @@ def get_config_variables():
         group='General',
         subgroup='System')
 
-    # CSV
+    # General export settings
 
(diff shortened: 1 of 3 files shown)


openslides/core/config_variables.py
```diff
@@ -126,7 +124,7 @@ def get_config_variables():
             {'value': 'left', 'display_name': 'Left'},
             {'value': 'center', 'display_name': 'Center'},
             {'value': 'right', 'display_name': 'Right'}),
-        weight=146,
+        weight=145,
```
normanjaeckel: Can we please change the values so that there is some space between them again. The intention is that plugins may go inbetween with their values.

Issue #3016, opened 2017-02-27
MotionCreateCtrl sends illegal PUT request to agenda


normanjaeckel:
Login as delegate and create a new motion. The client sends a PUT request to the agenda. The server sends 403 (which is correct in this case) but the client shouldn't try to send this request.

FinnStutzenstein:
This is the same issue as #2779 . The client tries to modify the created agenda item, but it has to wait for the motion (and so the agenda item) to be created. If the agenda item is more coupled with the containing object, this should be done on server side.

emanuelschuetze:
I moved this to 2.2 because #2779 has also 2.2 milestone.

emanuelschuetze:
Already fixed.

PR #3341, opened 2017-08-03
Supported server side setup of new agenda items (type and parent).

normanjaeckel:
@FinnStutzenstein First step here. Now we can implement the client sending extra fields like type and parent on creation of agenda related items like motion, motion block, topic and assignment. Do you want to do some things here like changeing client for e. g. assignment. Then we can go on with the server ...

See #3016, #2886, #2779.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,7 @@ Agenda:
 - Fixed wrong sorting of last speakers [#3193].
 - Fixed issue when sorting a new inserted speaker [#3210].
 - New permission for managing lists of speakers [#3366].
+- Fixed multiple request on creation of agenda related items [#3341].
 
 Motions:
 - New export dialog [#3185].
--- a/openslides/agenda/signals.py
+++ b/openslides/agenda/signals.py
@@ -14,16 +14,25 @@ def listen_to_related_object_post_save(sender, instance, created, **kwargs):
     Receiver function to create agenda items. It is connected to the signal
     django.db.models.signals.post_save during app loading.
 
-    Do not run caching and autoupdate if the instance as an attribute
-    skip_autoupdate (regardless of its truthy or falsy conent).
+    The agenda_item_update_information container may have fields like type,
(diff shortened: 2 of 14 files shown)


FinnStutzenstein:
@normanjaeckel If you create an assignment the field `agenda_type` and `agenda_parent_id` will be in the request data. On an update just the `agenda_type`will be provided. The server has to check the permissions to set this field. If this is OK, I will remove all uncommented stuff and change topics and motions.

normanjaeckel:
I prefer that we remove all agenda related stuff from the all update forms. (At the moment the form does not reflect the database status properly (the form field refers to 'is_hidden' and not to 'type') So we should offer this feature only for create views. @emanuelschuetze?

FinnStutzenstein:
OK for me, @emanuelschuetze ?

emanuelschuetze:
@normanjaeckel Please let us discuss about this pr.

emanuelschuetze:
> I prefer that we remove all agenda related stuff from the all update forms.

+1 (as discussed with @normanjaeckel)

normanjaeckel:
@FinnStutzenstein Please review and finisch assignment app (remove form fields in update form etc.).

normanjaeckel:
Note for me:
* MotionCreate, MotionUpdate
* The same for MotionBlock
* The same for Topic

normanjaeckel:
@FinnStutzenstein Ready for next test (I had only some time for a short test.). Please also have a look at my changes on the client.

Maybe we also have to fix TopicImport. Any idea here?

FinnStutzenstein:
@normanjaeckel Works fine for me with topics, motions, motion blocks and assignments. Just one request during creation and no agenda related settings in the update forms. With my last commit, this should be ready to merge.

normanjaeckel:
I fixed topic import. So please retest (also topic import). Then merge.

emanuelschuetze:
Fixed travis (missing semicolon).

PR #3342, opened 2017-08-04
Fixing strange umlauts when copy text from word

FinnStutzenstein:
This [unicode char](http://www.fileformat.info/info/unicode/char/0308/index.htm) is behind a 'u o a' to make the dots above.


FinnStutzenstein:
@normanjaeckel Added and simplified the replace function.

normanjaeckel:
Ok. Please think about CHANGELOG. Then merge.

emanuelschuetze:
There is already the pr number in the Changelog entry ("Bugfixes for PDF creation").

PR #3345, opened 2017-08-09
some changes in the example data plugin

FinnStutzenstein:
@normanjaeckel Whats about checking, if `user0` uo to `userXX` exists and then to create e.g. 100 new users `userXX+1` up to `userXX+100`? This happens, if the script is runned multiple times. As a quick bug fix I catch the IntegrityError, so the script doesn't break.

### DIFF
--- a/openslides/utils/settings.py.tpl
+++ b/openslides/utils/settings.py.tpl
@@ -23,6 +23,9 @@ OPENSLIDES_USER_DATA_PATH = %(openslides_user_data_path)s
 
 INSTALLED_PLUGINS += (
 #    'plugin_module_name',
+
+# Built-in plugins:
+#    'tests.example_data_generator',
 )
 
 INSTALLED_APPS += INSTALLED_PLUGINS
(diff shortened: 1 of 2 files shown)


openslides/utils/settings.py.tpl
```diff
@@ -23,6 +23,9 @@ OPENSLIDES_USER_DATA_PATH = %(openslides_user_data_path)s
 
 INSTALLED_PLUGINS += (
 #    'plugin_module_name',
+
+# Buildin plugins:
```
normanjaeckel: Typo: `Built-in`

emanuelschuetze: Fixed.

emanuelschuetze:
@normanjaeckel Can you give us your opinion:

> Whats about checking, if user0 uo to userXX exists and then to create e.g. 100 new users 
> userXX+1 up to userXX+100? This happens, if the script is runned multiple times. As a quick bug 
> fix I catch the IntegrityError, so the script doesn't break.

PR #3346, opened 2017-08-11
removing obsolete merge-before

FinnStutzenstein:
>  Basically, the merge-before-class would become obsolete and is replaced by os-split-before (and os-split-after), indicating that the given tags has been split by the diff and no extra margin / padding should be applied before / after it.

(From issue #3232 )

Just replace the merge-before in the pdf parser. Added a little helper function, to check classes in elements. Do not use the `e.getAtttribute('class') == 'my-class'` because if `e` has additional classes, the test will fail.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -74,7 +74,7 @@ General:
 - Switched from npm to Yarn [#3188].
 - Improved performance for PDF generation significantly (by upgrading
   to pdfmake 0.1.30) [#3278, #3285].
-- Bugfixes for PDF creation [#3227, #3251, #3279, #3286].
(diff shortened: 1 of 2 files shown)


emanuelschuetze:
Works good for paragraphs. But if you make changes in list elements there is still undesired margin on top and bottom...

emanuelschuetze:
Please add to commit message that #3232 is fixed.

emanuelschuetze:
Now, I get a "TypeError: Cannot read property 'value' of undefined" while generating pdf...

emanuelschuetze:
Sorry, my fault (old database with old configs)...

emanuelschuetze:
I added PR number to existing changelog entry.

PR #3348, opened 2017-08-11
Default sorting for users, mediafiles and assignments (fixes #3344)

### LINKED ISSUE #3344: Generic solution for order issues in list views
We need a generic solution for all list views to sort correctly (also user selected a custom sort parameter). See #3334  for motions.

FinnStutzenstein:


FinnStutzenstein:
@normanjaeckel Done.

PR #3351, opened 2017-08-11
Exclude special comments in comment field settings; fixing the inline…

FinnStutzenstein:
… editiing


emanuelschuetze:
Works good (filter and inline-editing). Wish: can you please hide the comment tile in motion detail template if no custom comment is configured? That means: if only some special (hidden) comments are set up the comment tile shoult not provide pdf and inline editing buttons -> please hide whole tile.

FinnStutzenstein:
@emanuelschuetze Try again.

PR #3355, opened 2017-08-16
Mediafile fullscreen (fixes #3169)

### LINKED ISSUE #3169: displaying PDF files always uses fullscreen
- upload a PDF to OS
- display it

result: OS uses always fullscreen. You can press the button to turn off fullscreen but nothing will happen.

FinnStutzenstein:
Done @emanuelschuetze two suggestions. Is the undo-image ok?


FinnStutzenstein:
Done.

PR #3356, opened 2017-08-16
Clear the identifier_number on identifier reset (fixes #3353)

### LINKED ISSUE #3353: Wrong motion identifier generation after reset state
How to reproduce:
* Set config for default motion worklow to 'complex workflow'
* Create new category (prefix "A").
* Create new motion.
* add category to motion.
* Set state from "published" to "permitted" -> motion gets identifier "A 1".
* Reset state (back to "published").
* Remove identifier via edit motion form.
* Set state again to "permitted" -> FAILURE: motion gets identifier "A 2" (instead of "A 1).

FinnStutzenstein:


normanjaeckel:
I am not sure whether this is the solution: Create two motions in complex workflow: A 1 and A 2. The clear identifier number of A 1 and click reset state and permit. Or even: clear identifier number of A 1 and rename A 2 to A 1, then click reset state and permit (for former A 1). Maybe we should first think about what we expect from this auto-numbering feature.

emanuelschuetze:
@normanjaeckel Note, admin users can also sort and renumber all motions of a category in the category sort view. So there is a workaround to fix the current wrong numbering. But it is very annoying if each new motion gets a wrong number... So for our case this fix works good!

FinnStutzenstein:
Yes, this is a special case. But I don't thing a general solution is efficient nor effective here.

normanjaeckel:
Sorry, this pull request does not solve the problem:

Fresh DB. Add two new motions in complex workflow and in category A. Then click permit for both. The click reset_state for A 1 and remove the identifier via edit form. The click for the first motion permit again. It get's now A 3.

Even worse: Add two permitted motions in complex workflow like above. Now edit A 2 and remove identifier. Then regret your decision and set identifier to `A 2x`. Now I would expect every new motion to get A 3 but if you get A 2 instead. This breaks the current (expected) behavior.

So I strongly recommend not to connect the edit form with the internal identifier number.

It might be possible to clear the indentifier_number in reset_state view or model_method but this has also unexpected side effects.

FinnStutzenstein:
@emanuelschuetze Whats your opinion here? Clearing the identifier_number in the reset_state method should work, too. Are you concerned about side effects here? If we do not want the current behavior, we should exactly say what OS should do here, so we can implement this right.

emanuelschuetze:
@normanjaeckel Let us discuss this pr.

emanuelschuetze:
@FinnStutzenstein 
As discussed with @normanjaeckel please try to clear the identifer number only in case of reset state (not in edit form) - in model method if possible.

FinnStutzenstein:
@emanuelschuetze Check, if this solution is ok.

openslides/motions/models.py
```diff
@@ -581,13 +581,20 @@ def set_state(self, state):
...
         'workflow' can be a workflow, an id of a workflow or None.
 
         If the motion is new and workflow is None, it chooses the default
         workflow from config.
         """
+
+        # The identifier is set automatically.
+        if config['motions_identifier'] is not 'manually':
+            self.identifier = ''
+            self.identifier_number = 0
```
emanuelschuetze: Now, the reset action removes the internal `identifier_number` AND the motion `identifier`. Manually removing of identifier in edit form is not required. Successfully tested. It's would be ok for me.
@normanjaeckel Can you live with this solution. As we discussed only the internal number should removed, right?

normanjaeckel: This is ok for me. But `identifier_number` should be set to `None` instead of `0`.

PR #3358, opened 2017-08-18
Add restricted_data_cache

ostcar:
Replace the startup cache with a restricted data cache


ostcar:
@emanuelschuetze do you want to test this or shut I merge it?

If you want to test it, you should test it with anonymous user and with and without redis.

PR #3360, opened 2017-08-18
Pin personal note.

FinnStutzenstein:
@emanuelschuetze Does this work for you?

### DIFF
--- a/openslides/core/static/templates/index.html
+++ b/openslides/core/static/templates/index.html
@@ -210,7 +210,7 @@
     </div>
 
     <!-- col1 -->
-    <div class="col1" ng-class="{
+    <div id="main-column" class="col1" ng-class="{
         'sidebar-max': isProjectorSidebar && operator.hasPerms('core.can_see_projector'),
         'sidebar-min': !isProjectorSidebar && operator.hasPerms('core.can_see_projector'),
         'sidebar-none': !operator.hasPerms('core.can_see_projector') }">
--- a/openslides/motions/static/js/motions/site.js
+++ b/openslides/motions/static/js/motions/site.js
@@ -1485,6 +1485,10 @@ angular.module('OpenSlidesApp.motions.site', [
             return Boolean(isAllowed);
         };
         // personal note
+        // For pinning the personal note container we need to adjust the width with JS. We
+        // do not use angular here, because on every window resize a digist cycle would trigger.
+        // This costs too much performance. We use JQuery here, because it is fast for DOM
+        // manipulation and very responsive.
         $scope.toggleStar = function () {
             if ($scope.motion.personalNote) {
                 $scope.motion.personalNote.star = !$scope.motion.personalNote.star;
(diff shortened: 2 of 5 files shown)


openslides/core/static/templates/index.html
```diff
@@ -210,7 +210,7 @@
     </div>
 
     <!-- col1 -->
-    <div class="col1" ng-class="{
+    <div id="col1" class="col1" ng-class="{
```
normanjaeckel: Can we please use a more explicit id (like sidebar for col2)? (main-column or something like this)

openslides/motions/static/js/motions/site.js
```diff
@@ -1469,6 +1469,22 @@ angular.module('OpenSlidesApp.motions.site', [
             }
             PersonalNoteManager.saveNote($scope.motion, $scope.motion.personalNote);
         };
+        $scope.personalNotePinned = false;
+        $scope.pinPersonalNote = function () {
+            $scope.personalNotePinned = !$scope.personalNotePinned;
+            if ($scope.personalNotePinned) {
+                resizePersonalNoteContainer();
+            } else {
+                $('#personalNote').css('width', '');
```
normanjaeckel: Please use AngularJS like DOM manipulation methods and not jQuery.

openslides/motions/static/js/motions/site.js
```diff
@@ -1469,6 +1469,22 @@ angular.module('OpenSlidesApp.motions.site', [
...
+            }
+        };
+        var resizePersonalNoteContainer = function () {
+            if ($scope.personalNotePinned) {
+                var width = $('#col1').width() - 20; // 2x10px margin
+                $('#personalNote').css('width', width + 'px');
```
normanjaeckel: Here the same.

openslides/motions/static/js/motions/site.js
FinnStutzenstein: Does angular has a nicer Way of doing this?

FinnStutzenstein: Using ng-style here is not very good: on every resize an angular digest cycle has to be started, so angular can update the directive. This causes many laggs.

normanjaeckel: Hm. I have to check this ...

FinnStutzenstein: One spontaneous idea:
<tag style="width: {{ width }}px;"></tag>
I'll check if this works better..

FinnStutzenstein: Ok, this is also angular related. If this attribute should change, a digest cycle is necessary.

I would prefer using just jquery, because always starting a new digest cycle slows down the performance a lot in fact that angular reloads everything in the background. JQuery is grea for just DOM manipulation that should happend fast and responsive and angular should not bother with the width of a fixed tile.

normanjaeckel: Ok. I still don't like it but we can do it that way. Please add a short comment at the top of the code block for this feature (before line 1487 with extra newline) and explain in three words that we use a JQLite instead of AngularJS stuff for simplification at this place. So we later can remember the reason for it.

FinnStutzenstein: Done.

FinnStutzenstein:
@emanuelschuetze Ready for merge.

PR #3361, opened 2017-08-18
Inline editing for motion reason

FinnStutzenstein:


FinnStutzenstein:
@emanuelschuetze Test again

PR #3370, opened 2017-08-19
Add typing

ostcar:


normanjaeckel:
Travis should be fixed, maybe as we discussed today (using a condition in isort call via env var refering to Python 3.4).

normanjaeckel:
Hm. Pull request looks ok but approvals are missing ... :-)

ostcar:
they cam via mail

PR #3371, opened 2017-08-22
Rewrite config to have id field

ostcar:
Refactor the config api to have a normal id field


ostcar:
@normanjaeckel can you look at the two tests I disabled?

tests/unit/config/test_api.py
```diff
@@ -23,11 +22,3 @@ def test_default_value_in_data(self):
             'test_default_value',
             "The value of config_variable.data['default_value'] should be the same "
             "as set as second argument of ConfigVariable()")
-
-
-class TestConfigHandler(TestCase):
```
normanjaeckel: This test should not be deleted. It should also work and test that `config['unknown']` leads to the specific error in server code.

openslides/core/config.py
```diff
@@ -122,6 +138,7 @@ def __setitem__(self, key, value):
         try:
             db_value = ConfigStore.objects.get(key=key)
         except ConfigStore.DoesNotExist:
+            # This should never happen. All DB entries are created at startup.
```
normanjaeckel: So we can remove this try-except-block?

openslides/utils/collection.py
```diff
@@ -41,16 +41,13 @@ def __init__(self, instance=None, deleted=False, collection_string=None, id=None
...
-                # For config objects we do not work with the pk but with the key.
-                self.id = instance.key
-            else:
-                self.id = instance.pk
+            self.id = instance.pk
         elif collection_string is not None and id is not None:
             # Collection element is created via values
             self.collection_string = collection_string
-            self.id = id
+            # if not isinstance(id, int):
```
normanjaeckel: Remove this comment.

tests/integration/core/test_viewset.py
```diff
@@ -120,16 +123,19 @@ class TestConfigDBQueries(TestCase):
...
+        config.save_default_values()
 
     @use_cache()
     def test_admin(self):
         """
         Tests that only the following db queries are done:
         * 2 requests to get the session an the request user with its permissions and
         * 1 requests to get the list of all config values
+
+        * 1 more that I do not understand
```
normanjaeckel: Hm. This is strange ...

tests/integration/utils/test_autoupdate.py
```diff
@@ -47,34 +47,34 @@ def test_change_many_elements(self):
...
-        """
-        assignment = Assignment.objects.create(title='test_assignment', open_posts=1)
-        poll = assignment.create_poll()
-        channel_layers[DEFAULT_CHANNEL_LAYER].flush()
-
-        inform_changed_data(poll)
-
-        channel_message = self.get_next_message('autoupdate.send_data', require=True)
-        self.assertEqual(len(channel_message['elements']), 1)
+    # def test_change_with_non_root_rest_elements(self):
```
normanjaeckel: What is the matter here?

openslides/core/config.py
```diff
@@ -130,43 +147,37 @@ def __setitem__(self, key, value):
...
+        item_index = dict((variable.name, variable) for variable in items)
+
         # Check that all ConfigVariables are unique. So no key from items can
         # be in already in self.config_variables
-        for key in new_items.keys():
-            if key in self.config_variables:
-                raise ConfigError(_('Too many values for config variable {} found.').format(key))
+        intersection = set(item_index.keys()).intersection(self.config_variables.keys())
+        if intersection:
+            raise ConfigError(_('Too many values for config variable {} found.').format(intersection))
```
normanjaeckel: Now we need plural: `Too many values for config variables {} found.`

openslides/core/config.py
FinnStutzenstein: Or raising an own error, telling that this should never happend and **why** this should never happend, so if this error is raised, the user know, that something is messed up and may correct this.

tests/integration/utils/test_autoupdate.py
ostcar: I was not able to fix this. Could you try it?

openslides/core/config.py
ostcar: I removed the try-except block. It is really impossible to get this error message, because of the try-except block at the beginning of the function.

openslides/core/config.py
ostcar: done

openslides/utils/collection.py
ostcar: done

tests/integration/core/test_viewset.py
ostcar: I know. But it does not matter in production, because all is cached...

tests/unit/config/test_api.py
ostcar: done

ostcar:
@normanjaeckel I was able to fix the tests, but the post_migrate signal does not work. Do you know why?

ostcar:
ok. I was able to fix the signal problem. Signal receivers have to be global functions.

Ready to merge

PR #3373, opened 2017-08-23
Fixing wrong motion comment fields if special comments are present

FinnStutzenstein:


FinnStutzenstein:
can be closed in favor of #3376 .

PR #3376, opened 2017-08-25
Rework on motion comments.

FinnStutzenstein:
Some important things changed:
- The configvalue has this layout: 
```
motions_comments = {
    1: {name: 'comment1', public: false},
    2: null, //deleted; we need to track the highest used id, so we cannot just delete empty entries.
    3: {name: 'comment2', public: true},
    4: null,
}
```
And comments in motions:
```
motion.comment = {
1: 'Text for comment1',
3: 'Text for comment2',
}
```
- The `MotionComment` factory has got these two new methods: `getCommentsFields` and `getNoSpecialCommentsFields`. The deleted and forbidden comments (forbidden if not public and the user has not `can_see_and_manage_comments`) are stripped out there.
- What a "special comment field" is is defined in `MotionComment` at only one place: `isSpecialCommentField(field)`. If a new type of special comment is added, just add it there.
- Added the migration path to index all motion comments and the configvalue `motions_comments`.
- All comment stuff is now handled with field **ids**. E.g. `getFieldNameForFlag` is changed to `getFieldIdForFlag`.

### DIFF
--- a/openslides/motions/migrations/0003_motion_comments.py
+++ b/openslides/motions/migrations/0003_motion_comments.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def change_motions_comments(apps, schema_editor):
+    """
+    Index the comments fields in the config. Changing from an array to a dict with the
+    ids as keys. CHange all motions from an array for comments to a dict with the comments
+    field id as key to link motion comments and comments fields.
+    """
+    # We get the model from the versioned app registry;
+    # if we directly import it, it will be the wrong version.
+    ConfigStore = apps.get_model('core', 'ConfigStore')
+    Motion = apps.get_model('motions', 'Motion')
+
+    try:
+        config_comments_fields = ConfigStore.objects.get(key='motions_comments').value
+    except ConfigStore.DoesNotExist:
+        config_comments_fields = []  # The old default: An empty list.
+
+    comments_fields = {}
+    for index, field in enumerate(config_comments_fields):
+        comments_fields[index+1] = field
+
+    max_index = len(config_comments_fields)-1
+
+    try:
+        db_value = ConfigStore.objects.get(key='motions_comments')
+    except ConfigStore.DoesNotExist:
+        db_value = ConfigStore(key='motions_comments')
+    db_value.value = comments_fields
+    # We cannot provide skip_autoupdate=True here, becuase this object is a fake object. It does *not*
+    # inherit from the RESTModelMixin, so the save() methos from base_model.py (django's default)
+    # gets called. This is because we are in the core app and try to save a core model. See
+    # comments in PR #3376.
+    db_value.save()
+
+    for motion in Motion.objects.all():
+        comments = {}
+        for index, comment in enumerate(motion.comments or []):
+            if index > max_index:
+                break
+            comments[index+1] = comment
+        motion.comments = comments
+        motion.save(skip_autoupdate=True)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('motions', '0002_misc_features'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            change_motions_comments
+        ),
+    ]
--- a/openslides/motions/static/js/motions/site.js
+++ b/openslides/motions/static/js/motions/site.js
@@ -910,11 +910,10 @@ angular.module('OpenSlidesApp.motions.site', [
         $scope.alert = {};
 
         // Motion comments
-        $scope.commentsFields = Config.get('motions_comments').value;
-        $scope.commentsFieldsNoSpecialComments = _.filter($scope.commentsFields, function (field) {
-            var specialComment = field.forState || field.forRecommendation;
-            return !specialComment;
-        });
+        $scope.noSpecialCommentsFields = MotionComment.getNoSpecialCommentsFields();
+        $scope.showCommentsFilter = function () {
+            return _.keys($scope.noSpecialCommentsFields).length > 0;
+        };
 
         // collect all states and all recommendations of all workflows
         $scope.states = [];
@@ -1289,13 +1288,10 @@ angular.module('OpenSlidesApp.motions.site', [
                 });
             }
         };
-        $scope.commentsFields = Config.get('motions_comments').value;
-        $scope.commentsFieldsNoSpecialComments = _.filter($scope.commentsFields, function (field) {
-            var specialComment = field.forState || field.forRecommendation;
-            return !specialComment;
-        });
-        $scope.commentFieldForState = MotionComment.getFieldNameForFlag('forState');
-        $scope.commentFieldForRecommendation = MotionComment.getFieldNameForFlag('forRecommendation');
+        $scope.commentsFields = MotionComment.getCommentsFields();
+        $scope.noSpecialCommentsFields = MotionComment.getNoSpecialCommentsFields();
+        $scope.commentFieldForStateId = MotionComment.getFieldIdForFlag('forState');
+        $scope.commentFieldForRecommendationId = MotionComment.getFieldIdForFlag('forRecommendation');
         $scope.version = motion.active_version;
         $scope.isCollapsed = true;
         $scope.lineNumberMode = Config.get('motions_default_line_numbering').value;
@@ -1402,14 +1398,14 @@ angular.module('OpenSlidesApp.motions.site', [
         // save additional state field
         $scope.saveAdditionalStateField = function (stateExtension) {
             if (stateExtension) {
-                motion["comment " + $scope.commentFieldForState] = stateExtension;
+                motion['comment_' + $scope.commentFieldForStateId] = stateExtension;
                 $scope.save(motion);
             }
         };
         // save additional recommendation field
         $scope.saveAdditionalRecommendationField = function (recommendationExtension) {
             if (recommendationExtension) {
-                motion["comment " + $scope.commentFieldForRecommendation] = recommendationExtension;
+                motion['comment_' + $scope.commentFieldForRecommendationId] = recommendationExtension;
                 $scope.save(motion);
             }
         };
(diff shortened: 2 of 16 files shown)


openslides/motions/migrations/0003_motion_comments.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    Index the comments fields in the config. Changing from an array to a dict with the
+    ids as keys. CHange all motions from an array for comments to a dict with the comments
+    field id as key to link motion comments and comments fields.
+    """
+    # We get the model from the versioned app registry;
+    # if we directly import it, it will be the wrong version.
+    Motion = apps.get_model('motions', 'Motion')
+
+    config_comments_fields = config['motions_comments']
```
FinnStutzenstein: @normanjaeckel I wonder why this works here to use the ConfigHandler?  Is this migration OK for you or should I do it different?

normanjaeckel: I am not sure but maybe we should use `apps.get_model('core', 'ConfigStore')` to be more save.

openslides/motions/migrations/0003_motion_comments.py
```diff
@@ -0,0 +1,68 @@
...
+
+def change_motions_comments(apps, schema_editor):
+    """
+    Index the comments fields in the config. Changing from an array to a dict with the
+    ids as keys. CHange all motions from an array for comments to a dict with the comments
+    field id as key to link motion comments and comments fields.
+    """
+    # We get the model from the versioned app registry;
+    # if we directly import it, it will be the wrong version.
+    ConfigStore = apps.get_model('core', 'ConfigStore')
```
FinnStutzenstein: @normanjaeckel Can you check this again? Now without using the ConfigHandler.

FinnStutzenstein:
@emanuelschuetze Fixed.

openslides/motions/migrations/0003_motion_comments.py
normanjaeckel: Looks good, not tested.

openslides/motions/migrations/0003_motion_comments.py
```diff
@@ -0,0 +1,68 @@
...
+    Motion = apps.get_model('motions', 'Motion')
+
+    try:
+        config_comments_fields = ConfigStore.objects.get(key='motions_comments').value
+    except ConfigStore.DoesNotExist:
+        config_comments_fields = get_motions_comments_default()
```
normanjaeckel: I think this is not useful. We know that the one and only value here is an empty dict.

openslides/motions/migrations/0003_motion_comments.py
```diff
@@ -0,0 +1,68 @@
...
+
+    comments_fields = {}
+    index = 0
+    for field in config_comments_fields:
```
normanjaeckel: Please use `enumerate()`

openslides/motions/migrations/0003_motion_comments.py
```diff
@@ -0,0 +1,68 @@
...
+        comments_fields[index] = field
+
+    try:
+        db_value = ConfigStore.objects.get(key='motions_comments')
+    except ConfigStore.DoesNotExist:
+        db_value = ConfigStore(key='motions_comments')
+    db_value.value = comments_fields
+    db_value.save()
+
+    max_index = index
```
normanjaeckel: This line can be just after the loop.

openslides/motions/migrations/0003_motion_comments.py
```diff
@@ -0,0 +1,68 @@
...
+    for field in config_comments_fields:
+        index += 1
+        comments_fields[index] = field
+
+    try:
+        db_value = ConfigStore.objects.get(key='motions_comments')
+    except ConfigStore.DoesNotExist:
+        db_value = ConfigStore(key='motions_comments')
+    db_value.value = comments_fields
+    db_value.save()
```
normanjaeckel: I am not sure whether skip_autoupdate is necessary here.

openslides/motions/migrations/0003_motion_comments.py
```diff
@@ -0,0 +1,68 @@
...
+
+    max_index = index
+
+    for motion in Motion.objects.all():
+        comments = {}
+        index = 0
+        for comment in motion.comments or []:
+            index += 1
+            if index > max_index:
+                break
```
normanjaeckel: Ah this is tricky. "Hidden" comments are now deleted. Are you sure?

openslides/motions/migrations/0003_motion_comments.py
```diff
@@ -0,0 +1,68 @@
...
+        db_value = ConfigStore(key='motions_comments')
+    db_value.value = comments_fields
+    db_value.save()
+
+    max_index = index
+
+    for motion in Motion.objects.all():
+        comments = {}
+        index = 0
+        for comment in motion.comments or []:
```
normanjaeckel: Use `enumerate()`.

openslides/motions/migrations/0003_motion_comments.py
FinnStutzenstein: Yes, because there does not exist a commebts field for them. If you start OS, you wouldn't see the comments.

openslides/motions/migrations/0003_motion_comments.py
FinnStutzenstein: The problem here are custom default values in custom OS versions. I can remove this here if you wish, but if so @emanuelschuetze should copy this code to replace this commit in custon OS versions.

normanjaeckel: Yes, please use customized migrations in customize OpenSlides builds.

openslides/motions/migrations/0003_motion_comments.py
normanjaeckel: Ok for me.

openslides/motions/migrations/0003_motion_comments.py
FinnStutzenstein: Ok

openslides/motions/migrations/0003_motion_comments.py
FinnStutzenstein: Done.

openslides/motions/migrations/0003_motion_comments.py
FinnStutzenstein: Done.

openslides/motions/migrations/0003_motion_comments.py
FinnStutzenstein: Done.

openslides/motions/migrations/0003_motion_comments.py
FinnStutzenstein: No: `TypeError: save() got an unexpected keyword argument 'skip_autoupdate'`

FinnStutzenstein:
@emanuelschuetze @normanjaeckel This is ready to test again.

openslides/motions/migrations/0003_motion_comments.py
normanjaeckel: Ok. So nothing to do ... :-)

openslides/motions/migrations/0003_motion_comments.py
```diff
@@ -0,0 +1,56 @@
...
+    db_value.save()
+
+    for motion in Motion.objects.all():
+        comments = {}
+        for index, comment in enumerate(motion.comments or []):
+            if index > max_index:
+                break
+            comments[index+1] = comment
+        motion.comments = comments
+        motion.save(skip_autoupdate=True)
```
normanjaeckel: Hm. But here we have skip_autoupdate. Why this?

openslides/motions/static/js/motions/site.js
```diff
@@ -1473,16 +1469,7 @@ angular.module('OpenSlidesApp.motions.site', [
...
-            if ($scope.commentsFields.length > 0) {
-                isAllowed = operator.hasPerms('motions.can_see_and_manage_comments') || _.find(
-                        $scope.commentsFields,
-                        function(field) {
-                            return field.public && !field.forState && !field.forRecommendation;
-                        }
-                );
-            }
-            return Boolean(isAllowed);
+            return _.keys($scope.noSpecialCommentsFields).length > 0;
```
normanjaeckel: Hm. This function is declared that is is about user permissions. But you only look into the none special comments fields list regardless of operator.hasPerms('motions.can_see_and_manage_comments') or whatever.

Maybe the name of the function has to be changed if it does something other than before.

openslides/motions/migrations/0003_motion_comments.py
normanjaeckel: Hm. I do not understand why this happens and not for motion, see below. I also think that we do not want to trigger autoupdate during migration (this caused many errors in former times ...).

FinnStutzenstein: Strange: The debugger stepped directly into djangos models/base.py. I've tried to override the save method in the ConfigStore class, but this doesn't work. So I added this to the ConfigStore:
```
def mysave(self, *args, **kwargs):
        return super().save(*args, **kwargs)
```
And calling `mysave` instead of `save`. Now I get `AttributeError: 'ConfigStore' object has no attribute 'mysave'`. It seems that during the migration (or because of getting the model with `apps.get_model`) the model is not loaded correctly. I do not have a clue why..

FinnStutzenstein:
```
(Pdb) db_value.__class__
<class '__fake__.ConfigStore'>
(Pdb) db_value.__class__.__bases__
(<class 'django.db.models.base.Model'>,)

(Pdb) motion.__class__
<class '__fake__.Motion'>
(Pdb) motion.__class__.__bases__
(<class 'openslides.utils.models.RESTModelMixin'>, <class 'django.db.models.base.Model'>)
```
....??

openslides/motions/static/js/motions/site.js
FinnStutzenstein: Renamed this function.

openslides/motions/migrations/0003_motion_comments.py
normanjaeckel: Hm. Maybe models from the core app (because we do migrations in the core app) cannot be loaded directly but as a Django Fake Model. But models from other apps can, because their migrations are already done ...

normanjaeckel: If you think this is the solution, please add a short comment in the code.

FinnStutzenstein: Ok.

FinnStutzenstein:
@normanjaeckel @emanuelschuetze all done.

PR #3379, opened 2017-08-25
New layout for agenda list slide, fixing long titles in agenda list v…

FinnStutzenstein:
…iew, fixing mediafiles in fullscreen

### DIFF
--- a/openslides/agenda/static/js/agenda/projector.js
+++ b/openslides/agenda/static/js/agenda/projector.js
@@ -77,6 +77,7 @@ angular.module('OpenSlidesApp.agenda.projector', ['OpenSlidesApp.agenda'])
         // class.
 
         // Bind agenda tree to the scope
+        var items;
         $scope.$watch(function () {
             return Agenda.lastModified();
         }, function () {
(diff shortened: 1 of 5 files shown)


FinnStutzenstein:
@emanuelschuetze Test again :)

openslides/agenda/static/js/agenda/projector.js
```diff
@@ -76,7 +76,44 @@ angular.module('OpenSlidesApp.agenda.projector', ['OpenSlidesApp.agenda'])
...
         // class.
 
+        // Converts the array of agendaitems into a real tree structure.
+        // Every node contains an array of children and the item. The return
+        // value is a simple array of nodes representing the top layer of nodes.
+        // requires an flat tree of agendaitems with a 'parentCount' that
+        // gives the number of parents above the item.
+        var getRealTree = function (items) {
+            var startNodes = [];
+            for (var i = 0; i < items.length; i++) {
```
normanjaeckel: Hm. This code looks slow: The loop runs over all items, also for the children of the children. The problem is that the lopp in line 105 also runs through all items again and again ...

Is this a performance issue or do you think it is ok even for a deeply nested agenda with 500 entries (one for each motion e. g.)? Maybe just test with 1000 somehow nested motions ...

FinnStutzenstein: You are right; I didn't see that we already implemented this sort of method: The `AgendaTree` has a method `getTree`. I'll use this.

FinnStutzenstein:
@normanjaeckel Please check again. Should has the same functionality.

normanjaeckel:
@emanuelschuetze Please test again (because of implementation change) then merge.

Issue #3377, opened 2017-08-25
Broken login with anonymous


FinnStutzenstein:
```2017-08-25 10:13:29,753 - ERROR - worker - Error processing message with consumer openslides.utils.autoupdate.ws_add_site:
Traceback (most recent call last):
  File "/home/finn/OpenSlides/.venv/lib/python3.5/site-packages/channels/worker.py", line 119, in run
    consumer(message, **kwargs)
  File "/home/finn/OpenSlides/.venv/lib/python3.5/site-packages/channels/sessions.py", line 192, in inner
    result = func(message, *args, **kwargs)
  File "/home/finn/OpenSlides/.venv/lib/python3.5/site-packages/channels/auth.py", line 73, in inner
    return func(message, *args, **kwargs)
  File "/home/finn/OpenSlides/.venv/lib/python3.5/site-packages/channels/sessions.py", line 70, in inner
    return func(message, *args, **kwargs)
  File "/home/finn/OpenSlides/.venv/lib/python3.5/site-packages/channels/auth.py", line 89, in inner
    return func(message, *args, **kwargs)
  File "/home/finn/OpenSlides/OpenSlides/openslides/utils/autoupdate.py", line 93, in ws_add_site
    restricted_data = access_permissions.get_restricted_data(collection, user)
  File "/home/finn/OpenSlides/OpenSlides/openslides/users/access_permissions.py", line 190, in get_restricted_data
    if full['user_id'] == user.id:
AttributeError: 'NoneType' object has no attribute 'id'

```

ostcar:
I have a branch "more_typings" where I add more type informations to functions. With this branch I also found this problem (one function can return None and the return value is used as iterabe). I will fix this issue in this branch in the next days

ostcar:
This error could not be found with our current typing annotations. To find it with mypy, we need type annotations for the entire openslides module, not only for openslides.utils. See: https://github.com/python/mypy/issues/3903

PR #3380, opened 2017-08-29
More typings

ostcar:


openslides/utils/autoupdate.py
```diff
@@ -92,6 +94,12 @@ def ws_add_site(message):
         access_permissions = collection.get_access_permissions()
         restricted_data = access_permissions.get_restricted_data(collection, user)
 
+        if restricted_data is None:
```
normanjaeckel: This can should never happen: In line 93 (old) / 95 (new) we get the restricted data using a Collection And not a single instance. In this case all restricted_data functions have to return an iterable (e. g. an empty list).

Is it possible to remove the two lines?

openslides/utils/access_permissions.py
```diff
@@ -1,46 +1,35 @@
...
+from typing import Any, Dict, List, Optional, Union
 
-from .collection import Collection
-from .dispatch import SignalConnectMetaClass
+from django.db.models import Model
+from rest_framework.serializers import Serializer
 
+from .collection import Collection, CollectionElement
 
-class BaseAccessPermissions(object, metaclass=SignalConnectMetaClass):
```
normanjaeckel: Why do you remove the metaclass here?  Maybe because the feature is not used any more ...?

openslides/utils/access_permissions.py
```diff
@@ -1,46 +1,35 @@
...
+Container = Union[CollectionElement, Collection]
+RestrictedData = Union[List[Dict[str, Any]], Dict[str, Any], None]
+
+
+class BaseAccessPermissions:
     """
-    Base access permissions container.
+    Base access permissions containser.
```
normanjaeckel: -1

openslides/utils/access_permissions.py
```diff
@@ -1,46 +1,35 @@
...
     Every app which has autoupdate models has to create classes subclassing
-    from this base class for every autoupdate root model. Each subclass has
-    to have a globally unique name. The metaclass (SignalConnectMetaClass)
-    does the rest of the magic.
+    from this base class for every autoupdate root model.
     """
-    signal = Signal()
 
-    def __init__(self, **kwargs):
+    def __init__(self) -> None:
```
normanjaeckel: The init method can be removed here.

openslides/utils/autoupdate.py
```diff
@@ -64,7 +66,7 @@ def format_for_autoupdate(collection_string, id, action, data=None):
 
 
 @channel_session_user_from_http
-def ws_add_site(message):
+def ws_add_site(message: Any) -> None:
```
normanjaeckel: Message is a specific type, isn't it?

openslides/utils/autoupdate.py
```diff
@@ -16,7 +18,7 @@
 from .collection import Collection, CollectionElement, CollectionElementList
 
 
-def send_or_wait(send_func, *args, **kwargs):
+def send_or_wait(send_func: Any, *args: Any, **kwargs: Any) -> None:
```
normanjaeckel: send_func should always be a channel's send function. Args and kwargs should only be the args and kwargs for such a function.

openslides/utils/cache.py
```diff
@@ -252,7 +269,7 @@ def use_redis_cache():
     return isinstance(caches['default'], RedisCache)
 
 
-def get_redis_connection():
+def get_redis_connection() -> Any:
```
normanjaeckel: Hm. Is it possible to set the correct type easily?

setup.cfg
```diff
@@ -18,7 +18,10 @@ multi_line_output = 3
 ignore_missing_imports = true
 strict_optional = true
 
-[mypy-openslides.utils.auth]
+[mypy-openslides.utils.dispatch]
+ignore_errors = true
```
normanjaeckel: Which errors remain?

openslides/utils/autoupdate.py
ostcar: I did not understand all of the functions you write. Please add a commit to the pull request where you add as match typings as possible.

It should be the goal, that there is no Any in the code. (But we will probably not  succeed.)

openslides/utils/autoupdate.py
ostcar: It is a Dict. Maybe you can use Dict[str, Any]. I don't belief that we can set an explicit type for the Value.

openslides/utils/autoupdate.py
ostcar: Yes, it is possible. The cast() call in the next line should solve this. But please be aware of this issue: https://github.com/OpenSlides/OpenSlides/issues/3377

My goal is still to remove the "container" object in the get_restricted_data methods and replace it with List[Dict[str, Any]] (maybe the value in the dict could also be str, or Unique[str, int, bool])

openslides/utils/access_permissions.py
ostcar: Yes. And I look forward of removing utils.dispatch anywhere. I think our new style, to explicit add objects to a list in app.raddy is match better.

openslides/utils/access_permissions.py
ostcar: done

openslides/utils/access_permissions.py
ostcar: done

openslides/utils/cache.py
ostcar: I don't know what type it is.

setup.cfg
ostcar: This file is a bit complicated and I do not understand it. As I wrote above, I would like to remove it. Therefore I don't want to spend any time on it.
But of cause, you can add typing information as you like

ostcar:
All errors done. Please merge this. It has no new feature. If you like to add more typings you can do this in another pull request.

ostcar:
@normanjaeckel I added the option ```check_untyped_defs``` to check code outside of utils. This option can be removed when the entire project has type annotations. I added one ```#TODO``` (without the space) with a question. What should we do in this case?

openslides/utils/autoupdate.py
normanjaeckel: Issue #3377 is fixed in current master, so please remove these lines.

openslides/utils/autoupdate.py
```diff
@@ -92,6 +94,12 @@ def ws_add_site(message):
         access_permissions = collection.get_access_permissions()
         restricted_data = access_permissions.get_restricted_data(collection, user)
 
+        if restricted_data is None:
+            continue
+
+        # At this point restricted_data has to be a list. So we have to tell it mypy
+        restricted_data = cast(List[Dict[str, Any]], restricted_data)
```
normanjaeckel: Please check if this is still required.

openslides/utils/autoupdate.py
ostcar: done

openslides/utils/autoupdate.py
ostcar: Yes it is. mypy can not know, that at this point restricted_data is a list. ```cast()``` is there for situations like this.

ostcar:
All done and rebased. I merge after travis is ok

PR #3382, opened 2017-08-29
Rewrite projector_element without dispatch

ostcar:
This is an example concept, how the projector element could be rewritten to remove the meta-class-dispatch system. 

For this to work, it has to be rebased on top of https://github.com/OpenSlides/OpenSlides/pull/3380 (after this is merged).

@normanjaeckel please tell me if you could approve this proposal, then I would rewrite all the other apps.


openslides/agenda/projector.py
```diff
@@ -136,3 +138,9 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
                         output.extend(self.get_requirements_as_collection_elements(config_entry))
                         break
         return output
+
+
+def get_all_slides() -> Generator[ProjectorElement, None, None]:
+    yield ItemListSlide()
+    yield ListOfSpeakersSlide()
+    yield CurrentListOfSpeakersSlide()
```
ostcar: Alternatively this could return Generator[Type[ProjectorElement], None, None] (the class and not the instance). Then get_all_projector_elements could return "new" instances on each call. What do you think?

openslides/utils/projector.py
```diff
@@ -104,3 +80,23 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
             else:
                 output = []
         return output
+
+
+projector_elements = []  # type: List[ProjectorElement]
```
ostcar: The type could be Dict[str, ProjectorElement] where the key is the name of the projector element. Currently we call get_all_projector_elements() three times and in any of this cases we create this dict.

What do you think?

openslides/agenda/projector.py
normanjaeckel: I think we do not need new instances on each call.

openslides/agenda/projector.py
```diff
@@ -136,3 +138,9 @@ def get_collection_elements_required_for_this(self, collection_element, config_e
                         output.extend(self.get_requirements_as_collection_elements(config_entry))
                         break
         return output
+
+
+def get_all_slides() -> Generator[ProjectorElement, None, None]:
```
normanjaeckel: Maybe just `get_projector_elements()`.

openslides/utils/projector.py
normanjaeckel: Ok.

normanjaeckel:
I like projector_element because some of them are no slides (like the clock or the countdown).

ostcar:
@normanjaeckel all done, please review

FinnStutzenstein:
Is this fixed with this PR?:
```
/home/finn/OpenSlides/OpenSlides/tests/example_data_generator/management/commands/create-example-data.py(117)handle()
-> self.create_motions(options)
  /usr/lib/python3.5/contextlib.py(30)inner()
-> return func(*args, **kwds)
  /home/finn/OpenSlides/OpenSlides/tests/example_data_generator/management/commands/create-example-data.py(156)create_motions()
-> motion.save(skip_autoupdate=True)
  /home/finn/OpenSlides/OpenSlides/openslides/motions/models.py(234)save()
-> super(Motion, self).save(skip_autoupdate=True, *args, **kwargs)
  /home/finn/OpenSlides/OpenSlides/openslides/utils/models.py(79)save()
-> return_value = super().save(*args, **kwargs)
  /home/finn/OpenSlides/.venv/lib/python3.5/site-packages/django/db/models/base.py(796)save()
-> force_update=force_update, update_fields=update_fields)
  /home/finn/OpenSlides/.venv/lib/python3.5/site-packages/django/db/models/base.py(833)save_base()
-> update_fields=update_fields, raw=raw, using=using)
  /home/finn/OpenSlides/.venv/lib/python3.5/site-packages/django/dispatch/dispatcher.py(191)send()
-> response = receiver(signal=self, sender=sender, **named)
  /home/finn/OpenSlides/OpenSlides/openslides/agenda/signals.py(21)listen_to_related_object_post_save()
-> Item.objects.create(content_object=instance)
  /home/finn/OpenSlides/.venv/lib/python3.5/site-packages/django/db/models/manager.py(85)manager_method()
-> return getattr(self.get_queryset(), name)(*args, **kwargs)
  /home/finn/OpenSlides/.venv/lib/python3.5/site-packages/django/db/models/query.py(399)create()
-> obj.save(force_insert=True, using=self.db)
  /home/finn/OpenSlides/OpenSlides/openslides/utils/models.py(81)save()
-> inform_changed_data(self.get_root_rest_element(), information=information)
> /home/finn/OpenSlides/OpenSlides/openslides/utils/autoupdate.py(318)inform_changed_data()
-> transaction.on_commit(lambda: send_autoupdate(collection_elements))
```
If a motion is created with `skip_autoupdate=True` the agenda item still causes an autoupdate. This is not necessary in the example-data-generator, because this causes the autoupdate system to crash, if many motions are created.

Also noticed: `inform_changed_data` is called when creating the builtin workflows during the migrations.

normanjaeckel:
Pull request is ok, please rebase.

ostcar:
@FinnStutzenstein no, I did not fix this. Maybe you need to create a ticket for you bug with an example how to reproduce it.

ostcar:
done. Merge after travis is ok

PR #3383, opened 2017-08-30
Custom Translations

FinnStutzenstein:


openslides/core/static/js/core/base.js
```diff
@@ -254,6 +254,42 @@ angular.module('OpenSlidesApp.core', [
     }
 ])
 
+// Hook into gettextCatalog to include custom translations by wrapping
```
normanjaeckel: This solution is ok for me but I think there might be a more AngularJS like way to do it: Use `provider` to hook into setup of the `gettextCatalog` service. It would also be ok for me if we open an issue for that.

FinnStutzenstein: Can you give a little example what you mean? My first idea was to create a gettextCatalog too and override the old one, but this would cause too much unnecessary code..

normanjaeckel: Hm. It looks like a [decorator](https://docs.angularjs.org/api/auto/service/$provide#decorator) (to decorate a service – using `provider` is wrong.

FinnStutzenstein: Yep, on an obect orientated language i would have create either a decorator or make a subclass, but this is not possible in angular.

normanjaeckel: It looks like. Follow the link ... :-)

FinnStutzenstein: Done.

PR #3384, opened 2017-08-30
Secure Mediafiles and check view permissions

FinnStutzenstein:


FinnStutzenstein:
Yes this is right, but I think if the system admin dismisses this, this is not our fault ;)
We may write a warning into the big mode readme.

normanjaeckel:
Hm. The system admin is not able to check out permissions and hidden flags of mediafiles in Nginx area ...

FinnStutzenstein:
He should not deliver mediafiles with nginx, only the staticfiles. So the url /media/ should be forewarded to OpenSlides.

normanjaeckel:
Ok.

PR #3400, opened 2017-09-12
Autoupdates on user login (fixes #3395, fixes #3392)

### LINKED ISSUE #3392: Prevent autoupdate on first login.
If a users is going to log in the first time, Django will update the hashed password in the database and an autoupdate will result: The [setter method](https://github.com/django/django/blob/master/django/contrib/auth/base_user.py#L103) will be called, because the hashed password is not up to date anymore ([hasher changed](https://github.com/django/django/blob/master/django/contrib/auth/hashers.py#L49)).

@normanjaeckel Can we update the encrypted password right at users creation time and do not trigger an autoupdate?

### LINKED ISSUE #3395: Disable last login field
We try to make a read-only OS instance, so no writing to the database is permitted. On login, the `last_login` field for the user is updated. Because the DB raises an error, the user cannot log in.
@normanjaeckel Whats about not to delete this field in the DB, but just to disconnect [this signal](https://github.com/django/django/blob/eedc88bd4a8468c2a0daa8346c9b57183dd77782/django/contrib/auth/apps.py#L23)? If this is ok for you, I'll make a PR.

FinnStutzenstein:
@normanjaeckel What should be the default for disabling the last login field? I would prefer `True` for disabling.

### DIFF
--- a/openslides/global_settings.py
+++ b/openslides/global_settings.py
@@ -106,7 +106,6 @@ PASSWORD_HASHERS = [
     'django.contrib.auth.hashers.Argon2PasswordHasher',
     'django.contrib.auth.hashers.BCryptSHA256PasswordHasher',
     'django.contrib.auth.hashers.BCryptPasswordHasher',
-    'django.contrib.auth.hashers.MD5PasswordHasher',  # MD5 is only used for initial passwords.
 ]
 
 
@@ -143,3 +142,6 @@ CHANNEL_LAYERS = {
         },
     },
 }
+
+# Enable updating the last_login field for users on every login.
+ENABLE_LAST_LOGIN_FIELD = False
--- a/openslides/users/apps.py
+++ b/openslides/users/apps.py
@@ -1,4 +1,6 @@
 from django.apps import AppConfig
+from django.conf import settings
+from django.contrib.auth.signals import user_logged_in
 
 from ..utils.collection import Collection
 from ..utils.projector import register_projector_elements
@@ -31,6 +33,10 @@ class UsersAppConfig(AppConfig):
         permission_change.connect(
             get_permission_change_data,
             dispatch_uid='users_get_permission_change_data')
+
+        # Disconnect the last_login signal
+        if not settings.ENABLE_LAST_LOGIN_FIELD:
+            user_logged_in.disconnect(dispatch_uid='update_last_login')
 
         # Register viewsets.
         router.register(self.get_model('User').get_collection_string(), UserViewSet)
--- a/openslides/utils/settings.py.tpl
+++ b/openslides/utils/settings.py.tpl
@@ -136,7 +136,6 @@ MEDIA_ROOT = os.path.join(OPENSLIDES_USER_DATA_PATH, 'media', '')
 # https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#module-django.contrib.auth.password_validation
 # AUTH_PASSWORD_VALIDATORS = []
 
-
 # Customization of OpenSlides apps
 
 MOTION_IDENTIFIER_MIN_DIGITS = 1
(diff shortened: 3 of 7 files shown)


openslides/users/apps.py
```diff
@@ -1,3 +1,4 @@
+import settings
```
FinnStutzenstein: @normanjaeckel Tests are failing here: No module named 'settings'. Any ideas?

normanjaeckel: from django.conf import settings

openslides/users/apps.py
```diff
@@ -20,6 +21,8 @@ def ready(self):
         from .signals import create_builtin_groups_and_admin, get_permission_change_data
         from .views import GroupViewSet, PersonalNoteViewSet, UserViewSet
 
+        from django.contrib.auth.signals import user_logged_in
```
normanjaeckel: Is it possible to use a global import for this?

openslides/users/apps.py
```diff
@@ -32,6 +35,10 @@ def ready(self):
             get_permission_change_data,
             dispatch_uid='users_get_permission_change_data')
 
+        # Disconnect the last_login signal
+        if 'DISABLE_LAST_LOGIN_FIELD' in dir(settings) and settings.DISABLE_LAST_LOGIN_FIELD:
```
normanjaeckel: Please use settings variable as shown [here](https://docs.djangoproject.com/en/1.11/topics/settings/#using-settings-in-python-code).

openslides/utils/settings.py.tpl
```diff
@@ -136,6 +136,8 @@ MEDIA_ROOT = os.path.join(OPENSLIDES_USER_DATA_PATH, 'media', '')
 # https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#module-django.contrib.auth.password_validation
 # AUTH_PASSWORD_VALIDATORS = []
 
+# Disable the last_login field for users.
```
normanjaeckel: We can move this settings to global settings because we don't have an existing use case at the moment to enable this field.

openslides/utils/settings.py.tpl
```diff
@@ -136,6 +136,8 @@ MEDIA_ROOT = os.path.join(OPENSLIDES_USER_DATA_PATH, 'media', '')
...
+DISABLE_LAST_LOGIN_FIELD = True
```
normanjaeckel: True is ok for me here.

openslides/utils/settings.py.tpl
FinnStutzenstein: Done.

openslides/users/apps.py
FinnStutzenstein: Done.

openslides/users/apps.py
FinnStutzenstein: Done.

openslides/utils/settings.py.tpl
FinnStutzenstein: Ok. Changed to `ENALBE_LAST_LOGIN_FIELD = False` to keep all default settings on `false`

emanuelschuetze:
@normanjaeckel All done. Ready for your last review. As discuss, I have to merge it now. Feel free to remark something later. I marked this pull request with new label "needs review" for you.

openslides/global_settings.py
```diff
@@ -143,3 +142,6 @@
         },
     },
 }
+
```
normanjaeckel: Please add an extra newline here.

openslides/utils/settings.py.tpl
```diff
@@ -136,7 +136,6 @@ MEDIA_ROOT = os.path.join(OPENSLIDES_USER_DATA_PATH, 'media', '')
 # https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#module-django.contrib.auth.password_validation
 # AUTH_PASSWORD_VALIDATORS = []
 
-
```
normanjaeckel: Please readd the newline here.

openslides/global_settings.py
emanuelschuetze: Done in #3411

openslides/utils/settings.py.tpl
emanuelschuetze: Done in #3411

PR #3407, opened 2017-09-16
Sort attributes before applying the diff - fixes #3402

### LINKED ISSUE #3402: Wrong change reco, if text has styles
If you change a word in a motion text with a style, not the single word gets marked as changed, but the full line. To reproduce make a new motion with text and change the foreground color of the full text to blue. Now there are span-tags with `style="color: #0000ff;"` between the p-tags and the text. With those extra span tags, the diff is wrong.
Removing the span-tags and make a new change recommendation, the diff is just one word as expected.

CatoTH:
The order of attributes of an html element should be normalized (sorted) before applying the diff, to prevent false positives.


emanuelschuetze:
@normanjaeckel I have to merge it today. A "needs review" label is added. Changlog entry is added by #3408.

normanjaeckel:
Ok.

PR #3408, opened 2017-09-16
Inline-Diff when the first word in a paragraph is replaced - fixes #3401

### LINKED ISSUE #3401: Complete line is marked, if first word is changed
If you just change one word, it should be marked red and the replacement should be next to it in green. Works fine for all words except the first. Easy to reproduce: make a motion with some text and make a change recommendation for just the first word. The full line gets marked.

@CatoTH Can you have a look?

CatoTH:
This fixes the problem that the paragraph-based diff is used instead of the inline-diff as a fallback when the first word of the paragraph is changed.

This affects another behavior: if a longer paragraph (containing multiple line breaks) is replaced by another text, the algorithm previously resulted in two paragraphs - one "deleted" paragraph with the old text, one "inserted" paragraph with the new text. Now, it will result in one paragraph, where each line is deleted individually, the line break and line number not being nested in the DEL-tag anymore, followed by the inserted text within an INS.
This hopefully should not negatively affect anything (like the PDF-export).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,7 +21,7 @@ Motions:
 - Fixed empty motion comment field in motion update form [#3194].
 - Fixed error on category sort [#3318].
 - Removed server side image to base64 transformation and
-  added local transformation [#3181]
+  added local transformation [#3181].
 - Added support for export motions in a ZIP archive [#3189].
 - Performance improvement for ZIP creation [#3251].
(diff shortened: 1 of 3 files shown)


emanuelschuetze:
@normanjaeckel I have to merge it today. A "needs review" label is added.

normanjaeckel:
Ok.

Issue #3409, opened 2017-09-16
Inline Editing inserts line breaks, colliding with diff


CatoTH:
If a motion is edited using the inline editing mode, new line breaks are inserted at an earlier point than where they should be.

FinnStutzenstein:
So this is a (wrong?) behavior of the CKEditor? Do you have an idea for a fix or what we maybe can change in the CKEditor config?

CatoTH:
For once, I've written a patch that handles data like that, once it's in the database like that. I'll have a look if we can normalize that when saving the HTML after inline-editing.

jsaalfeld:
Since we will eventually ditch CKeditor and rewrite the whole client, this won't hopefully be an issue anymore. If it persists, this issue should be reopened.

PR #3410, opened 2017-09-16
Line break algorithm should not be distracted by newline characters

CatoTH:
First part of the bugfix for #3409 - fixes the diff and line breaking algorithms to handle newlines in the database correctly.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -37,7 +37,7 @@ Motions:
 - Added config value for pagenumber alignment in PDF [#3327].
 - Bugfix: Several bugfixes regarding splitting list items in
   change recommendations [#3288].
-- Bugfix: Several bugfixes regarding diff version [#3407, #3408].
+- Bugfix: Several bugfixes regarding diff version [#3407, #3408, #3410].
 - Added inline Editing for motion reason [#3361].
 - Added multiselect filter for motion comments [#3372].
 - Added support for pinning personal notes to the window [#3360].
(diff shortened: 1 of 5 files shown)


emanuelschuetze:
@CatoTH How can I test your fix successfully? I tested the following use case:
* Create a motion with long dummy text.
* Activate inilne editing for motion text and added a linebreak by Shift+Enter in first line.
* Add a change recommendation for (new) first line (add a new word in the middle of the line).
* Show diff version: The whole first line is marked deleted+inserted (paragraph diff mode).
Expected behavior: Only the added word should show in insert (inline) mode.

CatoTH:
@emanuelschuetze I suppose, that was yet another edge case...
What I described, was actually even more of a problem, as entering a newline via Shift+Enter was not necessary to produce this problem. Simply editing and saving did add newline characters at the places where the line numbers are in the markup.
I added another fix to the problem with the added BR.

FinnStutzenstein:
How can I test this PR?

CatoTH:
- Create a motion with a long text
- Inline edit, just make a very little change and save
- BUG: At this point, maybe already the line numbering has changed. In the database, you should see newlines in the database.
- If you make a change recommendation for one line, adding one word in the middle, another bug occurrs: another word is marked as changed (deleted / added), as the newline character is recognized as a change.

Both bugs should be fixed in this branch.

emanuelschuetze:
@normanjaeckel Please review later. Thanks.

normanjaeckel:
Ok.

PR #3416, opened 2017-09-22
Fixing special comments fields.

FinnStutzenstein:


emanuelschuetze:
@normanjaeckel Please review later if you like. Thanks.

normanjaeckel:
Ok. But as we discussed I am not sure that the reconnection delay will really help.

PR #3419, opened 2017-09-26
Index and Webclient View caching

FinnStutzenstein:
The new gulp task is for delivering the three main views with a Webserver. For every request to `/`, `/projector` and `/real-projector` the matching staticfile should be delivered. You can test this, if you put e.g. `localhost:8000/static/html/index.html` in your browser. The Nginx configuration is:
```
location ~* ^/(?!ws|wss|webclient|core/servertime|users/whoami|users/login|users/logout|users/setpassword|motions/docxtemplate|projector|real-projector|static|media|rest).*$ {
    rewrite ^.*$ /static/templates/index.html;
} 
location ~* ^/projector.*$ {
   rewrite ^.*$ /static/templates/projector-container.html;
} 
location ~* ^/real-projector.*$ {
    rewrite ^.*$ /static/templates/projector.html;
}
```

For setups without a Webserver the templates are cached to raise performance in this case.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -85,6 +85,8 @@ Core:
 - Added custom translations in config [#3383].
 - Added dynamic webpage title [#3404].
 - Added 'go to top'-link [#3404].
+- Added caching for the index views. When using a Webserver for serving
+  static files see the example configuration in the PR [#3419].
 
 Mediafiles:
 - Fixed reloading of PDF on page change [#3274].
--- a/openslides/utils/views.py
+++ b/openslides/utils/views.py
@@ -1,5 +1,8 @@
 from typing import Any, Dict, List  # noqa
 
+from django.contrib.staticfiles import finders
+from django.core.exceptions import ImproperlyConfigured
+from django.http import HttpResponse
 from django.views.decorators.csrf import ensure_csrf_cookie
 from django.views.generic.base import View
 from rest_framework.response import Response
@@ -45,3 +48,22 @@ class APIView(_APIView):
     # Add the http-methods and delete the method "method_call"
     get = post = put = patch = delete = head = options = trace = method_call
     del method_call
+
+
+class IndexView(View):
+    """
+    A view to serve a single cached template file. Subclasses has to provide 'template_name'.
+    """
+    template_name = None  # type: str
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+        if self.template_name is None:
+            raise ImproperlyConfigured("'template_name' is not provided")
+
+        with open(finders.find(self.template_name)) as template:
+            self.template = template.read()
+
+    def get(self, *args: Any, **kwargs: Any) -> HttpResponse:
+        return HttpResponse(self.template)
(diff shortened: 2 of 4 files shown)


ostcar:
The nginx regex ```location ~* ^/projector/[0-9]+\/?$ {``` should also work.

What would be nice is, if collect_static could generate a static file for the WebClientView. This should be possible by writing a [finder](https://docs.djangoproject.com/en/1.11/ref/settings/#staticfiles-finders).

openslides/utils/views.py
```diff
@@ -45,3 +47,20 @@ def method_call(self, request: Any, *args: Any, **kwargs: Any) -> Any:
...
+
+class IndexView(View):
+    """
+    A view to serve a single cached template file. Subclasses has to provide 'template_name'.
+    """
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+        if 'template_name' not in dir(self):
+            raise AttributeError("'template_name' is not provided")
```
ostcar: I think the [ImproperlyConfigured](https://docs.djangoproject.com/en/1.11/ref/exceptions/#improperlyconfigured) Exception is better here.

I other places we don't use dir(self) but set the needed attribute as class attribute set to None. I think this helps mypy. So for example:
```python
class IndexView(View):
    template_name = None  # type: str
    def foo(...):
        if self.template_name is None:
            raise ImproperlyConfigured(...)
```

FinnStutzenstein:
@ostcar All Done. Updated the nginx config for more performance.

emanuelschuetze:
@normanjaeckel Please review later! I have to merge it now.

openslides/utils/views.py
```diff
@@ -45,3 +48,22 @@ def method_call(self, request: Any, *args: Any, **kwargs: Any) -> Any:
     # Add the http-methods and delete the method "method_call"
     get = post = put = patch = delete = head = options = trace = method_call
     del method_call
+
+
+class IndexView(View):
```
normanjaeckel: The wording `IndexView` is uses twice: Here for a general view class and in core/views for the real index view. Maybe change this to `TemplateView`.

openslides/utils/views.py
```diff
@@ -45,3 +48,22 @@ def method_call(self, request: Any, *args: Any, **kwargs: Any) -> Any:
...
+    """
+    A view to serve a single cached template file. Subclasses has to provide 'template_name'.
```
normanjaeckel: `have`

CHANGELOG
```diff
@@ -85,6 +85,8 @@ Core:
 - Added custom translations in config [#3383].
 - Added dynamic webpage title [#3404].
 - Added 'go to top'-link [#3404].
+- Added caching for the index views. When using a Webserver for serving
```
normanjaeckel: Please add a handy example configuration to a file like `DEVELOPMENT.rst` or something similar and remove this hint here.

openslides/utils/views.py
FinnStutzenstein: Done.

openslides/utils/views.py
FinnStutzenstein: Done.

CHANGELOG
FinnStutzenstein: Done.

FinnStutzenstein:
@normanjaeckel See pr #3429

normanjaeckel:
Ok.

PR #3436, opened 2017-10-10
Fixed all remarks

FinnStutzenstein:


FinnStutzenstein:
@normanjaeckel Check again.

PR #3442, opened 2017-10-13
Show motion identifier in (current) list of speakers

FinnStutzenstein:


emanuelschuetze:
Tested: motion identifier is still missing on (current) list of speakers SLIDE.

FinnStutzenstein:
@emanuelschuetze Check again

emanuelschuetze:
An other error:
Projekt the agenda. Then go to current list of speaker view (`agenda/speakers`). 

```
angular.js:13923 TypeError: Cannot read property 'content_object' of undefined
    at Scope.$scope.getTitle (site.js:708)
```

emanuelschuetze:
The first error is gone after server restart (sorry). But the second error (`Cannot read property 'content_object' of undefined`) exists if you project the agenda and open the curren list of speakers view (backend).

emanuelschuetze:
Sorry, not fixed:
Project agenda. Then go to current list of speakers view (not slide). => JS error:
```
TypeError: Cannot read property 'content_object' of undefined
    at Scope.$scope.getTitle (site.js:708)
```

FinnStutzenstein:
@emanuelschuetze Fixed special case for the current list of speakers, if there is no item to show the list of speakers.

openslides/agenda/static/js/agenda/projector.js
```diff
@@ -49,19 +50,48 @@ angular.module('OpenSlidesApp.agenda.projector', ['OpenSlidesApp.agenda'])
         $scope.updateCurrentListOfSpeakers = function () {
             $scope.agendaItem = CurrentListOfSpeakersItem.getItem($scope.currentListOfSpeakersReference);
         };
+
+        $scope.getTitle = function () {
```
normanjaeckel: Ca we please use a different approach here? The motion specific code should not live here but somewhere in motion app. So we can maybe do something like this:
```javascript
$scope.getTitle = function () {
    if ($scope.agendaItem.content_object.getAgendaTitle) {
        return $scope.agendaItem.conten_object.getAgendaTitle();
    } else {
        return ...
    }
};
```
So every app and even plugins can define the title itself.

FinnStutzenstein:
@normanjaeckel @emanuelschuetze I looked in the code and recognized, that the agenda item checks already, if the content_object specifies a custom title. Please check again.

emanuelschuetze:
Changelog entry added.

PR #3443, opened 2017-10-13
Reset Countdown when a speech was ended

FinnStutzenstein:
+ Layout changes for users with long names (no break after the microphone icon)


normanjaeckel:
Oh sorry. Maybe a CHANGELOG entry for this bugfix would be nice.

PR #3444, opened 2017-10-13
Go to top on a page change

FinnStutzenstein:
This is an example for the user list view. If this is approved, I'll impelment this also for the agenda, assignments, motions and mediafiles.

### DIFF
--- a/openslides/core/static/js/core/site.js
+++ b/openslides/core/static/js/core/site.js
@@ -1008,7 +1008,14 @@ angular.module('OpenSlidesApp.core.site', [
                 $scope.show = ($window.pageYOffset >= 150);
             });
         });
-        $scope.gotoTop = function () {
+    }
+])
+
+.run([
+    '$rootScope',
+    '$window',
+    function ($rootScope, $window) {
+        $rootScope.gotoTop = function () {
             $window.scrollTo(0, 0);
         };
     }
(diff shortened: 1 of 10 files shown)


FinnStutzenstein:
@emanuelschuetze Adapted for every list view except for mediafiles - there is no pagination yet.

openslides/core/static/js/core/site.js
```diff
@@ -1008,7 +1008,14 @@ angular.module('OpenSlidesApp.core.site', [
                 $scope.show = ($window.pageYOffset >= 150);
             });
         });
-        $scope.gotoTop = function () {
+    }
+])
+
+.run([
```
normanjaeckel: Hm. Of course this is a possible approach I better like a solution with an extra factory to be more explicit and more AngularJS like. Or what is your opinion?

emanuelschuetze:
Successfully tested.

openslides/core/static/js/core/site.js
FinnStutzenstein: Yeah, I was thinking about this, but: This is just one line of code. Putting this into a factory is some kind of too much. Also we would have to inject the factory into every list controller and they are already flooded with injections. So I decided to put this into the root scope.

normanjaeckel: Hm. You are right, the controllers are flooded with injections. I think this is because the code is not in a good structure at the moment. It would be better to put more pieces of code into a tree-like structure instead of merging everything together.

Nevertheless I think the code would be harder to understand *in general* if we start putting snippets of code outside our current "system". Here we really have a controller like control function even if it is only one line of code.

I recommend the other solution but it would be ok for me if you reject.

FinnStutzenstein:
@normanjaeckel I prefer not doing a factory for this. But notice issue #3453 .

normanjaeckel:
After issue #3453 is solved we can think about my suggestion again.

PR #3446, opened 2017-10-13
New motion identifier behavior

FinnStutzenstein:
@emanuelschuetze Please test all scenarios we have talked about

### DIFF
--- a/openslides/motions/models.py
+++ b/openslides/motions/models.py
@@ -338,21 +338,6 @@ class Motion(RESTModelMixin, models.Model):
             # Do not set an identifier.
             return
 
-        # The motion is an amendment.
-        elif self.is_amendment():
-            motions = self.parent.amendments.all()
-
-        # The motions should be counted per category.
-        elif config['motions_identifier'] == 'per_category':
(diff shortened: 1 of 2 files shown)


emanuelschuetze:
@normanjaeckel I have to merge it today. Please review it later. Thanks.

normanjaeckel:
Sorry but can you please describe the problem and use case here. There is no issue for this but I would like to understand why the set_identifier method should now rely on an existing identifier_number.

emanuelschuetze:
@normanjaeckel There are two typically use cases:

1. During an event a motion has to reset, because wrong end state was selected by admin user. 
Previously (wrong) behaviour: The motion identifier was removed by reset action. This is not good because the motion can be projected and then the motion identifier is removed also from projector. And the admin user has to find and set manually the same identifier.
Expected (right) behaviour (reached with this PR): Don't remove the identifier by reset and keep the same identifier after set a new state.

2. Remove the last motion during preparation phase. E.g. A2 of two motions A1 and A2. Create a new motion and set identifier. 
Old behaviour: A3 is set.
New behaviour: A2 is set.

PR #3450, opened 2017-10-15
Disable change ratio detection / Fix marking unchanged dangling line as changed

CatoTH:
The problem reported is that if a longer paragraph is added after another one by selecting the last line of the previous paragraph and adding the new one after it, lead to the selected line as being marked as removed and re-added:
bildschirmfoto 2017-10-15 um 14 49 29

This is because we use a change ratio detection for cases where a bigger part of a paragraph is replaced by other text, to avoid a rather confusing inline-diff. This pull request removes this mechanism and therefore fixes the reported problem, but therefore might introduce other strange formattings.

It might be possible to divide change recommendations and the original text to paragraphs, and perform the diff on the paragraphs separately; this would take a couple of hours and increase the computational ressources on the client side quite a bit, so I'm not doing it without further discussion.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -37,7 +37,8 @@ Motions:
 - Added config value for pagenumber alignment in PDF [#3327].
 - Bugfix: Several bugfixes regarding splitting list items in
   change recommendations [#3288].
-- Bugfix: Several bugfixes regarding diff version [#3407, #3408, #3410, #3440].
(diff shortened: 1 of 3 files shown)


emanuelschuetze:
@normanjaeckel Please review later. Thanks.

emanuelschuetze:
Added pr number to changelog entry.

normanjaeckel:
Ok.

PR #3451, opened 2017-10-17
try/catch for inconsistent data

CatoTH:
An attempt to fail less dramatically when inconsistent data occurs in the database (when the change recommendation refers a line that does not exist in the motion, e.g. if the motion has beed shortened afterwards)

### DIFF
--- a/openslides/agenda/static/js/agenda/site.js
+++ b/openslides/agenda/static/js/agenda/site.js
@@ -164,6 +164,7 @@ angular.module('OpenSlidesApp.agenda.site', [
         $scope.limitBegin = 0;
         $scope.pageChanged = function() {
             $scope.limitBegin = ($scope.currentPage - 1) * $scope.itemsPerPage;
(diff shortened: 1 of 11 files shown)


normanjaeckel:
I do not recommend this way of exception check but prefer a try-block to be as short as possible and the exception catcher as explicit as possible (which exception might be thrown in which case at which line: only catch this one a this line ...).

But this is in some area of code where I am not deep inside, so if you want to do it **at your own risk** (here I mean that the costs of debugging and code refactoring may increase more than you may like – and we already know that we have to refactor the client code (dropping AngularJS) some day) feel free.

emanuelschuetze:
@CatoTH Do you want to update your pull request as proposed by @normanjaeckel? (keep only minimal/relevant code inside of the try block).

FinnStutzenstein:
@CatoTH I'm confused. Why did you close this PR and opened the same one two times?

PR #3457, opened 2017-10-24
try/catch for inconsistent data

CatoTH:
An attempt to fail less dramatically when inconsistent data occurs in the database (when the change recommendation refers a line that does not exist in the motion, e.g. if the motion has beed shortened afterwards)

(I opened a new PR, as a strange commit somehow sneaked into my last one at #3451)


emanuelschuetze:
@CatoTH Why do you created two pull requests? #3456 and #3457 Please close one that I can test the other.

CatoTH:
The other one is deleted. I got some strange error messages from Github when I created the former one, so I created the second one.

PR #3458, opened 2017-10-25
Remove supervisord.conf volume from Dockerfile

boehlke:
Not needed any more.

### DIFF
--- a/openslides/core/management/commands/collectstatic.py
+++ b/openslides/core/management/commands/collectstatic.py
@@ -27,7 +27,7 @@ class Command(CollectStatic):
 
     def collect(self) -> Dict[str, Any]:
         try:
-            destination_dir = os.path.join(settings.STATICFILES_DIRS[0], 'js')
+            destination_dir = os.path.join(settings.STATIC_ROOT, 'js')
         except IndexError:
             # If the user does not want do have staticfiles, he should not get
             # the webclient files either.
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,5 +35,4 @@ RUN mkdir /data && chown openslides /data
 USER openslides
 EXPOSE 8000
 USER openslides
-VOLUME /supervisord.conf
 VOLUME /data
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,3 +36,4 @@ USER openslides
 EXPOSE 8000
 USER openslides
 VOLUME /data
+VOLUME /plugins


emanuelschuetze:
We have to wait until the changes of new multiinstance backend is merged und in production. Otherwise current instances does not work with updated Dockerfile.

boehlke:
@emanuelschuetze Bitte zusammen mit dem PR in multi-backend mergen. Ich teste das dann morgen/übermorgen auf dem Multi-Test-Server.

openslides/core/management/commands/collectstatic.py
```diff
@@ -27,7 +27,7 @@ def handle(self, **options: Any) -> str:
 
     def collect(self) -> Dict[str, Any]:
         try:
-            destination_dir = os.path.join(settings.STATICFILES_DIRS[0], 'js')
+            destination_dir = os.path.join(settings.STATIC_ROOT, 'js')
```
emanuelschuetze: @FinnStutzenstein Can you check if this is the best solution for OpenSlides core?

FinnStutzenstein: @normanjaeckel May have a look

normanjaeckel: I do not understand this solution. The command should collect the statics into the collected static dir and not into anything else.

boehlke: @normanjaeckel In my estimation, STATICFILES_DIRS[0] is the first directory where static files should be collected from. STATIC_ROOT is the directory to collect files into! See https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-STATIC_ROOT and https://docs.djangoproject.com/en/2.0/ref/settings/#staticfiles-dirs ... do you come to the same conclusion?

normanjaeckel: Hm. You are right. I mixed it up.

But the problem is, now the extra flags of the collectstatic command won't work properly (like --dry-run, --clear, --link, --ignore) and the STATIC_ROOT dir will get some stuff even if the user doesn't want this.

boehlke: The problem is probably that the whole server/client code mix-up is a hack :-D
Do you need the features dry-run, clear etc?

emanuelschuetze: @normanjaeckel Any concerns here? We need a solution that we can merge it soon for using with new multi-backend.

normanjaeckel: I wrote a last question to @boehlke via email to fully understand the problem.

normanjaeckel:
See #3804

PR #3459, opened 2017-10-27
Motion optimizations

FinnStutzenstein:
- Two column layout for the motion block slide and fixed the labels. Added some more space in the layout. Added number of motions in the title.
- navigation in the motion detail view
- 'undone' field in state filter; added divider between normal states and end states.

Question: I do like the disabled navigation buttons, if the motion is the first or last, but whats about the text? The other possibility would be to hide the button.

### DIFF
--- a/openslides/core/static/js/core/base.js
+++ b/openslides/core/static/js/core/base.js
@@ -1420,8 +1420,8 @@ angular.module('OpenSlidesApp.core', [
 ])
 
 // filters the requesting object (id=selfid) from a list of input objects
-.filter('notself', function() {
-    return function(input, selfid) {
+.filter('notself', function () {
+    return function (input, selfid) {
         var result;
         if (selfid) {
             result = [];
@@ -1437,6 +1437,26 @@ angular.module('OpenSlidesApp.core', [
         return result;
     };
 })
+
+// Wraps the orderBy filter. But puts ("", null, undefined) last.
+.filter('orderByEmptyLast', [
+    '$filter',
+    function ($filter) {
+        return function (array, sortPredicate, reverseOrder, compareFn) {
+            var falsyItems = [];
+            var truthyItems = _.filter(array, function (item) {
+                var falsy = item[sortPredicate] === void 0 ||
+                    item[sortPredicate] === null || item[sortPredicate] === '';
+                if (falsy) {
+                    falsyItems.push(item);
+                }
+                return !falsy;
+            });
+            truthyItems = $filter('orderBy')(truthyItems, sortPredicate, reverseOrder, compareFn);
+            return _.concat(truthyItems, falsyItems);
+        };
+    }
+])
 
 // Make sure that the DS factories are loaded by making them a dependency
 .run([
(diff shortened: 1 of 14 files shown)


FinnStutzenstein:
@emanuelschuetze Do we need the `orderByEmptyLast` filter in some other list view? I would guess in all

emanuelschuetze:
> @emanuelschuetze Do we need the orderByEmptyLast filter in some other list view?

Yes, e.g. for user number. So you can use it in all list views.

FinnStutzenstein:
Done.

FinnStutzenstein:
@emanuelschuetze Updated: Not used workflows are not shown in the state filter. If just one workflow is given, the workflow name is not shown.

openslides/core/static/js/core/base.js
```diff
@@ -1438,6 +1438,26 @@ angular.module('OpenSlidesApp.core', [
...
 })
 
+// Wraps the orderBy filter. But puts ("", null, undefined) last.
+.filter('orderByEmptyLast', [
+    '$filter',
+    function ($filter) {
+        return function (array, sortPredicate, reverseOrder, compareFn) {
+            var falsyItems = [];
+            var truthyItems = _.filter(array, function (item) {
+                var falsy = item[sortPredicate] === undefined ||
```
normanjaeckel: Please use `void 0` instead of `undefined`.

openslides/core/static/js/core/site.js
```diff
@@ -573,17 +573,31 @@ angular.module('OpenSlidesApp.core.site', [
...
             };
+            var existsStorageEntry = function () {
+                return $sessionStorage[tableName];
+            };
+            var storage = existsStorageEntry();
+            if (storage) {
+                self = storage;
+            }
+
+            self.existsStorageEntry = existsStorageEntry;
```
normanjaeckel: This attribute is not used in your new code. Is it used somewhere else? If not please remove it (and then maybe do not put this logic into an extra variable and function).

openslides/core/static/js/core/base.js
FinnStutzenstein: Done.

openslides/core/static/js/core/site.js
FinnStutzenstein: Done.

openslides/core/static/js/core/site.js
```diff
@@ -573,17 +573,30 @@ angular.module('OpenSlidesApp.core.site', [
...
-    function () {
-        var createInstance = function () {
+    '$sessionStorage',
+    function ($sessionStorage) {
+        var createInstance = function (tableName) {
             var self = {
                 column: '',
                 reverse: false,
             };
+            var existsStorageEntry = function () {
```
normanjaeckel: This can be simplified:
```javascript
var storage = $sessionStorage[tableName];
if (storage){
    self = storage;
}
```

FinnStutzenstein: Done.

emanuelschuetze:
Thanks!

PR #3460, opened 2017-10-27
Expose ports 8001-8008 for new multiinstance servers.

FinnStutzenstein:


boehlke:
This can be closed... we do not need it: Reason: The rkt networking binds 8000 in each container.
- instance1: 8000:PORT1
- instance1: 8000:PORT2
...

PR #3465, opened 2017-10-29
Better handling of inserted paragraphs

CatoTH:


normanjaeckel:
Ok, but the regex lines look very complex. Maybe add a short comment.

emanuelschuetze:
> Ok, but the regex lines look very complex. Maybe add a short comment.

@CatoTH You can do it in one of your next pull request if you like.

PR #3466, opened 2017-10-29
Bugfix no word wrap after insertion

CatoTH:
For once, I added a yarn-command "karma:watch" that runs the tests repetedly.

Then, this should fix a problem with the line numbering algorithm. Before, the line breaking only considered the current text node when deciding where to break. This lead to suboptimal results when the current node is an inline text node starting pretty late on a line and without a whitespace character at the beginning.
Now, a mechanism is added that can break a previous node.

### DIFF
--- a/DEVELOPMENT.rst
+++ b/DEVELOPMENT.rst
@@ -139,6 +139,9 @@ Run client tests by starting karma::
 
     $ yarn run karma
 
+Watch for file changes and run the tests automatically after each change::
+
+    $ yarn run karma:watch
 
 OpenSlides in big mode
 ======================
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -47,6 +47,7 @@ Motions:
 - Show motion identifier in (current) list of speakers [#3442]
 - Added navigation between single motions [#3459].
 - Improved the multiselect state filter [#3459]. 
+- Added karma:watch command [#3466].
 
 Elections:
 - Added pagination for list view [#3393].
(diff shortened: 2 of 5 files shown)


DEVELOPMENT.rst
```diff
@@ -135,10 +135,13 @@ To run some server tests see `.travis.yml
 b. Running AngularJS test cases
 '''''''''''''''''''''''''''''''
 
-Run client tests by starting karma::
```
normanjaeckel: Please reinsert the second colon, see RST defs.

DEVELOPMENT.rst
```diff
@@ -135,10 +135,13 @@ To run some server tests see `.travis.yml
...
+Run client tests by starting karma:
```
normanjaeckel: Please use two colons.

DEVELOPMENT.rst
```diff
@@ -135,10 +135,13 @@ To run some server tests see `.travis.yml
...
 
     $ yarn run karma
 
+Watch for file changes and run the tests automatically after each change:
```
normanjaeckel: Here too.

DEVELOPMENT.rst
emanuelschuetze: Done

DEVELOPMENT.rst
emanuelschuetze: Done

DEVELOPMENT.rst
emanuelschuetze: Done

Issue #3427, opened 2017-09-27
Bug in restricted_data_cache


FinnStutzenstein:
Only save cache in `add_element` if the cache exists.

FinnStutzenstein:
This can currupt one's cache. To recreate:
 - Delete the cache from a user.
 - Make an autoupdate that affects the user (e.g. project something)
 - reload the users browser. 

https://github.com/OpenSlides/OpenSlides/blob/4c9a8b4ec75882d6d6f02228327ba6a57eb1bd3a/openslides/utils/autoupdate.py#L82
This `build_cache_for_user` logic could be separated and called, if the user does not have a cache, if `add_element` is called.

ostcar:
The quickfix is to add a method:
```python
    def update_element(self, user_id: int, collection_string: str, id: int, data: object) -> None:
        if self.exists_for_user(user_id):
            self.add_element(user_id, collection_string, id, data)
```
To the restricted_data_cache an use it at this line:
https://github.com/OpenSlides/OpenSlides/blob/9fd0121132c8c0a69e7a3f63fbed8b19ef3ec33b/openslides/utils/autoupdate.py#L296

The problem is, that this method is not atomic so in very rare cases it is possible, that the error still occurs. The best solution would be to use a [lua script](https://redis.io/commands/eval) instead.

emanuelschuetze:
Done with a hotfix in #3483. Please find a better solution with implementing a cache api in #3428. Thanks.

Issue #3428, opened 2017-09-27
Extend the restricted_data_cache


FinnStutzenstein:
- Add `del_user()`, `del_all()` to delete caches if e.g. permission changes.
- Add `build_for_user()` to (re)build the cache for a users. This should be callable from the commandline, so the cache for a big assembly can be prebuild.

emanuelschuetze:
see also hoftix of #3427 in pr #3483

emanuelschuetze:
@ostcar When do you could implement this cache api? In the next weeks want to release 2.2.

ostcar:
I fixed the first part of this issue in #3600.

I don't think any more , that added `build_for_user()` to the cache is a good idea. If it would be added, it should not be used in `ws_add_site` because the current implementation not only creates the cache but also creates the `formatted_data`. So if the cache would be build in a separate function, the two for-loop have to be run twice. This would be unnecessary slow.

Another disadvantage would be, that all functions in [this code-block](https://github.com/OpenSlides/OpenSlides/blob/d185e05ab194f05214e677bb1a537d142fa75389/openslides/utils/autoupdate.py#L85) have to be imported into the cache-file.

So the solution is, that the person, who creates the management command copys all the linked lines into the management command.

ostcar:
Can this issue be closed after #3784 is merged?

PR #3483, opened 2017-11-08
Adds elements to restricted data cache only if cache already exists

emanuelschuetze:
Hotfix for #3427.
Prevents corrupt cache if restricted data cache is cleared while runtime.

Successfully tested with online instance.
@ostcar @FinnStutzenstein Please review.


emanuelschuetze:
@ostcar I'll merge this hotfix now. Please try to implement, as discussed, a better solution with a real cache api (see #3428).

PR #3487, opened 2017-11-10
Use native twisted for daphne

FinnStutzenstein:
Currently we are using daphne with a busy loop. We can use native twisted when enabling it. See this code: https://github.com/django/daphne/blob/master/daphne/server.py#L115

Warning: I do not know whether this is an improvement or just another buffer causing a bottleneck. So please DO NOT MERGE. I need to test this. Also I'm unsure about txredisapi. I needed to install this.

Does anyone have some experience using native twisted?

I hope not using a busy loop will raise performance much. On our multi-instance server all daphnes does have *much* cpu time, so they are heavily used. This may chill the server a bit and raise the throughput of daphne.

### DIFF
--- a/openslides/asgi.py
+++ b/openslides/asgi.py
@@ -7,3 +7,6 @@ from .utils.main import setup_django_settings_module
 setup_django_settings_module()
 
 channel_layer = get_channel_layer()
+
+# Use native twisted mode
+channel_layer.extensions.append("twisted")
--- a/requirements_big_mode.txt
+++ b/requirements_big_mode.txt
@@ -6,3 +6,4 @@ asgi-redis>=1.3,<1.4
 django-redis>=4.7.0,<4.8
 django-redis-sessions>=0.5.6,<0.6
 psycopg2>=2.7,<2.8
+txredisapi==1.4.4
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,7 @@ docs/_build/*
 *.egg-info
 build/*
 dist/*
-debug/
+debug/*
 .DS_Store
 .idea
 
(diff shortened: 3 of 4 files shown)


FinnStutzenstein:
I tested the serverperformance with `oswstest`: The average and min times are not affected much. But also in former tests the maximal time was very high. This may be caused by just a few clients which do not get connections because daphne is heavy loaded. Those values are just a bit higher then average.

I think when using a busy loop, some requests didn't get noticed as fast as using twisted mode (with maybe callbacks for new connections). For the first try-out this works very well.

normanjaeckel:
Hm. I do not have any experience here. Sorry. :-)

emanuelschuetze:
@ostcar What do you think about this pull request?

ostcar:
Hm. I do not have any experience here. Sorry. :-)

FinnStutzenstein:
@emanuelschuetze So then I would prefer merging this.

openslides/asgi.py
```diff
@@ -7,3 +7,5 @@
 setup_django_settings_module()
 
 channel_layer = get_channel_layer()
+# Use native twisted mode
```
normanjaeckel: Please add a newline before the comment.

requirements_big_mode.txt
```diff
@@ -6,3 +6,4 @@ asgi-redis>=1.3,<1.4
 django-redis>=4.7.0,<4.8
 django-redis-sessions>=0.5.6,<0.6
 psycopg2>=2.7,<2.8
+txredisapi==1.4.4
```
normanjaeckel: Please test (now or later) Windows Portable and install steps on fresh Ubuntu 16.04 so that this new package does not bring new headache on release day. :-)

openslides/asgi.py
emanuelschuetze: Done

requirements_big_mode.txt
emanuelschuetze: "pip install txredisapi" (v1.4.4) successfully tested on Windows (Portable) and Linux (arch, ubuntu 16.10).

PR #3490, opened 2017-11-11
Reduce autoupdates by adding a follow_recommendation view (fixes #348…

FinnStutzenstein:
…8, fixes #3489)

PR #3491, opened 2017-11-11
Autoupdate fixes: Bundle autoupdates to save resources and prevent race conditions

FinnStutzenstein:
The idea of this pull request has its origins in the following problem: Seeing in #3490 some views send multiple autoupdates for the same model. This causes race condition, because the "right" order of autoupdates is not guaranteed. So one update contains a "wip" like data of the model and the last autoupdate contains the actual ready-processed modeldata. If the order switches, the client gets corrupted data.
The other thing is, that e.g. autoupdates for an item, motion and user are fired. This results in three separate autoupdates. One could simply pack them together and send this one bundle of all changes to the clients. This should improve the performance of big solutions of OpenSlides.

Also seen in the linked PR: Its much work and effort to search all places where this can occur and fix this. I was thinking of a think like `with transaction.atomic():` where all changes are cached and at the end of the block one transaction is fired.

In this PR I try to reach three thinks:
a) Block all autoupdates while the server is in a speacial block of working on the models.
b) After this block, the new logic should have kept track what was changed and send a bundled autoupdate.
c) This should be easy to add to our existing code without changing much (compare to the linked PR)

I'm intrerested in hearing what you think about this approach. If you have another idea for the global variable `blocked_autoupdate_callback` and some other code structure, let me know!

As you see, this covers only `inform_changed_data`. The deleted data and collection element list functions are not implemented yet. This has to be integrated somehow clever.

I do have one problem: For what exactly is this `information` argument for the `inform_*_data` functions? Where is it used and can this be solved somehow else. The goal of the `set()` is to track all modified instances **once**, so they are unique. Whats about two times the same model with different informations, how should this be handled?

@ostcar @normanjaeckel @emanuelschuetze Please have a look at this new approach

Thanks!

### DIFF
--- a/openslides/global_settings.py
+++ b/openslides/global_settings.py
@@ -33,6 +33,7 @@ MIDDLEWARE = [
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'openslides.utils.autoupdate.AutoupdateBundleMiddleware',
 ]
 
 ROOT_URLCONF = 'openslides.urls'
--- a/openslides/motions/views.py
+++ b/openslides/motions/views.py
@@ -13,7 +13,7 @@ from rest_framework import status
 
 from ..core.config import config
 from ..utils.auth import has_perm
-from ..utils.autoupdate import inform_changed_data
+from ..utils.autoupdate import inform_changed_data, bundle_autoupdates
 from ..utils.collection import CollectionElement
 from ..utils.rest_api import (
     DestroyModelMixin,
@@ -149,6 +149,7 @@ class MotionViewSet(ModelViewSet):
         headers = self.get_success_headers(serializer.data)
         return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
 
+    @bundle_autoupdates
     def update(self, request, *args, **kwargs):
         """
         Customized view endpoint to update a motion.
@@ -298,6 +299,7 @@ class MotionViewSet(ModelViewSet):
         return Response({'detail': message})
 
     @detail_route(methods=['put'])
+    @bundle_autoupdates
     def set_state(self, request, pk=None):
         """
         Special view endpoint to set and reset a state of a motion.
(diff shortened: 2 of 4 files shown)


emanuelschuetze:
@ostcar @normanjaeckel I'm very interesting what do you think about this pr. Look very good in our tests.

ostcar:
I have no experience with that option. If it looks good in the tests, then you can use it

normanjaeckel:
Hm. This is why we invented the skip_autoupdate attribute.

What about using a Django middleware? Then we do not have to add any extra code to the apps and views. (I am really not a fan of global variables because it is different to debug side effects.)

See my comment in #2779 for a small part of this problem.

FinnStutzenstein:
But this type of merging autoupdates together is more intuitive and easy to use. Now we have to keep track which method causes an autoupdate and wich not. Maybe during a method some other model is saved (This happens, if a motion is updated. The agenda item is send and under special circumstances also the submitter is send to the client). This produces much code overhead and is complicated to maintain.

If we see, that all vies should get this decorator, we can easily add it to a base class. So there is one point, where this can be set.

In which case can a middleware helps us here? The autoupdate system is not managed through middlewares. Also I'm not a fan of global variables, but the autoupdate system in it's own is not encapsulated into a class. All functions are on the global scope, so this design-decision was not made by myself. The block variable would be a perfect class member, if we had classes here.

FinnStutzenstein:
I added a middleware to globally bundle autoupdates. Yes, this makes the `skip_autoupdate` obsolete. We can remove it, if this is merged. But first, we have to solve all open questions.
But first, do you support this new way of optimizing autoupdates?

ostcar:
The problem with middleware are, that they only work on http-requests. So you have to make sure that there are no other way how data can be changed. (For example in a migration, through websocket, etc.)

Also have a look at the kwargs ```immediately``` in the [channel send method](https://github.com/django/channels/blob/69c59ee8b4924cd346e0a2b1724979375ad9924c/channels/channel.py#L32). We disable that feature [here](https://github.com/OpenSlides/OpenSlides/blob/3161bdaa05ef547ab757955356601a6915c7cd62/openslides/utils/autoupdate.py#L40).

I don't say, that your way is not the best one, but maybe other projects have the same problem and found other solutions.

But please get rid of the global variables. That kills any possibility to make openslides thread save again. When there are two workers, that use the same global variable, then you will get very funny results.

normanjaeckel:
See my e-mail. Needs discussion.

normanjaeckel:
See #3539

PR #3492, opened 2017-11-12
Template improvements for motions and projector

emanuelschuetze:
- Use same title layout for (current) list of speakers slide.
- Set default choose for current list of speakers slide to 'project as slide'
  if more than one projectors available.
- Remove additional control bottons in list of speakers view to stop/end current/next speech.
  (User should better use the start/stop buttons behind the speaker names.)
- Fix typo in projection defaults.
- Set sidebox in motion slide to fix position (should not scroll).
- Fix order of amendments in motion detail template.
- Increase font size of motion block items on slide.


emanuelschuetze:
@FinnStutzenstein Please review later. I've to merge it now.

PR #3495, opened 2017-11-17
New change recommendation type other

FinnStutzenstein:


emanuelschuetze:
@FinnStutzenstein Please rebase
@normanjaeckel The "other" state is required for recommendations which does not require a modification of text lines, e.g. mark some lines with recommendation "Annahme als Material für $Gremium". So you can mix multiiple recommendations (insertion/deletion/replacement) with some special (free) recommendations.

normanjaeckel:
So this is not a *change* recommendation because nothing has changed ...? I do not like it but I will approve and merge nevertheless.

Issue #3253, opened 2017-05-12
Load css from plugins like js.


FinnStutzenstein:
This may be solved with #2209 together. Currently I have to put css into the templates with style-tags. This is not very efficient and clear.

A plugin could deliver a `css_files` array, like we do this with js already.

But a very nice way would be that OS automatically discovers js- and css-files and includes them into the index.html

FinnStutzenstein:
New Proposal: Because of the new build process for all css files with sass (see #3507 ) the gulp build process has to be run after every change in plugins. This could not be done e.g. with the windows portable build.

So I would recommend a js provider where each app can register its css files. The provider then creates link-tags and load the css file into the DOM. Each app can have its own build process may using sass, plain css or the tool in flavor of the plugin's author. He/She just have to register the one or many resulting css files in js.

normanjaeckel:
Sounds good.

jsaalfeld:
With the rework of the client this shouldn't be an issue anymore. If it still persists, please reopen it.

Issue #2209, opened 2016-06-29
Load css files for each app


emanuelschuetze:
Load css files per each app in angular - like js files.

FinnStutzenstein:
Can we then split up these massive app.js and projector.js? I would like this much..

emanuelschuetze:
> Can we then split up these massive app.js and projector.js?

You mean *.css... Yes, we should do it for 2.2

normanjaeckel:
We should setup Sass by the way. But first we should think about #3011 because jspm has also a css loader plugin which can be used easily.

FinnStutzenstein:
Why using a css loader, if we can just concat all css together?
```
gulp.task('css', function () {
    return gulp.src([
            path.join('openslides', '**', 'static', 'css', '**', '*.css'),
        ])
        .pipe(concat('openslides.css'))
        .pipe(gulp.dest(path.join(output_directory, 'css')));
});
```

emanuelschuetze:
@FinnStutzenstein I think we can close this issue because #3507 is merged, right?

PR #3507, opened 2017-12-01
split up css classes

FinnStutzenstein:
What do you think of this new way of managing all the css classes?

There will be two new gulp tasks that create two css files for either site and projector. The files included are given in the core/static/css/site.css (resp. projector.css) with import statements. The gulp task parses them and put everything together. Every app should get a site.css and  projector.css as well as css files for parts like the diff-view. These files gets imported by every app's site.css/projector.css.

This should be the the import order:
```
core/static/css/site.css
  - core/static/css/core/site.css
      - fonts.css
      - helper.css
      - mediaqueries.css
      - osTable.css
      - ...
  - motions/static/css/motions/site.css
      - diffView.css
  - users/static/css/users/site.css
  - ...
core/static/css/projector.css
  - core/static/css/core/projector.css
      - fonts.css
      - helper.css
      - slides.css
      - ...
  - motions/static/css/motions/projector.css
      - diffView.css
  - users/static/css/users/projector.css
  - ...
```


normanjaeckel:
Please check #3253, #2209 and #3011. Of course we can change the way we handle CSS and we do not have to use the jspm loader but concat all together. I only recommend to think about this for a moment.

gulpfile.js
```diff
@@ -102,6 +103,15 @@ gulp.task('templates', function () {
         .pipe(gulp.dest(path.join(output_directory, 'js')));
 });
 
+gulp.task('css_site', function () {
```
normanjaeckel: Please add docstring with short explanation.

FinnStutzenstein:
@normanjaeckel I switched to sass. Is this now ok? I'll do a _variables.scss with all things we use often (OS colors, alert boxes colors, sizes, fonts, ..).

normanjaeckel:
Ok, looks good. I do not understand why we use and underscore here but if you do it is ok for me.

FinnStutzenstein:
@normanjaeckel I do not like the underscore neither, but this is [sass syntax for partial stylesheets](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#Partials__partials):
`If you have a SCSS or Sass file that you want to import but don't want to compile to a CSS file, you can add an underscore to the beginning of the filename. This will tell Sass not to compile it to a normal CSS file. You can then import these files without using the underscore.`

FinnStutzenstein:
You can test the site. should be ok for the first splitting.

emanuelschuetze:
@normanjaeckel Ready for next review.

FinnStutzenstein:
Sass itself does neither globs nor variables in imports. This package enables glob support: https://www.npmjs.com/package/gulp-sass-glob

I'll try this

emanuelschuetze:
@FinnStutzenstein Please ping if this pr is ready to merge.

FinnStutzenstein:
@normanjaeckel I tried to use sass-glob, but somehow it does not work and I'm not exactly sure why. So lets lust merge this PR.

Issue #3510, opened 2017-12-07
Drag-and-drop file upload


serge1peshcoff:
Hello there.
I am the IT responsible of one big event (800+ people), where we are using OpenSlides to display presentations/videos on stage and on the Internet. And sometimes we need to upload a lot of files to the OpenSlides in a short time, and it takes a lot of time to fill all the values into the upload form, find the file in the filesystem etc.

I was thinking that it'd be cool to have the ability to upload new files by drag-and-dropping it. And I think that this could make a life more easier for a lot of OpenSlides users, because it's easier to drag file to the Files menu rather then pressing a button, filling in all the info and then clicking Save. And it should save some time if you need to upload a lot of time (especially if you don't have that much time, like in our case).

The way I see it, is to add the drag-and-drop area to `Files` area so the user can drag the file from the filesystem to it. The user would also be able to upload file the old way.

The new workflow would be something like that:
- user goes into `Files` tab and drags the file onto the area
- a `New file` dialog is displayed with the file path filled in
- user fills in other values, like file name, hidden/not hidden etc. and press `Save` and it's saved.

I want to implement it by myself, since it doesn't seem like a big deal to me (although I didn't look at the code structure thoroughly yet) and make a PR to this repository.

What do you think about this idea? Do you think that's a good idea and would you merge it into your repository if it'll be implemented properly?

(I've opened this issue as a platform for discussion, because I had some cases when I implemented some feature for some tool that I use and the author said that they didn't need it, making this work pointless, I don't want to implement something that won't be merged.)

Thanks in advance.

emanuelschuetze:
Nice feature request! Yes, please create a pull request. We will test and review it. If everything ok we will merge it into OpenSlides core.

serge1peshcoff:
Btw do you have an instruction how to run it locally for development and stuff like that? The thing is I didn't have a lot of Python experience so far, so I have no idea how to run it locally.

(Also it'd be cool to have this in README to make the life easier for the future developers, I can make a PR on that one I'll know how to do it)

emanuelschuetze:
Please read the [DEVELOPMENT.rst](https://github.com/OpenSlides/OpenSlides/blob/master/DEVELOPMENT.rst) which should help your for setup a local development installation based on master.

PR #3511, opened 2017-12-10
Drag-and-drop files upload

serge1peshcoff:
I've implemented some basics of drag-and-drop for files upload in "Files" tab.
I've decided that implementing this in a way that opens up a form with the "newFile" field already filled in is too complicated, so I've decided to implement it the other way: when the user drops the file onto the drop area, it is uploaded automatically with the 'name' field set as the filename and the 'uploader' field set as the current user.

NOTE: this PR is not for merging yet, it's mostly WIP, as I've implemented some basic stuff and there's a lot of things to be done before merging, for now it's just to show the idea of what I'm going to do.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,3 +26,4 @@ Authors of OpenSlides in chronological order of first contribution:
     Thomas Junk <lilith2k3@archlinux.us>
     Meinert Leinigen <openslides@leinigen.cc>
     Andreas Engler <engel.a@web.de> (Russian translation)
+    Sergey Peshkov <serge1peshcoff@gmail.com>
--- a/openslides/core/static/css/app.css
+++ b/openslides/core/static/css/app.css
@@ -2023,3 +2023,33 @@ tr.selected td {
         width: 100%;
     }
 }
+
+/* Styles for 'Files' page header' */
+
+.col1 .header .title-h1 {
+    display: table-cell;
+    padding-right: 20px;
+}
+
+.col1 .header .title-upload-wrapper {
+    display: table-cell;
+    width: 100%;
+    margin: 20px;
+}
+
+.col1 .header .title-upload-div {
+    border: 2px solid rgb(126, 122, 122);
+    color: rgb(126, 122, 122);
+    font-weight: bold;
+    text-align: center;
+    border-radius: 5px;
+}
+
+.col1 .header .title-upload-div:hover {
+    background-color: rgb(233, 232, 232);
+}
+
+.col1 .header .files-submenu {
+    display: table-cell;
+    padding-left: 20px;
+}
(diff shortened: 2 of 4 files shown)


normanjaeckel:
See #3510.

normanjaeckel:
When this PR is finished later please do not forget to add a line to AUTHORS file.

serge1peshcoff:
Sooo, I've implemented it finally, check this out. (It's not a final version though, see below.)

Two concerns before this is finalized:
1) I am not that good at CSS, can somebody help me improving the UX/UI of the dropzone?
2) I want to add another Config setting to set the default 'hidden' attribute for the files uploaded via dropzone, so the user would have the ability to upload hidden files by default. How can I add another Config option?

Thanks in advance.
@normanjaeckel @emanuelschuetze @FinnStutzenstein

FinnStutzenstein:
Hi, I would prefer to implement the dropzone into the `+new` dialog. See [this example](http://nervgh.github.io/pages/angular-file-upload/examples/simple/). Maybe adding a queue for all selected files, so the user can set the name and hidden flag for each file.
You can also mark the file-choose-field as a dropzone, too: The user can choose, if he wants to select files of drop them there. (File: core/static/templates/core/file.html)

serge1peshcoff:
@FinnStutzenstein AFAIK you can already drop the files to `Upload` button on the `New` dialog and that would work just as you described. I can do this also, but this would mean doing extra steps for the user, and my approach intends to save clicks and extra steps for the user.

Can you please describe the upload workflow you are thinking about? I was thinking about the following:
1) User opens the `Files` tab
2) User drops the file onto the dropzone
3) The `hidden` attribute of the mediafile-to-upload is set to either true or false, depends on the config.
4) The file is uploaded successfully/an error is displayed.

Also I was thinking about hiding the dropzone and displaying it only if the user is dragging something onto the browser window, but it should be a little bit harder to implement. What do you thing about?

emanuelschuetze:
@serge1peshcoff We discussed and tested your pull request. We don't like the drag&drop area in the mediafile list view - sorry, but it looks not very nice...
Suggestion: Please provide a multi select file chooser which can upload multiple selected files at once. Optional: you can add a preview table in create form of all selected files...

emanuelschuetze:
@serge1peshcoff What do you think about my suggestion (multi select file chooser)? Can you implement it in your pull request? Then we could merge it for 2.2.

serge1peshcoff:
@emanuelschuetze looks cool, but I don't have that much time lately unfortunately so I don't have time working on that :(
I'll try to get back on that when I can though.

emanuelschuetze:
Closed - see new pull request #3650

PR #3514, opened 2017-12-15
Remove HTML tags from description in assignment PDF.

emanuelschuetze:
Fix usage of clearfix in assignment templates.


emanuelschuetze:
@normanjaeckel I have to merge it now. Please review later. Thanks.

normanjaeckel:
Ok.

PR #3516, opened 2017-12-15
Allow to delete own motions

FinnStutzenstein:
Can be tested, just the validationerror has to be changed.

### DIFF
--- a/openslides/motions/views.py
+++ b/openslides/motions/views.py
@@ -65,7 +65,7 @@ class MotionViewSet(ModelViewSet):
         """
         if self.action in ('list', 'retrieve'):
             result = self.get_access_permissions().check_permissions(self.request.user)
-        elif self.action in ('metadata', 'partial_update', 'update'):
+        elif self.action in ('metadata', 'partial_update', 'update', 'destroy'):
             result = has_perm(self.request.user, 'motions.can_see')
             # For partial_update and update requests the rest of the check is
             # done in the update method. See below.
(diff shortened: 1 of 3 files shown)


emanuelschuetze:
@normanjaeckel Please review later. I have to merge it today. Thanks.

FinnStutzenstein:
@normanjaeckel see #3518 .

openslides/motions/views.py
```diff
@@ -65,7 +65,7 @@ def check_view_permissions(self):
         """
         if self.action in ('list', 'retrieve'):
             result = self.get_access_permissions().check_permissions(self.request.user)
-        elif self.action in ('metadata', 'partial_update', 'update'):
+        elif self.action in ('metadata', 'partial_update', 'update', 'destroy'):
```
normanjaeckel: Please extend comment in line below.

openslides/motions/views.py
```diff
@@ -85,6 +85,19 @@ def check_view_permissions(self):
...
         return result
 
+    def destroy(self, request, *args, **kwargs):
+        """
+        Destroy is allowed if the user has manage permissions, or he is the submitter and
+        the current state allows to edit the motion.
+        """
+        motion = self.get_object()
+
+        if (has_perm(request.user, 'motions.can_manage') or
```
normanjaeckel: Please switch if condition:
```
if <not allowed>:
    raise ...
return super()....
```

PR #3524, opened 2018-01-02
Add mechanism to check the code-style

jsaalfeld:
pycodestyle is formerly known as pep8 and one of the widest known style checks for python and jslint is a well-known javascript-linter. With that mechanisms in place it should be easier to maintain a good code quality.


gulpfile.js
```diff
@@ -66,6 +67,15 @@ gulp.task('js-libs', function () {
         .pipe(gulp.dest(path.join(output_directory, 'js')));
 });
 
+// Lints all js files
```
normanjaeckel: Please move this task down to our extra tasks.

gulpfile.js
```diff
@@ -26,12 +26,13 @@ var argv = require('yargs').argv,
     templateCache = require('gulp-angular-templatecache'),
     through = require('through2'),
     uglify = require('gulp-uglify'),
-    vsprintf = require('sprintf-js').vsprintf;
+    vsprintf = require('sprintf-js').vsprintf,
+    jshint = require('gulp-jshint'),
```
normanjaeckel: This line already exists in line 22. Please remove one line.

gulpfile.js
```diff
@@ -26,12 +26,13 @@ var argv = require('yargs').argv,
...
+    stylish = require('jshint-stylish');
```
normanjaeckel: Please use alphabetic order of imports.

gulpfile.js
```diff
@@ -214,7 +224,8 @@ gulp.task('default', [
         'fonts-libs',
         'ckeditor',
         'angular-chosen-img',
-        'translations'
+        'translations',
+	'js-lint'
```
normanjaeckel: Please do not run linter on default.

gulpfile.js
normanjaeckel: I see that we already have a js-hint task. So please use (and maybe modify) our task.

setup.cfg
```diff
@@ -27,3 +27,8 @@ disallow_any = unannotated
 
 [mypy-tests.*]
 ignore_errors = true
+
+[pycodestyle]
+max-line-length = 160
+statistics = True
+
```
normanjaeckel: Maybe this line can be removed. Last character in every file should be \n and not \n\n.

jsaalfeld:
We gonna talk about this topic in the next week at a meeting I think

PR #3525, opened 2018-01-09
Filter for email and more dialog to the user, if some users have no emails

FinnStutzenstein:


emanuelschuetze:
@normanjaeckel Please review later. I have to merge it today.

openslides/users/static/js/users/site.js
```diff
@@ -765,9 +774,54 @@ angular.module('OpenSlidesApp.users.site', [
             $http.post('/rest/users/user/mass_invite_email/', {
                 user_ids: user_ids,
             }).then(function (success) {
+                var numEmails = success.data.count;
```
normanjaeckel: Please wrap these (massive) message code into a nice separate function or module.

PR #3529, opened 2018-01-12
Hide identifier field in motion update form if the user is delegate

### LINKED ISSUE #3528: Submitter can see identifier field in motion edit form
As normal user (delegate) create a new motion (title, text reason).
Now edit this motion (which is allowed for submitter in complex workflow):
The user can see also the (empty) identifier field, which is not desired! But a filled field is not saved after sending request - so not really critical. 
BUT: We should hide this field in form for normal users without manage permissions.

FinnStutzenstein:
…fixes #3528)


emanuelschuetze:
@normanjaeckel Please review later. I have to merge it today.

normanjaeckel:
Ok.

PR #3535, opened 2018-01-20
Preparations for the SAML plugin; Fixed caching of main views.

FinnStutzenstein:


emanuelschuetze:
CHANGLOG entry and missing `$http` in Logout factory added. Rebased.

PR #3539, opened 2018-01-20
Added autoupdate bundle middleware.

normanjaeckel:


FinnStutzenstein:
@normanjaeckel Thought: When calling `infrom_changed_data` not during a request cycle, it currently does nothing. Do we have other cases like tests or extern events that needs to send autoupdate out of the rquest cycle, e.g. Notify? If so, I would prefer not to change the behaviour of inform_changed_data, so If we are not in a request cycle and infomr_changed_data is called, we just send the requested autoupdate.

FinnStutzenstein:
@emanuelschuetze Yes, the client sends two seperate requests. This causes two bundles to be send.

normanjaeckel:
Ready to merge.

PR #3564, opened 2018-02-02
Fixed email translations, motion sorting, undefined in DOCX, reduced …

FinnStutzenstein:
…motion search

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,6 @@ script:
   - coverage report --fail-under=44
 
   - DJANGO_SETTINGS_MODULE='tests.settings' coverage run ./manage.py test tests.integration
-  - coverage report --fail-under=74
+  - coverage report --fail-under=73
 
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -30,6 +30,6 @@ script:
   - coverage report --fail-under=44
 
   - DJANGO_SETTINGS_MODULE='tests.settings' coverage run ./manage.py test tests.integration
-  - coverage report --fail-under=74
+  - coverage report --fail-under=73
```
normanjaeckel: Ah, please do not decrease test coverage but add some tests. :-)

(But it would be no problem for me if you do not have time for this, but beware of future debts ...)

normanjaeckel:
Maybe add Pull Request number to the respective line in CHANGLOG.

emanuelschuetze:
Ok, but for me a changelog entry is not required (because just bug fixing of alread existing features/entries).

PR #3566, opened 2018-02-06
Added --no-template-caching

FinnStutzenstein:


FinnStutzenstein:
@normanjaeckel Hi, do you have an idea for this issue (see travis) to fix?
In "normal" execution, we set the `openslides.args` to the parser args. But this does not apply for the tests.

FinnStutzenstein:
@normanjaeckel Do you have an idea how to fix this?

normanjaeckel:
Hm. Looks like that the tests also need to add args attribute to openslides module.

Better: Only check args in views if it is not None.

FinnStutzenstein:
Done a new approach. Need more development.

FinnStutzenstein:
@normanjaeckel Done some rework. Please review the new approach

FinnStutzenstein:
@emanuelschuetze ready for merge

PR #3568, opened 2018-02-06
Custom fonts for pdf and projector

FinnStutzenstein:


openslides/core/apps.py
```diff
@@ -46,6 +47,9 @@ def ready(self):
         post_permission_creation.connect(
             delete_django_app_permissions,
             dispatch_uid='delete_django_app_permissions')
+        post_permission_creation.connect(
```
normanjaeckel: Please do not add migration process code here.

openslides/core/config.py
```diff
@@ -125,13 +125,14 @@ def __setitem__(self, key: str, value: Any) -> None:
...
+        if config_variable.input_type == 'static':
             if not isinstance(value, dict):
-                raise ConfigError(_('logo has to be a dict.'))
+                raise ConfigError(_('This has to be a dict.'))
             whitelist = (
                 'path',
+                'default',
                 'display_name',
-            )
+                )
```
normanjaeckel: Please remove some whitespaces.

openslides/core/config.py
```diff
@@ -144,7 +145,7 @@ def __setitem__(self, key: str, value: Any) -> None:
             for entry in value:
                 if not isinstance(entry, dict):
                     raise ConfigError(_('Every value has to be a dict, not {}.'.format(type(entry))))
-                whitelist = (
+                whitelist = (  # type: ignore
```
normanjaeckel: Type is set, not?

openslides/core/signals.py
```diff
@@ -36,6 +36,15 @@ def delete_django_app_permissions(sender, **kwargs):
     Permission.objects.filter(content_type__in=contenttypes).delete()
 
 
+def delete_logo_permission(sender, **kwargs):
+    """
```
normanjaeckel: See above.

openslides/core/static/js/core/base.js
```diff
@@ -860,6 +860,74 @@ angular.module('OpenSlidesApp.core', [
...
+            },
+            getForCss: function (key) {
+                var url = this.getUrl(key);
+                if (url) {
+                    var ext = _.last(url.split('.'));
+                    return "url('" + url + "') format('" +
+                        extensionFormatMap[ext] + "')";
+                }
+            },
+            isMediafileUsedAsFont: function (mediafile) {
```
normanjaeckel: Hm. Is it possible not to have mediafile code in core/base?

openslides/core/static/templates/projector.html
```diff
@@ -1,12 +1,40 @@
...
+<html lang="{{ selectedLanguage }}" ng-controller="LanguageAndFontCtrl" class="no-js">
 <meta charset="utf-8">
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
 <base href="/">
 <title>OpenSlides &ndash; Projector</title>
 <link rel="stylesheet" href="static/css/openslides-libs.css">
 <link rel="stylesheet" href="static/css/openslides-projector.css">
 <link rel="icon" href="/static/img/favicon.png">
+
+<style type="text/css">
```
normanjaeckel: Is it possible to move this style defs to an extra file?

openslides/core/apps.py
FinnStutzenstein: Done. moved to migrations

openslides/core/config.py
FinnStutzenstein: Mypy has a conflict with the `whitelist` in line 131. There it is a `Tuple[ str, str, str ]` and here a `Tuple[ str, str, str, str ]`....

openslides/core/static/js/core/base.js
FinnStutzenstein: Done. Refactored logo and font code.

openslides/core/static/templates/projector.html
FinnStutzenstein: I'm not a fan of this, because angular has to replace the `src` there.

FinnStutzenstein:

openslides/core/config.py
normanjaeckel: Then change whitelist in line 131 and here into a list.

openslides/core/migrations/0007_auto_20180130_1400.py
```diff
@@ -0,0 +1,41 @@
...
+def delete_old_logo_permission(apps, schema_editor):
+    """
+    Deletes the old 'can_manage_logo' permission which is replaced with
+    'can_manage_logos_and_fonts'. If this is a fresh database, no permission
+    will be deleted, in fact the old permission does not exist. Django creates
+    the permission after all migration and the old one is not generated.
+    If this is an old database, the new permission will be created and the old
+    one deleted.
+    """
+    Permission.objects.filter(codename='can_manage_logos').delete()
```
normanjaeckel: Now a migration admin can not manage logos any more.

(Sorry. But I nevertheless think it is the right way to change perms in this file. Maybe you have to do it manually: First get all groups with the old perm. Add the new perm to these groups. Then remove the old perm.)

openslides/core/static/templates/projector.html
normanjaeckel: Ok, this is hacky. Please open an issue to clean up this problem later.

openslides/core/migrations/0007_auto_20180130_1400.py
FinnStutzenstein: It's fine. First I thought this does not work, but it does. The reassignment of the permissions is a great idea!

FinnStutzenstein: Hmm, I cannot assign the new permission here. Doesn't django create the permissions after all migrations, does it? So I can filter all groups with the permission, but cannot assign the new one, because it does not exist yet..

FinnStutzenstein: @normanjaeckel whats your opinion about this: https://stackoverflow.com/questions/31735042/adding-django-admin-permissions-in-a-migration-permission-matching-query-does-n/31735043

normanjaeckel: Maybe this is the way.

FinnStutzenstein:
@normanjaeckel Hi, Im stuck at assigning the new permission to all necessary groups. See this part of the traceback:
```
File "/home/finn/OpenSlides/OpenSlides/openslides/core/migrations/0007_auto_20180130_1400.py", line 29, in delete_old_logo_permission
    
  File "/home/finn/OpenSlides/.venv/lib/python3.5/site-packages/django/db/models/fields/related_descriptors.py", line 881, in add
    self._add_items(self.source_field_name, self.target_field_name, *objs)
  File "/home/finn/OpenSlides/.venv/lib/python3.5/site-packages/django/db/models/fields/related_descriptors.py", line 1018, in _add_items
    (self.model._meta.object_name, obj)
TypeError: 'Permission' instance expected, got <Permission: core | config store | Can manage logos and fonts>
```
I don't know, why I cannot add this permission. The debugger says that `type(perm) -> <class 'django.contrib.auth.models.Permission'>`, which is correct. I've tried two ways: First (as propsed) let Django create all permissions and get the new one to assign. The second way is just create the neede permission.

Both ways lead to this error above.

To test this: Create a db in the current master (and maybe save it that you can just do the migration several times). Switch to this branch and migrate.

openslides/core/static/templates/projector.html
FinnStutzenstein: Ok.

normanjaeckel:
Sorry I do not have any time now to test this manually.

emanuelschuetze:
Successfully retested - with db migration.

PR #3569, opened 2018-02-06
Agenda DOCX export, motion log message when comment edited, motion se…

### LINKED ISSUE #3558: enhancement: return to actual page in agenda list
I have an agenda that is so long that it needs 2 pages.

- click on an agenda item on page 2 to modify it
- click now on the button "back to overview

result: page 1 of the agenda is displayed and not page 2 where the item appears

I found out that it would be better to return to the actual page where the item appears in the agenda because one often has the task to modify items one after each other. If you disagree, feel free to close this as WONTFIX.

FinnStutzenstein:
…quential number in detail view and PDF, save pagination state to session storage (fixes #3558).


FinnStutzenstein:
@normanjaeckel Can you check this travis error? Interestingly it does not occour on my computer. There 2 logs are created, one for update and one for the comment. May this be a cache issue?

normanjaeckel:
By the way: Thank you for the refactoring. Generic code looks much better now.

FinnStutzenstein:
@normanjaeckel I've created this test and on my machine it's fine. There should be two logs. One for creation and one for the updated comment.
Can you verify this test on your computer?
I'll put some debug output in the commit and let travis run again..

normanjaeckel:
Travis is ok. So please rebase.

emanuelschuetze:
Changelog entry added. Rebased.

PR #3570, opened 2018-02-06
Mark speakers

FinnStutzenstein:


openslides/agenda/views.py
```diff
@@ -57,14 +57,16 @@ def check_view_permissions(self):
...
-    @detail_route(methods=['POST', 'DELETE'])
+    @detail_route(methods=['POST', 'PATCH', 'DELETE'])
     def manage_speaker(self, request, pk=None):
         """
         Special view endpoint to add users to the list of speakers or remove
         them. Send POST {'user': <user_id>} to add a new speaker. Omit
         data to add yourself. Send DELETE {'speaker': <speaker_id>} or
         DELETE {'speaker': [<speaker_id>, <speaker_id>, ...]} to remove one or
         more speakers from the list of speakers. Omit data to remove yourself.
+        Send PATCH {'user': <user_id>} to toggle the marked attribute of the
```
normanjaeckel: Hm. I do not like a simple toggle view. Because one who calls this view can not be sure what result he gets. Please send something like {'user': <user_id>, mark: true}.

FinnStutzenstein: Done.

PR #3576, opened 2018-02-13
Custom format remover plugin for CKEditor.

FinnStutzenstein:
@emanuelschuetze Please check this. It is not ready yet, but the functionally should be there.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -117,6 +117,7 @@ Core:
 - Preparations for the SAML plugin; Fixed caching of main views [#3535].
 - Removed unnecessary OPTIONS request in config [#3541].
 - Added possibility to upload custom fonts for projector and pdf [#3568].
+- Use custom format cleanup plugin for CKEditor [#3576].
 
 Mediafiles:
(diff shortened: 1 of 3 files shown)


FinnStutzenstein:
Maybe another solution: Use the included removeformat-plugin and abuse the `filter` property. The Plugin traverses through all elements in the editor (but I do not understand the code...) and removes the attributes and tags for each node. But first - if specified - all nodes are put through filters, that the user (we) can register. The filter should return `true` for continue cleaning this object or `false` not to clean this. This is done without parsing the editor raw data but with CKEditor's own DOM implementation, so the changes are right in place rather then take all the content, modify it and replace.

We can use the filter to return always `false`, so no element is changed, but we do have the node in the filter, so we can manipulate it our way, namely removing the `color` and `background-color` attributes.

FinnStutzenstein:
@emanuelschuetze Please test again. Changed the whole plugin to the new approach ;)

emanuelschuetze:
@normanjaeckel I have to merge it now. Please review later. Thanks!

normanjaeckel:
Ok.

PR #3578, opened 2018-02-14
Patch userdetails without redundant declaration of users details

jsaalfeld:
Before this patch you'd have to send either the username, first_name or last_name with any given PATCH request to a user, even though you know his unique identifier: The User ID.

After this patch you just have to send any details and you can update the user information with just the knowledge of the users ID. This simplifies user information updates significantly, because you don't have to GET all information beforehand, if you just want to update one users details. Especially when an 3rd party system manages the users; it just has to map their user information to the ID in OpenSlides without any additional knowledge and is then able to update the users information.

Additionally: The mechanism in place was a bit awkward with patches: If you send, i.e. an update for the title and an update for the username, it changed both; therefore the requirement for having the username there in first place, is redundant and ineffective for a double-check; therefore I updated the check in a way that at least some information has to be in place, if you want to update a given user.

### DIFF
--- a/openslides/users/serializers.py
+++ b/openslides/users/serializers.py
@@ -57,18 +57,22 @@ class UserFullSerializer(ModelSerializer):
 
     def validate(self, data):
         """
-        Checks that first_name or last_name is given. Generates the
+        Checks if the given data is empty. Generates the
         username if it is empty.
         """
-        if not (data.get('username') or data.get('first_name') or data.get('last_name')):
-            raise ValidationError({'detail': _('Username, given name and surname can not all be empty.')})
 
-        # Generate username. But only if it is not set and the serializer is not
-        # called in a PATCH context (partial_update).
         try:
             action = self.context['view'].action
         except (KeyError, AttributeError):
             action = None
+
+        # Check if we are in Patch context, if not, check if we have the mandatory fields
+        if action != 'partial_update':
+            if not (data.get('username') or data.get('first_name') or data.get('last_name')):
+                raise ValidationError({'detail': _('Username, given name and surname can not all be empty.')})
+
+        # Generate username. But only if it is not set and the serializer is not
+        # called in a PATCH context (partial_update).
 
         if not data.get('username') and action != 'partial_update':
             data['username'] = User.objects.generate_username(


openslides/users/serializers.py
```diff
@@ -57,11 +57,19 @@ class Meta:
...
-        if not (data.get('username') or data.get('first_name') or data.get('last_name')):
-            raise ValidationError({'detail': _('Username, given name and surname can not all be empty.')})
+        # Check if one of the fields is actually filled
+        filled_field_keys = []
+        for field in data:
+            if data.get(field) != '':
+                filled_field_keys.append(field)
+
+        # Check if filled fields match to any of the given fields
+        intersecting_keys = list(set(filled_field_keys) & set(self.fields.keys()))
```
normanjaeckel: This solution does not work it one sends only some minor field in a PUT context.

I propose to do in line 74 (old file): Only if username is empty and it is not a PATCH context, then first validate that at least one of the names is given, then generate the username.

jsaalfeld: Do I understand correctly: Just execute my new batch of code, if we are in a PATCH context, but do the old stuff, if we are in a PUT context?

I tested the PUT context, iE User creation and it works as expected. I don't quiet get why username, first_name or last_name should be mandatory at that point, when one of the fields is sufficient to create a user in first place? Or should these fields be mandatory upon user creation?

jsaalfeld:
I changed the commit, so the new condition will only hit, when the request is a PATCH

openslides/users/serializers.py
```diff
@@ -57,19 +57,34 @@ class Meta:
...
+        if action == 'partial_update':
+            filled_field_keys = []
+            for field in data:
+                if data.get(field) != '':
+                    filled_field_keys.append(field)
+
+            # Check if filled fields match to any of the given fields
+            intersecting_keys = list(set(filled_field_keys) & set(self.fields.keys()))
+            if len(intersecting_keys) < 1:
+                raise ValidationError({'detail': _('You cannot send an empty instruction')})
```
normanjaeckel: Why is this a problem? The case that someone sends no valid fields is a standard case and is (I think so) already solved by the DjangoRESTFramework. I think what you only want to change is the case in the else clause. Right?

openslides/users/serializers.py
normanjaeckel: Sorry but I missed to answer this. My proposal is not to add much code for the PATCH context but to put some old code into an if clause in case of non-PATCH contexts.

openslides/users/serializers.py
jsaalfeld: The current issue is it is not possible to send a patch to a defined user ID, without either re-distributing the username, first or last name.

This could lead to extra steps in working with the rest frame work:
 * GET the Infos for the user ID first
 * Parse them
 * extract one of those: username, first or last name
 * build patch request
 * send PATCH

instead of just
 * build the patch request
 * send PATCH

Since there is objectively no need to redistribute one of the information during a patch, because an existing data object is changed, one should not be required to redeliver those, especially since those three fields are worthy of protection, since it can prevent the user from logging in.

On the issue of checking simply if one field is filled: It is just a convinience thing, since the class already holds a list of fields, I thought it would be more readable to just check if one is filled, that a change of data is truly happening.
Unfortunately the DjangoRESTFFramework does not prevent in sending an empty data set within a patch, it just won't set anything without a rejection. To prevent unnecessary calculations or workflow triggering this simply check would be sufficient, I think.

jsaalfeld: Done!

normanjaeckel:
@emanuelschuetze last test, then merge.

Issue #3574, opened 2018-02-09
ImportError: cannot import name 'build_endpoint_description_strings'


mwaeckerlin:
[Build / installation instructions for Linux](https://github.com/OpenSlides/OpenSlides/blob/master/README.rst#installation) are wrong.

Setup (in a Docker build) according to the documentation:
```
FROM mwaeckerlin/ubuntu-base
MAINTAINER mwaeckerlin

RUN apt-get update && apt-get install -y build-essential python3 python3-dev python3-pip python3-venv
#RUN ln $(which pip3) /usr/bin/pip
RUN mkdir /openslides
WORKDIR /openslides
RUN python3 -m venv .virtualenv
RUN . .virtualenv/bin/activate && pip3 install --upgrade setuptools pip && pip install openslides
CMD . .virtualenv/bin/activate && openslides
EXPOSE 8000
```
Result:
```
marc@jupiter:~/docker/openslides$ docker build --pull --rm --force-rm -t mwaeckerlin/openslides .
Sending build context to Docker daemon  3.072kB
Step 1/9 : FROM mwaeckerlin/ubuntu-base
latest: Pulling from mwaeckerlin/ubuntu-base
Digest: sha256:e9207a59d15739dec5d1b55412f15c0661383fad23f8d2914b7b688d193c0871
Status: Image is up to date for mwaeckerlin/ubuntu-base:latest
 ---> 66aa06d7a865
Step 2/9 : MAINTAINER mwaeckerlin
 ---> Using cache
 ---> 62e7ec9e8beb
Step 3/9 : RUN apt-get update && apt-get install -y build-essential python3 python3-dev python3-pip python3-venv
 ---> Using cache
 ---> c06ef14d1d39
Step 4/9 : RUN mkdir /openslides
 ---> Running in 7acc9a0cf93f
Removing intermediate container 7acc9a0cf93f
 ---> 6455c1486460
Step 5/9 : WORKDIR /openslides
Removing intermediate container 0944cd34180c
 ---> fcb5d355f482
Step 6/9 : RUN python3 -m venv .virtualenv
 ---> Running in a07630d930d4
Removing intermediate container a07630d930d4
 ---> 262562fdd3e2
Step 7/9 : RUN . .virtualenv/bin/activate && pip3 install --upgrade setuptools pip && pip install openslides
 ---> Running in 50e79be3ab8c
Collecting setuptools
[...]

Issue #3572, opened 2018-02-09
Check psycopg2 dependency


FinnStutzenstein:
After doing a `pip install -U -r requirements.txt` I get this warning when starting OpenSlides:
```
/home/finn/OpenSlides/.venv/lib/python3.5/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
  """)
```
Check this.

jsaalfeld:
Should be fixed with #3579

Issue #3522, opened 2017-12-20
Upgrade python requirements


emanuelschuetze:
Check release notes if we can easily upgrade our python requirements, e.g.:
* bleach to 2.x
* django to 2.0.x
* djangorestframework to 3.7.x
* jsonfield to 2.0.x
* setuptools to 38.x
* ...

emanuelschuetze:
@normanjaeckel Can you do this in the next weeks?

normanjaeckel:
Sorry. Maybe I do not have enough time. Lets have a phone call ...

jsaalfeld:
should be done now.

Task for the not-so-easy-upgrades is over at #3582

PR #3579, opened 2018-02-16
Updating requirements and locking versions

jsaalfeld:
Updating the versions (#3522) of  the requirements and locking channels and daphne, because it would break OpenSlides otherwise (#3574).

Also renaming psycopg2 to psycopg2-binary (#3572) to avoid further issues caused by the upcomming renaming.

### DIFF
--- a/requirements_production.txt
+++ b/requirements_production.txt
@@ -1,10 +1,11 @@
 # Requirements for OpenSlides in production in alphabetical order
-bleach>=1.5.0,<1.6
+bleach>=1.5.0,<2.2
 channels>=1.1,<1.2
+daphne<2
 Django>=1.10.4,<1.11
 djangorestframework>=3.4,<3.5
-jsonfield>=1.0,<1.1
-mypy_extensions>=0.3,<1.1
+jsonfield>=1.0,<2.1
+mypy_extensions>=0.3,<0.4
 PyPDF2>=1.26,<1.27
 roman>=2.0,<2.1
-setuptools>=29.0,<35.0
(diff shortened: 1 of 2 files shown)


requirements_production.txt
```diff
@@ -1,10 +1,11 @@
 # Requirements for OpenSlides in production in alphabetical order
-bleach>=1.5.0,<1.6
-channels>=1.1,<1.2
+bleach>=1.5.0,<2.2
+channels<1.2
```
emanuelschuetze: Please keep the lower version as our minimal supported version: `channels>=1.1,<1.2`

requirements_production.txt
```diff
@@ -1,10 +1,11 @@
...
+daphne<2
 Django>=1.10.4,<1.11
```
emanuelschuetze: Why do you not updated to django 2?

jsaalfeld: I tried. It breaks a lot of stuff. Even a small update to django 1.11 breaks some mutables and immutables which cause some trouble.

requirements_production.txt
jsaalfeld: adjusted this back to the original state.

requirements_production.txt
emanuelschuetze: Ok, see new issue #3582

emanuelschuetze:
@normanjaeckel I have to merge it now to get it in new beta version. Please review later. Thanks!

normanjaeckel:
Ok. but please check minimal version of daphne. Now there is no lower version bound.

PR #3590, opened 2018-02-23
Downgrade angular-pdf to 1.3.0 (Fixed #3552).

### LINKED ISSUE #3552: displaying PDF files leads to black projector
- upload this PDF to the files of OS:
[test.pdf](https://github.com/OpenSlides/OpenSlides/files/1658353/test.pdf)
- display it

result: the projector preview window doesn't show the PDF but is completely black

Since in the projector it is displayed but fullscreen, try to make it smaller:

result: using the zoom button in the projector preview has no effect

Using the zoom buttons in the presentation control elements works and when clicking to zoom out the projector preview is now working

- now rotate the PDF via the presentation control elements

result: this works

- now zoom in via the presentation control elements

result: the projector preview gets black again, the projector itself is correct.

- now press the reset zoom button in the presentation control elements

result: now also the projector is black

emanuelschuetze:
using pdfjs-dist#1.3.100
angular-pdf 1.3.1 does not work in Firefox with PDF documents >127kb.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "angular-formly-templates-bootstrap": "~6.2.0",
     "angular-gettext": "~2.3.7",
     "angular-messages": "~1.5.8",
-    "angular-pdf": "~1.3.0",
+    "angular-pdf": "1.3.0",
     "angular-sanitize": "~1.5.8",
     "angular-scroll-glue": "~2.0.7",
     "angular-ui-router": "~0.3.1",


normanjaeckel:
Please keep on tracking this problem for later versions of the problematic package.

emanuelschuetze:
> Please keep on tracking this problem for later versions of the problematic package.

see #3632

PR #3600, opened 2018-02-24
Cache api

ostcar:
Activate mypy and the restricted_data_cache again.

Add methods to delete one user or all users from the restricted_data_cache

### DIFF
--- a/openslides/users/models.py
+++ b/openslides/users/models.py
@@ -1,5 +1,4 @@
 import smtplib
-
 from random import choice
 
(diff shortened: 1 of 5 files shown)


normanjaeckel:
Ok. But del_all() and del_user() are still not in use, right?

emanuelschuetze:
> But del_all() and del_user() are still not in use, right?

@ostcar ping

ostcar:
no. They are not in use. I only build the api so it can be used at other places. But this is not part of this pr

PR #3637, opened 2018-03-07
Paragraph based amendments / Diff

CatoTH:
Current decisions:
- We reuse the ``motion``-table
- The amended paragraphs are stored as an array of strings (and null's for non-changing paragraphs) in a new JSONfield in the ``motionversion``-table.
- The ``Apply text for new amendments``-setting is changed into a drop-down, enabling the two previous behaviors, plus the new one (not implemented yet).
- Amendments are referencing motions - not motion versions. The latter would be more precise, but will be added at a later iteration.


FinnStutzenstein:
@CatoTH Nice work!
I've one request: Could you add a function to a motion returning an array of paragraphed based amendments? So this function would return `[motion, motion, ...]` or just the ids (this is not important) of all amendments. And additionally a function that returns the one selected paragraph. So that I can get the original paragraph, if I have an paragraph based amendment with the changes.
This would be very helpful implementing the overview table for all changes of all motions.

If I have overseen a feature that you already have implemented, please tell me how do you named this function.

Thanks!

CatoTH:
@FinnStutzenstein
The first function would be ``getParagraphBasedAmendments()``.
For the second one, calling ``.getAmendmentParagraphsByMode("original")[0]`` should return the selected paragraph of the base motion.

FinnStutzenstein:
Perfect. This should be helpful!

openslides/motions/static/js/motions/linenumbering.js
```diff
@@ -614,6 +614,36 @@ angular.module('OpenSlidesApp.motions.lineNumbering', [])
...
+        /**
+         * @param {string} html
+         */
+        this.getHeadingsWithLineNumbers = function (html) {
+            var fragment = this._htmlToFragment(html),
+                headings = [];
+            var headingNodes = fragment.querySelectorAll('h1, h2, h3, h4, h5, h6');
+            for (var i = 0; i < headingNodes.length; i++) {
+                var heading = headingNodes.item(i);
+                var linenumbers = heading .querySelectorAll('.os-line-number');
```
FinnStutzenstein: Does this work with the space there? :)

emanuelschuetze:
Rebased to one commit and fixed travis.

openslides/motions/models.py
```diff
@@ -265,7 +268,7 @@ def save(self, use_version=None, skip_autoupdate=False, *args, **kwargs):
         elif use_version is None:
             use_version = self.get_last_version()
             # Save title, text and reason into the version object.
```
normanjaeckel: Comment has to be updated.

openslides/motions/serializers.py
```diff
@@ -2,6 +2,7 @@
 
 from django.db import transaction
 from django.utils.translation import ugettext as _
+from rest_framework import serializers
```
normanjaeckel: Please use ValidationError from ..utils.rest_api and do not import something directly from rest_framework.

openslides/motions/serializers.py
```diff
@@ -344,12 +372,25 @@ class Meta:
...
             validated_comments[id] = validate_html(comment)
         data['comments'] = validated_comments
+
+        if 'amendment_paragraphs' in data:
+            data['amendment_paragraphs'] = list(map(lambda entry: validate_html(entry) if type(entry) is str else None,
+                                                    data['amendment_paragraphs']))
+            data['text'] = ''
+        else:
+            if 'text' in data and len(data['text']) == 0:
+                raise serializers.ValidationError({
```
normanjaeckel: Please use ValidationError here instead of serializers.ValidationError.

openslides/motions/serializers.py
```diff
@@ -344,12 +372,25 @@ class Meta:
...
+                    'text': _('This field may not be blank.')
```
normanjaeckel: I am not sure whether this string needs to be translated. Also the keyword should be 'detail' and not 'text', but I am not sure.

openslides/motions/models.py
FinnStutzenstein: Done.

openslides/motions/serializers.py
FinnStutzenstein: Done.

openslides/motions/serializers.py
FinnStutzenstein: Done.

openslides/motions/serializers.py
FinnStutzenstein: Done.

emanuelschuetze:
Rebased on top of master.
@normanjaeckel Please review again the remarks and approve that we can merge it. Thx.

PR #3638, opened 2018-03-08
Disable the agenda hierachy just for searchning, not for filtering

FinnStutzenstein:
This resolves the two issues:
- hide the collapse/expand all link for flat agendas
- You can collapse and expand if the filters are used


emanuelschuetze:

normanjaeckel:
Ok.

PR #3641, opened 2018-03-08
Adding new permission to groups during migration from 2.1.1 to 2.2

FinnStutzenstein:


FinnStutzenstein:
Todo: The contenttypes does not exist, if a new DB is created. I'll create them, if needed and django does not crash..

normanjaeckel:
> Todo: The contenttypes does not exist, if a new DB is created. I'll create them, if needed and django does not crash..

If a new DB is created we do not have to migrate permissions to some groups. so you do not have to create the contenttypes. Just skip ... Right?

FinnStutzenstein:
Yes right, if the permission is not found, we skip already

FinnStutzenstein:
@emanuelschuetze This is my os 2.11 db. The migration works fine, the admin group is not created and I can login with admin/admin. You have to remove the two last extensions get the database:
[db.sqlite3.os211.txt](https://github.com/OpenSlides/OpenSlides/files/1802501/db.sqlite3.os211.txt)

emanuelschuetze:
@normanjaeckel Ping

PR #3644, opened 2018-03-10
OL Offsets - fixes #3643

### LINKED ISSUE #3643: Change Recommendations in OLs with [start]
When a numbered list with a start-attribute greater than 1 is used in a motion and a change recommendation is made, the offset in numbering is not respected in the change recommendation and the following list item numbers in the diff view:
bildschirmfoto 2018-03-10 um 17 55 55

CatoTH:
After this patch, affected change recommendations need to be re-created, as the offset is hard-coded in the HTML of the recommendation.


emanuelschuetze:
Changelog entry added. Ready for merge.

normanjaeckel:
Ok.

PR #3646, opened 2018-03-11
Go to original view after deleting the Title CR...

### LINKED ISSUE #3645: Title-Change-Recommendations are not in every case possible
- open a motion and click on the plus sign before its title
- insert a change recommendation and save it
- delete the recommendation because you have e.g. made a typo
- try to create a new, this time correct, recommendation

result: this is not possible. One has first to go back to the motion list, then open the motion again.

Title-Change-Recommendations are new since #3626.

CatoTH:
…if it is the last one - fixes #3645


normanjaeckel:
Ok.

PR #3649, opened 2018-03-12
Fixed Nginx settings

FinnStutzenstein:


serge1peshcoff:
I have a question regarding this: shouldn't the `/media/file-path.extension` request be served as static files instead of being redirected to `/static/templates/index.html`? Or am I misunderstanding something?

emanuelschuetze:
@serge1peshcoff If you do not need a protection for media files (only authorized user with special  permission are allowed to download mediafiles) you can deliver all files as static files, yes. But by default we use the "protection mode" (delivered via django).

FinnStutzenstein:
Hint: See [the protected_serve](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/mediafiles/views.py#L74). It's like a regular server for staticfiles, but checks, if a user has the permission to see this file.

normanjaeckel:
Ok.

PR #3655, opened 2018-03-15
Fix group filtering in users list view

emanuelschuetze:
Use ":true" for angular filter to exact match for searching id (e.g. '2' not '12').
Order groups by id (only important for big mode with caching).


emanuelschuetze:

normanjaeckel:
Ok.

PR #3657, opened 2018-03-15
Add firstLine to cache definition - fixes #3656

### LINKED ISSUE #3656: Wrong line number caching on identical diffs
If two change recommendations are performing an identical change to two identical sections of the text, the line number is wrongly cached, i.e. the second change sees the same line numbers as the first one.
bildschirmfoto 2018-03-15 um 15 27 07

CatoTH:


normanjaeckel:
Ok.

PR #3658, opened 2018-03-16
Get the path of the settings.py file

FinnStutzenstein:
Needed for the saml plugin: If OS is started with `runworker` we do not have any chance of getting the path to the settings. Unfortunately `from django.conf import settings; print(settings.__file__);` does not work, so this workaround has to be taken.

### DIFF
--- a/openslides/utils/settings.py.tpl
+++ b/openslides/utils/settings.py.tpl
@@ -11,6 +11,7 @@ https://docs.djangoproject.com/en/1.10/ref/settings/
 import os
 from openslides.global_settings import *
 %(import_function)s
+SETTINGS_FILEPATH = __file__  # Do not change!
 
 # The directory for user specific data files
 


emanuelschuetze:
@normanjaeckel Can you please look at this pull request. Is this ok for your?

normanjaeckel:
> For a plugin I need an extra configuration file in JSON, so the plugin cannot use the settings.py, but the new file should be in the same directory as the settings.py. So for read and write, I need to have the path of the settings.

The Django settings concept is, as I mention, a encaspulated module which is loaded at some time of the startup (that's why sometimes Django calls it LazySettings). So the settings can live in any Python module a user wants to use.

Is it possible to find a solution for the plugin to use its extra configuration with something that has nothing to do with the settings path? I am not sure but I do not like the possibility to write something into the same dir as the settings file during runtime. Maybe I do not understand the problem ...

emanuelschuetze:
@normanjaeckel ping

FinnStutzenstein:
@normanjaeckel I will add to the plugin documentation, that this line has to be added manually.

PR #3679, opened 2018-03-24
fixed indentions in pdf, remove every tab in html (fixes #3678)

### LINKED ISSUE #3678: Remove all \t in al html
Just replace them with an enpty string. These \t breaks the diff, becuase somehow (i think during C&P) the \t get into the original motion text. When creating a changerecommendation the CKEditor does not allow \t and removes them. The result are red dashes over a whitespace. Luckyly they are not visible in OpenSlides, but in the PDF they are visible.
Putting this into `utils.validate.validate_html` should do the business.

red-dashed

FinnStutzenstein:


normanjaeckel:
Ok.

PR #3682, opened 2018-03-27
Changes for the voting plugin

FinnStutzenstein:
- new-style template hooks. See changes in the template hook documentation
- reordered user list view to get more space for extra things
- added template hooks.

### DIFF
--- a/openslides/core/static/js/core/base.js
+++ b/openslides/core/static/js/core/base.js
@@ -553,27 +553,57 @@ angular.module('OpenSlidesApp.core', [
 ])
 
 // Template hooks
-// 2 possible uses:
-// - { Id: 'myHookId', template: '<button>click me</button>' }
-// - { Id: 'myHookId', templateUrl: '/static/templates/plugin_name/my-hook.html' }
-// It is possible to provide a scope, that is merged into the scope of the templateHook.
-// This overrides functions/values of the parent scope, but there may are conflicts
-// with other plugins defining the same function/value. E.g.:
+// Possible uses:
+// 1. { id: 'myHookId', template: '<button>click me</button>' }
+// 2. { id: 'myHookId', templateUrl: '/static/templates/plugin_name/my-hook.html' }
+// 3. { id: 'myHookId' }
+//
+// Deprecated: Give the id with 'Id'. Please use 'id'.
+//
+// Option 3 is for just changing the scope (see below), but not the original content. This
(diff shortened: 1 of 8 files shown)


FinnStutzenstein:
Please review, so we can put this into the master. Especially the changes in the templatehooks are interesting for other plugins.

openslides/core/static/js/core/base.js
```diff
@@ -553,27 +553,57 @@ angular.module('OpenSlidesApp.core', [
...
+//
+// Deprecated: Give the id with 'Id'. Please use 'id'.
+//
+// Option 3 is for just changing the scope (see below), but not the original content. This
+// is usefull to alter a JS behavior, e.g. on a ng-click. In this case, override is false
+// for this template hook.
+//
+// It is possible to provide a scope, that is merged into the surrounding scope.
+// You can override functions or values of the surrounding scope by providing them:
 // { Id: 'hookId', template: '<button ng-click="customFn()">click me</button>',
```
normanjaeckel: Use `id` instead of `Id`. Also use `customOrOverwritten` instead of `customFn`, right?

openslides/core/static/js/core/base.js
```diff
@@ -553,27 +553,57 @@ angular.module('OpenSlidesApp.core', [
...
 //   scope: {
-//     customFn: function () { /*Do something */ },
+//     customOrOverwritten: function () { /*Do something */ },
 //   },
 // }
+// Or you provide a function that returns an object of functions/values to overwrite to
+// get access to the scope merged in:
+// { Id: 'hookId', template: '<button ng-click="customFn()">click me</button>',
```

openslides/core/static/js/core/base.js
```diff
@@ -553,27 +553,57 @@ angular.module('OpenSlidesApp.core', [
...
 .factory('templateHooks', [
     function () {
         var hooks = {};
         return {
             hooks: hooks,
             registerHook: function (hook) {
-                if (hooks[hook.Id] === undefined) {
-                    hooks[hook.Id] = [];
+                // Deprecated: Set the new style 'id', if 'Id' is given.
+                if (hook.id === undefined) {
```
normanjaeckel: Please use `void 0` instead of `undefined`. (I see that there was `undefined` before ... ;-))

normanjaeckel:
Ok, Only stylistic remarks.

PR #3683, opened 2018-04-04
Add fonts to the medifile serve view

FinnStutzenstein:


openslides/mediafiles/views.py
```diff
@@ -77,9 +77,11 @@ def protected_serve(request, path, document_root=None, show_indexes=False):
...
         return HttpResponseNotFound(content="Not found.")
 
-    if ((not has_perm(request.user, 'mediafiles.can_see') or
-            (mediafile.hidden and not has_perm(request.user, 'mediafiles.can_see_hidden'))) and
-            not mediafile.is_logo()):
+    can_see = has_perm(request.user, 'mediafiles.can_see')
+    is_special_file = mediafile.is_logo() or mediafile.is_font()
+    is_hidden_but_no_perms = mediafile.hidden and not has_perm(request.user, 'mediafiles.can_see_hidden')
+
+    if not can_see or (is_hidden_but_no_perms and not is_special_file):
```
normanjaeckel: I think the logic here is not the same as before:

Before is was `if (not can_see or is_hidden_but_no_perms) and not is_logo`.

Now the brackets are somewhere else ...

FinnStutzenstein:
Yes, the logic changed.
Before, everyone can get logos, but now can_see is the base requirement. Maybe you have logos, that only users should see or fonts, that you are not allowed to share.

PR #3685, opened 2018-04-06
Fixed motion slide not showing any diffs

FinnStutzenstein:
the motion id was used to search for change recommendations, not the active version id.


emanuelschuetze:

PR #3686, opened 2018-04-09
Reset scroll level for each new projection.

emanuelschuetze:


emanuelschuetze:
@FinnStutzenstein Please review carefully if my suggestion is ok for you.

openslides/core/views.py
```diff
@@ -249,6 +249,11 @@ def activate_elements(self, request, pk):
             raise ValidationError({'detail': 'Data must be a list.'})
 
         projector_instance = self.get_object()
+        # reset scroll level
```
FinnStutzenstein: This leads to not wanted behaviour:
The `activate_elements` is (I think only) used by stable projector elements. So the scroll level resets, if one projects a message or the current list of speakers overlay.
For the most non stable elements we use a combination of `clear_elements` and `prune_elements` (E.g. see in motions/base.js line 653 `project()`).
So just resetting the scroll level for pruning elements should be enough, or do you have a case where `activate_elements` should lead to a resetted scroll level?

emanuelschuetze: No, I was not sure if `active_elements` is used... I'll change it.

emanuelschuetze: Done. Reset scroll level was used in `prune_elements` only.

emanuelschuetze:
Changelog entry added.

emanuelschuetze:

openslides/core/views.py
```diff
@@ -272,6 +272,10 @@ def prune_elements(self, request, pk):
...
+        if (projector_instance.scroll != 0):
```
normanjaeckel: The scroll level reset should be done after line 278 (serialiver.save()) because we do not want to reset something in case of ValidationError or other exception.

PR #3690, opened 2018-04-10
Linebreak protection for motion sidebox (Fixed #3593).

### LINKED ISSUE #3593: missing linebreak protection in projection of election results
As you can see in this screenshot:
screenshot-localhost 8000-2018-02-23-19-44-32

OS breaks the line between the value and the % character. This looks ugly. There should be a protected space between the number and the unit.
The line must also not be broken after the opening and before the closing brace.

emanuelschuetze:


emanuelschuetze:

PR #3691, opened 2018-04-10
Use simple pdf export button for delegates.

emanuelschuetze:
Don't use drop down list with one entry only.
Show agenda docx entry for managers only.


emanuelschuetze:

Issue #3396, opened 2017-09-06
Motion cache is not updated


ostcar:
https://github.com/OpenSlides/OpenSlides/blob/37b2996b6cbc585988a921426e3c8fe613403b94/openslides/motions/serializers.py#L277

@normanjaeckel I think this is a bug. If you call
```python
state = motion.state
state.required_permission_to_see = 'some perm'
state.save()
```
then the root_rest_element of state (=Workflow) is updated through the autoupdate system and therefore only the cache of the Workflow is updated. But this means, that the cache of the motion is not updated, so there is still the old required_permission_to_see in it.

I think it is not a good idea to hardlink a string from one root_rest_element in a value from another. I do not understand why this field is needed at all. It should be possible to get the permission throw the state.

normanjaeckel:
I agree. The client should care about this by resolving motion -> state -> required_permission_to_see

 @FinnStutzenstein Can you have a look at the client code?

FinnStutzenstein:
@normanjaeckel @ostcar Yes, we should remove the `required_permission_to_see` from the motion model. But this is not a thing for the client. If you search for this property, I just get two hits: `motions/serializer.py` and in `motions/access_permissions.py`. So this field is not used on the client side, in fact the clinet already resolves the permission the way Norman described.

normanjaeckel:
Ok. You can remove it.

emanuelschuetze:
@FinnStutzenstein Currently, we use in some custom workflow states the `required_permission_to_see` attribute. So please let us discuss before you remove it.

emanuelschuetze:
Already reported in #3455:

Similar problem when a user gets a new permission (e.g. "motions.can_see_comments"). The user see the new comment fields but without content. Also reload does not helps because the restricted user cache contains was not updated (with new comment field values). Deleting the specific cache key helps. After reload the config values are visible in OpenSlides. - The REST api already shows the correct values.

Any hints how to fix this?

ostcar:
see https://github.com/OpenSlides/OpenSlides/pull/3600

You have to [remove the cache for the user](https://github.com/ostcar/OpenSlides/blob/39da8f942fb4ef43603cb570122007c4b7c72b45/openslides/utils/cache.py#L399), when the permissions change

emanuelschuetze:
> You have to remove the cache for the user, when the permissions change

@ostcar Do you have time to create a pull request?

PR #3696, opened 2018-04-11
removed motion.state_required_permission_to_see

FinnStutzenstein:
I removed the serializer field as discussed in #3396 . I might think I know why we had this attribute: It was never used in the client and so the caching is not the point here. Now we have to query the motions state every time we want to have restricted data to get the `required_permission_to_see`. Additionally I had to solve the circular imports of 'models<-->access_permissions` with the import in the `get_restricted_data` method.
If you don't like this, we just can close this and the issue.


openslides/motions/access_permissions.py
```diff
@@ -49,7 +49,8 @@ def get_restricted_data(
                     is_submitter = False
 
                 # Check see permission for this motion.
-                required_permission_to_see = full['state_required_permission_to_see']
+                from .models import State  # import it here to resolve circular imports
+                required_permission_to_see = State.objects.get(pk=full['state_id']).required_permission_to_see
```
normanjaeckel: I see: This is a problem. I think we do not want to have extra DB queries here. @ostcar What do you think?

emanuelschuetze: @normanjaeckel @ostcar ping

normanjaeckel: I think we do not want to have extra DB queries here.

PR #3697, opened 2018-04-11
Protect numbering of the agenda

FinnStutzenstein:


openslides/agenda/config_variables.py
```diff
@@ -9,6 +9,15 @@ def get_config_variables():
 
     It has to be evaluated during app loading (see apps.py).
     """
+    yield ConfigVariable(
+        name='agenda_enable_numbering',
+        label='Enable numbering for agenda items',
```
emanuelschuetze: Please add config string also as translateable string into site.js

emanuelschuetze:
Looks good. One remark. And add Changelog entry please.

emanuelschuetze:
rebased

normanjaeckel:
Ok, but this is also something that might be solved with a settings variable to spare some cockpit config buttons.

PR #3698, opened 2018-04-11
Proposal for top pagenavigation

FinnStutzenstein:
@emanuelschuetze If this is ok, I would like to adapt also assignments, motions, files and users.


emanuelschuetze:
Added `&:focus` to css. Added issue number.
Ready for merge.

normanjaeckel:
Ok.

PR #3702, opened 2018-04-11
new style for motion agenda title

FinnStutzenstein:


emanuelschuetze:
fixed typo ("thi" -> "this")

openslides/motions/static/js/motions/base.js
```diff
@@ -276,9 +276,8 @@ angular.module('OpenSlidesApp.motions', [
                 getListOfSpeakersTitle: function () {
                     var title = gettextCatalog.getString('Motion');
                     if (this.identifier) {
-                        title += ' ' + this.identifier + ':';
+                        title += ' ' + this.identifier;
                     }
-                    title += ' ' + this.getTitle();
```
FinnStutzenstein: If a motion does not have an identifier, the resoult would be `Motion`. We should return the motion title, if no identifier is given.

normanjaeckel:
Ok.

PR #3713, opened 2018-04-17
One request for each projection. Added some validation for clear_elem…

FinnStutzenstein:
…ents and prune_elemnts

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -132,6 +132,7 @@ Core:
 - Reset scroll level for each new projection [#3686].
 - Scroll to top on every state change [#3689].
 - Added pagination on top of lists [#3698].
+- New api route to project items with just one request needed [#3713].
 
 Mediafiles:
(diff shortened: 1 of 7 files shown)


FinnStutzenstein:
After merge: Create an issue for utils/main.py the `#type: ignore`. The origin is the upate of mypy from 0.580 to 0.590.

emanuelschuetze:
@normanjaeckel You can already start with your review.

FinnStutzenstein:
What do you get for a response text?

emanuelschuetze:
http://localhost:8000/rest/core/projector/project/
-> 400 Bad Request
{"detail":"The id \"{'projectorId': 1, 'mode': 'original'}\" has to be int."}

FinnStutzenstein:
@normanjaeckel Yes we may do this. But clean everything up in another PR.

PR #3722, opened 2018-04-25
precalculate projectorchanges in the client

FinnStutzenstein:


normanjaeckel:
Can you please explain this? I do not see the advantages at the moment.

FinnStutzenstein:
@emanuelschuetze

emanuelschuetze:
The purpose of this pull request: The user who controls the projector needs an _immediate_ feedback from OpenSlides as soon as he/she project a slide. Currently you have to wait until the autoupdate is received to get a blue project button - this could be take some seconds. Now with this pull request we change the project button state directly with the botton click - the autoupdate confirms this button state (nothing happens) or change it (if something goes wrong).
Summarizing: This PR improves the response time of the ui.

emanuelschuetze:
Addition: Without this pull request users clicked sometimes 2 times on project button because nothing happened after first click (in slow network environment). Then the slide was deprojected and the projector was empty. This case we want to eliminate.

FinnStutzenstein:
The request does respond quickly, but the button just changes after an autoupdate which is slow. So these "fake changes" are actually correct.

normanjaeckel:
This pull request changes the projector button immediatelly even before the server request is sent. So once again: What about adding a spinner?

emanuelschuetze:
I don't like a spinner over a projector button. The current situation is in my opinion the best solution for the end user who doesn't notice the (slow) duration of auto updates.
Please let us discuss it on phone if you have more arguments. Then I can explain the use cases.

PR #3734, opened 2018-05-04
Fixes for the item projection (fixes #3733)

### LINKED ISSUE #3733: Projection of agenda item including all sub items failed
* Create 3 agenda items A, B and C.
* Move B and C as sub items of A.
* Project A -> OK
* Project A - now with 'include all sub items' option -> Failed with JS error:
 `Error: "id" must be a string or a number!`

FinnStutzenstein:
There was a race between deprojection and projection when changing the subitem. Now, the client waits that the deprojection is finished before project the new one. This causes a flickering in the client (which is technically correct). This special case for changing the subitems projection state can be handled seperately, but not in this PR. THis just fixes the issue and eliminates the race between two requests.

### DIFF
--- a/openslides/core/static/js/core/base.js
+++ b/openslides/core/static/js/core/base.js
@@ -744,7 +744,7 @@ angular.module('OpenSlidesApp.core', [
                     element: {name: this.getResourceName(), id: this.id},
                 };
             }
-            ProjectHelper.project(requestData);
+            return ProjectHelper.project(requestData);
         };
         BaseModel.prototype.isProjected = function() {
             // Returns the ids of all projectors if there is a projector element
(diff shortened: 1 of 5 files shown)


openslides/core/static/js/core/base.js
```diff
@@ -1336,12 +1337,15 @@ angular.module('OpenSlidesApp.core', [
 
                 Projector.inject(projectorsChanged);
 
-                $http.post('/rest/core/projector/project/', data).then(null,
-                    function (error) {
+                return $q(function (resolve, reject) {
```
normanjaeckel: This $q is not necessary. This should work ...
```javascript
return $http.post(...).catch(function (error) {
    // revert the changed made earlier
    Projector.inject(originalProjectors);
    return $q.reject(error)
});
```

PR #3735, opened 2018-05-08
Fixed percentage for special values (fixes #3736), added config to couple the candidates with the list of speakers (closes #3737)

### LINKED ISSUE #3708: Move code to reset scroll level
Move some lines of code, see #3686

### LINKED ISSUE #3736: Wrong calculation of election results
- look at this screenshot:
screenshot-2018-5-9 wahl listenfindungskommission openslides
There were 49 valid ballot papers. I entered the yes votes and used -2 for No votes because we did not distinguish between No or absent.

As you can see OS calculates for every case -4.08%, no matter how many yes votes somebody got.

### LINKED ISSUE #3737: no way to prevent automatic speaker's list setting
- create a new election to elect 2 persons
- add 2 persons as candidates.
- since the 2 candidates are not at the assembly some other persons speak for them
- now start a ballot

result: OS puts the 2 candidates automatically on the speaker's list

There is no way to prevent this. This is a problem because in real assemblies some persons are not at the assembly. Therefore the speaker's list consist of candidates and other persons. For some elections the assembly rules are so that candidates are not speaking at all. (e.g. election to preside the assembly)
Therefore there must be a way to prevent the automatic addition to the speaker's list.

FinnStutzenstein:


normanjaeckel:
Ok.

PR #3742, opened 2018-05-16
Delete restricted data cache on permission changes (closes #3396)

### LINKED ISSUE #3396: Motion cache is not updated

FinnStutzenstein:


normanjaeckel:
Ok.

PR #3745, opened 2018-05-16
Autoupdate with affected users on group delete (fixes #3455)

### LINKED ISSUE #3455: Missing autoupdate on group delete
When deleting a group, all users in the deleted group doesn't update. A reload is necessary that all 'undefined' strings in the client disappear.

FinnStutzenstein:


normanjaeckel:
Ok.

PR #3746, opened 2018-05-16
Throw errors, if the pds generation fails, if an image could not be l…

### LINKED ISSUE #3730: OS is not giving feedback about missing files in PDF
In #3668 I had the case that I could not create a ballot paper because of a missing image file. Unfortunately OS did not warn/inform me.
This is a problem because I had to create a ballot paper using Excel during the assembly. This was lot of stress and I got negative feedback and critical people found them self confirmed that software like OS is not yet usable. (And there are always critical people, sic!)

Therefore OS should inform admins what the problem is:
- is a file is missing
- if an uncodable character in e.g. a motion text is used preventing the display or creation of a PDF/DOCX
- maybe there are other cases where OS could warn the admins

FinnStutzenstein:
…oaded (fixes #3730)


normanjaeckel:
Ok.

Issue #3739, opened 2018-05-09
OS must block to enter negative votes


donovaly:
OS allows to enter negative votes in general. OS should deny to enter any negative number, except of "-1" and "-2".

emanuelschuetze:
Negative votes are already blocker in assignment votes input form. Try to enter `Yes: -10, No: -10; Abstain: -10':
=> `Fehler: votes: Stelle sicher, dass dieser Wert größer oder gleich -2 ist.`
Only -1 and -2 are allowed as negative votes.

=> closed. Works for me.

PR #3750, opened 2018-05-22
no negative votes and better scrolling in forms with number inputs (c…

FinnStutzenstein:
…loses #3739)


emanuelschuetze:

normanjaeckel:
Ok.

PR #3751, opened 2018-05-22
adding working privacy policy

jsaalfeld:
In order to archive GDPR compliance we need a privacy policy.

I know that this way around is yet far from perfect. In further additions, we could add a standard-version of the GDPR to the translations to the *po files and let the translations happen there. But until then, we should have something valid for coming friday.

### DIFF
--- a/openslides/core/config_variables.py
+++ b/openslides/core/config_variables.py
@@ -58,6 +58,15 @@ def get_config_variables():
         subgroup='Event')
 
     yield ConfigVariable(
+        name='general_event_privacy_policy',
+        default_value='',
+        input_type='markupText',
+        label='Privacy policy',
+        weight=132,
+        group='General',
+        subgroup='Event')
+
+    yield ConfigVariable(
         name='general_event_welcome_title',
         default_value='Welcome to OpenSlides',
         label='Front page title',
--- a/openslides/core/static/js/core/site.js
+++ b/openslides/core/static/js/core/site.js
@@ -359,6 +359,15 @@ angular.module('OpenSlidesApp.core.site', [
                 },
             })
 
+            // privacy policy
+            .state('privacypolicy', {
+                url: '/privacypolicy',
+                controller: 'PrivacyPolicyCtrl',
+                data: {
+                    title: gettext('Privacy policy'),
+                },
+            })
+
             //config
             .state('config', {
                 url: '/config',
@@ -1179,6 +1188,9 @@ angular.module('OpenSlidesApp.core.site', [
         });
     }
 ])
+
+// Privacy Policy Controller
+.controller('PrivacyPolicyCtrl', function () {})
 
 // Config Controller
 .controller('ConfigCtrl', [
@@ -2053,6 +2065,7 @@ angular.module('OpenSlidesApp.core.site', [
         gettext('Event location');
         gettext('Event organizer');
         gettext('Legal notice');
+        gettext('Privacy policy');
         gettext('Front page title');
         gettext('Welcome to OpenSlides');
         gettext('Front page text');
--- a/openslides/core/static/templates/privacypolicy.html
+++ b/openslides/core/static/templates/privacypolicy.html
@@ -1,0 +1,9 @@
+<div class="header">
+  <div class="title">
+    <h1 translate>Privacy Policy</h1>
+  </div>
+</div>
+
+<div class="details">
+  <div ng-bind-html="config('general_event_privacy_policy') | translate"></div>
+</div>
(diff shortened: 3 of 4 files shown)


openslides/core/static/js/core/site.js
```diff
@@ -1180,6 +1189,18 @@ angular.module('OpenSlidesApp.core.site', [
...
 ])
 
+// Privacy Policy Controller
+.controller('PrivacyPolicyCtrl', [
+    '$scope',
+    '$http',
+    function ($scope, $http) {
+        $http.get('/core/version/').then(function (success) {
+            $scope.core_version = success.data.openslides_version;
+            $scope.core_license = success.data.openslides_license;
```
FinnStutzenstein: The `cure_url` is missing here.

openslides/core/config_variables.py
```diff
@@ -57,6 +57,44 @@ def get_config_variables():
         group='General',
         subgroup='Event')
 
+    yield ConfigVariable(
```
FinnStutzenstein: Why just not hardcode this into the `privacypolicy` page? I thing nobody should be allowed to change this.

openslides/core/static/js/core/site.js
jsaalfeld: '''cure_url'''? Can't find that in the code. Can you elaborate?

openslides/core/config_variables.py
jsaalfeld: You should be able to change this, to maybe change some details for additional plugins, other languages and to add administrator mails or other contacts if you wan't to. But the best reason to change this, is if you host it by yourself and you delete logs after 10 days or something like this, this should then be changend in the privacy policy as well. As you can see the privacy policy must or should be changed according to every users setup and/or environment.

FinnStutzenstein:
Hm, Why do we need the OpenSlides link (with url, license and version) on the bottom? It's the same as in the legal notice.
I would say we just leave this out. So just remove this and all code in the controller.

openslides/core/config_variables.py
FinnStutzenstein: Ok.

openslides/core/static/templates/privacypolicy.html
```diff
@@ -0,0 +1,16 @@
...
+  <div class="title">
+    <h1 translate>Privacy Policy</h1>
+  </div>
+</div>
+
+<div class="details">
+  <div ng-bind-html="config('general_event_privacy_policy') | translate"></div>
+  <hr>
+  <p>
+    <a ng-href="{{ core_url }}" target="_blank">
```
FinnStutzenstein: The `core_url` is not defined anywhere.

openslides/core/static/js/core/site.js
FinnStutzenstein: Sorry `core_url`. See comment below. Its used in the template but nowhere defined.

jsaalfeld:
I just thought i looked a bit nicer with something at the buttom. I changed all points.

openslides/core/static/js/core/site.js
```diff
@@ -1180,6 +1189,14 @@ angular.module('OpenSlidesApp.core.site', [
     }
 ])
 
+// Privacy Policy Controller
```
FinnStutzenstein: You can remove the injections and make this one a one-liner.

jsaalfeld:
I also think that the german hardcoded version is not the best option, but it is an O.K. option for friday. We should add a standard version for every language. The current text is written in the most open way possible where every environment should fit in. If you work or log differently, you can fix that settings. I think adding a standard text is helping our users to provide a useful template for them to use. Not every user has studied the GDPR in depth, so the text should help them to have a good guideline if they have a different setup. That's why I think a pre-written text is useful.

jsaalfeld:
I just phoned with @emanuelschuetze and he explained the issues adressed by @normanjaeckel  and @ostcar. I still think that it would be more user friendly to add a default text so that everyone doesn't have to dig as deep into OpenSlides. @emanuelschuetze will copy the "example-text" into the wiki, so users don't have to read as much but it will be strictly marked as example and we don't hold any liability.

emanuelschuetze:
@normanjaeckel Thanks for your feedback. Please review and approve later. Thanks.

emanuelschuetze:
I moved the DSGVO example text to this wiki page:
https://github.com/OpenSlides/OpenSlides/wiki/De:Datenschutzerkl%C3%A4rung

normanjaeckel:
Ok.

PR #3752, opened 2018-05-22
upgrade to node v10 and gulp 4 [WIP]

jsaalfeld:
Since Node 6 is going to be out of support in a couple of months (https://github.com/nodejs/Release#release-schedule) and node 10 only works with gulp 4, i upgraded both.


FinnStutzenstein:
Why did you include the not used privacy policy template?

jsaalfeld:
It was added by accident.

As I worked initially on #3751 I discovered, that my system runs on node 10.1 and therefore gulp 3.9 wasn't working anymore, so I digged into that issue. I rewrote some stuff and accidentally transfered this file into both branches. It is removed now.

gulpfile.js
```diff
@@ -220,7 +220,7 @@ gulp.task('ckeditor-lang', function () {
 });
 
 // Combines all CKEditor related tasks.
-gulp.task('ckeditor', ['ckeditor-defaults', 'ckeditor-skins', 'ckeditor-plugins', 'ckeditor-lang'], function () {});
+gulp.task('ckeditor', gulp.series('ckeditor-defaults', 'ckeditor-skins', 'ckeditor-plugins', 'ckeditor-lang'), function () {});
```
FinnStutzenstein: All task can run in parallel

gulpfile.js
```diff
@@ -233,7 +233,7 @@ gulp.task('translations', function () {
 });
 
 // Gulp default task. Runs all other tasks before.
-gulp.task('default', [
+gulp.task('default', gulp.series(
```
FinnStutzenstein: Same here

gulpfile.js
```diff
@@ -246,24 +246,28 @@ gulp.task('default', [
...
 
 
 /**
  * Extra tasks that have to be called manually. Useful for development.
  */
 
 // Watches changes in JavaScript and templates.
-gulp.task('watch', ['js', 'templates', 'pdf-worker', 'css-site', 'css-projector'], function () {
-    gulp.watch([
+gulp.task('watch', gulp.series(gulp.parallel('js', 'templates', 'pdf-worker', 'css-site', 'css-projector')), function () {
```
FinnStutzenstein: The `series` can be removed, too.

Does the watcher works for you? Here, it just do the tasks and finishes. I tried to put a `console.log` inside the function, but it won't be called, so the watcher isn't set up on my machine. The version:
```
node_modules/.bin/gulp --version
[08:04:17] CLI version 2.0.1
[08:04:17] Local version 4.0.0
```

gulpfile.js
```diff
@@ -246,24 +246,28 @@ gulp.task('default', [
...
     ], ['js']);
-    gulp.watch(path.join('openslides', '*', 'static', 'templates', '**', '*.html'), ['templates']);
-    gulp.watch(path.join('openslides', 'core', 'static', 'js', 'core', 'pdf-worker.js'), ['pdf-worker']);
+    watcher += (path.join('openslides', '*', 'static', 'templates', '**', '*.html'), ['templates']);
+    watcher += (path.join('openslides', 'core', 'static', 'js', 'core', 'pdf-worker.js'), ['pdf-worker']);
     // We cannot differentiate between all scss files which belong to each realm. So if
     // one scss file changes the site and projector css is rebuild.
-    gulp.watch(path.join('openslides', '*', 'static', 'css', '**', '*.scss'), ['css-site', 'css-projector']);
+    watcher += (path.join('openslides', '*', 'static', 'css', '**', '*.scss'), ['css-site', 'css-projector']);
+    watcher.on('all', function(event, path, stats) {
```
FinnStutzenstein: Are you sure this works? `watcher` is an array..

package.json
```diff
@@ -9,7 +9,7 @@
   "devDependencies": {
     "angular-mocks": "~1.5.11",
     "bower": "^1.8.0",
-    "gulp": "^3.9.1",
+    "gulp": "~4.0.0",
```
FinnStutzenstein: Can you also check in a modified `yarn.lock`. It will be generated, if you run yarn instead of npm.

gulpfile.js
jsaalfeld: You are right on the ```series``` since there is only one entry. I just moved along a tutorial for upgrading. But the official documentation says otherwise. Will dig into that.

Will also look at the watcher, I honestly didn't check it since it didn't throw any errors and the result was as expected. Will have a look at how to watch mutliple paths with the new gulp syntax and check that in as well.

jsaalfeld:
Since there was a new release, I switched branches and created a new PR at #3755

PR #3755, opened 2018-05-28
upgrade to node v10 and gulp 4

jsaalfeld:
Since Node 6 is going to be out of support in a couple of months (https://github.com/nodejs/Release#release-schedule) and node 10 only works with gulp 4, i upgraded both.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -30,8 +30,13 @@ var argv = require('yargs').argv,
     uglify = require('gulp-uglify'),
     vsprintf = require('sprintf-js').vsprintf;
 
+
 // Directory where the results go to
 var output_directory = path.join('openslides', 'static');
+
+
+// Container for all watchers
+var watchers = [];
 
 
 /**
@@ -40,11 +45,13 @@ var output_directory = path.join('openslides', 'static');
 
 // Catches all JavaScript files (excluded worker files) from all core apps and concats them to one
 // file js/openslides.js. In production mode the file is uglified.
+var js_src = [
+    path.join('openslides', '*', 'static', 'js', '**', '*.js'),
+    '!' + path.join('openslides', 'core', 'static', 'js', 'core', 'pdf-worker.js'),
+];
+
 gulp.task('js', function () {
-    return gulp.src([
-            path.join('openslides', '*', 'static', 'js', '**', '*.js'),
-            '!' + path.join('openslides', 'core', 'static', 'js', 'core', 'pdf-worker.js'),
-        ])
+    return gulp.src(js_src)
         .pipe(sourcemaps.init())
         .pipe(concat('openslides.js'))
         .pipe(sourcemaps.write())
@@ -53,9 +60,14 @@ gulp.task('js', function () {
         .pipe(gulp.dest(path.join(output_directory, 'js')));
 });
 
+watchers.push(function () {
+    gulp.watch(js_src, gulp.series('js'));
+});
+
+
 // Catches all JavaScript files from all bower components and concats them to
 // one file js/openslides-libs.js. In production mode the file is uglified.
-gulp.task('js-libs', function () {
+gulp.task('js_libs', function () {
     return gulp.src(mainBowerFiles({
             filter: /\.js$/
         }))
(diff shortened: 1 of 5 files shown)


gulpfile.js
```diff
@@ -40,25 +40,113 @@ var output_directory = path.join('openslides', 'static');
 
 // Catches all JavaScript files (excluded worker files) from all core apps and concats them to one
 // file js/openslides.js. In production mode the file is uglified.
-gulp.task('js', function () {
-    return gulp.src([
-            path.join('openslides', '*', 'static', 'js', '**', '*.js'),
-            '!' + path.join('openslides', 'core', 'static', 'js', 'core', 'pdf-worker.js'),
-        ])
+
+var paths = {
```
normanjaeckel: I would prefer not to put all path definitions in one big container but put everything together that belongs together:

Example:
```javascript
// Catches all JavaScript files (excluded worker files) from all core apps and concats them to one
// file js/openslides.js. In production mode the file is uglified.
var js_path = [path.join(...), '!' + path.join(...)];
var js_task = function () {
    return gulp.src(js_path)
        .pipe(...);
};
var js_watch = function () {
    gulp.watch(js_path, js_task);
};


...

gulp.task('default', ...);

gulp.task('watch', function () {
    js_watch();
    ...
}
```

jsaalfeld: I stuck to the examples and documentation from [https://github.com/gulpjs/gulp](the official repository). It seems that having all paths in a container is good practice and it feels very practical as well to have everything in one place.

gulpfile.js
```diff
@@ -40,25 +40,113 @@ var output_directory = path.join('openslides', 'static');
...
+          path.join('bower_components', 'ckeditor', 'lang', 'ru.js'),],
+    dest: path.join(output_directory, 'ckeditor', 'lang')
+  },
+  translations: {
+    src: path.join('openslides', 'locale', 'angular-gettext', '*.po'),
+    dest: path.join(output_directory, 'i18n')
+  }
+};
+
+function js() {
```
normanjaeckel: I also prefer function expressions before function declarations. (See some opinion [here](https://stackoverflow.com/questions/1013385/what-is-the-difference-between-a-function-expression-vs-declaration-in-javascrip#comment41375064_3344397).)

gulpfile.js
```diff
@@ -103,187 +188,132 @@ gulp.task('templates', function () {
...
         .pipe(gettext.compile({
             format: 'json'
         }))
-        .pipe(gulp.dest(path.join(output_directory, 'i18n')));
-});
+        .pipe(gulp.dest(paths.translations.dest));
+}
+
+// Combines all CKEditor related tasks.
+gulp.task('ckeditor', gulp.series(ckeditor_defaults, ckeditor_skins, ckeditor_plugins, ckeditor_lang));
```
normanjaeckel: Is it possible to run the tasks in parallel?

gulpfile.js
```diff
@@ -103,187 +188,132 @@ gulp.task('templates', function () {
...
-        'js',
-        'js-libs',
-        'pdf-worker',
-        'pdf-worker-libs',
-        'templates',
-        'css-site',
-        'css-projector',
-        'css-libs',
-        'fonts-libs',
+gulp.task('default', gulp.series(
```
normanjaeckel: Here the same. The tasks should ran in parallel.

jsaalfeld:
I won't alter the commit any further. The time that this has been in the pipeline has already caused a double-commit and some additional work (https://github.com/OpenSlides/OpenSlides/pull/3761). This is at a point where I think it fits best to the documentation and example from the gulp repository. I don't think it needs more alteration, even though I understand if someone isn't happy with functional grouping by "This is the Path" and "This are the function" and thinks a more objective approach in structure benefits the overview.

@normanjaeckel if you persist on your preferations, you are invited to add a commit on this pull request.

normanjaeckel:
I refactored the gulpfile and pushed a new commit.

@emanuelschuetze Please retest.

@jsaalfeld Ok with this? (I also reverted some probably unwanted changes, like removing extra options for CKEditor paths).

emanuelschuetze:
Successfully retested. Ok for me.

emanuelschuetze:
Why I get `Starting '<anonymous>'...` if I run `gulp watch`? Is there a gulp task name missing?

normanjaeckel:
Yes. I can fix this, if you like. There should be shown something like "watching task".

emanuelschuetze:
> Yes. I can fix this, if you like.

Yes please.

normanjaeckel:
Done.

emanuelschuetze:
Ok, successfully tested.
@jsaalfeld Ok for you? Then I'll rebase and merge.

jsaalfeld:
As long as it's finally upgraded and working, I'm fine!

PR #3767, opened 2018-06-22
make the privacy policy available for non logged in users (closes #3757)

### LINKED ISSUE #3757: Add Privacy Policy and Legal Notice Links to Login Page
While OpenSlides generally displays a footer with links to the privacy policy and other legal information, there are no such links on the login page (`/login`).  That means there is no way to obtain this information *before* logging in, and for visitors without accounts it is entirely unavailable.

According to the GDPR (and possibly other legislation) this is not acceptable.  Similarly, at least in Germany, the imprint must be public.

For these reasons, the same pages or links that are shown within OpenSlides should also be made available from its login screen.

FinnStutzenstein:
Note: If the policy is changed, the server has to be restarted, because the policy is now put as a constant in the main JS-View. This view is cached, so changes do not reflect there.

And yes, this is a very hacky, but fast solution. If we want to enable the `/privacypolicy` page, we need to refactor the whole client-state-authentication-system.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ openslides_*
 
 # Mypy cache for typechecking
 .mypy_cache
+
+# Development of a new client. Easier to switch branches with this entry
+client/*
--- a/openslides/core/apps.py
+++ b/openslides/core/apps.py
@@ -115,7 +115,13 @@ class CoreAppConfig(AppConfig):
             'name': 'OpenSlidesConfigVariables',
             'value': config_groups}
 
-        return [client_settings, config_variables]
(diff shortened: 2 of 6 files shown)


FinnStutzenstein:
@emanuelschuetze See my first node: the server has to be restarted. I messed the point that the client needs a reload, too. But if you restart both, it should work.

FinnStutzenstein:
@emanuelschuetze I'll merge this. You gave me verbal feedback that it works and I need the gitignore entry for further development ;)
You can review later.

normanjaeckel:
Ok.

PR #3769, opened 2018-06-26
Fix dockerfile to build with gulp4/node10.

emanuelschuetze:


emanuelschuetze:
@jsaalfeld Please review my changes. Note: It was required to run `COPY . /app` _before_ `yarn` command. Otherwise yarn failed because file from gulp task (e.g. pdf-worker.js) not found. With this dockerfile changes `docker build .`runs successfully for me.

jsaalfeld:
It should have run before because before the execution of ```yarn``` the files were copied (```COPY package.json yarn.lock bower.json gulpfile.js /app/```). But this way around it is much nicer.

Dockerfile
```diff
@@ -1,39 +1,33 @@
 FROM python:3.5
 RUN apt-get -y update && apt-get -y upgrade
-
 RUN apt-get install -y libpq-dev supervisor curl vim
```
jsaalfeld: we should consider removing ```vim```. Nice for debugging containers, but not useful for the runtime environemnt itself.

Dockerfile
```diff
@@ -1,39 +1,33 @@
...
+  ln -sf /tmp/node-v10.5.0-linux-x64/bin/node /usr/bin/node
 RUN mkdir /app
-COPY package.json yarn.lock bower.json gulpfile.js /app/
 WORKDIR /app
+COPY . /app
 RUN chown -R openslides /app
 USER openslides
 RUN curl -o- -L https://yarnpkg.com/install.sh | bash
 RUN $HOME/.yarn/bin/yarn --non-interactive
+RUN node_modules/.bin/gulp --production
```
jsaalfeld: Are you sure, that you should run gulp before getting the python requirements?

Dockerfile
emanuelschuetze: I prefer to keep `vim`. Sometimes it is required to do some hot fixes in running rkt containers. And this dockerfile are used for old rkt solution only.

Dockerfile
emanuelschuetze: Yes, it is independet in which order your run the setup (yarn or  pip install).

PR #3770, opened 2018-06-30
Add change id cache (not used yet)

ostcar:
This is a prove of concept of the a change id cache. Is written but not read yet.

If you use this branch and change some objects, you can get all elements that where changed with:

```python
from openslides.utils.cache import change_id_cache
change_id_cache.get_elements(0)  # 0 is the change id
change_id_cache.get_elements_full_data(0) # 0 is the change id
change_id_cache.get_elements_restricted_data(1, 0) # 1 is the user id, 0 is the change id
```


ostcar:
This is part of https://github.com/OpenSlides/OpenSlides/pull/3784 now

PR #3771, opened 2018-06-30
DS and models

FinnStutzenstein:
@tsiegleauq I try to keep everything type safe, this results in some too much code in my opinion. Please see the `TODO` in `user.ts`. Do you have an idea how to do the casts in the basemodel?
There are more TODOs to do. Please look at the structure and give feedback (But it's excatly like discussed). You can work on this, I think.


FinnStutzenstein:
Closed in favor of #3773

PR #3772, opened 2018-06-30
Custom workflows and states

FinnStutzenstein:


normanjaeckel:
@FinnStutzenstein I propose not to handle states via workflow viewset or workflow serializer. We can use a new API endpoint as we do for motion polls e. g. Please see my commit. This way you can go on. Or let's have a short phone call if there are any questions.

FinnStutzenstein:
@normanjaeckel You can do the first review.

openslides/motions/serializers.py
```diff
@@ -101,12 +101,43 @@ class WorkflowSerializer(ModelSerializer):
...
+    @transaction.atomic
+    def create(self, validated_data):
+        """
+        Customized create method. Creating a new workflow does always create a
+        new state which is used as first state.
+        """
+        workflow = super().create(validated_data)
+        first_state = State.objects.create(
+            name='new',
+            action_word='new',
```
normanjaeckel: This should be something like 'set to new' or 'set new'.

openslides/motions/serializers.py
```diff
@@ -101,12 +101,43 @@ class WorkflowSerializer(ModelSerializer):
     Serializer for motion.models.Workflow objects.
     """
     states = StateSerializer(many=True, read_only=True)
-    first_state = PrimaryKeyRelatedField(read_only=True)
+    # The first_state is checked in the update() method
+    first_state = PrimaryKeyRelatedField(queryset=State.objects.all(), required=False)
```
normanjaeckel: Do we really want the user to change the first state? I propose that the user can not change the first state but only modify it (rename etc.).

openslides/motions/views.py
```diff
@@ -893,6 +895,54 @@ def check_view_permissions(self):
             result = False
         return result
 
+    def create(self, *args, **kwargs):
+        try:
+            response = super().create(*args, **kwargs)
+        except WorkflowError as e:
```
normanjaeckel: Which WorkflowError do you expect here?

openslides/motions/views.py
```diff
@@ -893,6 +895,54 @@ def check_view_permissions(self):
...
+            raise ValidationError({'detail': e.args[0]})
+        return response
+
+
+class StateViewSet(CreateModelMixin, UpdateModelMixin, DestroyModelMixin, GenericViewSet):
+    """
+    API endpoint for workflow states.
+
+    There are the following views: update, partial_update and destroy.
```
normanjaeckel: Missing: create

openslides/motions/serializers.py
emanuelschuetze: @normanjaeckel I don't like it. I prefer to remove the action word and use always the state name! Action words are sometimes irritating for users, e.g. if a motion was accepted the use has to click on "accept" (action word) to get "accepted" (state name). But often users think alread in "accepted" if they search for the end state.
What do you think?

emanuelschuetze: issue #3776 created.

openslides/motions/serializers.py
FinnStutzenstein: I don't know. I think this should be possible. And currently it works ;)

openslides/motions/views.py
FinnStutzenstein: You currently can provide the `first_state_id` and the serializer will set it. This just can result in a WorkflowError, because if the state exists (checked by the serializer) it will be in the wrong workflow.
Dou you have an idea how to prevent this. Shall we just delete the key `first_state_id` here?

openslides/motions/views.py
FinnStutzenstein: Done.

FinnStutzenstein:
Updated the PR: Added tests for workflows. States are currently missing any tests. Added an error if one wants to create a motion without workflows. Funny thing: If you delete all workflows and restart the server, the default ones will be created.
Please check again  @emanuelschuetze .
@normanjaeckel Please check the answers to your comments above.

openslides/motions/views.py
normanjaeckel: I propose to set the first_state field in the serializer to read_only=True. Then we do not have this problems and the flow to create new workflows is clear.

FinnStutzenstein: THis just works, if we forbid to set the first state. Emanuel and I had a discussion about this and we may think, we should leave the first_state fixed, so no deletion of it and no change of the relationship. One can just change all attributes.
But we want this to be merged. I'll create another issue, because removing this functionality should be easy, but needs testing and I do not have the time today.

PR #3777, opened 2018-07-09
Update python requirements

ostcar:
I could not fix some tests. Please rewrite them or delete them.


ostcar:
There is also a new major release of channels (channels 2.0) after you have merged this, it could be a good idea also to migrate to channels 2.0. But this will probably by a bigger step.

emanuelschuetze:
@ostcar As discussed, please finish this pull request which should update django to 2.0. If all done we will merge this into master for next 2.3 release.

ostcar:
@emanuelschuetze all done. I used grep to find all places, where request.data is altered.

openslides/users/views.py
```diff
@@ -93,10 +93,15 @@ def update(self, request, *args, **kwargs):
...
             if str(request.user.pk) != self.kwargs['pk']:
                 self.permission_denied(request)
+
+            # This is a hack to make request.data mutable. Otherwise fields can not be deleted.
+            request.data._mutable = True
             # Remove fields that the user is not allowed to change.
             # The list() is required because we want to use del inside the loop.
             for key in list(request.data.keys()):
                 if key not in ('username', 'about_me'):
+                    # request.data is an immutable QueryDict. I can not see the
```
normanjaeckel: Please remove this comment here or - at your choice - add this comment also to all other places where we delete something in the request.

openslides/utils/migrations.py
```diff
@@ -20,7 +20,7 @@ def add_permission_to_groups_based_on_existing_permission(
 
     def function(apps: Any, schema_editor: Any) -> None:
         content_type = ContentType.objects.filter(model=model, app_label=app_label)
-        base_perm = Permission.objects.filter(codename=codename, content_type=content_type)
+        base_perm = Permission.objects.filter(codename=codename, content_type__in=content_type)
```
normanjaeckel: Maybe line 22 must be "get" instead of "filter"?

tests/unit/core/test_views.py
```diff
@@ -11,17 +11,6 @@ def setUp(self):
         self.viewset = views.ProjectorViewSet()
         self.viewset.format_kwarg = None
 
-    def test_activate_elements(self, mock_object):
```
normanjaeckel: Please (if there is no time ti fix the tests) only disable them and add an issue that we can fix them later.

normanjaeckel:
Some minor remarks.

openslides/users/views.py
ostcar: done

openslides/utils/migrations.py
ostcar: it has to be filter because it can be an empty list. Now it is the same as before.

tests/unit/core/test_views.py
ostcar: The problems is the mock object. I would suggest, that we no not use unit tests for views anymore.

PR #3783, opened 2018-07-14
Fixed the logo config migration (fixed #3782)

### LINKED ISSUE #3782: Migration 0008_changed_logo_fields.py failed if no logo set
Try to migrate a database (without any logo configuration) you get this python error:
```
...0008_changed_logo_fields.py", line 48, in move_old_logo_settings
    logo_pdf_L.value['path'] = logo_pdf.value.get('path', '')
TypeError: 'str' object does not support item assignment
```
@FinnStutzenstein Any hotfix?

FinnStutzenstein:
The default values might not be saved in the database yet, so the
dict-assignment fails. If there isn't a default loaded yet, get it and
set it to the database field before setting the new path-value.


FinnStutzenstein:
Closed, becuase it is fixed in #3786

Issue #2468, opened 2016-10-01
Extend autoupdate API for non-models.


normanjaeckel:
It should be possible to use the autoupdate API for fake models. See #2281.

ostcar:
I think this is fixed in the current master.

`inform_changed_elements()` (https://github.com/OpenSlides/OpenSlides/blob/e8a141387b05977a44eb8c60ef95ebff079e6936/openslides/utils/autoupdate.py#L87) accepts a "fake model" if it has the values id, collection_string and full_data.

In `handle_changed_elements()` the fake model can be saved into the cache, if it is registered as `Cachable` (https://github.com/OpenSlides/OpenSlides/blob/e8a141387b05977a44eb8c60ef95ebff079e6936/openslides/utils/cache_providers.py#L476)

So what you have to do to use a fake model is to create it with the methods `get_collection_string()`, `get_elements()` and `restrict_elements()` where the elements are a dict with at least the key `id` which is an int. Then you have to return this fake models with your AppConfig.get_startup_elements().

So I think this was fixed with #3974. If not, please reopen

PR #3784, opened 2018-07-15
[wip] Channels2

ostcar:
This branch uses channels2.

TODO:
* [ ] fix restricted_data_cache
* [ ] fix notify


openslides/users/models.py
```diff
@@ -72,6 +72,7 @@ def create_or_reset_admin_user(self):
         admin_group.permissions.add(Permission.objects.get(query_can_manage))
 
         admin, created = self.get_or_create(
+            pk=1,
```
ostcar: remove this and rewrite tests to use the username instead of the pk

ostcar:
@normanjaeckel I don't understand way isort fails on travis. It works fine on my computer. Can you check?

FinnStutzenstein:
Hi, I'm testing this branch. When I add a plugin (in my case `openslides-voting`) I get this error at the end of the migrations: https://pastebin.com/TFNaenUw
However I do not understand where the connection between the plugin and the creation of the default workflows is. This needs further investigation also with different plugins.

openslides/utils/consumers.py
```diff
@@ -0,0 +1,306 @@
...
+    async def connect(self) -> None:
+        """
+        Adds the websocket connection to a group specific to the projector with the given id.
+        Also sends all data that are shown on the projector.
+        """
+        user = self.scope['user']
+        projector_id = self.scope["url_route"]["kwargs"]["projector_id"]
+        await self.accept()
+
+        if not await database_sync_to_async(has_perm)(user, 'core.can_see_projector'):
```
FinnStutzenstein: Yes, we should keep this consistent with the site. Also sending an error code there would be great.

ostcar:
I changed the signatur of the method get_startup_elements(), https://github.com/OpenSlides/OpenSlides/pull/3784/files#diff-a5a428ba9327cc39ed3b6a6037e6d58cR48-R54

Now, the method has to return Cachables (for example RestModelMixins). Before the PR, it had to return Collection objects. This change was necessary but also helps to fix #2468.

I will fix this by making the Collection a Cachable. Then it is possible for plugins to return the Collection objects. On the long run, the plugins should be changed.

openslides/utils/consumers.py
ostcar: You mean the comment on line 136 thar we close the connection?

FinnStutzenstein: Yes

FinnStutzenstein:
Ok, I see. The "fix" is easy, but we should think about older plugins; What do you think about checking the returned type by `get_startup_elements` and if it's an collection, get the model of the collection and use it?

ostcar:
I changed the Collection to a Cachable.

I don't see the problem with older plugins. What could happen? They return a Collection and a Collection is a Cachable now.

openslides/core/config.py
```diff
@@ -37,21 +49,42 @@ def __init__(self) -> None:
...
         # Index to get the database id from a given config key
-        self.key_to_id = {}  # type: Dict[str, int]
+        self.key_to_id = None  # type: Optional[Dict[str, int]]
 
     def __getitem__(self, key: str) -> Any:
         """
         Returns the value of the config variable.
         """
-        # Build the key_to_id dict
-        self.save_default_values()
```
normanjaeckel: This method is not called here anymore. Where do you use it instead?

openslides/core/config.py
```diff
@@ -37,21 +49,42 @@ def __init__(self) -> None:
...
             self.get_collection_string(),
-            self.key_to_id[key]).get_full_data()['value']
+            self.get_key_to_id()[key]).get_full_data()['value']
+
+    def get_key_to_id(self) -> Dict[str, int]:
+        """
+        Returns the key_to_id dict. Builds it, if it does not exist.
+        """
+        if self.key_to_id is None:
+            async_to_sync(self.build_key_to_id)()
```
normanjaeckel: Maybe you have to ensure that all default values are in the database or in the cache before you can run this line.

openslides/core/signals.py
```diff
@@ -44,12 +45,12 @@ def get_permission_change_data(sender, permissions, **kwargs):
     for permission in permissions:
         if permission.content_type.app_label == core_app.label:
             if permission.codename == 'can_see_projector':
-                yield Collection(core_app.get_model('Projector').get_collection_string())
+                yield core_app.get_model('Projector')
```
normanjaeckel: Please rewrite docstring.

openslides/mediafiles/apps.py
```diff
@@ -37,4 +36,4 @@ def get_startup_elements(self):
         Yields all collections required on startup i. e. opening the websocket
         connection.
         """
-        yield Collection(self.get_model('Mediafile').get_collection_string())
+        yield self.get_model('Mediafile')
```
normanjaeckel: Please rewrite docstring.

openslides/motions/apps.py
```diff
@@ -63,5 +62,5 @@ def get_startup_elements(self):
         Yields all collections required on startup i. e. opening the websocket
         connection.
         """
-        for model in ('Category', 'Motion', 'MotionBlock', 'Workflow', 'MotionChangeRecommendation'):
-            yield Collection(self.get_model(model).get_collection_string())
+        for model_name in ('Category', 'Motion', 'MotionBlock', 'Workflow', 'MotionChangeRecommendation'):
```
normanjaeckel: Please rewrite docstring.

openslides/topics/apps.py
```diff
@@ -34,4 +33,4 @@ def get_startup_elements(self):
         Yields all collections required on startup i. e. opening the websocket
         connection.
         """
-        yield Collection(self.get_model('Topic').get_collection_string())
```
normanjaeckel: Please rewrite docstring.

openslides/users/apps.py
```diff
@@ -48,8 +47,8 @@ def get_startup_elements(self):
         Yields all collections required on startup i. e. opening the websocket
         connection.
         """
-        for model in ('User', 'Group', 'PersonalNote'):
-            yield Collection(self.get_model(model).get_collection_string())
+        for model_name in ('User', 'Group', 'PersonalNote'):
```
normanjaeckel: Please rewrite docstring.

openslides/utils/auth.py
```diff
@@ -46,6 +47,18 @@ def anonymous_is_enabled() -> bool:
     return config['general_system_enable_anonymous']
 
 
+async def a_anonymous_is_enabled() -> bool:
```
normanjaeckel: Please rename to something like `async_anonymous_is_enabled()`?

openslides/utils/auth.py
```diff
@@ -75,7 +88,10 @@ def user_to_collection_user(user: AnyUser) -> Optional[CollectionElement]:
             "Unsupported type for user. Only CollectionElements for users can be"
             "used. Not {}".format(user.collection_string))
     elif isinstance(user, int):
-        user = CollectionElement.from_values(User.get_collection_string(), user)
+        if user == 0:
```
normanjaeckel: In which case may user be 0?

openslides/utils/autoupdate.py
```diff
@@ -393,15 +41,20 @@ def inform_changed_data(instances: Union[Iterable[Model], Model], information: D
             pass
 
     # Put all collection elements into the autoupdate_bundle.
+    collection_elements = {}
```
normanjaeckel: The comment in line 43 is not correct. Can you please rewrite it? Please also extend the docstring because we do not use the AutoupdateBundleMiddleware in all cases now, do we?

openslides/utils/autoupdate.py
```diff
@@ -411,18 +64,21 @@ def inform_deleted_data(elements: Iterable[Tuple[str, int]], information: Dict[s
 
     The argument information is added to each collection element.
     """
-    # Put all stuff to be deleted into the autoupdate_bundle.
```
normanjaeckel: Here the same as in inform_changed_data().

openslides/utils/autoupdate.py
```diff
@@ -431,13 +87,16 @@ def inform_data_collection_element_list(collection_elements: List[CollectionElem
     Informs the autoupdate system about some collection elements. This is
     used just to send some data to all users.
     """
-    # Put all stuff into the autoupdate_bundle.
```
normanjaeckel: Here the same as in inform_changed_data().

openslides/utils/autoupdate.py
```diff
@@ -461,41 +120,48 @@ def __call__(self, request: Any) -> Any:
         response = self.get_response(request)
 
         bundle = autoupdate_bundle.pop(thread_id)  # type: Dict[str, CollectionElement]
-        # If currently there is an open database transaction, then the
-        # send_autoupdate function is only called, when the transaction is
-        # commited. If there is currently no transaction, then the function
-        # is called immediately.
-        transaction.on_commit(lambda: send_autoupdate(bundle.values()))
```
normanjaeckel: Please let us talk about the transaction problem via phone call so I can understand it better.

openslides/utils/utils.py
```diff
@@ -29,3 +35,43 @@ def to_roman(number: int) -> str:
...
+    if user is None:
+        user_id = 0
+    else:
+        user_id = user.id
+    return user_id
+
+
+def str_dict_to_bytes(str_dict: Dict[str, str]) -> Dict[bytes, bytes]:
+    """
+    Converts the key and the value from a dict from st to bytes.
```
normanjaeckel: Rewrite: `Converts the key and the value of a dict from str to bytes.`

openslides/utils/utils.py
```diff
@@ -29,3 +35,43 @@ def to_roman(number: int) -> str:
         return roman.toRoman(number)
     except (roman.NotIntegerError, roman.OutOfRangeError):
         return str(number)
+
+
+def get_element_id(collection_string: str, id: int) -> str:
```
normanjaeckel: This function name is a little bit missleading because it does not return the (database) id of an element (as used above in `element.id`) but a new unique OpenSlides object identifier composed of collection name and database id.

Proposal: Rename it to `get_element_uid()` for element unique identifier.

openslides/utils/autoupdate.py
```diff
@@ -461,41 +120,48 @@ def __call__(self, request: Any) -> Any:
...
-        yield from get_startup_elements()
+        if not getattr(settings, 'SKIP_CACHE', False):
+            # Hack for django 2.0 and channels 2.1 to stay in the same thread.
+            # This is needed for the tests.
+            change_id = await element_cache.change_elements(cache_elements)
+        else:
+            change_id = 1
+
+        channel_layer = get_channel_layer()
+        # TODO: don't await. They can be send in parallel
```
normanjaeckel: Maybe you can use https://docs.python.org/3/library/asyncio-task.html#asyncio.gather here. See https://docs.python.org/3/library/asyncio-task.html#example-parallel-execution-of-tasks

openslides/utils/consumers.py
```diff
@@ -0,0 +1,306 @@
...
+
+        If it is an anonymous user and anonymous is disabled, the connection is closed.
+
+        Sends the startup data to the user.
+        """
+        # TODO: add a way to ask for the data since a change_id and send only data that is newer
+        if not await a_anonymous_is_enabled() and self.scope['user'].id is None:
+            await self.close()
+        else:
+            await self.accept()
```
normanjaeckel: In channels docs it "is recommended that accept() be called as the last action in connect() if you choose to accept the connection." See https://channels.readthedocs.io/en/latest/tutorial/part_2.html. Is this possible for us?

openslides/utils/consumers.py
```diff
@@ -0,0 +1,306 @@
...
+        """
+        If we recieve something from the client we currently just interpret this
+        as a notify message.
+
+        The server adds the sender's user id (0 for anonymous) and reply
+        channel name so that a receiver client may reply to the sender or to all
+        sender's instances.
+        """
+        if notify_message_is_valid(content):
+            await self.channel_layer.group_send(
```
normanjaeckel: Can you send in parallel?

openslides/utils/consumers.py
normanjaeckel: I dont know.

openslides/utils/models.py
```diff
@@ -22,7 +27,7 @@ def formfield(self, **kwargs: Any) -> Any:
         return super(MinMaxIntegerField, self).formfield(**defaults)
 
 
-class RESTModelMixin:
+class RESTModelMixin():
```
normanjaeckel: Why do we need the brackets?

openslides/agenda/apps.py
```diff
@@ -51,4 +50,4 @@ def get_startup_elements(self):
         Yields all collections required on startup i. e. opening the websocket
         connection.
         """
-        yield Collection(self.get_model('Item').get_collection_string())
```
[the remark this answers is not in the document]
ostcar: Done (also in other apps)

openslides/assignments/apps.py
```diff
@@ -44,7 +43,7 @@ def get_startup_elements(self):
         Yields all collections required on startup i. e. opening the websocket
         connection.
         """
-        yield Collection(self.get_model('Assignment').get_collection_string())
+        yield self.get_model('Assignment')
```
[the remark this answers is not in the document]
ostcar: done

openslides/core/apps.py
```diff
@@ -69,8 +68,8 @@ def get_startup_elements(self):
         Yields all collections required on startup i. e. opening the websocket
         connection.
         """
-        for model in ('Projector', 'ChatMessage', 'Tag', 'ProjectorMessage', 'Countdown', 'ConfigStore'):
-            yield Collection(self.get_model(model).get_collection_string())
+        for model_name in ('Projector', 'ChatMessage', 'Tag', 'ProjectorMessage', 'Countdown', 'ConfigStore'):
```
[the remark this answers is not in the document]
ostcar: done

openslides/core/config.py
ostcar: It is called after a migration, see: https://github.com/OpenSlides/OpenSlides/blob/32e20aa6708b14f42fbcfcaa1e0792cb20af514b/openslides/core/apps.py#L57 https://github.com/OpenSlides/OpenSlides/blob/32e20aa6708b14f42fbcfcaa1e0792cb20af514b/openslides/core/apps.py#L127

This line was necessary before, because it made sure, that the dict key_to_id was set: https://github.com/OpenSlides/OpenSlides/blob/32e20aa6708b14f42fbcfcaa1e0792cb20af514b/openslides/core/config.py#L189

Now, get_key_to_id() is used, what does the same thing but uses the cache instead of the database.

openslides/core/config.py
ostcar: They are inserted in the database after each migration. If the elements are not in the cache, the cache is build with element_cache.get_all_full_data()

openslides/core/signals.py
ostcar: done

openslides/mediafiles/apps.py
ostcar: done

openslides/motions/apps.py
ostcar: done

openslides/topics/apps.py
ostcar: done

openslides/users/apps.py
ostcar: done

openslides/utils/auth.py
ostcar: ok. I thought the prefix a was common for an async method in python, see https://www.python.org/dev/peps/pep-0525/#aiter-and-anext-builtins 

But I changed it now.

openslides/utils/auth.py
ostcar: In some cases, we use 0 for anonymous user. See for example https://github.com/OpenSlides/OpenSlides/blob/32e20aa6708b14f42fbcfcaa1e0792cb20af514b/openslides/utils/autoupdate.py#L72 in the current master. I don't think, that there is any case in this branch, where this line is needed. If you like, I can remove it.

openslides/utils/autoupdate.py
ostcar: done

openslides/utils/autoupdate.py
ostcar: done

openslides/utils/autoupdate.py
ostcar: done

openslides/utils/autoupdate.py
ostcar: Sure. See also: https://github.com/django/channels/issues/1110

openslides/utils/utils.py
ostcar: done

openslides/utils/utils.py
ostcar: I see what you mean. I don't like uid. When I here it, I think of an user id. See also https://de.wikipedia.org/wiki/UID

I use the term 'element_id' at a lot of places. See the file cache.py. I thought element_id is ok, because we did not use this term in the code before.

openslides/utils/autoupdate.py
ostcar: I will add this performance improvement in another pull request. I am not sure if this can be used here, because django channels creates a lot of different event-loops with async_to_sync. This needs more testing

openslides/utils/consumers.py
ostcar: No. The connection has to be accepted, before data can be send. You get ```Socket has not been accepted, so cannot send over it``` when you change the order of the lines.

openslides/utils/consumers.py
ostcar: Maybe. But this is not important, because I want to delete the projector consumer anyway.

openslides/utils/models.py
ostcar: Typo. 

I tried to make the RESTModelMixin class(!) a Cachable be adding Cachable as a metaclass. This did not work but I did forget to remove the brackets.

FinnStutzenstein:
Note to change the merge target to the OpenSlides3 branch.

ostcar:
I opened another PR for the OpenSlides-3 branch #3796 

I close this one

PR #3785, opened 2018-07-16
Do not allow changing a workflow's first state (closes #3778)

### LINKED ISSUE #3778: Remove the customizability of the workflow's first state
After merge of #3772 : remove the dropdown for the first state. Set `first_state` to read only in the serializer. Update tests and do not catch the `WorkflowError` in the create view.

FinnStutzenstein:


openslides/motions/serializers.py
```diff
@@ -102,7 +102,7 @@ class WorkflowSerializer(ModelSerializer):
     """
     states = StateSerializer(many=True, read_only=True)
     # The first_state is checked in the update() method
-    first_state = PrimaryKeyRelatedField(queryset=State.objects.all(), required=False)
+    first_state = PrimaryKeyRelatedField(read_only=True)
```
normanjaeckel: We also need some changes in the update method. Maybe our custom update method can be removed and we can also remove this line and set read_only via [Meta option](http://www.django-rest-framework.org/api-guide/serializers/#specifying-read-only-fields).

emanuelschuetze:
@normanjaeckel As discussed, please review again.

emanuelschuetze:
Rebased.
@FinnStutzenstein Please review 2nd commit. Then I'll rebase and merge.

FinnStutzenstein:
@emanuelschuetze Ok.

PR #3786, opened 2018-08-08
support python 3.7

ostcar:
also fix new pep8 stype
also fix an issue with migrations introduced with #3772


ostcar:
I merge this. There is nothing harmful in here. But i can not continue coding with archlinux on python 3.7 without it.

.travis.yml
```diff
@@ -1,11 +1,13 @@
 language: python
-sudo: false
+dist: xenial
+sudo: true
```
normanjaeckel: Why is this required?

ostcar: See https://github.com/travis-ci/travis-ci/issues/9815

I hope, we can remove this change soon.

PR #3789, opened 2018-08-15
Use newer version of django-redis-sessions, new settings format

FinnStutzenstein:
I locked the version of django-redis-sessions to 1.6.1, because the settings format has changed and may depend on what version pip will install. This was an issue on our production servers. 

https://github.com/martinrusev/django-redis-sessions/blob/master/CHANGELOG.md

I also added the configuration for the session with the new format. The settings will be inconsistend, if the user only changes the `CACHES` section.

### DIFF
--- a/openslides/utils/settings.py.tpl
+++ b/openslides/utils/settings.py.tpl
@@ -79,6 +79,15 @@ DATABASES = {
 use_redis = False
 
 if use_redis:
+    # Redis configuration for django-redis-session. Keep this synchronized to
+    # the caching settings
+
+    SESSION_REDIS = {
+        'host': '127.0.0.1',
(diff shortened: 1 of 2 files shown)


openslides/utils/settings.py.tpl
```diff
@@ -79,6 +79,15 @@ DATABASES = {
 use_redis = False
 
 if use_redis:
+    # Redis configuration for redis-session. Keep this synchronized to the
```
emanuelschuetze: typo: 'django-redis-sessions' instead of 'redis-session'

FinnStutzenstein: Done

PR #3790, opened 2018-08-15
New item type internal.

FinnStutzenstein:
The old hidden type was used as internal, so everything is changed to
not be shown if the item is internal. hidden is "new", and actually
behaves as hidden now.


openslides/agenda/config_variables.py
```diff
@@ -59,6 +59,19 @@ def get_config_variables():
...
 
+    yield ConfigVariable(
+        name='agenda_new_items_default_visibility',
+        default_value='2',
+        input_type='choice',
+        choices=(
+            {'value': '1', 'display_name': 'Public item'},
+            {'value': '2', 'display_name': 'Internal item'},
+            {'value': '3', 'display_name': 'Hidden item'}),
+        label='Default visibility for new agenda items',
```
emanuelschuetze: Please add this string to site.js

FinnStutzenstein:
@emanuelschuetze I rebased the PR.

normanjaeckel:
@emanuelschuetze @FinnStutzenstein Please let's discuss this new feature in the following point: The default for new topics should be public. At the moment the default is caught from the config and the config default is internal.

FinnStutzenstein:
Lets discuss this tomorrow -- we had a longer discussion about that, too ;)

PR #3792, opened 2018-08-16
Layout changes, config for enabling amendments in the motions table

FinnStutzenstein:


emanuelschuetze:

normanjaeckel:
Ok (only one full stop in CHANGELOG is missing but I don't make any more noise here).

PR #3797, opened 2018-08-20
Fixed settings creation for windows.

normanjaeckel:
@emanuelschuetze Please check this. I got an AttributeError on current master without this patch after delting personal_data dir on Windows.


normanjaeckel:
@FinnStutzenstein Can you please check this commit (https://github.com/OpenSlides/OpenSlides/commit/10038b782f979df64dcc11553adc03f4901204e8).

ostcar:
The function seems a bit complicated. Maybe it can be rewritten with somethink like this: https://stackoverflow.com/questions/13184414/how-can-i-get-the-path-to-the-appdata-directory-in-python

PR #3798, opened 2018-08-22
New url schema

ostcar:
The goal is, to easier root requests from nginx. Currently, all urls (also from plugins) have to be set in nginx (see nginx-config in DEVELOPMENT.rst). With this pull request, all individual urls (also from plugins) have the prefix `views/`. This makes it a lot easier to configure nginx.

The prefix `views/` is open to discussion.

Set `ROOT_URLCONF = 'openslides.old_urls'` in your personal settings, to use the old url schema (needed if you want to use the old client).

When the new client is ready, the old_urls-file has to be removed.

@FinnStutzenstein you could change the url-files to that the urls.py uses the new client and the old_urls.py uses the old client automatically

### DIFF
--- a/openslides/old_urls.py
+++ b/openslides/old_urls.py
@@ -3,28 +3,22 @@ from django.conf.urls import include, url
 from django.contrib.staticfiles.urls import urlpatterns
 from django.views.generic import RedirectView
 
+from openslides.core import views as core_views
 from openslides.mediafiles.views import protected_serve
+from openslides.utils.plugins import get_all_plugin_urlpatterns
 from openslides.utils.rest_api import router
 
-from .core import views as core_views
 
-
-# Urls for /static/ are already in urlpatterns
+urlpatterns += get_all_plugin_urlpatterns()
 
 urlpatterns += [
-    # URLs for /media/
     url(r'^%s(?P<path>.*)$' % settings.MEDIA_URL.lstrip('/'), protected_serve, {'document_root': settings.MEDIA_ROOT}),
-
-    # When a url without a leading slash is requested, redirect to the url with
-    # the slash. This line has to be after static and media files.
     url(r'^(?P<url>.*[^/])$', RedirectView.as_view(url='/%(url)s/', permanent=True)),
-
-    # URLs for the rest system
     url(r'^rest/', include(router.urls)),
-
-    # Other urls defined by modules and plugins
-    url(r'^apps/', include('openslides.urls_apps')),
-
+    url(r'^agenda/', include('openslides.agenda.urls')),
+    url(r'^motions/', include('openslides.motions.urls')),
+    url(r'^users/', include('openslides.users.urls')),
+    url(r'^core/', include('openslides.core.urls')),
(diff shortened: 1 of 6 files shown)


openslides/old_urls.py
```diff
@@ -0,0 +1,24 @@
...
+    url(r'^%s(?P<path>.*)$' % settings.MEDIA_URL.lstrip('/'), protected_serve, {'document_root': settings.MEDIA_ROOT}),
+    url(r'^(?P<url>.*[^/])$', RedirectView.as_view(url='/%(url)s/', permanent=True)),
+    url(r'^rest/', include(router.urls)),
+    url(r'^agenda/', include('openslides.agenda.urls')),
+    url(r'^motions/', include('openslides.motions.urls')),
+    url(r'^users/', include('openslides.users.urls')),
+
+    # The urls.py of the core app has to be the last entry. It contains the
+    # main entry points for OpenSlides' browser clients.
+    url(r'^', include('openslides.core.urls')),
```
normanjaeckel: THis won't work because you remove the string 'core' for some views in the core/urls.

openslides/urls.py
```diff
@@ -1,23 +1,43 @@
...
     url(r'^rest/', include(router.urls)),
-    url(r'^agenda/', include('openslides.agenda.urls')),
-    url(r'^motions/', include('openslides.motions.urls')),
-    url(r'^users/', include('openslides.users.urls')),
 
-    # The urls.py of the core app has to be the last entry. It contains the
-    # main entry points for OpenSlides' browser clients.
-    url(r'^', include('openslides.core.urls')),
+    # Other urls defined by modules and plugins
+    url(r'^views/', include('openslides.urls_views')),
```
normanjaeckel: Proposal: Use apps instead of views.

openslides/urls_views.py
```diff
@@ -0,0 +1,13 @@
+from django.conf.urls import include, url
```
normanjaeckel: Also rename this file to urls_apps.py

openslides/urls_views.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+urlpatterns = [
+    url(r'^core/', include('openslides.core.urls')),
+    url(r'^agenda/', include('openslides.agenda.urls')),
+    url(r'^motions/', include('openslides.motions.urls')),
+    url(r'^users/', include('openslides.users.urls')),
+]
+
+urlpatterns += get_all_plugin_urlpatterns()
```
normanjaeckel: I think the plugins should live before the original apps.

openslides/old_urls.py
ostcar: This file is old. See my initial comment of the pr

ostcar:
@normanjaeckel done

openslides/urls_views.py
normanjaeckel: Ping.

normanjaeckel:
Please add CHANGELOG entry.

PR #3799, opened 2018-08-22
Updates files for big mode

ostcar:
Channels2 adds the possibility to use gunicorn with uvicorn instead of daphne. This helps to start and manage many server processes at the same time at the same port. It is also **a lot** faster.

This is a perfect alternative for geiss and it can be installed via pip.

### DIFF
--- a/DEVELOPMENT.rst
+++ b/DEVELOPMENT.rst
@@ -25,9 +25,6 @@ For Ubuntu 16.04 e. g. follow `Yarn installation instructions
 
     $ sudo apt-get install git nodejs nodejs-legacy npm build-essential python3-dev
 
-*Note: For Ubuntu 14.04 you have to update Node.js before. The distribution
-version is 0.10.25 which is not sufficient.*
-
 
 b. Get OpenSlides source code
 '''''''''''''''''''''''''''''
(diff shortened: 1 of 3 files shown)


DEVELOPMENT.rst
```diff
@@ -166,15 +166,8 @@ Then add database user and database. For Ubuntu 16.04 e. g. run::
     $ sudo -u postgres createdb --owner=openslides openslides
 
 
-2. Install additional packages
```
FinnStutzenstein: This should be at least somewhere mentioned.

requirements_big_mode.txt
```diff
@@ -6,3 +6,7 @@ channels-redis>=2.2,<2.3
 django-redis-sessions>=0.6.1,<0.7
 psycopg2-binary>=2.7.3.2,<2.8
 aioredis>=1.1.0,<1.2
+
+# Requirements for fast asgi server
+gunicorn>=19.9<19.10
```
jsaalfeld: Wouldnt ```<20``` be sufficient and more secure?

DEVELOPMENT.rst
ostcar: See requirements.txt and line 54

requirements_big_mode.txt
ostcar: For me, <20 is ok. @normanjaeckel ?

DEVELOPMENT.rst
```diff
@@ -190,17 +183,20 @@ Populate your new database::
...
 
-To start Daphne as protocol server run::
+To start gunicorn with uvicorn as protocol server run::
 
     $ export DJANGO_SETTINGS_MODULE=settings
     $ export PYTHONPATH=personal_data/var/
-    $ daphne openslides.asgi:channel_layer
+    $ gunicorn -w 4 -k uvicorn.workers.UvicornWorker openslides.asgi:application
 
+This example uses 4 instances. The recommendation is to use CPU cors * 2.
```
normanjaeckel: typo cores

PR #3803, opened 2018-08-22
Support for decimal places in motion and assignment polls

FinnStutzenstein:


FinnStutzenstein:
@normanjaeckel Done and rebased.

PR #3805, opened 2018-08-22
drop python 3.5

ostcar:


normanjaeckel:
Please also add CHANGELOG entry.

PR #3828, opened 2018-08-25
Faster then fastes big_mode (with tests)

ostcar:
fixes #3825


ostcar:
@jsaalfeld Please review the Dockerfile change and maybe also change the files in the openslides-docker repo

PR #3842, opened 2018-08-31
Remove motion version and comments rework

FinnStutzenstein:


openslides/motions/access_permissions.py
```diff
@@ -123,6 +101,40 @@ def get_serializer_class(self, user=None):
...
+        from .serializers import MotionCommentFieldSerializer
+
+        return MotionCommentFieldSerializer
+
+    def get_restricted_data(
+            self,
+            full_data: List[Dict[str, Any]],
+            user: Optional[CollectionElement]) -> List[Dict[str, Any]]:
+        """
+        Returns the restricted serialized data for the instance prepared for
```
normanjaeckel: Please update docstring.

openslides/motions/models.py
```diff
@@ -93,6 +98,7 @@ class Motion(RESTModelMixin, models.Model):
 
     This attribute is to get the current state of the motion.
     """
+    state_extension = models.TextField(blank=True, null=True)
```
normanjaeckel: Please add newline and docstring.

openslides/motions/models.py
```diff
@@ -102,6 +108,7 @@ class Motion(RESTModelMixin, models.Model):
     """
     The recommendation of a person or committee for this motion.
     """
+    recommendation_extension = models.TextField(blank=True, null=True)
```
normanjaeckel: Please add newline and docstring.

openslides/motions/models.py
```diff
@@ -777,6 +507,42 @@ def get_paragraph_based_amendments(self):
         return list(filter(lambda amend: amend.is_paragraph_based_amendment(), self.amendments.all()))
 
 
+class MotionCommentField(RESTModelMixin, models.Model):
```
normanjaeckel: Please do not use "Field" in the name of the class because it can be mixed up with a Django Field (like TextField etc.). Proposal: MotionCommentSection

openslides/motions/models.py
```diff
@@ -777,6 +507,42 @@ def get_paragraph_based_amendments(self):
...
+    access_permissions = MotionCommentFieldAccessPermissions()
```
normanjaeckel: Please add a class docstring.

openslides/motions/models.py
```diff
@@ -777,6 +507,42 @@ def get_paragraph_based_amendments(self):
...
+    read_groups = models.ManyToManyField(Group, blank=True, related_name='read_comments')
+
+    write_groups = models.ManyToManyField(Group, blank=True, related_name='write_comments')
+
+    class Meta:
+        default_permissions = ()
+
+
+class MotionComment(RESTModelMixin, models.Model):
+    comment = models.TextField()
```
normanjaeckel: Please add a class docstring.

openslides/motions/models.py
```diff
@@ -777,6 +507,42 @@ def get_paragraph_based_amendments(self):
         return list(filter(lambda amend: amend.is_paragraph_based_amendment(), self.amendments.all()))
 
 
+class MotionCommentField(RESTModelMixin, models.Model):
+    access_permissions = MotionCommentFieldAccessPermissions()
+
+    name = models.CharField(max_length=255)
+
+    read_groups = models.ManyToManyField(Group, blank=True, related_name='read_comments')
```
normanjaeckel: Is it possible to avoid an import of "Group"? See how we refer to user in motion submitter and motion supporter.

Try `settings.AUTH_USER_MODEL.groups.field.related_model`

openslides/motions/serializers.py
```diff
@@ -337,6 +290,37 @@ def validate(self, data):
...
+    class Meta:
+        model = MotionComment
+        fields = (
+            'id',
+            'comment',
+            'field',
+            'read_groups_id',)
+
+    def get_read_groups_id(self, comment):
+        return [group.id for group in comment.field.read_groups.all()]
```
normanjaeckel: This causes extra db queries so we need a special model manager (and some query tests).

openslides/motions/views.py
```diff
@@ -247,119 +233,82 @@ def update(self, request, *args, **kwargs):
...
-        """
-        Special view endpoint to permit and delete a version of a motion.
-
-        Send PUT {'version_number': <number>} to permit and DELETE
-        {'version_number': <number>} to delete a version. Deleting the
-        active version is not allowed. Only managers can use this view.
-        """
-        # Retrieve motion and version.
+    @detail_route(methods=['POST', 'DELETE'])
+    def manage_comments(self, request, pk=None):
```
normanjaeckel: Please add a docstring which describes also the HTTP methods.

openslides/motions/views.py
```diff
@@ -700,6 +645,30 @@ def create(self, request, *args, **kwargs):
...
+    queryset = MotionCommentField.objects.all()
+
+    def check_view_permissions(self):
+        """
+        Returns True if the user has required permissions.
+        """
+        print(self.action)
+        if self.action in ('list', 'retrieve'):
+            result = self.get_access_permissions().check_permissions(self.request.user)
+        elif self.action in ('create', 'destroy', 'update'):
```
normanjaeckel: partial_update is missing

openslides/motions/views.py
```diff
@@ -700,6 +645,30 @@ def create(self, request, *args, **kwargs):
...
+    There are the following views: update, partial_update and destroy.
+    """
+    access_permissions = MotionCommentFieldAccessPermissions()
+    queryset = MotionCommentField.objects.all()
+
+    def check_view_permissions(self):
+        """
+        Returns True if the user has required permissions.
+        """
+        print(self.action)
```
normanjaeckel: Please remove this line.

openslides/motions/views.py
normanjaeckel: What is about the delete case? Deleting a comment section that is used in any comment is not possible (see on_delete=PROTECT) but we do not want the user to get a HTTP 500 but a HTTP 400.

openslides/utils/auth.py
```diff
@@ -39,6 +39,41 @@ def has_perm(user: Optional[CollectionElement], perm: str) -> bool:
...
+def in_some_groups(user: Optional[CollectionElement], groups: Union[List[int], List[Group]]) -> bool:
+    """
+    Checks that user is in at least one given group. Groups can be given as a list
+    of ids or group instances.
+
+    User can be a CollectionElement of a user or None.
+    """
+
+    if len(groups) > 0:
+        if isinstance(groups[0], Group):
```
normanjaeckel: Here the same. Please try to use `get_user_model().groups.field.related_model`

normanjaeckel: This looks like ducktyping: What if the first element is a group but the second not ... ?

openslides/utils/auth.py
```diff
@@ -39,6 +39,41 @@ def has_perm(user: Optional[CollectionElement], perm: str) -> bool:
...
+        return False  # early end here, if no groups are given.
+
+    # Convert user to right type
+    # TODO: Remove this and make use, that user has always the right type
+    user = user_to_collection_user(user)
+    if user is None and not anonymous_is_enabled():
+        in_some_groups = False
+    elif user is None:
+        # Use the permissions from the default group with id 1.
+        in_some_groups = 1 in groups
```
normanjaeckel: Hm. This looks dangerous because in future we may change the meaning of Group pk 1 (superadmin group). Proposal: Let's introduce a upper case variable for this.

openslides/utils/models.py
```diff
@@ -49,6 +49,7 @@ def get_access_permissions(cls) -> BaseAccessPermissions:
         its corresponding viewset.
         """
         if cls.access_permissions is None:
+            print(cls.get_collection_string())
```
normanjaeckel: Please remove this line.

openslides/motions/models.py
FinnStutzenstein: The type of `settings.AUTH_USER_MODEL` is str, so no, this won't work.
Another possibility is `get_user_model().groups.field.related_model`, but then I get this error:
`django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.`
This is reasonable too, because `get_user_model()` just works, if the application is running.

So I'll stick to `Groups` until someone finds another solution

openslides/utils/auth.py
FinnStutzenstein: This is the thing with weak typed languages as python. I expect the user go give a valid input described by the `groups` parameter with this type: `Union[List[int], List[Group]]`. So either a list of integers or a list of groups.

openslides/motions/serializers.py
FinnStutzenstein: Added a comment manager. Also updated `test_motion_db_queries`. Works fine.

openslides/utils/auth.py
FinnStutzenstein: Where should i create it? in `auth.py` or somewhere in the users?

FinnStutzenstein:
@normanjaeckel I do not know, why all old tests are errored. On my machine it works fine. Can you check, if you have those error, too?
It's strange, because on the constants I didn't change anything. Just removed two config variables.

emanuelschuetze:
`$ pytest test/old/` works for me without errors!

emanuelschuetze:
But `$ pytest --cov --cov-fail-under=75` shows 33 errors - like travis!

FinnStutzenstein:
Narrowed it down: The combination of first integration/util tests and then old tests also doesn't work:

`pytest tests/integration/utils/ tests/old ` (`--cov --cov-fail-under=20` does not change anything)

emanuelschuetze:
both commands failed:
`$ pytest tests/integration/utils/ tests/old/`
`$ pytest tests/integration/utils/test_collection.py  tests/integration/utils/test_consumers.py tests/old/`

This works!
`$ pytest tests/integration/utils/test_consumers.py  tests/integration/utils/test_collection.py tests/old/`

FinnStutzenstein:
I do have a feeling: Something is cached across the tests.. This should be fixed, but not here. Let's create an issue, if this is mreged.

@normanjaeckel 3 things left (see in the comments above):
- Proposal: Let's introduce a upper case variable for this.
- ducktyping
- `settings.AUTH_USER_MODEL.groups.field.related_model`

openslides/utils/auth.py
normanjaeckel: I am not sure. Maybe somewhere in users.

openslides/motions/models.py
normanjaeckel: I think the correct solution is to add a new settings variable AUTH_GROUP_MODEL which is a string `'user.Group'` or something like this.

openslides/motions/serializers.py
```diff
@@ -337,6 +293,47 @@ def validate(self, data):
...
 
 
+class MotionCommentSectionSerializer(ModelSerializer):
+    """
+    Serializer for motion.models.MotionCommentSection objects.
+    """
+    read_groups = IdPrimaryKeyRelatedField(
+        many=True,
+        required=False,
+        queryset=Group.objects.all())
```
normanjaeckel: Here you can use get_user_model().groups...

openslides/utils/auth.py
FinnStutzenstein: removed this now.

openslides/utils/auth.py
FinnStutzenstein: According to #3859 I'll add it to auth.

openslides/motions/serializers.py
FinnStutzenstein: the type of `get_user_model().groups.field.realted_model` is `django.contrib.auth.Group`. I don't know what's wrong there and I won't search deeper. I just created a `get_group_model` in out `auth.py` that works exactly like the django's one for users.

tests/integration/motions/test_viewset.py
```diff
@@ -617,8 +617,8 @@ def setUp(self):
         self.client.login(username='admin', password='admin')
 
         self.admin = get_user_model().objects.get()
-        self.group_in = self.admin.groups.first()
```
FinnStutzenstein: @normanjaeckel Just FYI: `self.admin.groups.first()` has the type `django.contrib.auth.models.Group` and not our own group model. This seems to be broken..
Now, you cannot assign groups from users to comments..

FinnStutzenstein:
Can be merged after oral OK from @normanjaeckel

PR #3843, opened 2018-09-01
ensures test on startup

ostcar:
This pull request is necessary for the history mode.

It loads the cache on startup. It is not allowed to clear the cache when OpenSlides is running. You have to stop end restart openslides, when you want to flushall redis (the same, when you want to clear the sql-database)

### DIFF
--- a/openslides/core/apps.py
+++ b/openslides/core/apps.py
@@ -21,6 +21,7 @@ class CoreAppConfig(AppConfig):
         # Import all required stuff.
         from .config import config
         from ..utils.rest_api import router
+        from ..utils.cache import element_cache
         from .projector import get_projector_elements
         from .signals import (
             delete_django_app_permissions,
@@ -73,6 +74,13 @@ class CoreAppConfig(AppConfig):
         router.register(self.get_model('ConfigStore').get_collection_string(), ConfigViewSet, 'config')
         router.register(self.get_model('ProjectorMessage').get_collection_string(), ProjectorMessageViewSet)
         router.register(self.get_model('Countdown').get_collection_string(), CountdownViewSet)
+
+        # Sets the cache
+        try:
+            element_cache.ensure_cache()
+        except (ImproperlyConfigured, OperationalError):
+            # This happens in the tests or in migrations. Do nothing
+            pass
 
     def get_config_variables(self):
         from .config_variables import get_config_variables
--- a/openslides/utils/cache_providers.py
+++ b/openslides/utils/cache_providers.py
@@ -52,7 +52,7 @@ class BaseCacheProvider:
     def get_change_id_cache_key(self) -> str:
         return self.change_id_cache_key
 
-    def clear_cache(self) -> None:
+    async def clear_cache(self) -> None:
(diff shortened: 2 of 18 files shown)


openslides/core/apps.py
```diff
@@ -63,6 +66,13 @@ def ready(self):
         router.register(self.get_model('ProjectorMessage').get_collection_string(), ProjectorMessageViewSet)
         router.register(self.get_model('Countdown').get_collection_string(), CountdownViewSet)
 
+        # Sets the cache
+        try:
+            element_cache.ensure_cache()
+        except (ImproperlyConfigured, OperationalError):
```
ostcar: @normanjaeckel I don't like this but I don't know how this can be fixed. The OperationalError happens when migrations are run. In migrations, this code is run before the database is created. Therefore the database tables do not exist and the cache can not be created.

The ImproperlyConfigured error occurs in tests, where the database is deactivated. See: https://pytest-django.readthedocs.io/en/latest/database.html#enabling-database-access-in-tests

This works, but if an ImproperlyConfigured or OperationError happens for another reason, the cache will not be created and the error will be silenced.

ostcar:
This also fixes a problem with python manage create when redis is empty.

It is now also possible, to run the tests with the redis cache. You have to put `REDIS_ADDRESS = "redis://127.0.0.1"` into the tests/settings.py. But this uses the "normal" redis connection. Therefore all data inside your redis database will be deleted.

@normanjaeckel what do you think. Should we use the redis cache per default in travis? This would result in a higher test coverage and more realistic tests

openslides/core/apps.py
normanjaeckel: You may raise a customized [warning](https://docs.python.org/3.7/library/warnings.html).

openslides/utils/cache_providers.py
```diff
@@ -113,22 +130,28 @@ class RedisCacheProvider(BaseCacheProvider):
...
         """
-        Returns a redis connection.
+        Deleted all cache entries created with this element cache.
         """
-        if self.redis_pool is None:
-            self.redis_pool = await aioredis.create_redis_pool(self.redis_address)
-        return self.redis_pool
+        async with self.get_connection() as redis:
+            # TODO: Fix me. Do only delete keys, that are created with this cache.
+            await redis.flushall()
```
normanjaeckel: Ah, this should be fixed as soon as possible because developers may flush their redis instance accidentally. Please open an urgent issue for this.

normanjaeckel:
I think we should not use the standard redis connection/address for the tests that one may run via command line. It would be ok for me that Travis itself uses redis so the test should check a proper environment variable.

openslides/utils/cache_providers.py
ostcar: It is not possible to raise a warning because it would be printed many times in the tests and also in migration.

ostcar: I disabled the test with redis code. So currently the problematic function is not used anywhere. I will fix this in another pr

normanjaeckel: Ok.

PR #3859, opened 2018-09-07
Added new admin group which grants all permissions. All existing grou…

normanjaeckel:
…ps are reset during migration.

@emanuelschuetze The migration is very hard. Please have a look at my quick and dirty solution (delete and reset all groups and permissions, reset admin user with new group and password).


FinnStutzenstein:
Notes from telephone call: Replace delegate group (PK2): create new group, name it 'Delegate2' and copy all users and permission from old delegate group into the new one. Make the old delegate group the new superadmin group and add all members from admin group (PK4), if exists. Rename 'Delegate2'->'Delegate'.

Please also find a way to not have the constants with the group ids at different places. Maybe also available in the tests.

FinnStutzenstein:
@normanjaeckel Please see #3910 . This can be closed I think

normanjaeckel:
Ok.

PR #3889, opened 2018-09-20
Fix image plugin

FinnStutzenstein:


emanuelschuetze:
This is a first bug fix of 2.3 - could be used for a 2.3.1.

PR #3892, opened 2018-09-23
update the dev tools

ostcar:
mypy and pytest works fine in the current version.

The only problem is, that pytest 3.8 shows deprication warnings by default (pytest 3.6 hide them).

We don't have any deprication warnings, but jsonfield2 has some.

I don't know, if jsonfield2 will be fixed, but we could add the code from jsonfield2 into our repo (it is not so much code). When we switch to mongodb in a future version, it is not needed anymore. What do you think?


ostcar:
See https://github.com/rpkilby/jsonfield2/pull/5

PR #3894, opened 2018-09-24
New motion features

FinnStutzenstein:
- Added weight and sort_parent fields to the motion model
- Added motion sort view (adapted from agenda)
- Added statute-paragraph model and tests for it
- Added statute_paragraph foreign key to the motion model
- Created migrations for sorting and statute-paragraph


openslides/motions/access_permissions.py
```diff
@@ -138,6 +138,25 @@ def get_restricted_data(
         return data
 
 
+class ConstitutionParagraphAccessPermissions(BaseAccessPermissions):
```
normanjaeckel: I prefer `statute` for German "Satzung" (https://dict.leo.org/englisch-deutsch/statute)

openslides/motions/models.py
```diff
@@ -35,6 +36,26 @@
...
+
+    title = models.CharField(max_length=255)
+    """Title of the constitution paragraph."""
+
+    text = models.TextField()
+    """Content of the constitution paragraph."""
+
+    class Meta:
+        default_permissions = ()
+        ordering = ['title']
```
normanjaeckel: It is not self-evident that we can order this by title. In some cases we may have an alpha-numeric title prefix but there may also other cases. So I think we need a weight field too.

openslides/motions/views.py
```diff
@@ -256,6 +259,34 @@ def update(self, request, *args, **kwargs):
 
         return Response(serializer.data)
 
+    @list_route(methods=['post'])
+    def sort(self, request):
+        """
+        Sort motions. Also checks sort_parent field to prevent hierarchical loops.
```
normanjaeckel: Please extend docstring: Describe what data should be send (something with nodes) so that the sorting works properly.

FinnStutzenstein: I'm honest: I do not know this excatly, too. This is taken 1:1 from the agenda. I'll have a closer look at this code, if the client implements sorting of agenda or motions. Maybe we decide to pass different options, accumulated data, or something else. I also didn't test this yet. Is this ok for you?

openslides/motions/models.py
FinnStutzenstein: ok

openslides/motions/access_permissions.py
FinnStutzenstein: ok

emanuelschuetze:
@normanjaeckel Please review again!

openslides/motions/views.py
normanjaeckel: Hm. I am not happy, but it would be ok. Please add a comment here that we plan to refactor the sorting on client side and then also refactor this view.

normanjaeckel:

openslides/motions/views.py
FinnStutzenstein: Done.

PR #3897, opened 2018-09-27
Updating Angular to the new 7 beta

jsaalfeld:
This enables us to use Drag n Drop and other components, without
using any plugins.


jsaalfeld:
https://beta-angular-material-io.firebaseapp.com/cdk/drag-drop/overview

PR #3924, opened 2018-10-13
remove old client

ostcar:
TODOs:
- [ ] See TODOs in DEVELOPMENT.rst and README.rst
- [ ] Document, how the new client has to be installed ( I had to guess, that I have to call npm install in the client folder)
- [ ] Document, how the new client can be started (I had to guess, that I have to open localost:4200 in the browser)
- [ ] Document, how the client can be build for production in packed into the python-package


FinnStutzenstein:
@ostcar I added one commit. Should go now in the right direction

I added a dummy index.html. If you do a `yarn build`, these files will be delivered instead of the dummy one. The index view is now a static file server, but if a file was requested that does not exist, or a directory, the index.html is delivered.

Why do we have the redirect view that appends a slash to every route? Does this has a purpose?

ostcar:
> @ostcar I added one commit. Should go now in the right direction
> Why do we have the redirect view that appends a slash to every route? Does this has a purpose?
I am not quite sure. It could be, that the django views needs the slash. I could only be, that we add this redirect, because we like uniform urls. The redirect makes sure, that there are not urls like /rest/motions/motions in the wild but only /rest/motions/motions/

ostcar:
I checked the urlpattern. The /rest/ url still works without the static in urlpattern.

I did not understand why you added the CSRFMixin to the IndexView. It has only an effect on post, put and patch requests. So I removed it.

I opened other issues. See the initial issue-text

ostcar:
@normanjaeckel I removed tested code. This means that the test coverage is lower then before. I would have to write tests for functions, that are not part of this pull request, to fulfill the travis requirements. I don't like the test-coverage anymore. Could we remove it?

@FinnStutzenstein I rebased the PR on top of master. This then, the js-tests fails. Could you have a look at it?

DEVELOPMENT.rst
```diff
@@ -25,6 +25,8 @@ For Ubuntu 16.04 e. g. follow `Yarn installation instructions
 
     $ sudo apt-get install git nodejs nodejs-legacy npm build-essential python3-dev
 
+TODO: Please check, if this is still valid for the new client
```
normanjaeckel: Please open an issue for this and remove the TODO here.

DEVELOPMENT.rst
```diff
@@ -84,13 +73,25 @@ When debugging something email related change the email backend to console::
...
+'''''''''''''''''''''''''''''
 
-    $ python manage.py runserver
+Go in the client's directory in a second command-line interface::
+
+    $ cd client/
+
+Install all dependencies and start the development server::
+
+    $ npm install  # TODO: yarn install gave me errors a few weeks ago..
```
normanjaeckel: We should know if we want to use yarn or npm, not both.

DEVELOPMENT.rst
```diff
@@ -84,13 +73,25 @@ When debugging something email related change the email backend to console::
...
+    $ yarn start
 
-Use gulp watch in a second command-line interface::
+Now the client is available under ``localhost:4200``.
 
-    $ node_modules/.bin/gulp watch
+If you do not need to work with the client, you can build the client and let it be delivered by the server directly::
+
+    $ yarn build # TODO: this is not tested!
```
normanjaeckel: Please remove the TODO. @FinnStutzenstein Can you test this please?

DEVELOPMENT.rst
```diff
@@ -201,6 +206,11 @@ When using Nginx as a proxy for delivering staticfiles the performance of the se
 
 This is an example configuration for a single Daphne listen on port 8000::
 
+TODO: Rewrite this that /static/templats/index.html is the default and only
```
normanjaeckel: Here the same. Please remove the TODO and put it into an issue.

README.rst
```diff
@@ -167,6 +167,8 @@ OpenSlides uses the following projects or parts of them:
 
 * Several JavaScript packages (see ``bower.json``)
 
+TODO: Please check, if this is still valid for the new client
```
normanjaeckel: Here the same.

openslides/core/static/index.html
```diff
@@ -0,0 +1,3 @@
+<h1>Hello user!</h1>
+
+Use <a href="http://localhost:4200">This url</a> to get to the client. For further information see the DEVELOPMENT.rst
```
normanjaeckel: Why not add a Django redirect view?

DEVELOPMENT.rst
ostcar: I would say we open one issue to fix the README and DEVELOPMENT file before the 3.0 release.

DEVELOPMENT.rst
ostcar: I tested it. It works fine

openslides/core/static/index.html
ostcar: When you call yarn build, this index.html is overwritten and the actual client is returned.

Of cause, you could add some JS in this html file, that makes the redirect.

FinnStutzenstein: This may not be the right url. We better give hints how to set up the server and client rigt, because if one gets to the page, he did something wrong

DEVELOPMENT.rst
FinnStutzenstein: removed

FinnStutzenstein:
@ostcar This can be merged right?

PR #3925, opened 2018-10-13
Improve redis cache

ostcar:
* delete only keys with prefix
* Make redis_provider atomic with transactions and lua scripts
* improve lock
* generate change_id in redis to make sure it is uniq
* use miliseconds as starttime
* add argument use max_change_id to get_{full|resticted}_data

This changes are needed before the change_id system can be implemented.

### DIFF
--- a/openslides/utils/cache.py
+++ b/openslides/utils/cache.py
@@ -77,7 +77,8 @@ class ElementCache:
 
         # Start time is used as first change_id if there is non in redis
         if start_time is None:
-            start_time = int((datetime.utcnow() - datetime(1970, 1, 1)).total_seconds())
+            # Use the miliseconds (rounted) since the 2016-02-29.
+            start_time = int((datetime.utcnow() - datetime(2016, 2, 29)).total_seconds()) * 1000
         self.start_time = start_time
 
         # Contains Futures to controll, that only one client updates the restricted_data.
@@ -151,13 +152,7 @@ class ElementCache:
         if deleted_elements:
             await self.cache_provider.del_elements(deleted_elements)
 
-        # TODO: The provider has to define the new change_id with lua. In other
-        #       case it is possible, that two changes get the same id (which
-        #       would not be a big problem).
-        change_id = await self.get_next_change_id()
-
-        await self.cache_provider.add_changed_elements(change_id, elements.keys())
-        return change_id
+        return await self.cache_provider.add_changed_elements(self.start_time + 1, elements.keys())
 
     async def get_all_full_data(self) -> Dict[str, List[Dict[str, Any]]]:
         """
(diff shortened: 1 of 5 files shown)


openslides/utils/cache.py
```diff
@@ -77,7 +77,8 @@ def __init__(
 
         # Start time is used as first change_id if there is non in redis
         if start_time is None:
-            start_time = int((datetime.utcnow() - datetime(1970, 1, 1)).total_seconds())
+            # Use the miliseconds (rounted) since the 2016-02-29.
```
normanjaeckel: Nice. I like this date ...

openslides/utils/cache_providers.py
```diff
@@ -253,21 +252,48 @@ def zadd_args(change_id: int) -> Generator[Union[int, str], None, None]:
...
-                else:
-                    collection_string, id = split_element_id(element_id)
-                    changed_elements[collection_string].append(element_json)
-            return changed_elements, deleted_elements
+            # lua script that returns gets all element_ids from change_id_cache_key
+            # and then uses each element_id on full_data or restricted_data.
+            # It returns a list where the odd values are the change_id and the
+            # even values the element as json. The function wait_make_dict creates
+            # a python dict from the returned list.
+            elements: Dict[bytes, Optional[bytes]] = await aioredis.util.wait_make_dict(redis.eval(
```
normanjaeckel: Do you want to use Python's textwrap.dedent() or is this not required?

openslides/utils/cache_providers.py
```diff
@@ -530,3 +565,28 @@ def get_all_cachables() -> List[Cachable]:
             continue
         out.extend(get_startup_elements())
     return out
+
+
+lua_script_change_data = """
```
normanjaeckel: Here the same. I would prefer to indent the code for better readability.

openslides/utils/cache.py
FinnStutzenstein: Was something else there together with the leap day?
I would suggest taking the 2011-9-12

ostcar: It has its reason.

openslides/utils/cache_providers.py
ostcar: It is not required. Taps and leading white space have to meaning in lua. The only downside is, that it is unneeded bytes that have to send to redis each time. But the correct solution for that (in another pull request) is to use https://redis.io/commands/evalsha

openslides/utils/cache_providers.py
ostcar: I don't understand what you mean with indent. See line 575 for example

FinnStutzenstein:
Ok, is there something I can test?
Do you think this will become more lua code? If yes, we should put the lua-code in lua-files and read their content during the setup. For this two pieces, it's fine, but if they get extended, or more scripts, we should move them out.

ostcar:
You can not test with with this pull request. After it is merged and #3926 is merged, I have to add one more pr on top of both and then you can use it.

PR #3928, opened 2018-10-15
Strip out response.data.

normanjaeckel:


openslides/utils/rest_api.py
```diff
@@ -160,6 +162,14 @@ def get_serializer_class(self) -> Type[Serializer]:
             serializer_class = super().get_serializer_class()  # type: ignore
         return serializer_class
 
+    def finalize_response(self, request, response, *args, **kwargs):
+        """
+        Just remove all response data so nobody may get unrestricted data.
+        """
+        response = super().finalize_response(request, response, *args, **kwargs)
```
ostcar: I think this method is also called on get requests.

If we want to keep the rest api, they have to see the data.

FinnStutzenstein: Yes, If we do this, we can disable the rest api. I'm not a fan of just removing the data, but this may be the simplest solution.

We may provide a list of keys (default an empty list) with keys to remove on responses. So users can have `['session_auth_hash']`. Here, the all entries with keys in this blacklist will be removed. This should be a simple solution, too, because you just have to specify keys for each model to be blacklisted.

normanjaeckel: But how can I access the REST API with the new client? /rest does not work.

FinnStutzenstein: You need to use the old port (8000).

normanjaeckel:
Can someone help me with the mypy error? (Maybe I am doing it wrong but I just want to use the Mixin as mixin so of course the superclass does not exist for the mixin itself.

openslides/utils/rest_api.py
normanjaeckel: Ok, done.

FinnStutzenstein:
Many tests are failing, because in any case of an error, the error is stripped, too. Maybe also check for the status code?

openslides/utils/rest_api.py
ostcar: Are there other http methods that needs to return data? What about OPTIONS requests?

Is it necessary now to go throw all writing requests and look to the client if it needs a response?

normanjaeckel:
Is it possible to run tests without having redis installed?

openslides/utils/rest_api.py
normanjaeckel: I think we do not want to support OPTIONS requests any more.

And yes: Client delevopers should define which information they want to have as a result of writing operations. So I strongly recommend not to use a blacklist but a whitelist.

normanjaeckel:
@FinnStutzenstein Can you please check if the client still needs from the server:
- group OPTIONS request
- mass email response (number of emails)
- messages of some motion actions (category numbered, recommendation, state, support status)
- list of speakers set or removal response?

Thanks.

FinnStutzenstein:
We do not need any OPTION requests. The `GroupViewSetMetadata` can be removed. We need feedback on mass-* operations, like email sending and importing users. I think, these operations will become more in the future. Also we may need the ids of created polls (this was needed for the voting plugin; but we should take account for some views to return data to the client).

Currently, success messages are not shown and I think they won't be in the future. But we need every error message. Maybe we should whitelist the key `detail`.

Also there are some GET requests like WhoAmI and for login, but those are no rest routes, so this should not be a problem.

openslides/utils/rest_api.py
ostcar: Could you add one example to this pull request, how a rest-view can explicit return data?

normanjaeckel:
The first pull request was not good, so here is a differen approach. Please test and let me know your opinions. Now I only strip out create() and update() responses made by the DjangoRESTFramework. All custom responses will stay.

FinnStutzenstein:
Ok, by overriding `update` and `create` one can send data. This is ok. Special views are not affected: OK. Whats about errors during creation and updating. I think an error will be raised, so this is ok, too, right?

tests/integration/users/test_viewset.py
```diff
@@ -85,7 +85,6 @@ def test_simple_creation(self):
 
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
         new_user = User.objects.get(username='Test name keimeiShieX4Aekoe3do')
-        self.assertEqual(response.data['id'], new_user.id)
```
ostcar: There is no assert statement that uses new_user. Could you either add one or add a comment, that no assert statement is needed, because it only testes, that the call does not rais a DoNotExists exception.

tests/integration/motions/test_viewset.py
```diff
@@ -342,7 +342,7 @@ def test_amendment_motion(self):
             {'title': 'test_title_doe93Jsjd2sW20dkSl20',
              'text': 'test_text_feS20SksD8D25skmwD25',
              'parent_id': parent_motion.id})
-        created_motion = Motion.objects.get(pk=int(response.data['id']))
+        created_motion = Motion.objects.get(title='test_title_doe93Jsjd2sW20dkSl20'))
```
ostcar: Here is a syntax error. The last bracket has to be removed.

tests/integration/motions/test_viewset.py
```diff
@@ -394,9 +394,9 @@ def create_parent_motion(self):
         """
         response = self.client.post(
```
ostcar: The variable "response" is not used anymore. Please remove it.

tests/integration/users/test_viewset.py
ostcar: If you don't want to use the varialbe new_user, then please remove it.

FinnStutzenstein:
In PR #3950 I make all responses `void` for create/update/delete in the clinet adapting changes in this PR. I noticed somthing:
We often need the id (not more) when something gets created, to redirect to the detail view. May you just allow to send an object with just the id as response?

normanjaeckel:
Everything is fine now. Create requests will respond with id, update requests won't. Please do last test, then merge.

PR #3931, opened 2018-10-18
CSV export (example implementation in user list)

MaximilianKrambach:
This does the csv export for the user list

The csvExportService is generic and can be used from other places as well (see user-list.component.ts: csvExportUserList)

I use the `file-saver` package, which works well. I decided against the `angular-file-saver` package, which is only an aditional wrapper around file-saver and Blob. If we need to support a browser that knows nothing about Blob, we might switch to angular-file-saver dependency instead.

I did not care about permission checks prior to exporting, as my main focus was on the export itself, and the user export is just the random example I picked.

### DIFF
--- a/client/src/app/app.component.ts
+++ b/client/src/app/app.component.ts
@@ -54,6 +54,7 @@ export class AppComponent {
      * TODO: Overloading can be extended to more functions.
      */
     private overloadArrayToString(): void {
+
         Array.prototype.toString = function(): string {
             let string = '';
             const iterations = Math.min(this.length, 3);
--- a/client/src/app/core/services/csv-export.service.ts
+++ b/client/src/app/core/services/csv-export.service.ts
@@ -1,0 +1,119 @@
+import { BaseViewModel } from '../../site/base/base-view-model';
+import { Injectable } from '@angular/core';
+import { FileExportService } from './file-export.service';
+import { TranslateService } from '@ngx-translate/core';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class CsvExportService {
+    /**
+     * Constructor
+     */
+    public constructor(protected exporter: FileExportService, private translate: TranslateService) {}
+
+    /**
+     * Saves an array of model data to a CSV.
+     * @param data Array of Model instances to be saved
+     * @param columns Column definitions
+     * @param filename name of the resulting file
+     * @param options optional:
+     *      lineSeparator (defaults to \r\n windows style line separator),
+     *      columnseparator defaults to semicolon (other usual separators are ',' '\T' (tab), ' 'whitespace)
+     */
+    public export<T extends BaseViewModel>(
+        data: T[],
+        columns: {
+            property: keyof T; // name of the property used for export
+            label?: string;
+            assemble?: string; // (if property is further child object, the property of these to be used)
+        }[],
+        filename: string,
+        { lineSeparator = '\r\n', columnSeparator = ';' }: { lineSeparator?: string; columnSeparator?: string } = {}
+    ): void {
+        const allLines = []; // Array of arrays of entries
+        const usedColumns = []; // mapped properties to be included
+
+        // initial array of usable text separators. The first character not used
+        // in any text data or as column separator will be used as text separator
+        let tsList = ['"', "'", '`', '/', '\\', ';', '.'];
+
+        if (lineSeparator === columnSeparator) {
+            throw new Error('lineseparator and columnseparator must differ from each other');
+        }
+
+        tsList = this.checkCsvTextSafety(lineSeparator, tsList);
+        tsList = this.checkCsvTextSafety(columnSeparator, tsList);
+
+        // create header data
+        const header = [];
+        columns.forEach(column => {
+            const rawLabel: string = column.label ? column.label : (column.property as string);
+            const colLabel = this.capitalizeTranslate(rawLabel);
+            tsList = this.checkCsvTextSafety(colLabel, tsList);
+            header.push(colLabel);
+            usedColumns.push(column.property);
+        });
+        allLines.push(header);
+        // create lines
+        data.forEach(item => {
+            const line = [];
+            for (let i = 0; i < usedColumns.length; i++ ){
+                const property = usedColumns[i];
+                let prop: any = item[property];
+                if (columns[i].assemble){
+                    prop = item[property].map(subitem => this.translate.instant(subitem[columns[i].assemble])).join(',');
+                }
+                tsList = this.checkCsvTextSafety(prop, tsList);
+                line.push(prop);
+            };
+            allLines.push(line);
+        });
+
+        // assemble lines, putting text separator in place
+        if (!tsList.length) {
+            throw new Error('no usable text separator left for valid csv text');
+        }
+
+        const allLinesAssembled = [];
+        allLines.forEach(line => {
+            const assembledLine = [];
+            line.forEach(item => {
+                if (typeof item === 'number') {
+                    assembledLine.push(item.toString(10));
+                } else if (item === null || item === undefined || item === '') {
+                    assembledLine.push('');
+                } else if (item === true) {
+                    assembledLine.push('1');
+                } else if (item === false) {
+                    assembledLine.push('0');
+                } else {
+                    assembledLine.push(tsList[0] + item + tsList[0]);
+                }
+            });
+            allLinesAssembled.push(assembledLine.join(columnSeparator));
+        });
+        this.exporter.saveFile(allLinesAssembled.join(lineSeparator), filename);
+    }
+
+    /**
+     * Checks if a given input contains any of the characters defined in a list
+     * used for textseparators. The list is then returned without the 'special'
+     * characters, as they may not be used as text separator in this csv.
+     * @param input any input to be sent to CSV
+     * @param tsList The list of special characters to check.
+     */
+    public checkCsvTextSafety(input: any, tsList: string[]): string[] {
+        if (input === null || input === undefined ) {
+            return tsList;
+        }
+
+        const inputAsString = String(input);
+        return tsList.filter(char => inputAsString.indexOf(char) < 0);
+    }
+
+    private capitalizeTranslate(input: string): string {
+        const temp = input.charAt(0).toUpperCase() + input.slice(1);
+        return this.translate.instant(temp);
+    }
+}
--- a/client/src/app/core/services/file-export.service.ts
+++ b/client/src/app/core/services/file-export.service.ts
@@ -1,0 +1,33 @@
+import { Injectable } from '@angular/core';
+import { saveAs } from 'file-saver';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class FileExportService {
+    /**
+     * Constructor
+     */
+    public constructor() {}
+
+    /**
+     * Saves a file
+     * @param file
+     * @param filename
+     */
+    public saveFile(file: BlobPart, filename: string): void {
+        const blob = new Blob([file]);
+        saveAs(blob, filename, { autoBOM: true });
+        // autoBOM = automatic byte-order-mark
+    }
+
+    /**
+     * Validates a file name for characters that might break.
+     * @param filename A name used to save a file
+     */
+    protected validateFileName(filename: string): boolean {
+        // everything not containing \/?%*:|"<> and not ending on a dot
+        const pattern = new RegExp(/^[^\\\/\?%\*:\|\"\<\>]*[^\.]+$/i);
+        return pattern.test(filename);
+    }
+}
--- a/client/src/app/site/users/components/user-list/user-list.component.ts
+++ b/client/src/app/site/users/components/user-list/user-list.component.ts
@@ -1,6 +1,7 @@
 import { Component, OnInit } from '@angular/core';
 import { Title } from '@angular/platform-browser';
 import { TranslateService } from '@ngx-translate/core';
+import { CsvExportService } from '../../../../core/services/csv-export.service';
 
 import { ViewUser } from '../../models/view-user';
 import { UserRepositoryService } from '../../services/user-repository.service';
@@ -17,6 +18,7 @@ import { Router, ActivatedRoute } from '@angular/router';
     styleUrls: ['./user-list.component.scss']
 })
 export class UserListComponent extends ListViewBaseComponent<ViewUser> implements OnInit {
+
     /**
      * The usual constructor for components
      * @param repo the user repository
@@ -28,7 +30,8 @@ export class UserListComponent extends ListViewBaseComponent<ViewUser> implement
         protected titleService: Title,
         protected translate: TranslateService,
         private router: Router,
-        private route: ActivatedRoute
+        private route: ActivatedRoute,
+        protected csvExport: CsvExportService
     ) {
         super(titleService, translate);
     }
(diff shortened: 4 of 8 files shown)


MaximilianKrambach:
ping @FinnStutzenstein for review

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,118 @@
...
+    public constructor() {}
+
+    /**
+     * Saves an array of model data to a CSV.
+     * @param data Array of Model instances to be saved
+     * @param columns Column definitions (model property; optional csv title)
+     * @param filename name of the resulting file
+     * @param options
+     */
+    public saveAsCsv(
```
FinnStutzenstein: `public save<T extends BaseViewModel>(data: T[], columns: { [P in keyof T]?/* the property*/: string /* the column title*/}, filename: string, ...rest)`
Does it work to make `{ineSeparator = '\r\n', columnSeparator = ';'}` optional. If you do not want to provide values, you have to pass an empty object; this is not nice.

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,118 @@
...
+        if (lineSeparator === columnSeparator) {
+            throw new Error('TODO: lineseparator and columnseparator must differ from each other');
+        }
+
+        ts_list = this.checkCsvTextSafety(lineSeparator, ts_list);
+        ts_list = this.checkCsvTextSafety(columnSeparator, ts_list);
+
+        // create header data
+        const header = [];
+        for (const column of columns) {
```
FinnStutzenstein: Here you have to interate through the key-values in the columns object.

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,118 @@
...
+            if (column.title) {
+                ts_list = this.checkCsvTextSafety(column.title, ts_list);
+                header.push(column.title);
```
FinnStutzenstein: Should this be translated?

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,118 @@
...
+                ts_list = this.checkCsvTextSafety(item[property], ts_list);
+                line.push(item[property]);
+            }
+            all_lines.push(line);
+        }
+
+        // assemble lines, putting text separator in place
+
+        const final_lines = [];
+        if (ts_list.length) {
```
FinnStutzenstein: swap this to reduce indentions:
```
if (!ts_list.length) {
    throw ...
}
...
```

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,118 @@
...
+                        // TODO export 0/1 for typeof === 'boolean' ?
+                    } else {
+                        const newstring = ts_list[0] + item + ts_list[0];
+                        newline.push(newstring);
+                    }
+                }
+                final_lines.push(newline.join(columnSeparator));
+            }
+            this.exporter.saveFile(final_lines.join(lineSeparator), filename);
+            return true;
```
FinnStutzenstein: remove this and make the function return void. If somethings happens, an error can be thrown.

client/src/app/core/services/file-export.service.ts
```diff
@@ -0,0 +1,45 @@
...
+     * empty constructor TODO
+     */
+    public constructor() {}
+
+    /**
+     * Saves a file
+     * @param file
+     * @param filename
+     * TODO: Potential encoding issues and path/filename oddities on different browsers
+     * also: default filename
```
FinnStutzenstein: I would suggest making the filename not optional.

client/src/app/core/services/file-export.service.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * Validates a file name for characters that might break.
+     * @param filename A name used to save a file
+     */
+    protected validateFileName(filename: string): boolean {
+        // everything not containing \/?%*:|"<> and not ending on a dot
+        const pattern = new RegExp(/^[^\\\/\?%\*:\|\"\<\>]*[^\.]+$/i);
+        if (pattern.test(filename)) {
```
FinnStutzenstein: just do `return pattern.test(filename);`.

client/src/app/site/users/components/user-list/user-list.component.ts
```diff
@@ -99,4 +101,18 @@ export class UserListComponent extends ListViewBaseComponent<ViewUser> implement
     public onPlusButton(): void {
         this.router.navigate(['./new'], { relativeTo: this.route });
     }
+
+    // TODO save all data from the dataSource
+    public saveUserList(): void {
+        this.csvExport.saveAsCsv(
+            this.dataSource.data,
+            [
```
FinnStutzenstein:
```
{
    fullName: 'Name',
    groups: 'Groups',
    isActive: 'Presence'
}
```

client/src/app/core/services/csv-export.service.ts
MaximilianKrambach:
```
{ lineSeparator = '\r\n', columnSeparator = ';' }:{ lineSeparator?: string; columnSeparator?: string } = {}
```
This should work (at least in es2016 this is how destructuring works with an default of 'no data submitted'; I'm testing it now)

client/src/app/core/services/csv-export.service.ts
MaximilianKrambach: So, we'll lose the 'label for this column is optional'. If that's fine, I'll continue with this

client/src/app/core/services/csv-export.service.ts
MaximilianKrambach: I would not translate it during csv export, but rather on the other end (the listView), where a translationService is already supposed to be in place

client/src/app/core/services/csv-export.service.ts
FinnStutzenstein: Perfect!

client/src/app/core/services/csv-export.service.ts
FinnStutzenstein: yeah. If we have too much of the same names, you can define the object values as `string|null`, null meaning "take the property value".

client/src/app/core/services/csv-export.service.ts
FinnStutzenstein: You can just put the `TranslateService` into the constructor. Then we have one central place where the values get translated and the views do not have to care about this (should be less code there then).

client/src/app/site/users/components/user-list/user-list.component.ts
MaximilianKrambach: (direct discussion: we keep it an array of objects: Order will be preserved, and label can be optional (capitalized version of the property will be translated and used, if no label present)

client/src/app/core/services/csv-export.service.ts
```diff
@@ -109,4 +116,13 @@ export class CsvExportService {
         }
         return ts_list;
     }
+
+    private capitalizeTranslate(input: any): string {
```
FinnStutzenstein: `input: string` and remove the check below.

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,128 @@
...
+            throw new Error('lineseparator and columnseparator must differ from each other');
+        }
+
+        ts_list = this.checkCsvTextSafety(lineSeparator, ts_list);
+        ts_list = this.checkCsvTextSafety(columnSeparator, ts_list);
+
+        // create header data
+        const header = [];
+        columns.forEach(column => {
+            const rawLabel = column.label ? column.label : column.property;
```
FinnStutzenstein: then declare `rawLabel` as string and cast the property: `column.property as string`.

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,124 @@
+import { BaseViewModel } from '../../site/base/base-view-model';
+import { Injectable } from '@angular/core';
+import { FileExportService } from './file-export.service';
+import { TranslateService } from '@ngx-translate/core';
+
+/**
+ * Handles all incoming and outgoing notify messages via {@link WebsocketService}.
```
FinnStutzenstein: Isn't this a bit off? ;)

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Saves an array of model data to a CSV.
+     * @param data Array of Model instances to be saved
+     * @param columns Column definitions
+     * @param filename name of the resulting file
+     * @param options optional:
+     *      lineSeparator (defaults to \r\n windows style line separator),
+     *      columnseparator defaults to semicolon (other usual separators are ',' '\T' (tab), ' 'whitespace)
+     */
+    public saveAsCsv<T extends BaseViewModel>(
```
FinnStutzenstein: Just name it `save` or `export`.

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,124 @@
...
+        data: T[],
+        columns: {
+            property: keyof T;
+            label?: string;
+        }[],
+        filename: string,
+        { lineSeparator = '\r\n', columnSeparator = ';' }: { lineSeparator?: string; columnSeparator?: string } = {}
+    ): void {
+        const all_lines = []; // Array of arrays of entries
```
FinnStutzenstein: Please make all variables camelCase

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,124 @@
...
+            const rawLabel: string = column.label ? column.label : (column.property as string);
+            const colLabel = this.capitalizeTranslate(rawLabel);
+            ts_list = this.checkCsvTextSafety(colLabel, ts_list);
+            header.push(colLabel);
+            used_columns.push(column.property);
+        });
+        all_lines.push(header);
+        // create lines
+        data.forEach(item => {
+            const line = [];
```
FinnStutzenstein:
```
 const line = used_columns.map(property => {
    ts_list = this.checkCsvTextSafety(item[property], ts_list);
    return item[property];
});
```

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,124 @@
...
+            });
+            all_lines.push(line);
+        });
+
+        // assemble lines, putting text separator in place
+        const final_lines = [];
+        if (!ts_list.length) {
+            throw new Error('no usable text separator left for valid csv text');
+        }
+        all_lines.forEach(function(line: any[]): void {
```
FinnStutzenstein: `const finalLines = allLines.map(line => {... return newline.join(columnSeparator)});`

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,124 @@
...
+            const newline = [];
+            line.forEach(function(item: any): void {
```
FinnStutzenstein: `line.forEach(item => {`

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Checks if a given input contains any of the characters defined in a list
+     * used for textseparators. The list is then returned without the 'special'
+     * characters, as they may not be used as text separator in this csv.
+     * @param input any input to be sent to CSV
+     * @param ts_list The list of special characters to check.
+     */
+    public checkCsvTextSafety(input: any, ts_list: string[]): string[] {
```
FinnStutzenstein: type of input: `string`. Cast above if accessing `item[property]`.

The complete function may be cut down to `return tsList.filter(char => input.indexOf(char) < 0);`. This needs testing but should work.

MaximilianKrambach: At this type I should not be required to rely on it being a string: For later parsing, I will need to escape texts into a textseparator (which I'm finding out at this stage). If I cast is to a string above, I'll lose the ability to check for numbers or booleans further in the line assembly

client/src/app/site/users/components/user-list/user-list.component.ts
```diff
@@ -69,4 +93,17 @@ export class UserListComponent extends ListViewBaseComponent<ViewUser> implement
     public onPlusButton(): void {
         this.router.navigate(['./new'], { relativeTo: this.route });
     }
+
+    // TODO save all data from the dataSource
+    public saveUserList(): void {
```
emanuelschuetze: Please use 'csv' in the function name, e.g. `csvExportUserList()`, because later we'll export the user list also in PDF. See function names in 2.3.x (`pdfExportUserList()`, `pdfExportUserAccessDataList()`, `csvExport())`: https://github.com/OpenSlides/OpenSlides/blob/stable/2.3.x/openslides/users/static/templates/users/user-list.html#L67

Also the menu button should show 'CSV', e.g. 'Export (CSV)'.

client/src/app/site/users/components/user-list/user-list.component.ts
```diff
@@ -69,4 +93,17 @@ export class UserListComponent extends ListViewBaseComponent<ViewUser> implement
...
+        this.csvExport.export(
+            this.dataSource.data,
+            [
+                { property: 'full_name', label: 'Name' },
+                { property: 'groups', label: 'Groups' },
+                { property: 'is_active', label: 'Presence' }
```
emanuelschuetze: Please use the same csv columns as in 2.3.x:
https://github.com/OpenSlides/OpenSlides/blob/stable/2.3.x/openslides/users/static/js/users/csv.js#L14

emanuelschuetze:
@MaximilianKrambach This pull requst ist for user csv export only, right? Should be mentioned in commit message / pr title. Still open: agenda and motions csv export.

client/src/app/site/users/components/user-list/user-list.component.ts
MaximilianKrambach: Excepting 'number' (which is not part of the data I can access in the user list), I included all data now

FinnStutzenstein: it's named `participant_number`.

MaximilianKrambach:
I updated according to your comments

ping @FinnStutzenstein @emanuelschuetze

MaximilianKrambach:
- Okay, a new type of menu was sneaked in; fixed.

- The other problem is that an openslide-specific overload destroyed `Array.join()`. I commented out  the overload, and we need to talk about how to deal with it (ping @FinnStutzenstein ).

- The other hiccup I stumbled on is that user.groups is not a string or number, but an array of sub-items. I think I fixed handling there, but it may not be clear what to do.

client/src/app/app.component.ts
```diff
@@ -54,22 +54,24 @@ export class AppComponent {
      * TODO: Overloading can be extended to more functions.
      */
     private overloadArrayToString(): void {
-        Array.prototype.toString = function(): string {
```
FinnStutzenstein: I checked and this was not a problem to comment in. Please verify

client/src/app/core/services/csv-export.service.ts
```diff
@@ -0,0 +1,123 @@
...
+            const line = [];
+            for (let i = 0; i < usedColumns.length; i++ ){
+                const property = usedColumns[i];
+                let prop: any = item[property];
+                if (columns[i].assemble){
+                    const subproperties = [];
+                    item[property].forEach(subitem => {
+                        subproperties.push(subitem[columns[i].assemble]);
+                    });
+                    prop = subproperties.join(' ');
```
FinnStutzenstein: `prop = item[property].map(subitem => this.translate.instant(subitem[columns[i].assemble])).join(',');`
Also remove the `,` from tsList.

MaximilianKrambach: So, the assumptions are: The data of the assembled subitems are always strings to be translated, and they will be comma separated. Well then.

client/src/app/app.component.ts
MaximilianKrambach: Okay, with your (slightly different) ways of assembling a list we indeed avoid this problem; I am still unhappy with sneakily overwriting natiive functions.

MaximilianKrambach:
I'll rebase tomorrow.

client/src/app/core/services/csv-export.service.ts
FinnStutzenstein: yes. Maybe ad a toString then; but we can change this, if it is the case.

MaximilianKrambach:
@FinnStutzenstein your remarks have been addressed as desired, and I rebased and squished the commits. Mergeable in my opinion

PR #3938, opened 2018-10-19
Public change id api

ostcar:
Please review. I fix the tests as soon as I find more time.

### DIFF
--- a/openslides/utils/consumers.py
+++ b/openslides/utils/consumers.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 from typing import Any, Dict, List, Optional
+from urllib.parse import parse_qs
 
 import jsonschema
 from asgiref.sync import sync_to_async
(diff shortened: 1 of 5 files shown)


openslides/utils/consumers.py
```diff
@@ -100,14 +102,34 @@ class SiteConsumer(ProtocollAsyncJsonWebsocketConsumer):
...
-            await self.accept()
-            data = await startup_data(self.scope['user'])
+            return
+
+        query_string = parse_qs(self.scope['query_string'])
+        if b'change_id' in query_string:
+            try:
+                change_id = int(query_string[b'change_id'][0])
+            except ValueError:
+                await self.close()  # TODO: Find a way to send an error code
```
FinnStutzenstein: Maybe we should define our own status codes beginning at 4000 (https://tools.ietf.org/html/rfc6455#section-7.4.2).
like one fpr OK, bad query string, ...

FinnStutzenstein:
For my part, you can do this, so I can take care about the client.

PR #3943, opened 2018-10-22
Diff Bugfix: inserted/deleted list items

CatoTH:
This fixes one or two of the given problems regarding inserted / deleted list items.

The deleted section is actually dead code that is not active anymore since a much earlier revision and was therefore only slowing down the execution.


normanjaeckel:
Ok.

PR #3949, opened 2018-10-26
Move websocket protocoll in apps

ostcar:
This makes it easier to add new websocket messages.

With this pull request, it is also possible for plugins, to add new websocket messages


ostcar:
There is something wrong with this pull request. I reopen it

PR #3953, opened 2018-10-27
Add test for second getElement request

ostcar:
This does not change anything. Only adds tests


ostcar:
I merge without review, because it does not change anything.

PR #3959, opened 2018-10-30
Split up autoupdate functions (sync and async).

normanjaeckel:


ostcar:
I will properply remove the CollectionElement Class before 3.0. If I do, I will rename you function. But currently, the name is correct.

PR #3966, opened 2018-11-02
Remove utils.collections.Collection class and other cleanups

ostcar:
* Activate restricted_data_cache on inmemory cache
* Use ElementCache in rest-api get requests
* Get requests on the restapi return 404 when the user has no permission
* Added async function for has_perm and in_some_groups
* changed Cachable.get_restricted_data to be an ansync function
* rewrote required_user_system
* changed default implementation of access_permission.check_permission to
  check a given permission or check if anonymous is enabled


openslides/utils/rest_api.py
```diff
@@ -225,15 +225,10 @@ def retrieve(self, request: Any, *args: Any, **kwargs: Any) -> Response:
             response = super().retrieve(request, *args, **kwargs)
         else:
             lookup_url_kwarg = self.lookup_url_kwarg or self.lookup_field
-            collection_element = CollectionElement.from_values(
```
normanjaeckel: Is the docstring in line 214 still valid?

ostcar: no. I removed it

PR #3978, opened 2018-11-04
Set websocket url to /ws/

ostcar:
The routing is desided by protocol


ostcar:
@FinnStutzenstein can you fix the js bug?

ostcar:
If it is not possible, or hard to solve, I have no problem with using /ws/

FinnStutzenstein:
Changing to this url would be sufficient and easy to do I guess.

PR #3979, opened 2018-11-04
[WIP] Remove Django Permissions

ostcar:


FinnStutzenstein:
@ostcar After a long period of nothing, I had time to make some migrations. I (re)added the django auth app, so we must not throw away old migrations and commented in some imports. It seems like it works. Please have a look.

FinnStutzenstein:
@ostcar Is tis still a thing?

ostcar:
I think, I don't want to do it anymore. Maybe if the new openslides server does not work out, I will look at it again.

PR #4002, opened 2018-11-08
first work on the call list

FinnStutzenstein:
@emanuelschuetze Please look at this. One thing to decide: should changes be saved immediately (like the list of speakers) or do we want a save button (the one I put there will be removed then and exchanged with a nice button in the head bar)?

Also this is just for a planar list. Tree-like sorting is a huge todo for OS as a whole.


emanuelschuetze:
Looks good. Save button is ok for the moment. But please show also the motion identifier in the sort view. Currently there is the motion title only.

FinnStutzenstein:
@emanuelschuetze Ok, added identifier and make the save button nicer. The last commits are just style things; If the call list is ok, we can rebase.

FinnStutzenstein:
@jsaalfeld All done.

jsaalfeld:
The wrong save button is used. If changed, this will be merged.

FinnStutzenstein:
@jsaalfeld Done and rebased

Issue #3958, opened 2018-10-30
Change default of internal motion change recommendation.


normanjaeckel:
See https://github.com/OpenSlides/OpenSlides/pull/3940#pullrequestreview-169731288

emanuelschuetze:
@CatoTH The default should be `internal=false` (new change recommendations should directly published)

PR #4013, opened 2018-11-15
Change default of change recommendation internal value to public

CatoTH:
For #3958


openslides/motions/models.py
```diff
@@ -740,7 +740,7 @@ class MotionChangeRecommendation(RESTModelMixin, models.Model):
     rejected = models.BooleanField(default=False)
     """If true, this change recommendation has been rejected"""
 
-    internal = models.BooleanField(default=True)
+    internal = models.BooleanField(default=False)
```
FinnStutzenstein: Please also change in the migration file.

CatoTH:
Ah, ok. Done.

PR #4016, opened 2018-11-15
Do not try to access the database during migrations

FinnStutzenstein:


jsaalfeld:
This seems to fix the issue in #4007

ostcar:
Is is possible to do this differently and only hit the database when "python manage.py runserver" was called? Then the try-except block could be removed.

jsaalfeld:
> Is is possible to do this differently and only hit the database when "python manage.py runserver" was called? Then the try-except block could be removed.

You mean, rather to check if `migrate` is in `sys.argv` to check if `runserver` was in there and then just execute it? Wouldn't there be sideeffects with the pypi package or running os via `gunicorn`?

FinnStutzenstein:
I changed this. @jsaalfeld or @ostcar please check with unicorn

jsaalfeld:
A server may be started in different ways. It is not sufficient to check if `runserver` is in `argv`. You may also check for:

* `['/your/complete/path/to/.venv/bin/gunicorn', '-w', '4', '-b', '0.0.0.0:8000', '-k', 'uvicorn.workers.UvicornWorker', 'openslides.asgi:application']`
* `['/your/complete/path/to/.venv/bin/daphne', '-b', '0.0.0.0', '-p', '8000', 'openslides.asgi:application']`
* `['manage.py', 'runserver']`

FinnStutzenstein:
Another issue (discovered by @jsaalfeld ) fixed with this: During migrations, `ensure_cache` tries to fill the cache and throws an error. But there are some items in the cache, so on the next actual start of the server, the cache is neither flushed and rebuild nor updated. So the invalid cache is used and produces errors.

PR #4043, opened 2018-11-29
Bugfix for initial values of merge_amendment_into_final - #3975

CatoTH:
@emanuelschuetze please have a look if the values seem correct. Note the difference between -1 and 0: if an amendment has a status with an merge_amendment_into_final-value of 0, then the recommendation will be considered. If -1, then it will not be in the final version no matter what the recommendation is.

### DIFF
--- a/openslides/motions/signals.py
+++ b/openslides/motions/signals.py
@@ -26,17 +26,19 @@ def create_builtin_workflows(sender, **kwargs):
                       workflow=workflow_1,
                       recommendation_label='Acceptance',
                       css_class='success',
-                      merge_amendment_into_final=True)
+                      merge_amendment_into_final=1)
     state_1_2.save(skip_autoupdate=True)
     state_1_3 = State(name=ugettext_noop('rejected'),
                       workflow=workflow_1,
                       recommendation_label='Rejection',
-                      css_class='danger')
+                      css_class='danger',
+                      merge_amendment_into_final=-1)
     state_1_3.save(skip_autoupdate=True)
     state_1_4 = State(name=ugettext_noop('not decided'),
                       workflow=workflow_1,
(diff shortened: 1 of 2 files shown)


openslides/motions/signals.py
```diff
@@ -57,31 +58,36 @@ def create_builtin_workflows(sender, **kwargs):
...
+                                     css_class='default',
+                                     merge_amendment_into_final=-1)
     state_2_8 = State.objects.create(name=ugettext_noop('refered to committee'),
                                      workflow=workflow_2,
                                      action_word='Refer to committee',
                                      recommendation_label='Referral to committee',
-                                     css_class='default')
+                                     css_class='default',
+                                     merge_amendment_into_final=-1)
     state_2_9 = State.objects.create(name=ugettext_noop('needs review'),
```
normanjaeckel: needs review should also be -1

CatoTH: If the status is "needs review" and the recommendation is "Accepted", sure that it shouldn't be in the final version?

normanjaeckel: If the status is need_review, it is not possible to jump to accepted because in this workflow needs_review is a final state (see next_states field). Needs review means something like adjourned here.

CatoTH:
https://github.com/OpenSlides/OpenSlides/commit/0a823877c2cb4972a555747b335ec0f2c66c2104 should (hopefully) fix the Travis failure here, too.

FinnStutzenstein:
@CatoTH Its merged, so please rebase

openslides/motions/signals.py
emanuelschuetze: Ok for me - after discussion with @CatoTH

Issue #4045, opened 2018-11-30
Include prettier into travis


FinnStutzenstein:
@jsaalfeld This might be a task for you integrating the prittier into travis (after linting and before tests) to raise an error, if the code is not formatted correctly

jsaalfeld:
Code shouldn't be altered when commited.

I don't know what prettifying should do, what linting doesn't?

jsaalfeld:
Did a small local test by just adding the line `const this = "that";` in some file. Ran the linter: no error. The pretty gave back an error and didnt alter the code. So will add a task for that.

jsaalfeld:
As soon as we have #4031 in place, we could switch to use that dockerfile to run the tests. That should significantly speed things up, since we don't need a whole re-build of dependecies (and so on) for every build-task.

PR #4047, opened 2018-11-30
naming tasks and add format check for client

jsaalfeld:
fixing #4045


jsaalfeld:
Sure. Just made an error, therefore to open-close chaos

Issue #4064, opened 2018-12-14
Add further server test to travis


jsaalfeld:
Like the current tests steps, please add a further test (maybe called "Server: Basic-Migration-Test" or something like that) that executes:

```
python manage.py createsettings
python manage.py migrate
```

This should help prevent issues like #4063

Issue #4070, opened 2018-12-17
Additional Server Checks in travis


jsaalfeld:
To prevent errors like in #4069, we should add additional server checks in startup (like test-migration, createsettings and a startup).

jsaalfeld:
duplicate of #4064

Issue #4069, opened 2018-12-17
Serializer Error on startup


jsaalfeld:
With the commit of #3985 the openslides server-start routine was broken.

When you try to startup the server, the following error will greet you:

```
% python manage.py runserver
Unhandled exception in thread started by <function check_errors.<locals>.wrapper at 0x7f2d793c79d8>
Traceback (most recent call last):
  File "/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/.virtualenv/lib/python3.7/site-packages/django/utils/autoreload.py", line 225, in wrapper
    fn(*args, **kwargs)
  File "/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/.virtualenv/lib/python3.7/site-packages/django/core/management/commands/runserver.py", line 109, in inner_run
    autoreload.raise_last_exception()
  File "/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/.virtualenv/lib/python3.7/site-packages/django/utils/autoreload.py", line 248, in raise_last_exception
    raise _exception[1]
  File "/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/.virtualenv/lib/python3.7/site-packages/django/core/management/__init__.py", line 337, in execute
    autoreload.check_errors(django.setup)()
  File "/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/.virtualenv/lib/python3.7/site-packages/django/utils/autoreload.py", line 225, in wrapper
    fn(*args, **kwargs)
  File "/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/.virtualenv/lib/python3.7/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/.virtualenv/lib/python3.7/site-packages/django/apps/registry.py", line 120, in populate
    app_config.ready()
  File "/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/openslides/core/apps.py", line 90, in ready
    element_cache.ensure_cache()
  File "/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/openslides/utils/cache.py", line 102, in ensure_cache
    for element in cachable.get_elements():
  File "/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/openslides/utils/models.py", line 122, in get_elements
[...]

PR #4071, opened 2018-12-17
travis check for createsettings and migration

jsaalfeld:
In reaction of #4069 the ticket #4070 was created.

Currently this version can sanely check if `createsettings` and `migrations` wents through. The `runserver` command is more or less for readability and for checking the logs, since it always returns `0`. A flag for the `runserver` command would be nice, that returns `-1` when it hits failure state.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,25 @@ matrix:
         - python -m mypy openslides/ tests/
         - python -W ignore -m pytest --cov --cov-fail-under=70
 
+    - language: python
+      name: "Server: Tests Startup Routine Python 3.7"
+      cache:
+        pip: true
+      python:
+        - "3.7"
+      install:
+        - python --version
+        - pip install --upgrade setuptools pip
+        - pip install --upgrade --requirement requirements/development.txt
+        - pip install --upgrade .[big_mode]
+        - pip freeze
+      script:
+        - set -e
+        - python manage.py createsettings
+        - python manage.py migrate
+        - python manage.py runserver --noreload & (sleep 15 && kill $(ps aux | grep 'manage.py runserver' | head -n -1 | awk '{print $2}'))
+        - set +e
+
     - language: node_js
       name: "Client: Linting"
       node_js:
--- a/DEVELOPMENT.rst
+++ b/DEVELOPMENT.rst
@@ -70,6 +70,25 @@ When debugging something email related change the email backend to console::
 
     $ python manage.py start --debug-email
 
+
+d.a Debugging the Server
+''''''''''''''''''''''''
+
+If you wish to have even further debugging, enable `django-extensions
+<https://django-extensions.readthedocs.io/>`_,
+in the ``settings.py``  by adding ``django_extensions`` to the list of
+``INSTALLED_PLLUGINS``. Make sure, you add the following lines to your
+``requirements/development.txt``::
+
+    Werkzeug
+    pyparsing
+    pydot
+    django-extensions
+
+and install them via pip. You can start the enhanced debugging-server via::
+
+    $ python manage.py runserver_plus
+
 e. Setup and start the client
 '''''''''''''''''''''''''''''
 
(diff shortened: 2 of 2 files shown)


jsaalfeld:
This branch is currently on purpose on track with the broken master, to make test the failing easier.

FinnStutzenstein:
@jsaalfeld the --noreload works. I added an Error, becuase the issue with the users did not come up every time. Please have a look, remove the error, rebase and merge.

ostcar:
I don't understand if this pull request is just a test or if you really want to get it merged.

If you wan't to get it merged, I don't understand, why there needs to be the new requirements. Before you merge this, please add me as a reviewer.

jsaalfeld:
This is _not_ a test PR.

The "need" for new requirements is for the `django_extensions` plugin. The requirements listed in the `development.txt` are for some features for that. I've documented some of them in the `README` but it is better if you take a look at them by yourself: https://django-extensions.readthedocs.io/

openslides/core/apps.py
```diff
@@ -116,6 +116,8 @@ def get_angular_constants(self):
 
         constants: Dict[str, Any] = {}
 
+        raise NotImplementedError()
```
ostcar: This seams wrong

openslides/utils/settings.py.tpl
```diff
@@ -26,6 +26,7 @@ INSTALLED_PLUGINS += (
 
 # Built-in plugins:
 #    'tests.example_data_generator',
+    'django_extensions',
```
ostcar: Can you explain why we need this?

If we need it, it shoul'd be in in openslides.global_settings and not in the template.

requirements/development.txt
```diff
@@ -6,3 +6,7 @@ pytest
 pytest-django
 pytest-asyncio
 pytest-cov
+Werkzeug
+pyparsing
+pydot
+django-extensions
```
ostcar: Please explain why we need all of this new requirements

openslides/utils/settings.py.tpl
jsaalfeld: The explanation to this is in my comment before.

requirements/development.txt
jsaalfeld: The explanation to this is in my comment before.

openslides/core/apps.py
jsaalfeld: ping @FinnStutzenstein

FinnStutzenstein: > I added an Error, becuase the issue with the users did not come up every time. Please have a look, remove the error, rebase and merge.
So just remove. This was to test this pr.

jsaalfeld: Since I don't have the branch locally anymore, can you do this quickly and alter your commits on this PR?

FinnStutzenstein: I can do it but not quickly since I'm on my mobile here ;)

ostcar:
I still don't see why django_extensions has to be included into the repo.

Of cause, you can install any packages you want on your virtualenv and you can add any django app into your personal settings. But where is the need to install it for any developer?

And as I wrote before, your code does not work in production. When you add 'django_extensions' into the settings template, then it will be added to any settings created with `python manage.py createsettings`, this means on any new installation. But in production, only requirements/production.txt is installed. So you add a plugin as default, that is not installed as default.

jsaalfeld:
This fixes #4064

FinnStutzenstein:
@ostcar I added django_extensions as an opt-in, like the create_example_data.

ostcar:
Could you add an own requirements.txt for this? Like make/requirments.txt

ostcar:
And I still don't like this in the (productive) settings template. Even when this is commented out. This can confuse users.

Please first convince me, why any developer needs this. We can talk about this next week.

normanjaeckel:
I also prefer not to add development stuff to the productive settings.py (even if commented) if not necessary.

jsaalfeld:
Since there isn't any progress here, I will bow to the request and just documented the debugging, instead of enabling the developer to use it directly.

Please re-Review @ostcar

PR #3985, opened 2018-11-05
Refactor assess_permission

ostcar:
* Remove access_permission.get_full_data()
* Remove access_permission.gcet_serializer_class()
* get serializer class from model

PR #4073, opened 2018-12-17
fix keyerror introduced by #3985

ostcar:


normanjaeckel:
Can someone give me an explanation here? Why isn't the serializer imported during runtime? Is this an early run of this part of the code. The hack looks really hacky ...

ostcar:
It is. The real fix would be, that apps.ready() would be called in two parts. First, all files from all apps would be imported and then the code from apps.ready() would be called. So the problem is, that core.apps.Apps.ready() calls code, that require, that the code from other apps are imported. But this other code was not imported, because other_app.apps.Apps.ready() was not called.

normanjaeckel:
So what about changing dependency order, maybe in Django's INSTALLED_APPS. First run the dependencies and then run the other apps.

PR #4080, opened 2018-12-22
Take motion configuration variables, some layout-fixes

CatoTH:
- Config variables for line length, default line numbering mode and default change recommendation mode are now evaluated.
- If no changes exist, the original view is displayed.
- Fix Dialog./.TinyMCE-Problems with the scroll bars.

### DIFF
--- a/client/src/app/site/motions/components/motion-detail-diff/motion-detail-diff.component.ts
+++ b/client/src/app/site/motions/components/motion-detail-diff/motion-detail-diff.component.ts
@@ -1,5 +1,5 @@
 import { AfterViewInit, Component, ElementRef, EventEmitter, Input, Output } from '@angular/core';
-import { ViewMotion } from '../../models/view-motion';
+import { LineNumberingMode, ViewMotion } from '../../models/view-motion';
 import { ViewUnifiedChange, ViewUnifiedChangeType } from '../../models/view-unified-change';
 import { DomSanitizer, SafeHtml, Title } from '@angular/platform-browser';
 import { MotionRepositoryService } from '../../services/motion-repository.service';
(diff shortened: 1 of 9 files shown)


CatoTH:
@emanuelschuetze the two points should be solved now.

client/src/app/site/motions/services/motion-repository.service.ts
```diff
@@ -43,6 +44,16 @@ import { Mediafile } from 'app/shared/models/mediafiles/mediafile';
     providedIn: 'root'
 })
 export class MotionRepositoryService extends BaseRepository<ViewMotion, Motion> {
+
+    // The line length; comes from the config variable motions_line_length
+    private lineLength = 80;
```
emanuelschuetze: The default value of [motions_line_length](https://github.com/OpenSlides/OpenSlides/blob/master/openslides/motions/config_variables.py#L84) is 90! 
@CatoTH Should we change it here? Maybe a better way: you can get the default value of the config instead hard coding?

client/src/app/site/motions/components/motion-detail-diff/motion-detail-diff.component.ts
```diff
@@ -1,5 +1,5 @@
 import { AfterViewInit, Component, ElementRef, EventEmitter, Input, Output } from '@angular/core';
-import { ViewMotion } from '../../models/view-motion';
+import {LineNumberingMode, ViewMotion} from '../../models/view-motion';
```
emanuelschuetze: Add whitespace

client/src/app/site/motions/services/motion-repository.service.ts
```diff
@@ -28,6 +28,7 @@ import { ViewMotionAmendedParagraph } from '../models/view-motion-amended-paragr
 import { CreateMotion } from '../models/create-motion';
 import { MotionBlock } from 'app/shared/models/motions/motion-block';
 import { Mediafile } from 'app/shared/models/mediafiles/mediafile';
+import {ConfigService} from "../../../core/services/config.service";
```
emanuelschuetze: Add whitespace

CatoTH:
Ok, done,

PR #4081, opened 2018-12-25
New projector

ostcar:
This is my suggestion for the new projector system as discussed.

@normanjaeckel: I don't understand the old projector system. What is a ProjectorDefault? Why is the config field in the projector write_only?

@FinnStutzenstein: Please tell me, if you can work with it.

I created some slides and they are not finished. When @normanjaeckel has answered my questions, so I can fix the broken tests, then this should be merged. The missing slides have to be added when the client is ready.


openslides/agenda/models.py
```diff
@@ -274,17 +274,6 @@ class Meta:
     def __str__(self):
         return self.title
 
-    def delete(self, skip_autoupdate=False, *args, **kwargs):
```
normanjaeckel: This method cares that a projector element is disabled if the respective database instance is removed. I think we will still need it. (Also the same for all other apps.)

openslides/agenda/projector.py
```diff
@@ -1,54 +1,78 @@
...
+    return {'user_ids': user_ids}
 
 
-def get_projector_elements() -> Generator[Type[ProjectorElement], None, None]:
-    yield ItemListSlide
-    yield ListOfSpeakersSlide
-    yield CurrentListOfSpeakersSlide
+def register_projector_elements() -> None:
+    register_projector_element('agenda/item-list', items)
+    register_projector_element('agenda/list-of-speakers', list_of_speakers)
```
normanjaeckel: Please readd the current-list-of-speakers projector element to be more consistent with the old code. We can remove it later if necessary.

openslides/assignments/projector.py
```diff
@@ -1,42 +1,19 @@
...
-        data = None
-        try:
-            assignment = Assignment.objects.get(pk=self.config_entry.get('id'))
-        except Assignment.DoesNotExist:
-            # Assignment does not exist, so just do nothing.
-            pass
-        else:
-            data = {'agenda_item_id': assignment.agenda_item_id}
-        return data
+    Assignment slide.
```
normanjaeckel: Please add a short hint on the poll slide.

openslides/core/projector.py
```diff
@@ -1,40 +1,52 @@
...
 
-class Clock(ProjectorElement):
-    """
-    Clock on the projector.
-    """
-    name = 'core/clock'
+# Important: All functions have to be prune. This means, that thay can only
+#            access the data, that they get as argument and do not have any
+#            side effects. They are called from an async context. So they have
+#            to be fast!
```
normanjaeckel: Please readd 'core/clock' projector element.

openslides/core/serializers.py
```diff
@@ -51,12 +51,12 @@ class ProjectorSerializer(ModelSerializer):
     """
     Serializer for core.models.Projector objects.
     """
-    config = JSONSerializerField(write_only=True)
+    config = JSONSerializerField()
```
normanjaeckel: The idea was the config is a write-only field and elements is the read-only counterpart. So the server is able to validate the projector config and clients may see this without any "connection" to the projector itself.

openslides/core/websocket.py
```diff
@@ -102,3 +103,46 @@ class AutoupdateWebsocketClientMessage(BaseWebsocketClientMessage):
...
         else:
             await consumer.channel_layer.group_discard('autoupdate', consumer.channel_name)
+
+
+class ListenToProjectors(BaseWebsocketClientMessage):
+    """
+    The client tells, to which projector it listens.
+
+    Therefor it sends a list of projector ids. If it sends an empty list, it does
+    not wont to get any projector informations.
```
normanjaeckel: Typo: want

normanjaeckel: Typo: information

openslides/utils/projector.py
```diff
@@ -1,66 +1,56 @@
...
+            projector_element = projector_elements.get(projector_config['name'])
+            if projector_element is None:
+                projector_data[projector_id][uuid] = {'error': 'Projector element {} does not exist'.format(projector_config['name'])}
+            else:
+                projector_data[projector_id][uuid] = projector_element(projector_config, all_data)
+    return projector_data
 
 
-def get_all_projector_elements() -> Dict[str, ProjectorElement]:
+def get_config(all_data: AllData, key: str) -> Any:
```
normanjaeckel: Please move this function to the tests or remove it.

openslides/agenda/models.py
ostcar: You are right. I don't know what I was thinking.

But I don't like, that we override the delete method of any model with nearly the same code. I try to find another way to solve the problem.

openslides/assignments/projector.py
ostcar: done

openslides/core/projector.py
ostcar: I added a todo. I wan't to talk about this with the client devs.

openslides/core/serializers.py
ostcar: I thought the reason was, that you wanted to add other data in the elements-dict that is not part of the config-dict.

The server can still validate the config value in the http request. Am I missing something?

openslides/core/websocket.py
ostcar: done

openslides/utils/projector.py
ostcar: It will be needed in the 3.0 release. See the mail from emanuel, that internal agenda_items should be shown on the projector, when a specific config is set.

So I would like this function to stay in this pull request. We use it as soon as the slide functions are implemented.

openslides/agenda/models.py
FinnStutzenstein: This might go in the direction of https://github.com/OpenSlides/OpenSlides/pull/4061

ostcar: I think it is a different problem. The issue #4061 is about autoupdating data if something else changes. So you have a valid database but you have to inform the clients in an efficient way (you could send all data on any change. That would solve #4061).

The problem here is, that the database itself gets corrupted. When an element is deleted, the id stays in the projector config. So a projector (that was valid a moment ago) shows an error `element not found` in the next moment. Maybe this is the correct behavior.

Django solves this for ForeignKeys with the `on_delete` keyword: https://docs.djangoproject.com/en/2.1/ref/models/fields/#django.db.models.ForeignKey.on_delete

Maybe we should decide what result we want before we implement something. The possibilities are:
* remove the projector element from all projectory -> The projector would still show a message or a countdown but not the element anymore.
* show an error, or at least send an error to the client, so the client can decide what to do. The error would be `element not found` and not `element was deleted`.
* Switch to a default projector element set if an element is deleted. So a countdown or a message also disappears. 
* The same as before but also switch to an default projector element set if an element is not found.

I would like the error-message, because it is the easiest to implement and the client is free to make a decision. For example an admin client how receives this message could automatically remove the projector element or shows a pop-up to the admin, that he has to do something.

ostcar: see https://github.com/OpenSlides/OpenSlides/issues/4105

PR #4086, opened 2019-01-06
Fix pytest to a version lower 4.1 until a bug in pytest-asyncio is fixed

ostcar:
The is a bug in pytest-asyncio. Until it is fixed, we have to use a lower version of pytest.

I merge this immediately, because this issue lets travis fail and all other pull requests would be blocked.

### DIFF
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -2,7 +2,7 @@ coverage
 flake8
 isort
 mypy
-pytest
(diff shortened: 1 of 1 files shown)


normanjaeckel:
Ok.

ostcar:
I look into the requirements files very often. I will notice it every time, because it is the only requirement with a pin in the developer file.

PR #4096, opened 2019-01-09
fix login not responding to errors

MaximilianKrambach:
Solves the issue of login errors not arriving at the login mask. As the error is converted into a string early on (http.service), I'm juggling this message down to the login mask, always keeping it a string.

This hurts a bit, because I like Error objects, but I don't want to break stuff *somewhere* that may already rely on it being a string.

The easier way was to have the login mask rely on 'if there is an error arriving, something is wrong, but I just need to display it'.

@FinnStutzenstein, because I was told the errorHandling is yours.

### DIFF
--- a/client/src/app/site/login/components/login-mask/login-mask.component.ts
+++ b/client/src/app/site/login/components/login-mask/login-mask.component.ts
@@ -7,7 +7,6 @@ import { OperatorService } from 'app/core/services/operator.service';
 import { MatSnackBar, MatSnackBarRef, SimpleSnackBar } from '@angular/material';
 import { FormGroup, Validators, FormBuilder } from '@angular/forms';
 import { TranslateService } from '@ngx-translate/core';
-import { HttpErrorResponse } from '@angular/common/http';
 import { environment } from 'environments/environment';
 import { OpenSlidesService } from '../../../../core/services/openslides.service';
 import { LoginDataService } from '../../../../core/services/login-data.service';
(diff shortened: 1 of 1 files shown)


client/src/app/core/services/auth.service.ts
```diff
@@ -57,9 +57,13 @@ export class AuthService extends OpenSlidesComponent {
             username: username,
             password: password
         };
-        const response = await this.http.post<LoginResponse>(environment.urlPrefix + '/users/login/', user);
-        this.operator.user = new User(response.user);
-        return response;
+        try {
```
FinnStutzenstein: This is not necessary. Please revert.

client/src/app/core/services/http.service.ts
```diff
@@ -158,7 +158,11 @@ export class HttpService {
      * @returns A promise holding a generic
      */
     public async post<T>(path: string, data?: any, queryParams?: QueryParams, header?: HttpHeaders): Promise<T> {
-        return await this.send<T>(path, HTTPMethod.POST, data, queryParams, header);
+        try {
```
FinnStutzenstein: This is not necessary. Please revert.

client/src/app/site/login/components/login-mask/login-mask.component.ts
```diff
@@ -137,14 +139,12 @@ export class LoginMaskComponent extends BaseComponent implements OnInit, OnDestr
...
             this.router.navigate([redirect]);
-        } catch (e) {
-            if (e instanceof HttpErrorResponse) {
-                this.loginForm.setErrors({
-                    notFound: true
-                });
-                this.loginErrorMsg = e.error.detail;
-                this.inProcess = false;
-            }
+        } catch (e) { // TODO currently is an error message, not an actual error
```
FinnStutzenstein: Please remove the comment. This is per design.

MaximilianKrambach:
fixed. @FinnStutzenstein

MaximilianKrambach:
rebase done @FinnStutzenstein

PR #4099, opened 2019-01-10
fixed motion title in agenda

FinnStutzenstein:


emanuelschuetze:
> the identifier can be null, but is this really wanted? 

Yes. I can explain it later if you like.

PR #4102, opened 2019-01-11
fix vulnerable npm dependency angular-devkit

MaximilianKrambach:
Ten minutes after my last fix, a similar npm 'high risk security warning' popped up, which is auto-fixable by now.

I did not test it in depth, yet (I'm not expecting anything, as this is deep down the node dependencies' rabbit hole)

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -51,7 +51,7 @@
         "zone.js": "^0.8.26"
     },
     "devDependencies": {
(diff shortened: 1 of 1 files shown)


jsaalfeld:
```
added 1410 packages from 1308 contributors and audited 46063 packages in 34.916s
found 0 vulnerabilities
```
:+1:

PR #4103, opened 2019-01-11
Client projector ui (WIP)

FinnStutzenstein:
This is indipendent to the serverwork. The purpose is to create some functionality in for the client to connect this to the new data structure and server changes in a second step. Style and layout is done later; this is purly functional.

What is left todo?
- Projectorbuttons
- Some kind of slide registering and a slide-service


ostcar:
The python files look

PR #4104, opened 2019-01-12
Used the same timestamp for multiple history objects.

normanjaeckel:
@FinnStutzenstein Please have a look if this is the desired behavior.


normanjaeckel:
I merge it as discussed.

PR #4132, opened 2019-01-19
adding viewPermissions

MaximilianKrambach:
- fix canFollowRecommendation
- add can_create_amendments
- reused permissions from 2.3


emanuelschuetze:
Some permission checks in template still missing:
- motion list: show menu entries only for 'can-manage' permissions: categories, motion blocks, statue, comment fields, export as csv -> osPerms should be enough to hide in template
- motion detail: show edit submitter icon button only for 'can manage' permission
- motion detail: allow changing state/recommendation only for 'can_manage_metadata' or 'can-manage' permisson
- motion detail: show whole majority line for can-manage users only
- motion detail: show motion log button for can-manage-users only

PR #4148, opened 2019-01-19
Squashed old migrations.

normanjaeckel:


normanjaeckel:
Please test migration from v2.3 with data. Please also check config (logo, fonts and thelike).

FinnStutzenstein:
This squashing can not be done at some places, where we have to seperate data and schema migrations for Postgresql. @normanjaeckel please adapt this PR, if you want to have the migrations squashed, but leave migration files seperated, if they are supposed to be seperated (indicated by the postfix `_<number>`).

normanjaeckel:
Rebased. Please test. If a PostgreSQL using instance is on v2.3 the migration to 3.0-dev should work even if I squashed older migration files with unseparated data and schema migrations, right?

emanuelschuetze:
I have no migration errors with the same 2.3.x-database on current master.

normanjaeckel:
If there are no errors it can be merged.

emanuelschuetze:
@normanjaeckel I think you misunderstand my last comment: this  PR does NOT work successfully! Can you please test it again with a database created with 2.3.x?

emanuelschuetze:
As discussed with @normanjaeckel we don't want to squash old migrations anymore.

PR #4154, opened 2019-01-20
Fixes for using pypi package

emanuelschuetze:


emanuelschuetze:
Added insecure option after discuss with @ostcar and @normanjaeckel

PR #4155, opened 2019-01-20
activate start command

ostcar:


openslides/utils/settings.py.tpl
```diff
@@ -23,9 +23,6 @@ OPENSLIDES_USER_DATA_DIR = %(openslides_user_data_dir)s
 
 INSTALLED_PLUGINS += (
 #    'plugin_module_name',
-
-# Built-in plugins:
-#    'tests.example_data_generator',
```
emanuelschuetze: @ostcar Why do you remove the built-in plugin? I want to prefere to readd this line for testing.

PR #4158, opened 2019-01-20
Show username if full_name or last_name is empty.

emanuelschuetze:
important for create-example_data


emanuelschuetze:
travis fixed

PR #4164, opened 2019-01-21
set/read state+recommendation extensions

MaximilianKrambach:
If show_state_extension_field is set, it will be displayed if the workflow allows for it. 
User with manage rights on the motion can also change this extension field.

Same for recommendation_extension.


emanuelschuetze:
rebased

emanuelschuetze:
@tsiegleauq You can review later if you like.

PR #4165, opened 2019-01-21
shorten Motion title (wip)

MaximilianKrambach:
Idea is to have a motion title stripped down to one line in the list view.

This approach is a bit ugly, but maybe we can use it.

- hard coded 'max pixels per char' is not a good idea, I guess.
- digging in the entrails of the rendered stuff is not the most elegant solution
- It does not work flawlessly in all instances the browser's size/zoom changes.

If we end up using it, long user names and other list views may want to use it, too.

(Some ninja fix: the 'favorite' star should only show if it is a favorite)


FinnStutzenstein:
Whats about this: https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

MaximilianKrambach:
Sounds a lot more elegant. Does not work inmediately due to some css magic, but I'll follow that approach instead, closing this PR

PR #4170, opened 2019-01-22
change initial focus of imports

MaximilianKrambach:
As requested by @emanuelschuetze 

Initial focus on user and agenda import is now on the text import area; to ease usability for quickly adding users by typing them in.

(click import, begin typing, typing return for next line, then tab to preview, return to start preview.)


MaximilianKrambach:
Order changed @emanuelschuetze

PR #4177, opened 2019-01-23
add a validation error for false template invitation email properties (server)

MaximilianKrambach:
The translations currently arrive messed up, and it seems nice to return some validation error feedback here.

### DIFF
--- a/openslides/users/models.py
+++ b/openslides/users/models.py
@@ -232,10 +232,20 @@ class User(RESTModelMixin, PermissionsMixin, AbstractBaseUser):
                 "password": self.default_password,
             }
         )
-        message = message.format(**message_format)
+        try:
+            message = message.format(**message_format)
+        except KeyError as err:
+            raise ValidationError(
+                {"detail": f"Invalid property {err}."}
+            )
 
         subject_format = format_dict({"event_name": config["general_event_name"]})
-        subject = subject.format(**subject_format)
+        try:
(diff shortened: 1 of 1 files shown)


FinnStutzenstein:
Ok, good to check for issues here. The reason why we got this error is another thing. I'll create an issue for the reason.

openslides/users/models.py
```diff
@@ -232,10 +232,20 @@ def __missing__(self, key):
                 "password": self.default_password,
             }
         )
-        message = message.format(**message_format)
+        try:
+            message = message.format(**message_format)
+        except KeyError as e:
```
ostcar: Please use a longer variable name then `e`. For example `err`

openslides/users/models.py
```diff
@@ -232,10 +232,20 @@ def __missing__(self, key):
...
+            raise ValidationError(
+                {"detail": f"Invalid property {e}."}
+            )
 
         subject_format = format_dict({"event_name": config["general_event_name"]})
-        subject = subject.format(**subject_format)
+        try:
+            subject = subject.format(**subject_format)
+        except KeyError as e:
```
ostcar: The same here

MaximilianKrambach:
I fixed the two short variable names and rebased it all on the current master (this time the commit is not that changing that a rebase would confuse anyone)

PR #4179, opened 2019-01-24
filter mediafiles by hidden status, and by 'pdf/not pdf'

MaximilianKrambach:
as in title.

file type is determined on upload, didn't check the reliability there.


MaximilianKrambach:
'uploader' sorting removed

PR #4181, opened 2019-01-24
cleanup

FinnStutzenstein:


client/src/app/site/agenda/models/view-speaker.ts
```diff
@@ -43,7 +43,7 @@ export class ViewSpeaker extends BaseViewModel implements Selectable {
     }
 
     public get name(): string {
-        return this.user.full_name || this.user.username;
```
emanuelschuetze: Then you get an empty string in the list of speakers if a user has no first/lastname.

Issue #4009, opened 2018-11-13
Running OpenSlides via gunicorn doens't work and produces 500


jsaalfeld:
We need to start the OpenSlides Server from our virtual environment via

```
DJANGO_SETTINGS_MODULE=settings
PYTHONPATH=/app/personal_data/var
gunicorn -w 4 -b 0.0.0.0:8000 -k uvicorn.workers.UvicornWorker openslides.asgi:application
```

The server is now reachable at `localhost:8000`.

Once you start the server normally via `npm start` from the `client`folder, you can reach the website at `http://localhost`. You'll notice an Browser error on network side on `/whoami/`

```
KeyError at /apps/users/whoami/
'core/config'

Request Method: GET
Request URL: http://localhost:8000/apps/users/whoami/
Django Version: 2.1.2
Python Executable: /home/jochen/VCS/GIT/jsaalfeld/OpenSlides/.virtualenv/bin/python3
Python Version: 3.7.1
Python Path: ['/home/jochen/.config/openslides', '/home/jochen/VCS/GIT/jsaalfeld/OpenSlides', '/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/.virtualenv/bin', '/usr/lib/python37.zip', '/usr/lib/python3.7', '/usr/lib/python3.7/lib-dynload', '/home/jochen/VCS/GIT/jsaalfeld/OpenSlides/.virtualenv/lib/python3.7/site-packages']
Server time: Tue, 13 Nov 2018 11:18:48 +0100
Installed Applications:
['openslides.core',
 'openslides.users',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.staticfiles',
 'rest_framework',
 'channels',
 'openslides.agenda',
 'openslides.topics',
 'openslides.motions',
 'openslides.assignments',
 'openslides.mediafiles']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'openslides.utils.autoupdate.AutoupdateBundleMiddleware']


Traceback:

[...]

PR #4184, opened 2019-01-24
Opens the websocket connection before closing it

ostcar:
This is a workaround for #4009

Gunicorn only fails, when the websocket connection is rejected before opening it. So this helps.

@FinnStutzenstein If we merge this the client maybe has to be changed. This is only relevant for the anonymous user when anonymous user is deactivated. Before this PR, the connection was not opened. Now, it is opened but closed immediately.

@jsaalfeld could you test, if the big mode works again with this workaround?


ostcar:
@jsaalfeld ping

PR #4188, opened 2019-01-25
agenda hierarchies

MaximilianKrambach:
Reintroduces the sorting and parent/child connections for topics, similar to motions' callListWeight'

### DIFF
--- a/client/src/app/site/agenda/components/agenda-list/agenda-list.component.ts
+++ b/client/src/app/site/agenda/components/agenda-list/agenda-list.component.ts
@@ -81,8 +81,9 @@ export class AgendaListComponent extends ListViewBaseComponent<ViewItem> impleme
     public ngOnInit(): void {
         super.setTitle('Agenda');
         this.initTable();
-        this.filterService.filter().subscribe(newAgendaItem => {
-            this.dataSource.data = newAgendaItem;
+        this.filterService.filter().subscribe(newAgendaItems => {
+            newAgendaItems.sort((a, b) => a.agendaListWeight - b.agendaListWeight);
+            this.dataSource.data = newAgendaItems;
             this.checkSelection();
         });
         this.config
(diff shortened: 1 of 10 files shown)


client/src/app/site/agenda/components/agenda-list/agenda-list.component.ts
```diff
@@ -226,4 +227,26 @@ export class AgendaListComponent extends ListViewBaseComponent<ViewItem> impleme
...
+    // dirty, not observing changes
+    public getHierarchyLevel(item: ViewItem): number {
+        if (!item.item.parent_id) {
+            return 0;
+        }
+        return 1 + this.getHierarchyLevel(this.repo.getViewModel(item.item.parent_id));
+    }
+
+    // TODO Docu
+    public getHierarchyClass(item: ViewItem): string {
```
FinnStutzenstein: Please do not use classes here, because this indention might be infinite. Use [ngStyle] and set margin and font per item.

MaximilianKrambach: Most browsers won't support infinite indentation. I'd rather have 'for or five levels defined, the rest in a 'further class'. But as said, it's still a dummy, I'll wait for design feedback

emanuelschuetze:
First tests looks good!

MaximilianKrambach:
I moved the observation of the hierarchy to the `getViewModelListObservable` of agenda items. So it stays accurate even if some parent changes level.

A viewItem subscribed on its' own (`getViewModelObservable`) won't have any data of it's hierarchy, though; I'm not sure if that's a problem we can ignore. @tsiegleauq @FinnStutzenstein

client/src/app/site/agenda/components/agenda-list/agenda-list.component.ts
MaximilianKrambach: In the end, I did as you wanted

MaximilianKrambach:
I cannot reproduce this. (hidden/internal items won't get a number from server side, the others receive the correct numbering)

emanuelschuetze:
You're right. Now, with public items it works as expected.
So please rebase.

Issue #4193, opened 2019-01-27
Remove motion_changerecommendation as RootRestElement


ostcar:
Currently the motion_changerecommendation is an own RootRestElement. This is not necessary. Each changerecommendation has always one motion it belongs to. There can not be one changerecommendation without its motion.

To fix this, the corresponding viewset has to be deleted an a rest-view to alter the changerecommendation has to be created.

emanuelschuetze:
> @emanuelschuetze Thinking of some instances with long motions (3000+ lines) and some more hundrets 
> of CRs, this seems not to be good.

The number of CRs per motion are usually limited (generally 1-10, rarely more). 
But possible are a huge amount of _amendments_ which are displayed as changes within the parent motion. But these (amendment) changes are not part of the parent motion rest element. 
So I think @ostcar's changes should work also for big amount of data.

ostcar:
I would suggest, that we merge #4194 and see what happens. We do not have to change the client or the server on any other place. So it is not a big deal to revert that change.

Then we can create a big test-database and see if this is a problem. If it is a problem, we revert it and only use ids in the model full_data. If it is not a problem, then we remove motion_changerecommendation as root-rest-model for the 3.1

normanjaeckel:
Ok.

ostcar:
This is not relevant anymore.

PR #4194, opened 2019-01-27
Projector V

ostcar:
Changes to the motion slide and other parts of the prjector system.

I changed the wording of "slide" and "element", changed the order of argument of slide functions and added an projector exception.

This PR has one big change. It adds the change_recommendation into the full_data of each motion. It does not remove the change_recommendation as own root_rest_model, so this change should not be a problem. An alternative would be to add only the ids of the change_recommendations to the motion-full-data. See also #4193

### DIFF
--- a/openslides/agenda/projector.py
+++ b/openslides/agenda/projector.py
@@ -1,7 +1,11 @@
 from collections import defaultdict
 from typing import Any, Dict, List, Tuple
 
-from ..utils.projector import AllData, register_projector_element
+from ..utils.projector import (
+    AllData,
+    ProjectorElementException,
+    register_projector_slide,
+)
 
 
 # Important: All functions have to be prune. This means, that thay can only
@@ -41,7 +45,7 @@ def get_tree(
     return get_children(children[parent_id])
 
 
-def items(element: Dict[str, Any], all_data: AllData) -> Dict[str, Any]:
+def items_slide(all_data: AllData, element: Dict[str, Any]) -> Dict[str, Any]:
     """
     Item list slide.
 
@@ -63,7 +67,9 @@ def items(element: Dict[str, Any], all_data: AllData) -> Dict[str, Any]:
     return {"items": agenda_items}
 
 
-def list_of_speakers(element: Dict[str, Any], all_data: AllData) -> Dict[str, Any]:
+def list_of_speakers_slide(
+    all_data: AllData, element: Dict[str, Any]
+) -> Dict[str, Any]:
     """
     List of speakers slide.
 
@@ -76,7 +82,7 @@ def list_of_speakers(element: Dict[str, Any], all_data: AllData) -> Dict[str, An
     try:
         item = all_data["agenda/item"][item_id]
     except KeyError:
-        return {"error": f"Item {item_id} does not exist"}
+        raise ProjectorElementException(f"Item {item_id} does not exist")
 
     user_ids = []
     for speaker in item["speakers"]:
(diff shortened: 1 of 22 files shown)


openslides/users/projector.py
```diff
@@ -16,5 +16,16 @@ def user_slide(all_data: AllData, element: Dict[str, Any]) -> Dict[str, Any]:
     return {"error": "TODO"}
 
 
+def get_user_short_name(all_data: AllData, user_id: int) -> str:
```
emanuelschuetze: please use always the full_name for submitters.

ostcar: Can you tell ne what data is in the full_name?

Is the short_name needed at any place on the projector or can I remove this function?

emanuelschuetze: You can remove "no." because we can't translate it.

openslides/users/projector.py
```diff
@@ -9,14 +9,26 @@
...
+    Returns the short name for an user_id.
+    """
+    user = all_data["users/user"][user_id]
+    name_parts: List[str] = []
+    for name_part in ("title", "first_name", "last_name"):
+        if user[name_part]:
+            name_parts.append(user[name_part])
+    if not name_part:
+        name_parts.append(user["username"])
+    return " ".join(name_parts)
```
FinnStutzenstein: Why not returning an object as before?

emanuelschuetze: As discussed with @ostcar yesterday the server should send the final user string (in full name). The client can use it directly - without building a full name string.

emanuelschuetze: Users on projector should show always as full name (e.g. submitter, speaker, candidate, user slide). So the server can do the job to make the full name ready-to-use.

openslides/users/projector.py
ostcar: I am not sure if I understand you correctly. If not, you can either tell me again what to change, or you merge this pr and fix it in another pr

openslides/core/projector.py
```diff
@@ -51,7 +47,7 @@ def message(
         return {"error": f"Message {message_id} does not exist"}
 
 
-def register_projector_elements() -> None:
-    register_projector_element("core/countdown", countdown)
-    register_projector_element("core/projector-message", message)
+def register_projector_slides() -> None:
+    register_projector_slide("core/countdown", countdown_slide)
+    register_projector_slide("core/projector-message", message_slide)
     # TODO: Add clock slide
```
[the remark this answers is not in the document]
ostcar: See #4187

openslides/agenda/projector.py
```diff
@@ -76,14 +82,14 @@ def list_of_speakers(element: Dict[str, Any], all_data: AllData) -> Dict[str, An
     try:
         item = all_data["agenda/item"][item_id]
     except KeyError:
-        return {"error": f"Item {item_id} does not exist"}
+        raise ProjectorElementException(f"Item {item_id} does not exist")
```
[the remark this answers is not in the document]
ostcar: I the wording I use, an element is one "element" on the projector. A slide is a type of an element.

So you could add two countdowns on one projector. You would have to elements of the type/slide countdown. So the countdown slide would be called two times. If one countdown is invalid, you get an exception for this element and not for the countdown slide.

openslides/users/projector.py
emanuelschuetze: Ok for me. Thx.

emanuelschuetze:
Ok for now. I'll merge it now to use it for building slides on client-side.

PR #4210, opened 2019-01-29
agenda pdf export

MaximilianKrambach:
Agenda pdf export, mostly from 2.3

not 100% sure the going through a tree is overkill now that we have the agenda hierarchy already in place, but I don't know a good way of telling 'is a parent hidden' there.

### DIFF
--- a/client/src/app/site/agenda/services/agenda-pdf.service.ts
+++ b/client/src/app/site/agenda/services/agenda-pdf.service.ts
@@ -1,0 +1,103 @@
+import { Injectable } from '@angular/core';
+
+import { TranslateService } from '@ngx-translate/core';
+
+import { TreeService, OSTreeNode } from 'app/core/services/tree.service';
+import { ViewItem } from '../models/view-item';
+
+/**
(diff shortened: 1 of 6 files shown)


client/src/app/site/agenda/services/agenda-pdf.service.ts
```diff
@@ -0,0 +1,113 @@
...
+                    {
+                        width: level * 15,
+                        text: ''
+                    },
+                    {
+                        width: 60,
+                        text: nodeItem.item.itemNumber
+                    },
+                    {
+                        text: nodeItem.item.getListTitle()
```
emanuelschuetze: getListTitle contains also the itemNumber which is already in your first column!

emanuelschuetze: Please use a bigger font size for items - like in 2.3 agenda pdf.

MaximilianKrambach:
The font size could only be roughly adapted (different font), but I think it's better now @emanuelschuetze

MaximilianKrambach:
rebased @tsiegleauq

PR #4234, opened 2019-01-31
Set use_https to true

emanuelschuetze:
used for password reset urls in emails by django

@normanjaeckel as discussed...


normanjaeckel:
See #4233

PR #4235, opened 2019-01-31
Added new flag for states to hide motions from submitters.

normanjaeckel:


normanjaeckel:
We have to fix the caching problem first, see my email @ostcar.

emanuelschuetze:
Rebased, now with caching fix #4237

emanuelschuetze:
Successfully tested. Works with:
``` 
required_permission_to_see="motions.can_manage",
hidden_for_submitter=True
```

But as discussed with @normanjaeckel we want to improve the permission check by introducing a new combined choice field which replaces both flags (required_permissions_to_see and hidden_for_submitter):
1. all users (with 'motions.can_see') can the the motion in this state
2. (1) without normal users
3. (2) without submitters
4. (3) without 'motions.can_manage_metadata'
5. only 'motions.can_manage'

emanuelschuetze:
I'll test and rebase it (to fix travis errors)

PR #4238, opened 2019-02-01
navigate to detail view after setting the history mode

MaximilianKrambach:
(if a detailView is existant)


emanuelschuetze:
Rebased

PR #4255, opened 2019-02-05
sort 'previous/next motion' by identifier in detail view

MaximilianKrambach:
using localecompare, which should be able to do advanced 'natural' sorting instead of dubious string comparision (avoiding things like 1, 10, 100, 2, 3...).
In some corner cases the sorting might change if the user changes language (I doubt this will ever occur)


emanuelschuetze:
@tsiegleauq Changes ok for you?

MaximilianKrambach:
done @tsiegleauq

PR #4265, opened 2019-02-05
Fixed login problem for non admin users.

emanuelschuetze:


openslides/core/apps.py
```diff
@@ -185,7 +185,7 @@ def required_users(element: Dict[str, Any]) -> Set[int]:
     """
     Returns all user ids that are displayed as chatters.
     """
-    return set(element["user_id"])
+    return set((element["user_id"]))
```
normanjaeckel: Are you sure that `()` is enough? It should be `(  ,)` or `[]` I think.

emanuelschuetze: It works, but I added a `, `

ostcar: I thought about this a bit more. Fore the future we should use the form `return {element["user_id"]}`

PR #4268, opened 2019-02-06
Fixes the initial loading of motions and their states

### LINKED ISSUE #4262: Load state from store only when it is initialized.
1. Open incognito window and go to the startpage of OpenSlides (local)
2. Navigate straight forward to the motions and click on an item.
This produce the following error:

`
MotionDetailComponent.html:225 ERROR TypeError: Cannot read property 'length' of null
    at Object.eval [as updateDirectives] (MotionDetailComponent.html:225)
    at Object.debugUpdateDirectives [as updateDirectives] (core.js:23813)
    at checkAndUpdateView (core.js:23209)
    at callViewAction (core.js:23450)
    at execEmbeddedViewsAction (core.js:23413)
    at checkAndUpdateView (core.js:23210)
    at callViewAction (core.js:23450)
    at execEmbeddedViewsAction (core.js:23413)
    at checkAndUpdateView (core.js:23210)
    at callViewAction (core.js:23450)
` 

This should be caused by `motion.nextStates.length > 0`.

GabrielInTheWorld:
Fixes #4262
- If the workflow is changed, then the state of the motion will be updated, too

### DIFF
--- a/client/src/app/site/motions/models/view-motion.ts
+++ b/client/src/app/site/motions/models/view-motion.ts
@@ -405,6 +405,7 @@ export class ViewMotion extends BaseProjectableModel {
     public updateWorkflow(workflow: Workflow): void {
         if (this.motion && workflow.id === this.motion.workflow_id) {
             this._workflow = workflow;
+            this._state = workflow.getStateById(this.state_id);
         }
     }
 


emanuelschuetze:
I can't reproduce the TypeError (#4262) with a fresh master checkout/database. But I see the error in an existing demo database.
@GabrielInTheWorld Can you give me a step-by-step instrustion 
OR
@tsiegleauq Can you confirm the fix? Then merge.

GabrielInTheWorld:
> @GabrielInTheWorld Can you give me a step-by-step instrustion

Yes:
1. Open an Incognito-window and navigate to the OpenSlides-site.
2. Log in.
3. Go to -> Motions and click on one item to navigate to the detailview.

This page should not correctly be displayed and in the browser-console the reason for this error is an object, that is undefined.
It could be, that you get an other error than in #4262 described, but the reason would be the same.

PR #4269, opened 2019-02-06
export personal note pdf

MaximilianKrambach:
'Comments' could use the same logic; but I'm unsure how the model works there (and most doc is incorrectly claiming to be a `category` there)


emanuelschuetze:
Please add a title 'Personal note' to pdf - like in 2.3.

MaximilianKrambach:
(just a note: personal note pdf export seems to be broken on demo.openslides)

MaximilianKrambach:
done

Issue #4266, opened 2019-02-06
Postgres migration from OpenSlides 2.3 to 3.x fails


jsaalfeld:
When migrating from OpenSlides 2.3 to the current OpenSlides version the migration of data from `sqlite` works fine, but it fails with migrations when `postgres` is chosen as a database-backend.

The sources tried had `postgres` running in version 9.6, since that is the version that was latest as 2.3 was often deployed.

It can be recreated with the data from the OpenSlides demo instance: [demo_data_2.3.zip](https://github.com/OpenSlides/OpenSlides/files/2835644/demo_data_2.3.zip)

The process is as followed (assuming database was named `openslides` and the database user is `openslides` as well) (ignore first step, if you are using the demo-data):

  1. export the database via `pg_dump -O -d openslides > dumpfile.sql` and copy it to the new system
  2. startup the new OpenSlides environment once and let migrations run to get an empty instance
  3. shut down OpenSlides, except for the `postgres` service
  4. drop the old `openslides` table with `dropdb openslides`
  5. create new database with `createdb -U openslides openslides`
  6. import the 2.3 dump with `psql -f dumpfile.sql -U openslides openslides`

The import will most likely generate some warnings for the grants, which should not matter. To finally migrate the data now, you can restart the run the migration process with `python manage.py migrate`. It will generate the following error:

```
Operations to perform:
  Apply all migrations: agenda, assignments, auth, contenttypes, core, mediafiles, motions, sessions, topics, users
Running migrations:
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
psycopg2.OperationalError: cannot ALTER TABLE "motions_motion" because it has pending trigger events


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "manage.py", line 11, in <module>
    exit(main())
  File "/app/openslides/__main__.py", line 41, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 316, in run_from_argv
[...]

PR #4272, opened 2019-02-07
seperate motion version migration operations

FinnStutzenstein:


jsaalfeld:
should this fix #4266 ?

jsaalfeld:
It does fix #4266 ! Thank you!

PR #4275, opened 2019-02-07
sort categories (+ motions) in dropdowns etc.

MaximilianKrambach:


MaximilianKrambach:
The categories should be sorted now; the code is still full of duplications and workarounds

emanuelschuetze:
Only one wrong sorting issue:
sort by _name_ in motion list does not work (the list is always sorted by prefix)

MaximilianKrambach:
sorting should be correct now

PR #4289, opened 2019-02-08
Rework the reset password component (fixes #4079)

### LINKED ISSUE #4079: set vs. reset password in user repo
When (re)setting a password different behaviours are wanted:

- Multiselect: generate random pw and set it as default pw.
- Change own pw: Query two times the new pw, change it.
- admin changes a user pw: just one query, displayes the default pw and sets only the real pw.

TODO: Verify, that the client does exactly that. Add boolean parameter to reset_password route, to automatically change the default pw to avoid two consecutive requests.

FinnStutzenstein:
There might be some followup:
@normanjaeckel We may remove the update_default_password from the reset password view and make a multi_reset_password view. During a normal reset, this field should not update. When doing a multiselect-reset-password action, this should reset.

### DIFF
--- a/client/src/app/core/repositories/users/user-repository.service.ts
+++ b/client/src/app/core/repositories/users/user-repository.service.ts
@@ -132,11 +132,15 @@ export class UserRepositoryService extends BaseRepository<ViewUser, User> {
      *
      * @param user The user to update
      * @param password The password to set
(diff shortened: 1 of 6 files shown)


emanuelschuetze:
Rebased. Ready for merge.

emanuelschuetze:
Moved question to #4293.

PR #4299, opened 2019-02-11
offer alternative sortings of list of speakers

MaximilianKrambach:
Some assemblies require their speakers to keep  a certain order. This offer an sorting option for a list of speakers, which sorts the users either by participant number, given name or last name. 

More advanced sorting, such as alternating between two or more categories of users, is not supported by this.


MaximilianKrambach:
closed because of misunderstanding what was wanted

PR #4304, opened 2019-02-11
export motion comments (pdf) and sequential numbers (pdf, csv)

MaximilianKrambach:
Functionality for the export of motion comments

edit: also handles the 'sequential number' for both pdf and csv export as dialog  option


emanuelschuetze:
needs a fix for #4326 first.

PR #4320, opened 2019-02-12
full text search for media file, some permission checks

MaximilianKrambach:
- hide filter 'hidden' for delegates
 - don't show empty menu for delegates
- fix full text filter for mediafiles (filter by file title (name given during import)


emanuelschuetze:
@tsiegleauq I'll merge it now. You can review later if you like.

PR #4322, opened 2019-02-12
direct pdf export and hidden menus for unprivileged users

MaximilianKrambach:
@emanuelschuetze  Please check if I chose the correct permission checks everywhere....


MaximilianKrambach:
Not adressed here:
> edit user profil: show only editable fields - at least (as in old client) it should be username, email, about me (-> we can check it later if the server error for saving is fixed)

> and add 'project' menu entry as in motion detail

additional things done:

- fixed an error in motion detail permission checks if no submitters are present
- fix display of empty email in user edit form

PR #4323, opened 2019-02-12
fix wrong array handling

MaximilianKrambach:
Code created invalid arrays (length: 0; item at position -1) here. Curious that this can happen at all


MaximilianKrambach:
I think it's the same as #4314

FinnStutzenstein:
Ok for now, we should do something generic there..

emanuelschuetze:
> Ok for now,

@FinnStutzenstein Does it mean that you approve this PR, I can merge it and you will do something generic later?

PR #4329, opened 2019-02-13
fix submitter updating

MaximilianKrambach:
reverts one  error from #4323


FinnStutzenstein:
Why was this necessary? The old code looks right. If you had the issue, that there were no submitters, please look at line 484. It seems, that it is wrong. Compare to the if-clause of supporters..

emanuelschuetze:
Submitters were not updated after editing user; now fixed in #4330

PR #4334, opened 2019-02-13
Topic slide

emanuelschuetze:


client/src/app/site/agenda/models/view-item.ts
```diff
@@ -137,4 +138,17 @@ export class ViewItem extends BaseViewModel {
         }
         return false;
     }
+
+    public getSlide(): ProjectorElementBuildDeskriptor {
```
FinnStutzenstein: If you want to project the content object, maybe a simple `return this.contentObject.getSlide()` is enough. Two things: Make shure, that contentObject is not null and that `isProjectable(this.contentObject)` is true.

FinnStutzenstein: Whats about the case, if one of the assertions above fails? If you return null, make sure to catch this in the projector button, etc.

client/src/app/site/agenda/models/view-item.ts
```diff
@@ -2,6 +2,7 @@ import { BaseViewModel } from '../../base/base-view-model';
 import { Item } from 'app/shared/models/agenda/item';
 import { Speaker } from 'app/shared/models/agenda/speaker';
 import { BaseAgendaViewModel, isAgendaBaseModel } from 'app/site/base/base-agenda-view-model';
+import { ProjectorElementBuildDeskriptor } from '../../base/projectable';
 
 export class ViewItem extends BaseViewModel {
```
FinnStutzenstein: `extends baseProjectableViewModel`.

emanuelschuetze:
@FinnStutzenstein All done. Please review again.

PR #4335, opened 2019-02-13
user sorting by config

MaximilianKrambach:
Should cover all drop down lists, and default sorting in the list view


MaximilianKrambach:
rebased

PR #4369, opened 2019-02-19
Added history information for motion change recommendations.

normanjaeckel:


normanjaeckel:
@emanuelschuetze Do you want to get different messages for general update and the change of the internal flag only?

emanuelschuetze:
> Do you want to get different messages for general update and the change of the internal flag only?

I need information about created / updated / deleted and (if possible) published/unpublished.

normanjaeckel:
MotionChangeRecommendation is a root rest element. So I can not easily simulate a changed motion. If you like I can add the motion id to the history information string.

normanjaeckel:
> 
> 
> > Do you want to get different messages for general update and the change of the internal flag only?
> 
> I need information about created / updated / deleted and (if possible) published/unpublished.

Extra information for published/unpublished needs some more afford. Please open an extra issue for this.

PR #4385, opened 2019-02-21
Added anchor column

GabrielInTheWorld:
Added an anchor column, so the user can open a separate tab.

prettified

Changed name of variable

### DIFF
--- a/client/src/app/site/motions/components/motion-list/motion-list.component.ts
+++ b/client/src/app/site/motions/components/motion-list/motion-list.component.ts
@@ -41,12 +41,12 @@ export class MotionListComponent extends ListViewBaseComponent<ViewMotion, Motio
     /**
      * Columns to display in table when desktop view is available
      */
-    public displayedColumnsDesktop: string[] = ['identifier', 'title', 'state'];
+    public displayedColumnsDesktop: string[] = ['identifier', 'title', 'state', 'anchor'];
 
     /**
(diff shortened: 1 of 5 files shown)


client/src/app/site/motions/components/motion-list/motion-list.component.scss
```diff
@@ -61,11 +61,33 @@
         }
     }
 
+    /** Style for the anchor column */
```
emanuelschuetze: I think you should move it to style.scss for better reuse in other tables.

client/src/app/site/motions/components/motion-list/motion-list.component.ts
```diff
@@ -41,12 +41,12 @@ export class MotionListComponent extends ListViewBaseComponent<ViewMotion, Motio
     /**
      * Columns to display in table when desktop view is available
      */
-    public displayedColumnsDesktop: string[] = ['identifier', 'title', 'state'];
+    public displayedColumnsDesktop: string[] = ['identifier', 'title', 'anchor', 'state'];
```
emanuelschuetze: Why you put 'anchor' not as last column - as in other places (e.g. users)?

GabrielInTheWorld: Oh, my mistake. I first thought, that this will resolve the order of items to render. Put it as last column, now.

PR #4389, opened 2019-02-22
current speaker chyron

FinnStutzenstein:
Moves most of the settings to the projector models to make the config smaller and let the user control these things individually per projector.


client/src/app/shared/components/projector/projector.component.ts
```diff
@@ -130,13 +134,11 @@ export class ProjectorComponent extends BaseComponent implements OnDestroy {
...
 
     // Some settings for the view from the config.
-    public enableHeaderAndFooter = true;
-    public enableTitle = true;
-    public enableLogo = true;
     public eventName;
     public eventDescription;
     public eventDate;
     public eventLocation;
+    public backgroundImage;
```
FinnStutzenstein: TODO: remove

openslides/core/models.py
```diff
@@ -82,6 +82,14 @@ class Projector(RESTModelMixin, models.Model):
     width = models.PositiveIntegerField(default=1024)
     height = models.PositiveIntegerField(default=768)
 
+    background_color = models.CharField(max_length=7, default="#ffffff")
+    header_background_color = models.CharField(max_length=7, default="#317796")
+    header_font_color = models.CharField(max_length=7, default="#f5f5f5")
+    header_h1_color = models.CharField(max_length=7, default="#317796")
+    show_header_footer = models.BooleanField(default=True)
+    show_title = models.BooleanField(default=True)
+    show_logo = models.BooleanField(default=True)
```
emanuelschuetze: @FinnStutzenstein I think about your changes and I don't like to move all projector config options to projector model.
The following options should we keep as global config options (they should be managed on the settings for all projectors):
- header_background_color
- header_font_color
- header_h1_color
Also very useful is to reset to default value which is available for config options only.

Only the following options are useful as projector model options because it is useful to manage these options separately for each projector:
- background_color
- show_header_footer
- show_title
- show_logo

Can you please change this in your pull request and updated the migration file? Thanks.

FinnStutzenstein:
I'm not sure. First, I had the exact split of configs as you said. But it felt strange, because some are here, some are there. This is more consistent and you can have the setup more individually. I can add the resetbutton, this is not the problem..

emanuelschuetze:
Rebased

openslides/core/models.py
emanuelschuetze: As discussed with @normanjaeckel and @FinnStutzenstein we can merge this

openslides/agenda/projector.py
```diff
@@ -193,12 +178,69 @@ def current_list_of_speakers_slide(
...
+    """
+    reference_projector = get_reference_projector(all_data, projector_id)
+    item_id = get_current_item_id_for_projector(all_data, reference_projector)
     if item_id is None:  # no element found
         return {}
 
     return get_list_of_speakers_slide_data(all_data, item_id)
 
 
+def current_speaker_chyron_slide(
```
normanjaeckel: That is "chyron"?

FinnStutzenstein: A "Bauchbinde". The thing on e.g. news tv where the speaker and news are shown. I'm not responsible for this name ;)

PR #4394, opened 2019-02-25
increasing default upload size

jsaalfeld:
The standard upload size is 2.5mb, which results in a 413, even though we normally allow upload sizes of up to 100mb in our standard-setup.

reference: https://docs.djangoproject.com/en/2.1/ref/settings/#data-upload-max-memory-size

### DIFF
--- a/openslides/utils/settings.py.tpl
+++ b/openslides/utils/settings.py.tpl
@@ -48,6 +48,9 @@ EMAIL_PORT = 587
 EMAIL_HOST_USER = ''
 EMAIL_HOST_PASSWORD = ''
 
+# Increasing Upload size to 100mb (default is 2.5mb)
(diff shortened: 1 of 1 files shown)


FinnStutzenstein:
Just as an info: when this limit was introduced? I remember uploading iso files of 5 GB in OS2.3 to test the upload and this was never a problem..

jsaalfeld:
Apperently since [Django 1.10](https://docs.djangoproject.com/en/1.10/ref/settings/#data-upload-max-memory-size). Before that, there was [no hard limit](https://docs.djangoproject.com/en/1.9/ref/settings/#file-upload-max-memory-size), but a maximum where files had to be streamed directly to the file-system. I don't know why we didn't ran in this issue before, maybe some internal handling has changed in the last couple of months. The thing i know is: The line fixes the issue ;)

PR #4396, opened 2019-02-25
Fixed error message that viewmodel is not projectable.

emanuelschuetze:
@FinnStutzenstein Please review if this check is correct to prevent errors.

### DIFF
--- a/client/src/app/core/core-services/projector.service.ts
+++ b/client/src/app/core/core-services/projector.service.ts
@@ -269,7 +269,7 @@ export class ProjectorService {
     ): T {
         this.assertElementIsMappable(element);
         const viewModel = this.viewModelStore.get<T>(element.name, element.id);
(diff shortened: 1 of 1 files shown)


client/src/app/core/core-services/projector.service.ts
```diff
@@ -269,8 +269,10 @@ export class ProjectorService {
     ): T {
         this.assertElementIsMappable(element);
         const viewModel = this.viewModelStore.get<T>(element.name, element.id);
-        if (!isProjectable(viewModel)) {
-            console.error('The view model is not projectable', viewModel, element);
+        if (viewModel){
+            if (!isProjectable(viewModel)) {
```
FinnStutzenstein: `if (viewModel && !isProjectable(viewModel)) {`

emanuelschuetze:
Thanks. Fixed and rebased. I'll merge it now.

PR #4399, opened 2019-02-25
bulk import of users

MaximilianKrambach:
Faster import for users using the mass_import API. Tested for smaller data sets; with bigger data I run into problems as described in #4356. Error feedback in that case would be still open (server not responding at all, for quite a while).

### DIFF
--- a/client/src/app/core/repositories/users/user-repository.service.ts
+++ b/client/src/app/core/repositories/users/user-repository.service.ts
@@ -1,20 +1,21 @@
 import { Injectable } from '@angular/core';
 import { Observable, BehaviorSubject } from 'rxjs';
+import { TranslateService } from '@ngx-translate/core';
 
 import { BaseRepository } from '../base-repository';
-import { ViewUser } from 'app/site/users/models/view-user';
-import { User } from 'app/shared/models/users/user';
-import { Group } from 'app/shared/models/users/group';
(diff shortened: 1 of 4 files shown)


MaximilianKrambach:
this is most likely a timing issue (and I'm unsure yet how to handle this). Mass_import gives no success feedback; so I have to look into the dataStore afterwards if the entries are there. It seems that they don't arrive 'on time'.

FinnStutzenstein:
There is an undocumented feature in the view: add "importTrackId" (the key) to each user. The value may be the index of the list or so. The server returns an array of trackIds, that were imported successfully. In the ui, you can mark imported users with this information.

I will try to check, where the bottleneck is during the import.

MaximilianKrambach:
@FinnStutzenstein  Thanks, it occurred to me this morning (you told me before, a while back), I did not yet come to implement it.

MaximilianKrambach:
I updated with the trackId feedback, results should be valid now.

Still not happy with the performance. On initial parsing, we lose 90%ish of the time to some cdK table render events; on import we don't lose that much time, but a feedback 'please wait, doing import' would be nice.

MaximilianKrambach:
The performance should be improved a bit now:

- only send 100 requests at once (avoiding timeouts)
- trying to make the initial table rendering less painful (a bit of cargo cult magic here, could probably be improved; paginating did not work as I thought it would).

PR #4414, opened 2019-02-27
marking and repairing translations

MaximilianKrambach:
Some translatable strings that were never done or lost their functionality.

@emanuelschuetze My system became slow, I'm not sure if it is something on the system or if my changes somehow affected performance.


MaximilianKrambach:
Sorry, that was another push.

The check for validity of motions on csv import (has a title and a text) somehow was gone. Server keeps complaining (if there is no text or title)

emanuelschuetze:
Can't see any performance issues in this branch...

PR #4421, opened 2019-02-27
fix identifier setting on mass setstate

MaximilianKrambach:
Problem: Identifier was not set and updated correctly for each updated motion in complex workflows.

(only line 673 is needed. Others are to reduce overhead, unsure if it really does this)


emanuelschuetze:
@normanjaeckel I've to merge it today. Please review later!

normanjaeckel:
I prefere a pretty modified solution. See #4448.

PR #4427, opened 2019-02-28
motion projection mode is optional and sensitive for the config

FinnStutzenstein:
@CatoTH Can you plug-in the logic required by `getSlide()`?


FinnStutzenstein:
@CatoTH @emanuelschuetze I added CRs as a dependency to motions -> in view motions one can now access CRs (Tobias: is this helpful for you in some other cases?)

CatoTH:
@FinnStutzenstein Would the same be possible with the amendments to a motion? In that case, querying these in the motion detail view could be solved slightly more elegant...

emanuelschuetze:
This PR already works for amendments because an amendment has no CRs. Thr mode buttons are hide.

CatoTH:
From my understanding, a motion that has amendments, but no CRs, should still have the buttons (contrary to a motion without none at all). For that, having the amendments in the ViewMotion would be necessary.

PR #4455, opened 2019-03-04
Disable the future-lock when updating the restircted data cache

ostcar:
Before this commit, there where two different locks when updating the restricted
data cache. A future lock, what is faster but only works in the same thread. The
other lock is in redis, it is not so fast, but also works in many threads.

The future lock was buggy, because on a second call of update_restricted_data
the same future was reused. So on the second run, the future was already done.

I don't see any way to delete. The last client would have to delete it, but there
is no way to find out which client the last one is.

### DIFF
--- a/openslides/agenda/models.py
+++ b/openslides/agenda/models.py
@@ -15,8 +15,8 @@ from openslides.utils.exceptions import OpenSlidesError
 from openslides.utils.models import RESTModelMixin
 from openslides.utils.utils import to_roman
 
-from ..utils.models import CASCADE_AND_AUTOUODATE, SET_NULL_AND_AUTOUPDATE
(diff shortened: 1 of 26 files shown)


emanuelschuetze:
Rebased.

normanjaeckel:
Sorry I do not understand why we sort all imports with one . before the imports from the higher level now (two ..).

ostcar:
The new version of isort did it. This was my first commit after the new release and as it seems I am the only one how calls isort.

I had the isort change in a separate commit, but emanuel merged it together.

If you don't like the stile, please change the config or open an issue on the isort project.

FinnStutzenstein:
Reverted with #4468..

PR #4459, opened 2019-03-05
set allow_null for integerfields, when they are not required

FinnStutzenstein:
Then `required=False`, the value might be `None`. The `IntegerField` validators doesn't allow null (must be between 1 and 3 in this case), so adding `Allow_null=True` let `None` pass throgh this validation step (https://stackoverflow.com/questions/19780731/django-rest-framework-serializer-field-required-false)

The client changes are for the anonymous user.

### DIFF
--- a/client/src/app/site/motions/services/local-permissions.service.ts
+++ b/client/src/app/site/motions/services/local-permissions.service.ts
@@ -109,6 +109,7 @@ export class LocalPermissionsService {
                         motion.state.allow_submitter_edit &&
                         motion.submitters &&
                         motion.submitters.length &&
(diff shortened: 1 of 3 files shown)


normanjaeckel:
What os about the IntegerField `agenda_parent_id`? Isn't this the same here?

FinnStutzenstein:
Interesting question. Somehow we do not get any errors fom the serializers there.. I do not understand this.. I guess we should leave this as is and see whether some issues in the future will pop up.

normanjaeckel:
Hm. Maybe the client sends data so the server does not complain about a null value.

PR #4477, opened 2019-03-08
projector scroll and contdown offset

FinnStutzenstein:
- 100px scroll (instead of 50px)
- faster scroll (500px)
- countdown is aware of header of projector


client/src/app/shared/components/slide-container/slide-container.component.ts
```diff
@@ -137,7 +138,7 @@ export class SlideContainerComponent extends BaseComponent {
     private updateScroll(): void {
         if (this.slideOptions.scrollable) {
             let value = this.scroll;
-            value *= -50;
+            value *= -100;
```
FinnStutzenstein: Please undo the swapping of "Down" and "Up" and do this here.

emanuelschuetze: I used better comments in projector detail component html.

PR #4484, opened 2019-03-08
Add amendments as dependencies in ViewMotion

FinnStutzenstein:
@CatoTH You requested to add amendments as dependencies. If this PR is helpful for you, please use this commit. You can say, what should happen with this PR...


emanuelschuetze:
I rebased this pull request and added a check that the projection dialog shows the slide option if amendments are available for the current motion.

@CatoTH Now, if you project a motion slide with amendments you get the same wrong diff which you already fixed in #4513 yesterday. Can you look a the motion slide please?

PR #4499, opened 2019-03-11
Some performance analysis (Do not merge)

FinnStutzenstein:
@emanuelschuetze `perfPrintStats()` in your devconsole prints all stats.


FinnStutzenstein:
@emanuelschuetze I'll close this as we won't put this upstream. Maybe we should consider adding this to a possible debug mode. But this can be done later.

PR #4510, opened 2019-03-15
Bugfixes: Highlighting wo CRs, better scrolling

CatoTH:
- Bugfix: Line highlighting when there is no CR
- Improvement: when highlighting a line, the line should not be scrolled below the fixed header anymore
- Issue 2 from @emanuelschuetze 's email should be fixed (amendments not being shown when opening a motion for the first time)


CatoTH:
@emanuelschuetze I couldn't really reproduce it; I suspect there was some special case around the line breaking algorithm; you don't by any chance still have this example??

PR #4518, opened 2019-03-20
Added new permission to see motions in internal state (so called auth…

normanjaeckel:
…orized users).


normanjaeckel:
@emanuelschuetze Please also check new wording. Transations are missing of course.

PR #4521, opened 2019-03-20
Refactored state access level by renaming state field to restriction.

normanjaeckel:
Please change client view and then test.


FinnStutzenstein:
Please test again, there was a migration conflict.

PR #4531, opened 2019-03-23
Fixes problems with config.key_to_id

ostcar:
Make the projector-system explicit async to call async functions

Also lock config.build_key_to_id

### DIFF
--- a/openslides/agenda/projector.py
+++ b/openslides/agenda/projector.py
@@ -12,11 +12,10 @@ from ..utils.projector import (
 
 # Important: All functions have to be prune. This means, that thay can only
 #            access the data, that they get as argument and do not have any
-#            side effects. They are called from an async context. So they have
-#            to be fast!
+#            side effects.
 
 
-def get_sorted_agenda_items(all_data: AllData) -> List[Dict[str, Any]]:
+async def get_sorted_agenda_items(all_data: AllData) -> List[Dict[str, Any]]:
     """
     Returns all sorted agenda items by id first and then weight, resulting in
     ordered items, if some have the same weight.
@@ -27,7 +26,7 @@ def get_sorted_agenda_items(all_data: AllData) -> List[Dict[str, Any]]:
     )
 
 
-def get_flat_tree(all_data: AllData, parent_id: int = 0) -> List[Dict[str, Any]]:
+async def get_flat_tree(all_data: AllData, parent_id: int = 0) -> List[Dict[str, Any]]:
     """
     Build the item tree from all_data.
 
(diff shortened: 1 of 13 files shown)


ostcar:
You don't have to flush redis. Please test again.

emanuelschuetze:
I retested. Same error after server restart (I keep the clients without reload; just restart server):

```
WebSocket HANDSHAKING /ws/ [127.0.0.1:53342]
Exception inside application: 'NoneType' object is not subscriptable
  File "/home/emanuel/projects/openslides/src/OpenSlides/.venv/lib/python3.7/site-packages/channels/sessions.py", line 179, in __call__
    return await self.inner(receive, self.send)
  File "/home/emanuel/projects/openslides/src/OpenSlides/.venv/lib/python3.7/site-packages/channels/middleware.py", line 41, in coroutine_call
    await inner_instance(receive, send)
  File "/home/emanuel/projects/openslides/src/OpenSlides/.venv/lib/python3.7/site-packages/channels/consumer.py", line 59, in __call__
    [receive, self.channel_receive], self.dispatch
  File "/home/emanuel/projects/openslides/src/OpenSlides/.venv/lib/python3.7/site-packages/channels/utils.py", line 52, in await_many_dispatch
    await dispatch(result)
  File "/home/emanuel/projects/openslides/src/OpenSlides/.venv/lib/python3.7/site-packages/channels/consumer.py", line 73, in dispatch
    await handler(message)
  File "/home/emanuel/projects/openslides/src/OpenSlides/.venv/lib/python3.7/site-packages/channels/generic/websocket.py", line 175, in websocket_connect
    await self.connect()
  File "/home/emanuel/projects/openslides/src/OpenSlides/openslides/utils/consumers.py", line 33, in connect
    if not await async_anonymous_is_enabled() and not self.scope["user"]["id"]:
  File "/home/emanuel/projects/openslides/src/OpenSlides/openslides/utils/auth.py", line 162, in async_anonymous_is_enabled
    (await config.async_get_key_to_id())["general_system_enable_anonymous"],
  'NoneType' object is not subscriptable
WebSocket DISCONNECT /ws/ [127.0.0.1:53342]
```

If I reload my both client tabs the server error is gone... But we have no control about all browser connectoins and cannot ask all users to reload... Any other idea to fix it  -without flushing redis?

ostcar:
I can not understand, how this is possible. This part of the code has nothing to do with the connected clients.

How do you restart the server without reconnecting the clients? When you restart the server, all clients have to reconnect. So please explain farther what you are doing.

emanuelschuetze:
- stop server (clients gets offline message)
- start server (client reconnect)
- server shows "toggeling" log messages between HANDSHAKING/DISCONNECT  -> error
- reload client (F5) -> server shows correct HANDSHAKING without error

ostcar:
ok I found another typo... Someone should write more tests...

ostcar:
I think the mix async and sync code, that django channels provides is broken. The async-package is not thread safe. It says so many times in the docs. But django channels creates a lot of threads when calling async_to_sync. So when you use an API in a way that the documentation says, you should not use it, then you should expect errors.

In this case, it has nothing to do with it. But it is nearly impossible to be sure, there are no race conditions. So the error, above could still happen in rare cases.

emanuelschuetze:
@ostcar Can you please rebase and solve the conflict in core/config.py? Thanks.

ostcar:
done.

I errors happens, there are probably missing await or async statements

PR #4542, opened 2019-03-26
Remove ChatMessage

FinnStutzenstein:
~~TODO: Maybe remove the chat permissions. Migrations are missing~~


normanjaeckel:
Yes, why do you want to remove this feature?

FinnStutzenstein:
@normanjaeckel The client does not have a chat and will not get one. If you think, someone else uses this feature, I can leave the server untouched.

normanjaeckel:
I propose to let the server untouched. Maybe we implement the client part later.

ostcar:
If the client don't uses it and there are no plan to use in in the near future, I would suggest to remove the code from the server. We are using git, so the code is not lost.

There are other (probably better) ways to implement the chat. For example the notify-system could be used.

FinnStutzenstein:
@normanjaeckel Is this OK for you? If so, please merge.

normanjaeckel:
Hm. I nevertheless propose to let the server code there and readd the feature in v3.1, then we do not need more and more migration files.

FinnStutzenstein:
Isn't the amount of migrations irrelevant? This should not be a reason not to change code and remove unused functionality. Until this is decided, I'll just remove the client code.

ostcar:
@FinnStutzenstein If you want to remove this, maybe you can convince norman, if you implement a chat system in the client using the notify-system, so the server code is not needed anymore ;)

FinnStutzenstein:
@normanjaeckel I do not see the point about any discussion here. If you want to leave it there, ok. I guarantee you, that the client will not get any chat system on this base anymore. So it is dead code, except someone else will use this. But will someone do this? I guess not.

If I should guess, I would say, there will be a chat system added later. But it has a low priority and will use another technology as Oskar said.

normanjaeckel:
For me it's a pity that we loose this simple but (in some usecases) nice feature. Of course if no one of the client devs will have time and fun to implement it in v3.x, we do not get it. This I must accept.

(If you will add it in v3.1 or v3.2 then I strongly recommend to keep the code so that we do not have more and more migration files (at the moment the migration path is extremely slow) and that users won't loose their former data.)

I do not think that we should add some p2p chat system because it might not work properly in case of networking problems. Because of this we decided to use this simple (but nice) implementation.

Conclusion: Remove the code if you think it is out of the question that some nice and small client code will be added in v3.x.

PR #4543, opened 2019-03-27
Use redis connection pool

ostcar:
This uses a channels connection pool but it makes sure, there are no more then 100 connections at the same time.

This should lower the open sockets but it could feel slower.

You can change the allowed parallel connections in line 19.

### DIFF
--- a/openslides/utils/redis.py
+++ b/openslides/utils/redis.py
@@ -1,5 +1,7 @@
+import asyncio
 from typing import Any
 
+from channels_redis.core import ConnectionPool
 from django.conf import settings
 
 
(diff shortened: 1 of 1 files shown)


ostcar:
I tested it locally and it was even faster then the master. The performance script took only have the time.

FinnStutzenstein:
Am I right, that previously for every access a new connection to redis was established? And now these connections are pooled and limited to 100?

Or are these connection to clients?

ostcar:
Your first assumption is correct. This are two new features. The connections *to redis* are reused *and* they are limited to 100.

I am not sure it 100 is the best value. If you could test other values in an live environment, it would be nice.

PR #4548, opened 2019-03-29
decode only the needed data when calculating the required users

ostcar:
Please test this on a test instance where the delegates can not see user names but can see the agenda and change speakers on the agenda

### DIFF
--- a/openslides/utils/access_permissions.py
+++ b/openslides/utils/access_permissions.py
@@ -85,16 +85,17 @@ class RequiredUsers:
         """
         user_ids: Set[int] = set()
 
-        all_full_data = await element_cache.get_all_full_data()
         for collection_string in collection_strings:
+            collection_full_data = await element_cache.get_collection_full_data(
+                collection_string
+            )
             # Get the callable for the collection_string
             get_user_ids = self.callables.get(collection_string)
-            elements = all_full_data.get(collection_string, {})
-            if not (get_user_ids and elements):
+            if not (get_user_ids and collection_full_data):
                 # if the collection_string is unknown or it has no data, do nothing
                 continue
 
-            for element in elements:
+            for element in collection_full_data.values():
                 user_ids.update(get_user_ids(element))
 
         return user_ids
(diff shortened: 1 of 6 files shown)


openslides/utils/access_permissions.py
```diff
@@ -85,16 +85,15 @@ def add_collection_string(
         """
         user_ids: Set[int] = set()
 
-        all_full_data = await element_cache.get_all_full_data()
         for collection_string in collection_strings:
+            full_data = await element_cache.get_collection_full_data(collection_string)
```
normanjaeckel: Maybe plese rename this to something like "collection_full_data".

ostcar: done

ostcar:
wait, I add some changes...

ostcar:
please test again

ostcar:
ok. One last time...

openslides/utils/cache_providers.py
```diff
@@ -204,6 +209,23 @@ def get_change_id_cache_key(self) -> str:
...
+    async def get_collection_data(
+        self, collection: str, user_id: Optional[int] = None
+    ) -> Dict[bytes, bytes]:
+        """
+        Returns all elements for a collection from the cache.
+        """
+        if user_id is None:
+            cache_key = self.get_full_data_cache_key()
+        else:
+            cache_key = self.get_restricted_data_cache_key(user_id)
```
normanjaeckel: Why do you add this second branch/case für restricted data? This pull request does not use it.

ostcar: Thats true. I added if for the future. And this is a bad thing to do.

normanjaeckel: Do you want to add the fix for the restricted data cache here directly?

ostcar: No. If you don't like the different cache keys at this place, I can remove it. But I would prefer to merge it as it is.

normanjaeckel:
@emanuelschuetze This Pull request should be tested (finally) and merged so I can do some steps upon it.

PR #4551, opened 2019-03-30
Fixed bug that motion reason is required if respective config value i…

normanjaeckel:
…s set.


normanjaeckel:
@emanuelschuetze Please test this (later) with different permissions and with and withput amendments.

PR #4559, opened 2019-04-02
Redesign offline indicator on projector

emanuelschuetze:


emanuelschuetze:
already tested in production

PR #4564, opened 2019-04-02
More flexible parsing of font sizes and margins

francisduvivier:
I noticed that I got NaN for font-size in case my text style was for example "font-size: 12px".
That is because +'12px' results in NaN
When using parseFloat instead, 12px would just become 12 instead of NaN.
parseFloat is available on all browsers according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat , so its safe to use...

### DIFF
--- a/client/src/app/core/ui-services/html-to-pdf.service.ts
+++ b/client/src/app/core/ui-services/html-to-pdf.service.ts
@@ -603,11 +603,11 @@ export class HtmlToPdfService {
                 if (styleDefinition.length === 2) {
                     switch (key) {
                         case 'padding-left': {
-                            styleObject.margin = [+value, 0, 0, 0];
+                            styleObject.margin = [parseFloat(value), 0, 0, 0];
                             break;
                         }
                         case 'font-size': {
-                            styleObject.fontSize = +value;
+                            styleObject.fontSize = parseFloat(value);
                             break;
                         }
                         case 'text-align': {


tsiegleauq:
how to reproduce the bug you found?

francisduvivier:
I reproduced it when when using the code to convert summernote generated
html-code.

I'm not really familiar with openslides, so I don't know if its
reproducible in the current context of openslides. It just looked like a
general  opportunity to improvement unless producing NaN is wanted
behaviour.

I could write a unit test if that would help.

Grts,
Francis

tsiegleauq:
we usually get the font size from a config variable passed from the server. That should just return a number as string, unless something else is wrong.
But since `parseFloat('12px')` would safely return "12", your solution seems more robust.

> I reproduced it when when using the code to convert summernote generated
html-code.

That's interesting, I suppose you came from the [HTML to PDF issue from PDFmake](https://github.com/bpampuch/pdfmake/issues/205) ?
If you happen to find a a reliable way to convert multiple nested `<ul>` `<ol>` `<li>` list's, please let me know! ;-)
That's one of my current todos, before I can (hopefully) clean everything up and trying to bring it upstream.

tsiegleauq:
@francisduvivier according to our policies, new contributors should add their credentials in the [AUTHORS](https://github.com/OpenSlides/OpenSlides/blob/master/AUTHORS) file.
Please do so and I can merge this commit.

Otherwise I would need your permission to add your changes under my name (since we are MIT licensed)

tsiegleauq:
I replaced plus with `parseInt(x, 10)`

PR #4567, opened 2019-04-03
export recommendations with extra labels

MaximilianKrambach:


client/src/app/site/motions/services/motion-csv-export.service.ts
```diff
@@ -52,8 +57,7 @@ export class MotionCsvExportService {
                 { property: 'title' },
                 {
                     label: 'recommendation',
-                    map: motion =>
-                        motion.recommendation ? this.translate.instant(motion.recommendation.recommendation_label) : ''
+                    map: motion => (motion.recommendation ? this.motionRepo.getExtendedRecommendationLabel(motion) : '')
```
emanuelschuetze: this works, but it is for the call list only. Please add your fix also for the motion list export - for extended state and recommenation label.

MaximilianKrambach:
should work now @emanuelschuetze

emanuelschuetze:
CSV works good now. But the new XLSX export needs the same fix. Maybe @tsiegleauq can help you.

MaximilianKrambach:
@tsiegleauq I think I adapted the xlsx export, too (see second commit), but I am unsure if the motionRepo is 'allowed' there.

PR #4574, opened 2019-04-06
Update django and other dependencys

ostcar:
The update to django was not a problem at all. 

See https://www.djangoproject.com/download/ for all supported django versions.

Django 2.0 is not supported anymore. We could drop django 1.11 and 2.0 before the release. What do you think?

### DIFF
--- a/requirements/big_mode.txt
+++ b/requirements/big_mode.txt
@@ -1,7 +1,7 @@
 # Requirements for Redis and PostgreSQL support
 channels-redis>=2.2,<2.4
 django-redis-sessions>=0.6.1,<0.7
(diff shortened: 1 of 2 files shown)


emanuelschuetze:
> Django 2.0 is not supported anymore. We could drop django 1.11 and 2.0 before the release. 

Why you don't drop django 1.11 and 2.0 in this PR? It would be ok for me.

ostcar:
I would also say, that we do this. @normanjaeckel 

In debian is only django 1.11. But users probably need other dependencys anyway. So i think this does not matter.

ostcar:
looks good

PR #4578, opened 2019-04-08
Implements a mechanism for a fallback theme

GabrielInTheWorld:
- If the previously selected theme is not available the default OpenSlides theme will be displayed.
- Before the current selected theme was loaded, no theme is displayed.

### DIFF
--- a/client/src/app/app.component.html
+++ b/client/src/app/app.component.html
@@ -1,3 +1,4 @@
 <div class="content">
     <router-outlet></router-outlet>
+    <os-global-spinner></os-global-spinner>
 </div>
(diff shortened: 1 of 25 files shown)


emanuelschuetze:
> If the previously selected theme is not available the default OpenSlides theme will be displayed.

Successfully tested.

> Before the current selected theme was loaded, no theme is displayed.

Mmh, I don't like "white empty theme" whilte the theme is loaded. Can you try to prevent this? Maybe a  global spinner with background could help to make the loading phase nicer? Whould great if you can improve it.

PR #4580, opened 2019-04-09
Fixes error from `mediafile-list.component` when logging out

GabrielInTheWorld:
- If the user is seeing the mediafiles and wants to log out, an error occurred causing by querying an attribute from undefined object.


emanuelschuetze:
I can't reproduce the error on master. Maybe somebody can retest master and this PR.

PR #4583, opened 2019-04-09
Protect element fields from the projector model. Fixes creation of

FinnStutzenstein:
projectors.

@emanuelschuetze Clients should work as before (incl. creating projectors)


normanjaeckel:
Hm. You know that you do not get autoupdate for projector model if projection defaults changed, do you?

PR #4584, opened 2019-04-09
Validates creating a new user

GabrielInTheWorld:
- If a new user is created and neither the username, first name nor last name is set, a hint is displayed and the user is still in the creating mode.
- Also prevents that the client queries a user that does not exist.

### DIFF
--- a/client/src/app/shared/validators/one-of-validator.ts
+++ b/client/src/app/shared/validators/one-of-validator.ts
@@ -1,0 +1,27 @@
+import { ValidationErrors, AbstractControl } from '@angular/forms';
+
+/**
+ * Custom validator class
+ * Several strings can be passed as `keys` of the form group.
+ * This validator ensures that at least one of the fields behind the given keys has a value.
+ */
+export class OneOfValidator {
+    /**
+     * Function to confirm that at least one of the given form controls has a valid value.
+     *
+     * @param keys Undefined amount of form control names.
+     *
+     * @returns An error if all of these form controls have no valid values or null if at least one is filled.
+     */
+    public static validation(...keys: string[]): (control: AbstractControl) => ValidationErrors | null {
+        return (control: AbstractControl): ValidationErrors | null => {
+            const formControls = keys.map(key => control.get(key));
+
+            const noOneSet = formControls.every(
+                formControl => (formControl && formControl.value === '') || !formControl
+            );
+
+            return noOneSet ? { noOneSet: true } : null;
+        };
+    }
+}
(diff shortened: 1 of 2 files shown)


client/src/app/shared/validators/one-of-validator.ts
```diff
@@ -0,0 +1,30 @@
+import { ValidationErrors, AbstractControl } from '@angular/forms';
+
+/**
+ * Custom validator class
```
FinnStutzenstein: ...and what does it do?

client/src/app/shared/validators/one-of-validator.ts
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Function to confirm that at least one of the given form controls has a valid value.
+     *
+     * @param keys Undefined amount of form control names.
+     *
+     * @returns An error if all of these form controls have no valid values or null if at least one is filled.
+     */
+    public static validation(...keys: string[]): ValidationErrors | null {
+        return (control: AbstractControl): ValidationErrors | null => {
+            const formControls = [];
```
FinnStutzenstein: `... = keys.map(key => control.get(key))`. Is it guaranteed, that every of these controls exist?

client/src/app/shared/validators/one-of-validator.ts
```diff
@@ -0,0 +1,30 @@
...
+            let noOneSet = true;
+            for (const key of keys) {
+                formControls.push(control.get(key));
+            }
+
+            for (const formControl of formControls) {
```
FinnStutzenstein: `const noOneSet = formControls.every(control => control.value === '');`

client/src/app/shared/validators/one-of-validator.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export class OneOfValidator {
+    /**
+     * Function to confirm that at least one of the given form controls has a valid value.
+     *
+     * @param keys Undefined amount of form control names.
+     *
+     * @returns An error if all of these form controls have no valid values or null if at least one is filled.
+     */
+    public static validation(...keys: string[]): ValidationErrors | null {
```
FinnStutzenstein: Return type is `(control: AbstractControl) => ValidationErrors | null`. I'm not sure why the linter doesn't noticed this.

client/src/app/shared/validators/one-of-validator.ts
GabrielInTheWorld: I added a description for what it does.

Issue #4579, opened 2019-04-09
additional assignment information: elected candidates


MaximilianKrambach:
The metainformation in assignment  could use a list of persons elected during the assignment 

I imagine a finished assignment with 200 candidates; and three elected in the course of three polls. It would require a lot of searching to find these persons right now.
what do you think @emanuelschuetze ?

MaximilianKrambach:
Ah, I see that the 'elected' status is set in the first poll, too, so everybody who is elected is visible in the first poll (is this intentional?)

MaximilianKrambach:
@normanjaeckel:
If I update the is_elected status in Poll 2 of an election, the candidate becomes elected in Poll 1, too. Is this desired behavior (and the `elected` status a per-assignment information)?
The result is a bit confusing: In poll 1 you'll see that the user has not enough votes, but is marked as 'elected', although that happens in another (later) poll)

normanjaeckel:
Hm. The poll should not have any "elected" status field. The elected status is saved in the assignment related user. The "is_elected" field in the poll serializer is read only and only referes to this. So if a candidate is elected in poll 2 it only can be marked as "elected" globally. The all polls show "is_elected=True". Maybe we should change or remove this extra field.

MaximilianKrambach:
> Hm. The poll should not have any "elected" status field. The elected status is saved in the assignment related user. 

Okay, it's just a minor confusion then and working as intended. The question remains 'should we show all elected candidates in one place', which might be an UI decision.

PR #4588, opened 2019-04-10
Add list of elected candidates (#4579)

MaximilianKrambach:
Should be done/adapted with #4577 (layout).
For now it is just the functionality


FinnStutzenstein:
This can be done better with #4594 . Should we wait for this or do you need this now?

MaximilianKrambach:
this will not be implemented now

PR #4592, opened 2019-04-12
bulk set/unset favorite status

MaximilianKrambach:
Adding the `bulkSetStar` was less hassle than adapting `savePersonalNote` to multiple models. If this is bad, we can still change it.

### DIFF
--- a/client/src/app/core/ui-services/personal-note.service.ts
+++ b/client/src/app/core/ui-services/personal-note.service.ts
@@ -60,6 +60,42 @@ export class PersonalNoteService {
         }
 
         pnObject.notes[model.collectionString][model.id] = content;
+        this.savePersonalNoteObject(pnObject);
+    }
+
+    /**
+     * Sets the 'favorite' status for several models of a type in bulk
+     *
+     * @param models The model the content belongs to
+     * @param star The new 'favorite' status
+     */
+    public async bulkSetStar(models: (BaseModel | BaseViewModel)[], star: boolean): Promise<void> {
+        if (!models.length) {
+            return;
+        }
+        const pnObject: Partial<PersonalNoteObject> = this.personalNoteObject || {};
+        if (!pnObject.notes) {
+            pnObject.notes = {};
+        }
+        for (const model of models) {
+            if (!pnObject.notes[model.collectionString]) {
+                pnObject.notes[model.collectionString] = {};
+            }
+            if (pnObject.notes[model.collectionString][model.id]) {
+                pnObject.notes[model.collectionString][model.id].star = star;
+            } else {
+                pnObject.notes[model.collectionString][model.id] = { star: star, note: '' };
+            }
+        }
+        await this.savePersonalNoteObject(pnObject);
+    }
+
+    /**
+     * Sends an updated personal note to the server
+     *
+     * @param pnObject a partial (if new) or complete personal note object
+     */
+    private async savePersonalNoteObject(pnObject: Partial<PersonalNoteObject>): Promise<void> {
         if (!pnObject.id) {
             await this.http.post('rest/users/personal-note/', pnObject);
         } else {
--- a/client/src/app/site/motions/services/motion-multiselect.service.ts
+++ b/client/src/app/site/motions/services/motion-multiselect.service.ts
@@ -2,20 +2,22 @@ import { Injectable } from '@angular/core';
 
 import { TranslateService } from '@ngx-translate/core';
 
-import { ViewMotion } from '../models/view-motion';
-import { ChoiceService } from 'app/core/ui-services/choice.service';
-import { PromptService } from 'app/core/ui-services/prompt.service';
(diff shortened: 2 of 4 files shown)


client/src/app/core/ui-services/personal-note.service.ts
```diff
@@ -66,4 +66,38 @@ export class PersonalNoteService {
...
+        for (const model of models) {
+            if (!star && pnObject.notes[colString][model.id] && pnObject.notes[colString][model.id].star) {
+                pnObject.notes[colString][model.id].star = false;
+            } else if (star && !pnObject.notes[colString][model.id]) {
+                pnObject.notes[colString][model.id] = { star: true, note: '' };
+            } else if (star && !pnObject.notes[colString][model.id].star) {
+                pnObject.notes[colString][model.id].star = true;
+            }
+        }
+        if (!pnObject.id) {
```
FinnStutzenstein: Please make a `private async savePersonalNoteObject(personalNoteObject: PersonalNoteObject): Promise<void>` out of this, because it is udes multiple times

client/src/app/core/ui-services/personal-note.service.ts
```diff
@@ -66,4 +66,38 @@ export class PersonalNoteService {
...
+     */
+    public async bulkSetStar(models: BaseModel[] | BaseViewModel[], star: boolean): Promise<void> {
+        if (!models.length) {
+            throw new Error('No models selected');
+        }
+        const pnObject: Partial<PersonalNoteObject> = this.personalNoteObject || {};
+        if (!pnObject.notes) {
+            pnObject.notes = {};
+        }
+        const colString = models[0].collectionString;
```
FinnStutzenstein: Find out the collection string per model, so move this into the for-loop.

client/src/app/core/ui-services/personal-note.service.ts
```diff
@@ -66,4 +66,38 @@ export class PersonalNoteService {
...
+        if (!pnObject.notes[colString]) {
+            pnObject.notes[colString] = {};
+        }
+        for (const model of models) {
+            if (!star && pnObject.notes[colString][model.id] && pnObject.notes[colString][model.id].star) {
```
FinnStutzenstein: Can be simplified to:
```
if (pnObject.notes[colString][model.id]) {
    pnObject.notes[colString][model.id].star = star;
} else {
    pnObject.notes[colString][model.id] = { star: star, note: '' };
}
```
So if there isn't a pnObject, create the default one. If there is one, we assume it is an object, so set the star attribute to the given value.

client/src/app/core/ui-services/personal-note.service.ts
```diff
@@ -66,4 +66,38 @@ export class PersonalNoteService {
...
+
+    /**
+     * Sets the 'favorite' status for several models of a type in bulk
+     *
+     * @param models The model the content belongs to
+     * @param star The new 'favorite' status
+     */
+    public async bulkSetStar(models: BaseModel[] | BaseViewModel[], star: boolean): Promise<void> {
+        if (!models.length) {
+            throw new Error('No models selected');
```
FinnStutzenstein: do a `return` instead of an error.

MaximilianKrambach: Which does not make the error handling more consistent. A function which should return a Promise just returns. Smells bad imo, but I'll do it.

FinnStutzenstein: Is it an error, if I provide an empty array and the method does nothing? I would say it is correct: For every item provided, the star is updated. Also: Because this is an async method, `return` wil limplicitly return a promise.

client/src/app/site/motions/services/motion-multiselect.service.ts
```diff
@@ -299,4 +303,22 @@ export class MotionMultiselectService {
             }
         }
     }
+
+    /**
+     * Bulk sets/unsets the favorite status (after a confirmation dialog)
+     *
+     * @param motions The motions to set/unset the favorite status for
+     */
+    public async bulkToggleFavorite(motions: ViewMotion[]): Promise<void> {
```
FinnStutzenstein: `bulkSetFavorite`

client/src/app/core/ui-services/personal-note.service.ts
```diff
@@ -60,6 +60,42 @@ export class PersonalNoteService {
...
+        this.savePersonalNoteObject(pnObject);
+    }
+
+    /**
+     * Sets the 'favorite' status for several models of a type in bulk
+     *
+     * @param models The model the content belongs to
+     * @param star The new 'favorite' status
+     */
+    public async bulkSetStar(models: BaseModel[] | BaseViewModel[], star: boolean): Promise<void> {
```
FinnStutzenstein: `models: (BaseModel | BaseViewModel)[]`

PR #4601, opened 2019-04-15
call list improvements

MaximilianKrambach:
- add extended recommendation label
- remove motion block column, use the column for 'notes'
- don't break up tables


MaximilianKrambach:
These 'fixes' were requested by users (dropping the motion block column perhaps being controversial)

PR #4609, opened 2019-04-17
sort import/export columns for motions

MaximilianKrambach:
@tsiegleauq please have a look if this is more or less how we could do it (configurability and extending to other exports/import later)


MaximilianKrambach:
@tsiegleauq mergeable or still something to add?

PR #4620, opened 2019-04-24
copy parent's tags to a new amendment

MaximilianKrambach:
A newly created amendment should, by default, inherit the tags of their parent motion.

This will fix empty 'subheader lines' in call lists (which are really empty subheader lines of 'no tag')


MaximilianKrambach:
tags are now copied in freestyle mode, too @emanuelschuetze

PR #4625, opened 2019-04-24
Assignment/Assignment poll slide

MaximilianKrambach:
basic assignment and assignment-poll slides

### DIFF
--- a/client/src/app/slides/assignments/poll/poll-slide.component.ts
+++ b/client/src/app/slides/assignments/poll/poll-slide.component.ts
@@ -1,5 +1,6 @@
 import { Component, Input } from '@angular/core';
 
+import { AssignmentPollService, SummaryPollKey } from 'app/site/assignments/services/assignment-poll.service';
 import { BaseSlideComponent } from 'app/slides/base-slide-component';
 import { PollSlideData } from './poll-slide-data';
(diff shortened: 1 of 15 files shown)


client/src/app/slides/assignments/poll/poll-slide.component.ts
```diff
@@ -10,21 +11,87 @@ import { SlideData } from 'app/core/core-services/projector-data.service';
...
 
-    public constructor() {
+    public constructor(private pollService: AssignmentPollService) {
         super();
     }
+
+    public getVoteString(rawValue: string): string {
+        const num = parseFloat(rawValue);
+        if (!isNaN(num)) {
+            return this.pollService.getSpecialLabel(num);
```
MaximilianKrambach: it would be neat if poll.service.ts could use `toLocaleString(this.translate.currentLang)`, so we have proper decimal separators, but apparently, `TranslateService` is undefined (if put into constructor and used there) when coming from a slide.
Is there a way to get there (basically 'getting the current language setting) @FinnStutzenstein

MaximilianKrambach:
@emanuelschuetze this is ready and reviewable as a first approach (I would appreciate feedback) - contains slides for an assignment; and each poll can be projected, too.

PR #4630, opened 2019-04-25
include global abstain/no option

MaximilianKrambach:
election polls with 'one vote per candidate' should have global no/abstain options

### DIFF
--- a/client/src/app/site/assignments/services/assignment-poll.service.ts
+++ b/client/src/app/site/assignments/services/assignment-poll.service.ts
@@ -30,7 +30,7 @@ export class AssignmentPollService extends PollService {
     /**
      * list of poll keys that are numbers and can be part of a quorum calculation
      */
-    public pollValues: CalculablePollKey[] = ['votesvalid', 'votesinvalid', 'votescast'];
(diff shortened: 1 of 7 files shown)


client/src/app/site/assignments/services/assignment-poll.service.ts
```diff
@@ -83,6 +88,7 @@ export class AssignmentPollService extends PollService {
                     if (Math.min(...yes) < 0) {
                         return null;
                     } else {
+                        // TODO: Counting 'No (and possibly 'Abstain') here?
```
MaximilianKrambach: this is the discussion point I need: how do 'no to  all' and 'abstain' influence the percent base?

MaximilianKrambach: in 2.3 client, these values don't affect the calculation. I'm unsure if they should.

FinnStutzenstein: @emanuelschuetze Ping

MaximilianKrambach: e clarified in direct discussion: We will continue as in 2.3

MaximilianKrambach:
@emanuelschuetze reviewable now

PR #4638, opened 2019-04-26
Fix restriction check for motions

FinnStutzenstein:


FinnStutzenstein:
@emanuelschuetze Please confirm, that the issue with the user is solved. I couldn't reproduce it with this fix.

FinnStutzenstein:
@emanuelschuetze Ping

normanjaeckel:
Please check this again. I do not understand the code change: 

https://github.com/OpenSlides/OpenSlides/blob/master/openslides/motions/models.py#L1020

If the restriction field contains "managers_only" then I think we should not check other possibilities (is_submitter and thelike). I see that my old code was not correct because of the order of the check but the new code does not match with the definition given in the models.py.

emanuelschuetze:
> If the restriction field contains "managers_only" then I think we should not check other possibilities 
> (is_submitter and thelike).

No, we rework your definition. The check is required because we allow multiple permissions (e.g. only one restriction has to be valid. We should "managers only" rename to "managers" because it is equal to other restrictions and can be combined together; e.g. managers + is_submitters).

So for me it works correctly now.

FinnStutzenstein:
Then we should change the definition there. "At least one is needed" is easier to understand than this special case...

PR #4640, opened 2019-04-26
Poll calculations

MaximilianKrambach:
Simplifies the poll percent calculations for assignments.


MaximilianKrambach:
@emanuelschuetze for testing:
- assignment and motion polls should now always use the same calculations in all percentage displays/exports (slides; detail views; pdf export (?))

PR #4651, opened 2019-04-30
Assignment ballots (and refactoring motion ballots)

MaximilianKrambach:
edit: 

- refactored ballot pdf creation for motions, using a new `PollPdfService`
- added ballot pdf creation for assignments

as previously, we don't have a solution for large amounts of candidates (approximate limit depending on vote method: 15 or 29 candidates). Pages might generate blank with more candidates.


MaximilianKrambach:
state here: with some long names, we cannot print identical ballots with the old ballot amounts per page; thus the ballots per page are slightly reduced (and there might still be cases where that is too high). Else, it is reviewable (especially the 'refactoring ballots' part)

MaximilianKrambach:
@tsiegleauq for request, as you're the most experienced in pdf wonderland

MaximilianKrambach:
@tsiegleauq I changed ballots per page back to 8 for two-option + 'no'  ballots.

MaximilianKrambach:
I rebased it and resolved the new merge conflict.

PR #4656, opened 2019-05-02
fixing bulk sort trees from list view

MaximilianKrambach:
This repairs functionality that broke about three months ago (server side changed); and includes new options for the tree service (reattaching branches elsewhere), which might be useful in other cases as well.

@FinnStutzenstein Feedback appreciated if this looks okayish

### DIFF
--- a/client/src/app/core/ui-services/tree.service.ts
+++ b/client/src/app/core/ui-services/tree.service.ts
@@ -218,40 +218,134 @@ export class TreeService {
     }
 
     /**
-     * Reduce a list of items to nodes independent from each other in a given tree
+     * Searches a tree for a list of given items and fetches all branches that include
+     * these items and their dependants
      *
-     * @param tree the tree to traverse
-     * @param items the items to check
-     * @returns the selection of items that belong to different branches
+     * @param tree an array of OsTreeNode branches
+     * @param items the items that need to be included
+     *
+     * @returns an array of OsTreeNodes with the top-most item being included
+     * in the input list
      */
-    public getTopItemsFromTree<T extends Identifiable & Displayable>(tree: OSTreeNode<T>[], items: T[]): T[] {
-        let results: T[] = [];
+    public getBranchesFromTree<T extends Identifiable & Displayable>(
+        tree: OSTreeNode<T>[],
+        items: T[]
+    ): OSTreeNode<T>[] {
+        let results: OSTreeNode<T>[] = [];
         tree.forEach(branch => {
-            const i = this.getTopItemsFromBranch(branch, items);
-            if (i.length) {
-                results = results.concat(i);
+            if (items.some(item => item.id === branch.item.id)) {
+                results.push(branch);
+            } else if (branch.children && branch.children.length) {
+                results = results.concat(this.getBranchesFromTree(branch.children, items));
             }
         });
         return results;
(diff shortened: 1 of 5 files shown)


MaximilianKrambach:
I cannot reproduce this error. However, I had a similar error in an earlier state (not sure if I committed back then). Is your version of the branch current? @FinnStutzenstein

FinnStutzenstein:
Retestat again to verify: With commit b2a5ea46f I can reproduce this error. With the 4 motions, select the middle ones, "move in call list", select A, press "Insert after".
Did you make a fresh db to achive an equal weight of 10000 for every motion? (maybe this is related or triggers the issue..)

MaximilianKrambach:
> Retestat again to verify: With commit [b2a5ea4](https://github.com/OpenSlides/OpenSlides/commit/b2a5ea46ffad1c808ec87c129202aa81391117fa)
I can reproduce this error. With the 4 motions, select the middle ones, "move in call list", select A, press "Insert after".
> Did you make a fresh db to achive an equal weight of 10000 for every motion? (maybe this is related or triggers the issue..)

Same commit, fresh database, reloaded the client twice with ctrl-F5; created 4 motions A B C D, did as you suggested. So maybe there's something else amiss

MaximilianKrambach:
@FinnStutzenstein It was that case of 'bno parent there at all', and also wrong slicing. Now it should work

client/src/app/core/ui-services/tree.service.ts
```diff
@@ -218,43 +218,130 @@ export class TreeService {
...
+                            }
+                        } else {
+                            branch.children = [...branch.children, ...branches];
+                        }
+                    }
+                }
+            }
+            return tree;
+        } else {
+            throw new Error('This should not happen. Invalid sorting items given');
```
FinnStutzenstein: This happens, if you e.g. select every motion and select something in the (empty) dialog. Maybe let the user do not open the dialog, if there is nothing, he can select...

MaximilianKrambach: I will throw an error (resulting in a error message: 'No items left o chose from')

MaximilianKrambach:
@FinnStutzenstein I changed it; now the dialog should not even open if is nothing to select

FinnStutzenstein:
Hmm, Is this a desired behavior?:
A
--B
C

If I select C, I cannot select B as a parent.

MaximilianKrambach:
> Hmm, Is this a desired behavior?:
> A
> --B
> C
> 
> If I select C, I cannot select B as a parent.

Not desired; fixed now.

FinnStutzenstein:
If I select some items, open the dialog and click on `set as parent` or `insert` after I get `Error: This should not happen. Invalid sorting items given` or `TypeError: Cannot read property 'parent_id' of undefined`. 

Expected: These buttons should not work. Also: There may be an option (an additional radio-field) or a new button, to set the parent to `null` (-> "moving" all selected items to the root of the tree)

MaximilianKrambach:
everything fixed @FinnStutzenstein

MaximilianKrambach:
rebased

emanuelschuetze:
@FinnStutzenstein I'll merge it now. Feel free to review again later.

PR #4660, opened 2019-05-06
Logging

FinnStutzenstein:


normanjaeckel:
@ostcar Can you have a look at this please?

ostcar:
on a short look, It seams ok. Be careful when you use asyncio.Lock. It is not Thread-Safe and it django channels starts new thready every time async_to_sync is called. Since we use async consumers, I think django channel does not call async_to_sync in this case.

FinnStutzenstein:
@ostcar You are right. I'll do some more research about the actual usage of threads by asyncio and channels to check, if async requests (http as well as WebSocket -> channel message handing) uses one thread or a threadpool. Or do you have a good place to look this up?

ostcar:
I don't know. You could search for async_to_sync in the channels, channels-redis and daphne code

PR #4665, opened 2019-05-07
Fixes scrollbar in old firefox versions

GabrielInTheWorld:


client/src/app/site/site.component.ts
```diff
@@ -99,6 +99,22 @@ export class SiteComponent extends BaseComponent implements OnInit {
         //      console.log('translation of motions in the target language: ' + res);
         //  });
 
+        // TODO: Remove this, when the ESR version of Firefox >= 64.
```
FinnStutzenstein: we should not remove this. Maybe someone has an older browser version...

PR #4687, opened 2019-05-10
Fix doulbe create requests for personal notes.

FinnStutzenstein:
When this is done, the clients needs handle this error @tsiegleauq

### DIFF
--- a/openslides/users/views.py
+++ b/openslides/users/views.py
@@ -16,6 +16,7 @@ from django.contrib.sites.shortcuts import get_current_site
 from django.core import mail
 from django.core.exceptions import ValidationError as DjangoValidationError
 from django.db import transaction
+from django.db.utils import IntegrityError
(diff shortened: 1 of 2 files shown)


openslides/users/views.py
```diff
@@ -447,7 +448,14 @@ def perform_create(self, serializer):
         Customized method to inject the request.user into serializer's save
         method so that the request.user can be saved into the model field.
         """
-        serializer.save(user=self.request.user)
+        try:
+            serializer.save(user=self.request.user)
+        except IntegrityError:
```
normanjaeckel: Is it possible to to use this globla IntegrityError check but a more concret check whether the object exists in the database?

FinnStutzenstein: This can be done, but you have to carefully put this into one transaction. The integrity error should be clear enough, because the only integrity, that can be broken is the unique user.

PR #4692, opened 2019-05-10
Refresh projector data after *every* (re)connect (closes #4691)

### LINKED ISSUE #4691: Autoupdate does not work, if projector full screen page is reloaded
To reproduce open projector in full screen, then reload with F5. Then change slide. The projector does not change. Workaround is the switch back to site and then back to projector. Then autoupdate works until someone presses F5.

FinnStutzenstein:


normanjaeckel:
@tsiegleauq Urgent fix. Please review later.

PR #4702, opened 2019-05-11
Bulk views motions

normanjaeckel:
@ client devs: Please change client side and let me know, if the server side views are ok.


MaximilianKrambach:
I did the client side. We still need some method to 'delete' the categories for each (`None` as 'no category/motion block' ); as this functionality will be lost elsewise. @FinnStutzenstein (?)

PR #4704, opened 2019-05-13
Request for settings a permission in a group (closes #4698)

### LINKED ISSUE #4698: Fast clicking in permission matrix
Currently the permission array is temporary changed and send to the server, so by changing permissions from the same group fast, old changes will be overwritten, if the response is not fast enough (via autoupdate). This can be avoided by adding a special rest view taking the group, permission and a set:True/False-flag and atomicly change this one permission.

FinnStutzenstein:


emanuelschuetze:
@normanjaeckel Please review later. Thx.

PR #4707, opened 2019-05-13
fix tree sorting

FinnStutzenstein:
Assigns the weight in the preorder traversal of the tree. Now one without every object (e.g. missing motions/items) still have the correct sorting.

This might destry the sorting in the list views. If so, you just have to slick "sort" again and the server will assign correct weights.

### DIFF
--- a/client/src/app/core/repositories/agenda/item-repository.service.ts
+++ b/client/src/app/core/repositories/agenda/item-repository.service.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@angular/core';
 
-import { tap, map } from 'rxjs/operators';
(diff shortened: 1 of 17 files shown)


MaximilianKrambach:
Will #4656 be affected? @FinnStutzenstein

FinnStutzenstein:
No this should not affect your changes. This PR changes the way, the server assigns the weight. This results in a more robust ordering, meaning that elements can be missing now (e.g. a user cannot see every motion/item) and the order in the list views stays unaffected of missing entries.

emanuelschuetze:
@FinnStutzenstein Is this mergeable for you? (Do you have test it successfully as discussed today?)

Issue #4711, opened 2019-05-14
Tree sorting error


FinnStutzenstein:
- Fresh DB
- Create Topics: A B C D E F
- Go to sort and move E below D, so E is a child of D:
  A
  B
  C
  D
   -- E
  F
- Move F behind E. Expected:
  A
  B
  C
  D
  -- E
  -- F

But I get this error:
```
AgendaSortComponent.html:48 ERROR TypeError: Cannot read property 'level' of undefined
    at SortingTreeComponent.push../src/app/shared/components/sorting-tree/sorting-tree.component.ts.SortingTreeComponent.moveItemToTree (sorting-tree.component.ts:748)
    at SortingTreeComponent.push../src/app/shared/components/sorting-tree/sorting-tree.component.ts.SortingTreeComponent.onDrop (sorting-tree.component.ts:531)
    at Object.eval [as handleEvent] (SortingTreeComponent.html:13)
    at handleEvent (core.js:23107)
    at callWithDebugContext (core.js:24177)
    at Object.debugHandleEvent [as handleEvent] (core.js:23904)
    at dispatchEvent (core.js:20556)
    at core.js:22046
    at SafeSubscriber.schedulerFn [as _next] (core.js:13527)
    at SafeSubscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.SafeSubscriber.__tryOrUnsub (Subscriber.js:192)
```

PR #4712, opened 2019-05-14
fix sorting tree condition #4711

MaximilianKrambach:
Simple reordering of a condition. (just first three lines changed, rest is indentation only)


MaximilianKrambach:
there were no objections, so I'll just merge

PR #4715, opened 2019-05-14
Support other text versions in CSV

CatoTH:
Alternative to #4714 . Should support the same methods as in the PDF export for motions, although it returns plain HTML, which still is awkward in combination with CSV.
The "liberalization" of the motion-export-dialog needs to be done separately

### DIFF
--- a/client/src/app/site/motions/services/motion-csv-export.service.ts
+++ b/client/src/app/site/motions/services/motion-csv-export.service.ts
@@ -3,13 +3,17 @@ import { Injectable } from '@angular/core';
 import { TranslateService } from '@ngx-translate/core';
 
 import {
-    CsvExportService,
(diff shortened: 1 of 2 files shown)


client/src/app/site/motions/services/motion-csv-export.service.ts
```diff
@@ -51,6 +102,11 @@ export class MotionCsvExportService {
                     label: 'state',
                     map: motion => this.motionRepo.getExtendedStateLabel(motion)
                 };
+            } else if (option === 'text') {
+                return {
+                    label: 'Final text',
```
emanuelschuetze: I would prefere to use just 'Text' because it could be also the original version - which is not the final text.

CatoTH:
ok, done

PR #4718, opened 2019-05-17
more sorting tree view filters

MaximilianKrambach:
- refactored agendaSort
- call list sort - filter by category or tags
<s>- category sort - filter by tags</s>


MaximilianKrambach:
state here: ugly code, probably could be made a lot nicer and shorter; and functionality hidden away in the three-dots menu. But functionality works for call list sorting  and category sorting

- no possibility to see which filter is active
- layout pending

MaximilianKrambach:
State: Rebased, category sort WIP is removed from this branch; layout adaptions were done. Ready and reviewable from my point of view @emanuelschuetze

PR #4726, opened 2019-05-20
Adds config-variable for pdf-export

GabrielInTheWorld:
- Adds the page-size 'A5' for pdf-export

### DIFF
--- a/client/src/app/core/ui-services/pdf-document.service.ts
+++ b/client/src/app/core/ui-services/pdf-document.service.ts
@@ -165,12 +165,14 @@ export class PdfDocumentService {
         this.initFonts();
         this.imageUrls = imageUrls ? imageUrls : [];
         pdfMake.vfs = await this.initVfs();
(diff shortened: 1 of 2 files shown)


client/src/app/core/ui-services/pdf-document.service.ts
```diff
@@ -165,12 +165,14 @@ export class PdfDocumentService {
         this.initFonts();
         this.imageUrls = imageUrls ? imageUrls : [];
         pdfMake.vfs = await this.initVfs();
+        const pageSize = this.configService.instant('general_export_pdf_pagesize');
+        const defaultMargins = pageSize === 'A4' ? [75, 90, 75, 75] : [45, 45, 45, 45];
```
FinnStutzenstein: is `pageSize` is `null`, the assigned page size below will be `'A4'`, but the `defaultMargins` are `[45, 45, 45, 45]`. Please change the logic, so that in an error case (`pageSize=null`), the default margins are for A4.

PR #4728, opened 2019-05-21
Internal flag for motion blocks

FinnStutzenstein:
@tsiegleauq Please add the update diealog for a motion block to set the `internal` flag. Also enhace the `lock`-mat-icon, and maybe add one to the list view.


GabrielInTheWorld:
The `<mat-icon>` and `text` snippets will be separated to a custom component in the future, if you mean that also.

emanuelschuetze:
closed, because #4745 contains also the client side improvements for internal motion blocks.

PR #4739, opened 2019-05-24
Fix permissions for the change-own-password-view

FinnStutzenstein:


client/src/app/site/site-routing.module.ts
```diff
@@ -45,8 +45,9 @@ const routes: Routes = [
             },
             {
                 path: 'users',
-                loadChildren: './users/users.module#UsersModule',
-                data: { basePerm: 'users.can_see_name' }
+                loadChildren: './users/users.module#UsersModule'
+                // No baseperm, because change own password is ok, even if the
```
elanfer: Shouldn't this be ` 'users.can_change_password'`  ?

FinnStutzenstein: No. There is no base permission here (so everyone can load the user module). The views are protected seperately in `user-routing.module.ts` (see line 21 for `/users/password`)

PR #4743, opened 2019-05-27
Improves the SpinnerComponent

GabrielInTheWorld:
- The AppComponent listens to `stable`-state and `user` to determine whether the spinner should be hidden.

### DIFF
--- a/client/src/app/app.component.ts
+++ b/client/src/app/app.component.ts
@@ -16,6 +16,7 @@ import { PrioritizeService } from './core/core-services/prioritize.service';
 import { PingService } from './core/core-services/ping.service';
 import { SpinnerService } from './core/ui-services/spinner.service';
 import { Router } from '@angular/router';
+import { ViewUser } from './site/users/models/view-user';
 
 /**
  * Angular's global App Component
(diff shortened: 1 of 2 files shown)


client/src/app/app.component.ts
```diff
@@ -88,10 +99,24 @@ export class AppComponent {
...
-                spinnerService.setVisibility(false);
+            .subscribe(stable => {
+                // check the stable state only once.
+                if (!this.isStable && stable) {
+                    this.isStable = true;
+                    this.checkConnectionProgress();
+                }
             });
+        // subscribe, to get the user.
+        operator.getViewUserObservable().subscribe(user => {
```
FinnStutzenstein: Please be careful here! I guess, that if the anonymous is enabled, `user` will always stay `null`... Please try this out. If so, I can give you hints, where to get these information (users loaded finished) from somewhere else.

GabrielInTheWorld: You're right, the user will always be `undefined`... Can you give me any hints for this, please? Thanks!

GabrielInTheWorld: **Update**: I have pushed a second commit - how do you think about my solution for this?

GabrielInTheWorld:
@emanuelschuetze Done

PR #4750, opened 2019-05-29
Multi move motions in category

MaximilianKrambach:
if there is a multiselection, a 'move to' button will be active, allowing for the selected items to be inserted before or after a given element

- the button should get some better wording and position (jointly with the fitllers (TODO) in this view?)


MaximilianKrambach:
@emanuelschuetze for first feedback (ui button could/should be improved later with filtering)

PR #4751, opened 2019-05-29
Disable the animations

GabrielInTheWorld:
- All animations are disabled
- The file `animation.ts` is still existing, if this will be needed in the future.

### DIFF
--- a/client/src/app/app.module.ts
+++ b/client/src/app/app.module.ts
@@ -1,6 +1,6 @@
 // angular modules
 import { BrowserModule } from '@angular/platform-browser';
-import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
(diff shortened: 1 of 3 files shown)


GabrielInTheWorld:
I've tested a scenario | Login -> Motions -> First motion | with animation on and off.
The time I needed to select the first motion is quite the same in both settings.
Maybe the setting `animation off` differs from the other by one second...

emanuelschuetze:
+1 for using OpenSlides without (or with reduced) animation.
Tested: 
good: Open group matrix page -> much faster than with animation
not good: Opening motion list with >100 motions -> table is empty for 1-3 sec while list is loding (this works better with animation I think)

Maybe @tsiegleauq has an idea to find a way between animation/no-animation...

emanuelschuetze:
> not good: Opening motion list with >100 motions -> table is empty for 1-3 sec while list is loding

with production build and pagination fix #4758 it works better!

emanuelschuetze:
Closed, because #4781 is merged now.

PR #4757, opened 2019-06-03
Enhances the tile view

GabrielInTheWorld:
- Adds subscriptions to the subscriptions-array from parent to clear them, after destroying the view
- Removes filter and sort service, if the selected view is the tile-view
Adds a second variable to hold the current selected view by user
- Adds a property for color
- Increases the size of the badge


GabrielInTheWorld:
@emanuelschuetze The color property is temporarly to see the difference between the two options. This could be removed, before it will be merged into the repository.

emanuelschuetze:
@GabrielInTheWorld Do you want to review my changes? Otherwise I'll rebase and merge it tomorrow.

emanuelschuetze:
Rebased

PR #4762, opened 2019-06-04
Fixes correct rendering of images and styles at startpage

GabrielInTheWorld:
- Images are decreased, if their dimensions are greater than the one of the container.
- Styles of text will be shown, now.

### DIFF
--- a/client/src/app/site/common/components/start/start.component.html
+++ b/client/src/app/site/common/components/start/start.component.html
@@ -8,6 +8,6 @@
     <div class="app-content" translate>
         <h1>{{ welcomeTitle | translate }}</h1>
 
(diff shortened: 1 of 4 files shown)


client/src/app/site/common/components/start/start.component.html
```diff
@@ -8,6 +8,6 @@ <h2 translate>Home</h2>
     <div class="app-content" translate>
         <h1>{{ welcomeTitle | translate }}</h1>
 
-        <div [innerHTML]="welcomeText | translate"></div>
```
FinnStutzenstein: This still needs to be translated, but this can be done then getting the config value, before sanitizing it.

PR #4766, opened 2019-06-04
Implementation of exporting submitters and recommendation

GabrielInTheWorld:
The user can choose the option to export the informtion about submitters and the recommendation including the extension of motions to the 'table of contents'.

### DIFF
--- a/client/src/app/site/motions/services/motion-pdf-catalog.service.ts
+++ b/client/src/app/site/motions/services/motion-pdf-catalog.service.ts
@@ -6,7 +6,8 @@ import { ViewMotion, LineNumberingMode, ChangeRecoMode } from '../models/view-mo
 import { MotionPdfService, InfoToExport } from './motion-pdf.service';
 import { ConfigService } from 'app/core/ui-services/config.service';
 import { ViewCategory } from '../models/view-category';
-import { PdfError, PdfDocumentService, StyleType } from 'app/core/ui-services/pdf-document.service';
+import { PdfError, PdfDocumentService, StyleType, BorderType } from 'app/core/ui-services/pdf-document.service';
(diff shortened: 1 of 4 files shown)


GabrielInTheWorld:
**Currently**
CurrentViewExportMotionList

**With grey subtitle text**
ViewExportMotionListWithGreySubtitle

client/src/app/site/motions/services/motion-pdf-catalog.service.ts
```diff
@@ -116,13 +118,15 @@ export class MotionPdfCatalogService {
...
 
-        if (!sorting) {
-            sorting = this.configService.instant<string>('motions_export_category_sorting');
-        }
-        const exportCategory = sorting === 'identifier' || sorting === 'prefix';
+        // Initialize the header and the layout for border-style.
+        const header = exportSubmitterRecommendation ? this.getTocHeaderDefinition() : undefined;
+        const layout = exportSubmitterRecommendation ? BorderType.LIGHT_HORIZONTAL_LINES : BorderType.DEFAULT;
 
-        if (exportCategory && categories) {
```
emanuelschuetze: `exportCategory` was always false because sorting config name was wrong. So the toc was printed always _without_ categories. Now I removed the sorting part and pdfmake failed if categories are set and the config option to show submitters/recommendation is enabled.

tsiegleauq:
Fixed two small errors.

There was a spread operator in line 151 which converted the array to an object. A table line has to be an array.
`tocBody.push(...this.appendSubmittersAndRecommendation(motion, StyleType.CATEGORY_SECTION));
to
` tocBody.push(this.appendSubmittersAndRecommendation(motion, StyleType.CATEGORY_SECTION));`

in line 129:
`if (categories) {`
to
`if (categories && categories.length) {`

because
`!![] === true;`

PR #4767, opened 2019-06-05
Refactored OpenSlides history (HistoryInformation is not a root rest …

normanjaeckel:
…element anymore.).

@emanuelschuetze You can now try to change the client side. In first step we have only the search by element id.


FinnStutzenstein:
@emanuelschuetze Done. Please test again.

PR #4785, opened 2019-06-17
Quicklink to history (closes #4777)

### LINKED ISSUE #4777: Followup history-refactor: Quicklink to history
There should exist a quicklink from motions to the history. Followup to #4767

FinnStutzenstein:


emanuelschuetze:
Currently, the permission `core.can_see_history` does not work because user has to be superadmin to get history information from server, right?

PR #4836, opened 2019-07-10
Requires agenda.can_manage for settings agenda item information

FinnStutzenstein:
…tent object creation


FinnStutzenstein:
@emanuelschuetze Done.

PR #4838, opened 2019-07-10
Recover-strategy to detect an updated server without a reload

FinnStutzenstein:
@emanuelschuetze This should allow the client to fetch new data, if the server was restarted and updated(migrated).


emanuelschuetze:
Tavis (server tests) failed

PR #4842, opened 2019-07-11
Removes the second row for the headbar

GabrielInTheWorld:
- Instead of the second row there is a toggle-button at the bottom of the side-nav


emanuelschuetze:
add margin-top for mat-cards (e.g. topic detail)

GabrielInTheWorld:
> add margin-top for mat-cards (e.g. topic detail)

Margin added. Should be okay, now.

PR #4845, opened 2019-07-12
Motion PDF improvements

emanuelschuetze:
- show motion export dialog for all users
- print recommendation in table of content lines in italic
- only print categories of filtered motions in table of content
- add page break after each (sub)category
- reduce font size for footer date


emanuelschuetze:
@tsiegleauq Please review/refactor later.

PR #4848, opened 2019-07-12
Fix mediafiles for non superadmins

FinnStutzenstein:


FinnStutzenstein:
@emanuelschuetze Done.

PR #4860, opened 2019-07-17
Implements a custom sorting for categories

GabrielInTheWorld:
- If the motion-list is sorted by categories, the category is internal sorted by the `category_weight` of the motion
- Adds a slot to pass a custom sorting function to the `sort-filter-bar.component`

### DIFF
--- a/client/src/app/core/ui-services/base-sort-list.service.ts
+++ b/client/src/app/core/ui-services/base-sort-list.service.ts
@@ -20,6 +20,7 @@ export interface OsSortingDefinition<V> {
 export interface OsSortingOption<V> {
     property: keyof V;
     label?: string;
+    sortFn?: (itemA: V, itemB: V, ascending: boolean, intl?: Intl.Collator) => number;
 }
 
 /**
@@ -57,7 +58,7 @@ export abstract class BaseSortListService<V extends BaseViewModel> {
     /**
      * The sorting function according to current settings.
      */
-    private sortFn: (a: V, b: V) => number;
+    public sortFn?: (a: V, b: V, ascending: boolean, intl?: Intl.Collator) => number;
 
     /**
      * Set the current sorting order
@@ -202,16 +203,6 @@ export abstract class BaseSortListService<V extends BaseViewModel> {
     }
 
     /**
-     * Sorts an array of data synchronously, using the currently configured sorting
-     *
-     * @param data Array of ViewModels
-     * @returns the data, sorted with the definitions of this service
-     */
-    public sortSync(data: V[]): V[] {
-        return data.sort(this.sortFn);
-    }
-
-    /**
      * Helper function to determine false-like values (if they are not boolean)
      * @param property
      */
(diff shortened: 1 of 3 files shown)


client/src/app/core/ui-services/base-sort-list.service.ts
```diff
@@ -76,6 +77,24 @@ export abstract class BaseSortListService<V extends BaseViewModel> {
         return this.sortDefinition.sortAscending;
     }
 
+    /**
+     * Setting the sorting function
+     *
+     * @param sortFn The function to customize the sorting.
+     */
+    public set sortFn(sortFn: (itemA: V, itemB: V, ascending: boolean, intl?: Intl.Collator) => number) {
```
FinnStutzenstein: The getter and setter are not necessary, because both are public. Then you just can make the member `_sortFn` public and rename it to `sortFn`.

client/src/app/core/ui-services/base-sort-list.service.ts
```diff
@@ -57,7 +58,7 @@ export abstract class BaseSortListService<V extends BaseViewModel> {
     /**
      * The sorting function according to current settings.
      */
-    private sortFn: (a: V, b: V) => number;
+    private _sortFn: (a: V, b: V, ascending: boolean, intl?: Intl.Collator) => number;
```
FinnStutzenstein: Seems to be optional, please add `?`

client/src/app/shared/components/sort-filter-bar/sort-filter-bar.component.ts
```diff
@@ -134,6 +134,12 @@ export class SortFilterBarComponent<V extends BaseViewModel> {
     }
 
     public set sortOption(option: OsSortingOption<V>) {
+        // If the option has a custom sorting function
+        if (option.sortFn) {
```
FinnStutzenstein: `this.sortService.sortFn = option.sortFn || null;` if you want to make sure, that it is null and not `undefined`, `0`, ...

client/src/app/site/motions/services/motion-sort-list.service.ts
```diff
@@ -71,4 +71,30 @@ export class MotionSortListService extends BaseSortListService<ViewMotion> {
...
+     * Custom function to sort the categories internal by the `category_weight` of the motion.
+     *
+     * @param itemA The first item to sort
+     * @param itemB The second item to sort
+     * @param intl The localizer to compare strings
+     * @param ascending If the sorting should be in ascended or descended order
+     *
+     * @returns {number} The result of comparing.
+     */
+    private getCategorySort(itemA: ViewMotion, itemB: ViewMotion, ascending: boolean, intl: Intl.Collator): number {
```
FinnStutzenstein: Naming: `categorySortFn`.

client/src/app/shared/components/sort-filter-bar/sort-filter-bar.component.ts
FinnStutzenstein: YOu can remove the complete if-clause and replace it with this one line.

PR #4878, opened 2019-07-26
Reworked config

FinnStutzenstein:
- Move config meta data into REST
- seperate views for config groups

TODO:
- [x] Better UI @tsiegleauq
- [x] Back button
- [x] Reset to default (Server request)
- [x] Bulk save for a subpage (Server request; But the UI should be
there first)

Sean: Please also add a dot-menu with the entry "reset to default" for
the complete list and every subpage

### DIFF
--- a/client/src/app/core/repositories/config/config-repository.service.ts
+++ b/client/src/app/core/repositories/config/config-repository.service.ts
@@ -2,9 +2,9 @@ import { Injectable } from '@angular/core';
 
 import { TranslateService } from '@ngx-translate/core';
 import { BehaviorSubject, Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
 
 import { CollectionStringMapperService } from 'app/core/core-services/collection-string-mapper.service';
-import { ConstantsService } from 'app/core/core-services/constants.service';
 import { DataSendService } from 'app/core/core-services/data-send.service';
 import { DataStoreService } from 'app/core/core-services/data-store.service';
 import { HttpService } from 'app/core/core-services/http.service';
@@ -13,28 +13,8 @@ import { ViewModelStoreService } from 'app/core/core-services/view-model-store.s
 import { BaseRepository } from 'app/core/repositories/base-repository';
 import { Identifiable } from 'app/shared/models/base/identifiable';
 import { Config } from 'app/shared/models/core/config';
+import { ConfigItem } from 'app/site/config/components/config-list/config-list.component';
 import { ConfigTitleInformation, ViewConfig } from 'app/site/config/models/view-config';
-
-/**
- * Holds a single config item.
- */
-interface ConfigItem {
-    /**
-     * The key of this config variable.
-     */
-    key: string;
-
-    /**
-     * The actual view config for this variable.
-     */
-    config: ViewConfig;
-
-    /**
-     * The config variable data given in the constants. This is hold here, so the view
-     * config can be updated with this data.
-     */
-    data: any;
-}
 
 /**
  * Represents a config subgroup. It can only holds items and no further groups.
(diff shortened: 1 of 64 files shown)


client/src/app/core/repositories/config/config-repository.service.ts
```diff
@@ -13,27 +13,7 @@ import { Identifiable } from 'app/shared/models/base/identifiable';
...
-     */
-    config: ViewConfig;
-
-    /**
-     * The config variable data given in the constants. This is hold here, so the view
-     * config can be updated with this data.
-     */
-    data: any;
-}
+import { map } from 'rxjs/operators';
```
FinnStutzenstein: Order of imports

client/src/app/core/repositories/config/config-repository.service.ts
```diff
@@ -160,151 +134,66 @@ export class ConfigRepositoryService extends BaseRepository<ViewConfig, Config,
...
+            this.configsSubject.next(this.configs);
         }
     }
 
-    /**
-     * Getter for the config structure
-     */
-    public getConfigStructure(): ConfigGroup[] {
-        return this.configs;
+    public getAvailableGroupsOberservable(): Observable<string[]> {
```
FinnStutzenstein: move to getters; docs

client/src/app/core/repositories/config/config-repository.service.ts
```diff
@@ -160,151 +134,66 @@ export class ConfigRepositoryService extends BaseRepository<ViewConfig, Config,
...
             }
-        }
+        });
+
+        this.configsSubject.next(groups);
     }
 
     /**
-     * Saves a config value.
+     * Saves a config value. The server nneds the key instead of the id to fetch the config variable.
```
FinnStutzenstein: typo

client/src/app/site/config/components/config-list/config-list.component.ts
```diff
@@ -1,10 +1,12 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, OnInit, OnDestroy } from '@angular/core';
 import { Title } from '@angular/platform-browser';
 
 import { TranslateService } from '@ngx-translate/core';
 
 import { ConfigRepositoryService, ConfigGroup } from 'app/core/repositories/config/config-repository.service';
 import { BaseComponent } from 'app/base.component';
+import { ActivatedRoute } from '@angular/router';
```
FinnStutzenstein: order of imports

FinnStutzenstein:
@GabrielInTheWorld Rebased and I added `/rest/core/config/bulk_update/` which accepts `[{key: <key>, value: <value>}, ...]` in a post request. It returns `{errors: {<key>: <error message>}}`, if one or more keys gave an error during update. Please make sure, that the assosicated fields show this message and the user will not be navigated away until resolving these errors.

openslides/core/views.py
```diff
@@ -436,6 +451,40 @@ def update(self, request, *args, **kwargs):
...
+                raise ValidationError(
+                    {"detail": "Invalid input. Config value is missing."}
+                )
+
+        errors = {}
+        for entry in request.data:
+            try:
+                config[entry["key"]] = entry["value"]
+            except ConfigError as err:
+                errors[key] = str(err)
```
GabrielInTheWorld: The last error overwrites the previous errors. So only one error is returned, even if several fields contain errors.

FinnStutzenstein: Yes, it should be: errors[entry["key"]] = str(err)

FinnStutzenstein:
Two small things, than merge.

GabrielInTheWorld:
> * The tiles should be replaced with a simple (android like) list
The decision working with tiles is inspired by our designer.

> * instead of the "question mark-circle-icon" right next to the options, please use "mat-hint" (where possible) to display the description of the options
This was decided, because the hints are different number of lines long and make it unable to deal with it.

PR #4882, opened 2019-07-29
Saves the value of the local search in 'list-view-table'

GabrielInTheWorld:
- Builds a new component 'rounded-input' to have a input-field with rounded borders.
- Saves the input on every change in the local storage.
- In the `OnInit`-function this value is restored.


FinnStutzenstein:
Please rebase on current master and run `npm run cleanup`

PR #4909, opened 2019-08-12
Improves the layout of motion-block-list

GabrielInTheWorld:
- Full size
- Adds the `sort-filter-bar.component`.
- `sort-filter-bar.component` displays the current and total count of filtered items by default, now.


emanuelschuetze:
@GabrielInTheWorld There is a conflict...
@tsiegleauq Please review

PR #4918, opened 2019-08-14
[WIP] Server performance

FinnStutzenstein:
One consumer per worker accepting changes from the autoupdate system.
Accesses the cache just twice per worker (autoupdate and projector) and
not for every SiteConsumer.


normanjaeckel:
See email.

FinnStutzenstein:
I'll close this due to uncertencies with the async code and updates in the websockets and uvicorn libraries, that makes the server more robust and faster. Currently there is no need for this improvements.

PR #4920, opened 2019-08-14
Rework config defaults.

emanuelschuetze:
show help text for config fields as tooltip

### DIFF
--- a/openslides/agenda/config_variables.py
+++ b/openslides/agenda/config_variables.py
@@ -9,14 +9,30 @@ def get_config_variables():
 
     It has to be evaluated during app loading (see apps.py).
     """
+
+    # General
+
+    yield ConfigVariable(
+        name="agenda_start_event_date_time",
+        default_value=None,
+        input_type="datetimepicker",
+        label="Begin of event",
+        help_text="Input format: DD.MM.YYYY HH:MM",
+        weight=200,
+        group="Agenda",
+        subgroup="General",
+    )
+
+    # Numbering
+
     yield ConfigVariable(
         name="agenda_enable_numbering",
         label="Enable numbering for agenda items",
         input_type="boolean",
         default_value=True,
-        weight=200,
+        weight=205,
         group="Agenda",
-        subgroup="General",
+        subgroup="Numbering",
     )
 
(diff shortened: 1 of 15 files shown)


openslides/agenda/config_variables.py
```diff
@@ -55,30 +55,27 @@ def get_config_variables():
...
 
     yield ConfigVariable(
-        name="agenda_hide_internal_items_on_projector",
-        default_value=True,
-        input_type="boolean",
-        label="Hide internal items when projecting subitems",
-        weight=225,
+        name="agenda_item_creation",
+        label="Add to agenda",
+        default_value="default_no",
```
normanjaeckel: default_yes

openslides/assignments/config_variables.py
```diff
@@ -33,7 +33,7 @@ def get_config_variables():
 
     yield ConfigVariable(
         name="assignments_poll_100_percent_base",
-        default_value="YES_NO_ABSTAIN",
+        default_value="VALID",
```
normanjaeckel: This assumes that one adds values to the valid field. But in ad hoc elections this is not always true. So I prefer YES_NO_ABSTAIN.

openslides/motions/config_variables.py
```diff
@@ -75,7 +60,7 @@ def get_config_variables():
 
     yield ConfigVariable(
         name="motions_default_line_numbering",
-        default_value="none",
+        default_value="outside",
```
normanjaeckel: I don't like it, but feel free to change it.

openslides/agenda/config_variables.py
emanuelschuetze: Ok, changed.

openslides/assignments/config_variables.py
emanuelschuetze: Ok, changed.

openslides/motions/config_variables.py
emanuelschuetze: As discussed, we will change the default to 'outside'.

emanuelschuetze:
All remarks done. Rebased.

PR #4959, opened 2019-08-28
Adds config to see a max number of next speakers

GabrielInTheWorld:
- Also updates the 'list of speakers' on the `overlay-slide`.


FinnStutzenstein:
@emanuelschuetze I added a -1 to disable this feature (=show all next speakers). 0 now shows no next speakers.

We might want to do a better UI there: A checkbox for "Limit displaying next speakers". If it is checked, a number input will be accessible. We might do this in #4878

PR #4979, opened 2019-09-04
Implements the possibility to sort trees by a given property

GabrielInTheWorld:
- Separates the base sorting function to a lower service, called `base-sort.service`.
- Adds the `tree-sort.service`.

If the user wants to sort the call-list, he will be asked, if he really want to do this.
If so, the call-list is sorted by the identifier of the motions and some properties, like level, expandable and position, are reset.

I've swapped out the main sort-function to a separated file to make only a few changes and I think the `base-sort-list.service` exists only to sort real list-views.


emanuelschuetze:
@tsiegleauq Please review
@GabrielInTheWorld Please resolve conflict.

PR #5005, opened 2019-09-11
Hotfix: Fresh login does not get the real user.

GabrielInTheWorld:


FinnStutzenstein:
@GabrielInTheWorld What exactly is the problem here? This hotfix breakes e.g. the redirection from the logn page to the start page, if the user is logged in, but accidentally navigated to /login.

PR #5015, opened 2019-09-12
Move big mode requirements to production.txt

emanuelschuetze:
OpenSlides required already channels-redis and websockets in 'small production mode'. So we should use only one production file which is also easier for pypi package installations.

BTW: pip install ... on Windows 10 works now without additional installation of Visual C++


normanjaeckel:
I think the solution is to fix the utils/redis.py

See https://github.com/OpenSlides/OpenSlides/pull/4543

emanuelschuetze:
Closed. see #5018

PR #5035, opened 2019-09-17
Quick sorting the reference of motions by their identifier

GabrielInTheWorld:


emanuelschuetze:
Please remove motion-detail sorting - already done in master

PR #5036, opened 2019-09-18
Adds some more space for the 'speaker-button'

GabrielInTheWorld:
- This will fix the cutting of the badge-icon.
- This will fix also an odd 'tooltip' `record_voice_over`.


GabrielInTheWorld:
There is a problem with travis again - the 'client testing'-section fails because of a default timeout.

PR #5049, opened 2019-09-24
Improves travis a little bit

GabrielInTheWorld:
This will cache downloaded and installed dependencies. So they have not to be installed every step again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,36 +1,54 @@
 dist: xenial
 sudo: true
 
+cache:
+  - directories:
+      - $HOME/.cache/pip
+      - client/node_modules
+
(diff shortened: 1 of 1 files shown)


GabrielInTheWorld:
> It seems, that in this configuration, Travis executes the tests one after another. The total test time is then 17 minutes, in contrast to ~7-8 minutes using the old configuration

travis_old_configuration

Like shown in the screenshot in the old configuration all processes download and install every dependency once again. In the summary this needs up to ~18min.

travis_new_configuration

With the "new" configuration the dependencies are downloaded and installed only once and every process afterwards needs ~1-2 min less than before to finish.

PR #5062, opened 2019-10-01
Use Proxies for ViewModels

FinnStutzenstein:
Also reworked creation of assignments, motions and users.


FinnStutzenstein:
@emanuelschuetze I missed one important return statement... please check again.

emanuelschuetze:
Retested: I get an error if I open a projector detail view:

TypeError: Cannot read property 'length' of undefined
    at Object.eval [as updateDirectives] (ProjectorDetailComponent.html:108)

FinnStutzenstein:
@tsiegleauq @emanuelschuetze What is the status here? Did you test this? Is it fast enough in the amendment-list with caching?

PR #5068, opened 2019-10-07
Added frontend testing for login, password reset and motion detail

jsangmeister:


client/cypress/integration/login.spec.js
```diff
@@ -0,0 +1,123 @@
...
+    it('can make password visible', () => {
+        cy.get('@pw').should('have.attr', 'type', 'password');
+        cy.contains('.mat-icon', 'visibility').click();
+        cy.get('@pw').should('have.attr', 'type', 'text');
+    });
+
+    it('removes error classes', () => {
+        // test (hopefully) wrong user/pw combination
+        cy.get('@user').type(
+            '00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
```
FinnStutzenstein: you are not submitting the form, so please choose another username.

client/cypress/integration/login.spec.js
```diff
@@ -0,0 +1,123 @@
...
+            { delay: 0 }
+        );
+        cy.get('@pw').type('wrong_password', { delay: 0 });
+        cy.get('@user-ff').should('not.have.class', 'mat-form-field-invalid');
+        cy.get('@pw-ff').should('not.have.class', 'mat-form-field-invalid');
+    });
+
+    it('rejects wrong login data', () => {
```
FinnStutzenstein: Ok, one questions here: Do you have the server running while testing? If not you will get the same error message.

I strongly propose not to include the server in testing to seperate these two parts. We want to test the client not the server and sometimes we want to take control of responses (e.g. what happens, if the server sends garbage?) and the autoupdate system.

What is missing here is the capability to hook into requests and websocket-connections. I would suppose that a test must assert the correct data is send in a request and the test can specify the response, so it can be tested, if responses are handled correctly.

Another speciality from OpenSlides: The requests just initiate some changes. Responses are mostly empty and just contains a status. The new/changed/deleted data comes through an autoupdate.

So it would be nice, if we can do this in tests (example):
- I expect a request to /apps/users/login with data `{"username": "xx", "password": "yy"}`
- For this request I want to give `{"some_not_expected_data": "value"}` as a response with status code 400.
- Inject some (fake) autoupdate into the autoupdate service/websocket service.
- I expect the form to show error XXX, because of the response or the autoupdate data.

The autoupdate is more important for data-manipulating requests, e.g. to the rest-API. (not especially for login)

jsangmeister:
Closing this PR now since its not coming with 3.2. Autoupdate remains a problem, could possibly be tackeld again for 4.0

PR #5082, opened 2019-10-21
Reworks the MetaTextBlockComponent

GabrielInTheWorld:
- The component has a slightly different look.
- I can pass a text and the component automatically let edit me that text.
- The components `personal-note.component` and `motion-comments.component` are not used anymore.

- There are added two more fonts: The default font with `font-weight: 300` and a font for the developer-theme.


emanuelschuetze:
@GabrielInTheWorld @tsiegleauq Should we close this unfinished PR? Maybe we can start a new try for 4.x?

PR #5100, opened 2019-10-30
majorities in polls

FinnStutzenstein:


openslides/poll/serializers.py
```diff
@@ -9,8 +19,70 @@
...
+        new_100_percent_base = self.norm_100_percent_base_to_pollmethod(
+            validated_data["onehundred_percent_base"], validated_data["pollmethod"]
+        )
+        if new_100_percent_base is not None:
+            validated_data["onehundred_percent_base"] = new_100_percent_base
+        return super().create(validated_data)
+
+    def update(self, instance, validated_data):
+        """ Prevent from updating the motion """
+        validated_data.pop("motion", None)
```
jsangmeister: update is defined in subclasses, and "motion" is only applicable to MotionPolls

tests/integration/assignments/test_polls.py
```diff
@@ -103,8 +103,9 @@ def test_simple(self):
                 "pollmethod": "YNA",
                 "type": "named",
                 "assignment_id": self.assignment.id,
+                "onehundred_percent_base": "YN",
```
jsangmeister: I would rather use the constants instead of the string, since that's what we have them for (same for all similar lines below)

PR #5103, opened 2019-11-01
Added a view to change the reference projector in one api call

jsangmeister:


openslides/core/views.py
```diff
@@ -329,6 +329,18 @@ def set_scroll(self, request, pk):
             {"detail": "Setting scroll to {0} was successful.", "args": [request.data]}
         )
 
+    @detail_route(methods=["post"])
+    def set_default_projector(self, request, pk):
+        """
+        REST API operation to set the projector with the given pk as the new default.
```
emanuelschuetze: The wording "default" is confused! This view changed the reference projector (for the current list of speakers) - and does NOT changed the projection defaults ('default projector' sounds like a projector which contains all projection defaults). Please change doc strings and function name.

FinnStutzenstein:
Ah right. @jsangmeister please change all names with "default" to "reference".

jsangmeister:
Done in #5111

PR #5105, opened 2019-11-01
Fixed a UniqueViolation error

jsangmeister:
Added tests for user creation and try-catch for a probably race-based IntegrityError


openslides/users/views.py
```diff
@@ -98,6 +99,13 @@ def check_view_permissions(self):
             result = False
         return result
 
+    # catch IntegrityError, probably being caused by a race condition
+    def perform_create(self, serializer):
+        try:
+            super().perform_create(serializer)
+        except IntegrityError as e:
+            raise ValidationError(str(e))
```
FinnStutzenstein: `...({"detail": str(e)})`

PR #5111, opened 2019-11-05
Change wording from 'default' to 'reference' projector

jsangmeister:
As requested in #5103


FinnStutzenstein:
Please also change the word in the client.

PR #5123, opened 2019-11-08
Fixed mediafiles

FinnStutzenstein:
original_filename must be unique for files. Some little fixes in the client showing `NaN` and `undefined`.


FinnStutzenstein:
@emanuelschuetze Done

PR #5132, opened 2019-11-12
Adds the chart and dialog for analog voting

GabrielInTheWorld:
Currently everything is still in progress.

- Implements some "nice-to-have"-features.
- Implements a library for charts for evaluating it.


client/src/app/site/motions/modules/motion-poll/motion-poll-detail/motion-poll-detail.component.ts
```diff
@@ -103,6 +158,34 @@ export class MotionPollDetailComponent extends BaseViewComponent implements OnIn
         }
     }
 
+    public getResult(): { name: string; value: number }[] {
+        const model = this.poll.getModel();
+
+        return [
+            ...Object.entries(model.options[0])
```
FinnStutzenstein: Does this work for `YN` pollmethod? It seems, that the `A` value is also copied. I would suggest to be more explicit here and just give `Y`, `N` and conditionally `A` in the array.

FinnStutzenstein:
The breadcromb navigation looks promising; but there is no indicator which state is active. Please also auto-open the dialog instead of sending a stop to the server, when the user tries to stop the poll (analog polls are auto-stopped when entering votes and with #5133 the /stop/-request is forbidden for analog polls)

Also a partial update is missing, if the poll is on a state different than created.

client/src/app/site/motions/models/view-motion-poll.ts
```diff
@@ -17,6 +19,47 @@ export class ViewMotionPoll extends BaseProjectableViewModel<MotionPoll> impleme
         return this._model;
     }
 
+    public get nextStates(): string[] {
+        switch (this.state) {
+            case 1:
```
FinnStutzenstein: Please use the values from the enum here.

FinnStutzenstein:
@GabrielInTheWorld Please fix the client tests.

client/src/app/site/motions/modules/motion-detail/components/motion-poll/motion-poll-manager/motion-poll-manager.component.ts
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @returns The created respectively updated `ViewMotionPoll`.
+     */
+    public async savePoll(data: any, model?: ViewMotionPoll): Promise<ViewMotionPoll> {
+        if (!model) {
+            const id = await this.pollRepo.create(data);
+            const poll = new MotionPoll(data);
+            poll.id = id.id;
+            poll.state = 1;
+            return new ViewMotionPoll(poll);
```
FinnStutzenstein: This is bad. Do not create ViewModels by hand. You will get the correct viewModel with the next autoupdate, but after creation you cannot do anything with it. If you wnt, the server can return an id of the created poll, if you navigate to a detail view.

client/src/app/site/motions/modules/motion-detail/components/motion-poll/motion-poll-manager/motion-poll-manager.component.ts
```diff
@@ -0,0 +1,99 @@
...
+        } else {
+            const update = {
+                // title: data.title,
+                onehundred_percent_base: data.onehundred_percent_base,
+                majority_method: data.majority_method
+            };
+            console.log('update', update);
+            await this.pollRepo.update(update, model);
+            return model;
```
FinnStutzenstein: This is also error-prone. You wont get the updated model this way.

client/src/app/site/motions/modules/motion-detail/components/motion-poll/motion-poll-preview/motion-poll-preview.component.ts
```diff
@@ -1,56 +1,135 @@
...
 
-    public enterAnalogVotes(): void {
-        throw new Error('TODO');
+    /**
+     * Searches for the parent-motion of this poll.
+     *
+     * @returns a `ViewMotion`.
+     */
+    public getMotionOfPoll(): ViewMotion {
+        return this.motionRepo.getViewModel(this.poll.motion_id);
```
FinnStutzenstein: Please add a new relation in the motion poll repo and assignmenr poll repo to link polls to motions/assignments.

PR #5134, opened 2019-11-13
Voting delegate view

jsangmeister:
- added `MotionPollVoteComponent` as a view for delegates to vote
- PollOptions are no longer nested and instead own models, since we need to have the access option -> poll
   - adjusted autoupdate tests and server code accordingly
- added a `BannerService` to manage the display of banners at the top of the page
   - OSStatus and Offline services are using it now to display their banners
- added a `VotingService` to check poll votability
- added `VotingBannerService` to display a banner while a user can vote on a poll
- added `ViewBasePoll` and `BasePollRepositoryService`
- added a server view for passing votes when creating an analog poll
- added ability to filter by functions to `BaseFilterListService`
- changed `ListViewtable` to not need a full repository but instead only an implementation of `HasViewModelListObservable`
- added a shared list for motion and assignment polls
- and much more


client/src/app/site/motions/modules/motion-poll/motion-poll-detail/motion-poll-detail.component.html
```diff
@@ -35,6 +35,10 @@ <h1>{{ poll.title }}</h1>
             <div>{{ 'Poll method' | translate }}: {{ poll.pollmethodVerbose | translate }}</div>
             <div>{{ 'Majority method' | translate }}: {{ poll.majorityMethodVerbose | translate }}</div>
             <div>{{ '100% base' | translate }}: {{ poll.percentBaseVerbose | translate }}</div>
+            <button mat-button (click)="setState(1)">1</button>
+            <button mat-button (click)="setState(2)">2</button>
+            <button mat-button (click)="setState(3)">3</button>
+            <button mat-button (click)="setState(4)">4</button>
```
jsangmeister: Debugging purposes, can be removed when #5132 is merged

jsangmeister:
@FinnStutzenstein please check, should be ready to merge

client/src/app/core/ui-services/banner.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Replaces a banner with another. Convenience method to prevent flickering
+     * @param toAdd the banner to add
+     * @param toRemove the banner to remove
+     */
+    public replaceBanner(toRemove: BannerDefinition, toAdd: BannerDefinition): void {
+        const newArray = Array.from(this.activeBanners.value);
+        const idx = newArray.findIndex(banner => banner === toRemove);
```
FinnStutzenstein: This is too optimistic: Please raise an error, if `toRemove` wan't found.

client/src/app/core/ui-services/voting.service.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * checks all polls for votable ones and displays banners for them
+     * @param polls the updated poll list
+     * @param mode motion or assignment
+     */
+    private checkForVotablePolls(polls: ViewAnyPoll[], mode: 'motion' | 'assignment'): void {
+        if (this.OSStatus.isInHistoryMode) {
+            polls = []; // display no active polls if in history mode
```
FinnStutzenstein: please do the cleanup here and finish the block with an return. So it is clear, that the special case was fully handled.

openslides/poll/access_permissions.py
```diff
@@ -36,13 +36,11 @@ class BasePollAccessPermissions(BaseAccessPermissions):
                     del poll["votesvalid"]
                     del poll["votesinvalid"]
                     del poll["votescast"]
+                    # add user_has_voted instead of voted_id array so user can check if he has voted yet
+                    poll["user_has_voted"] = user_id in poll["voted_id"]
```
FinnStutzenstein: One thing here: managers do not get this field. You need to modify all polls before the permission check above.

jsangmeister:
> Hmm, we should hide the banner if someone has voted (even if he can change his vote).

I discussed this with @emanuelschuetze, we came to the conclusion that the banner should stay. If not, we should IMO not do this as a banner but instead as a snackbar.
If needed, we can still change this later, or add a hint in the banner a là "You already voted on this poll", but that can be done outside this PR.

client/src/app/site/assignments/components/assignment-poll-detail/assignment-poll-detail.component.html
```diff
@@ -0,0 +1,78 @@
...
+            <div>{{ 'Poll type' | translate }}: {{ poll.typeVerbose | translate }}</div>
+            <div>{{ 'Poll method' | translate }}: {{ poll.pollmethodVerbose | translate }}</div>
+            <div>{{ 'Majority method' | translate }}: {{ poll.majorityMethodVerbose | translate }}</div>
+            <div>{{ '100% base' | translate }}: {{ poll.percentBaseVerbose | translate }}</div>
+        </div>
+
+        <div *ngIf="poll.state === 3 || poll.state === 4">
+            <h2 translate>Result</h2>
+
+            <div *ngIf="poll.type === 'named'" style="display: grid; grid-template-columns: auto repeat({{ poll.options.length }}, max-content);">
```
GabrielInTheWorld: Please write style for styling components in the stylesheets.

PR #5179, opened 2020-01-21
Minor Server changes

jsangmeister:
default motion poll method set
changed permission from 'can_manage_metadata' to 'can_manage'


jsangmeister:
@tsiegleauq After this merged, the client will no longer have to send the pollmethod on motion polls, so the filed can be safely removed from the form

PR #5181, opened 2020-01-21
Updated action spec.

normanjaeckel:


docs/interfaces/action-service.txt
```diff
@@ -1,27 +1,19 @@
...
  * Executes multiple actions in the conext of the user given by the user_id.
+ * All actions are successful if no exception is thrown, else no action was
+ * processed.
  *
- * @throws ActionNotFound
+ *
+ * @throws ActionException
  */
-execute(actions: Action[], user_id: number): ExecuteResult;
+handle_request (payload: Action[], user_id: number): void
```
FinnStutzenstein: There will be actions, that produces some results, like creating a model (the id), or some status updates for the users. E.g. on mass-actions, we sometimes uses "track-ids" to visualize in the client, which objects have been successfully changed. There must be a way to give data back to the client.

Also: The name `handle_request` is misleading. This has nothing to do with requests in any case, but wich actions, that need to be performed.

docs/interfaces/action-service.txt
```diff
@@ -1,27 +1,19 @@
...
 interface Action {
-    name: string;
-    data: object;
-}
-
-interface ActionError {
-    error: string;
-    argumens: string[];
+    action: string;
+    data: object[];
```
FinnStutzenstein: Why must the data be an array and cannot be an object? Maybe you should provide examples, of how you designed some specific actions.

docs/interfaces/action-service.txt
normanjaeckel: Hm. I can send back something like a list of results like `[None, None, {"some":"result"}, None]", ok?

"handle_request" seems correct to me because it is about handling a request to the action service by this handler (and its bunch of action components behind it). It is not just performing them (because if the user has no permission, nothing will be performed).

docs/interfaces/action-service.txt
normanjaeckel: Data should be an array because you should be able to do every action as mass/bulk action.
For every action there is a json schema in the code the defines the expected payload, e. g. https://github.com/OpenSlides/openslides-backend/blob/master/openslides_backend/actions/topic/create.py#L16

docs/interfaces/action-service.txt
```diff
@@ -1,27 +1,19 @@
...
 
-// TODO: More Exceptions?
-Exception ActionNotFound(name: string);
+Exception ActionException(message: string);
 
 
 /**
  * Executes multiple actions in the conext of the user given by the user_id.
+ * All actions are successful if no exception is thrown, else no action was
+ * processed.
```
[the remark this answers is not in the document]
normanjaeckel: I thought the event store does not provide the possibility that different actions are sent together and may handled independently.

Email is a really special case:

If the request user asks me to send three mails I would try to send them all. If all were successful (means just successfully handled to the mail transfer agent), I can try to save the email_sent timestamp to the user model. If this fails ... bad luck. If not ... ok.

If the first email is successful and the second not, what should I do? Should I try to save the timestamp? Should I first tra to sent the third email? I do not have different7independent workers or threads so we must decide the order of what to do.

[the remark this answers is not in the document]
FinnStutzenstein: > I thought the event store does not provide the possibility that different actions are sent together and may handled independently.

It does! You just have to change your execution (during the dispatching): Instead of collectiong all events from all actions and sending them once, send the events after each action.

The order is given by the client: all actions are evaluated in the order given. There are more cases: You can select all users to send an email, but some users doesn't even have an email in the system. This is not limited to users: Most bulk actions may send multiple models, for which the action is not allowed. This is a huge overhead in the client to query and filter, but easily done, if the actions are performed independent of eachother.

docs/interfaces/action-service.txt
FinnStutzenstein: Ok

docs/interfaces/action-service.txt
FinnStutzenstein: Ah, I had something like this in mind. Maybe you document this here with a comment and e.g. a `Type ActionData=object` and using `ActionData` as a clear name.

FinnStutzenstein:
@normanjaeckel Ping

docs/interfaces/action-service.txt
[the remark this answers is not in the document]
normanjaeckel: Changed.

docs/interfaces/action-service.txt
normanjaeckel: Changed.

docs/interfaces/action-service.txt
```diff
@@ -5,15 +5,21 @@ Exception ActionException(message: string);
 
 /**
  * Executes multiple actions in the conext of the user given by the user_id.
- * All actions are successful if no exception is thrown, else no action was
- * processed.
+ * All actions are processed independently.
+ *
+ * TODO: Check if we need an interface or flag to process actions all together.
```
FinnStutzenstein: This depends on the actions you implement. But I'm sure, that we will need this. I think about actions, that cannot be covered with a catch-all-update-action, like resetting a motion's state. In a client update view, a user may change some properties of the motion and resets the state. This would result in two _coupled_ actions, that must be performed together atomically.

normanjaeckel: So let's see when such actions appear. Until this we can merge this PR.

docs/interfaces/action-service.txt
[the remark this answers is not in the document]
normanjaeckel: I want this to change back, see my email.

docs/interfaces/action-service.txt
FinnStutzenstein: Ok

PR #5190, opened 2020-01-27
Reworks the banner showing if there are polls

GabrielInTheWorld:
- Makes it higher on mobilephones
- Changes title, if there is only one poll


GabrielInTheWorld:
@emanuelschuetze Removed the underline in the banner's description.

PR #5197, opened 2020-01-29
Fixed some errors

jsangmeister:


openslides/assignments/apps.py
```diff
@@ -1,6 +1,6 @@
 from typing import Any, Dict, Set
-
 from django.apps import AppConfig
+from asgiref.sync import async_to_sync
```
FinnStutzenstein: Please undo (and the removed line two lines above). Travis will fail ;)

PR #5220, opened 2020-02-06
Added first draft for restriction and presenter interfaces.

normanjaeckel:


FinnStutzenstein:
Hi, I pushed a new commit for the presenter service. I would like to get the specs for it done, that I can use it in the client. If you can check the presenter, it would be nice. If it is OK, I'll create a new PR with just the presenter interface

normanjaeckel:
I prefer to use an exception also for 4xx like errors so I can throw it anywhere inside the code and do not have to wrap it once again (the same thing as I do in actions service).

FinnStutzenstein:
First, you can use exceptions in your code if you want to. E.g. when looping through the presenters, each one can throw some exceptions that are collected in the loop and aggregated into an array of (error) responses.

The difference between the exception and errortype was intentional. If the user does something wrong (e.g. wrong input data) in one presenter, another one must not fail and must be executed. All presenters are independend of each other. An exception means that the current control flow is aborted, so if there is an exception in one presenter, all following ones are not executed. I guess we have two options here:
1) Just do one presenter at a time. Than you can throw exceptions for user errors and "programming errors" (e.g. some other serivce errors and so on).
2) Stay at error responses per presenter and keep multiple presenters (my favorite).

normanjaeckel:
I am not finished with this but would like to merge this pull request so that we already have a first draft in our docs.

PR #5221, opened 2020-02-08
Defined back references (part 1)

normanjaeckel:
I would like to document the (I call them) back references.

E. g. in `models/meeting.py` I want to define the `committee_id` field with a foreign key to a committee. The related name is also given, e. g. `meeting_ids`. This related field will be automatically added to the committee model but I do not want to define it in `models/committee.py`.

So the quest is to find out, which fields a real relation fields and which are back references. Both belong to the interface of course but I would like to build the back references read only of course.

This pull request is only a first draft with some new sorting. Maybe someone can continue here ...

### DIFF
--- a/docs/models.txt
+++ b/docs/models.txt
@@ -15,6 +15,10 @@ values for `<some_variable>` that fits into the format of a field. E.g.
 `amendment_paragraph_3`. Note that it might be possible that no kind of
 this field exists in a model.
 
+Note to so called back references: These fields are read only. The actual
+relation is provided by the related field. See models definition in backend
+service.
+
 Interface projector {
     id: number;
     scale: number;
@@ -42,6 +46,7 @@ Interface projector {
     projectiondefault_ids: projectiondefault[];
     meeting_id: meeting;
 }
+
 Interface projection {
     id: number;
     projector_id: projector;
@@ -50,6 +55,7 @@ Interface projection {
     element: Fqid;
     options: JSON;
 }
+
 Interface projectiondefault {
     id: number;
     name: string;
@@ -58,6 +64,7 @@ Interface projectiondefault {
     projector_id: projector;
     meeting_id: meeting;
 }
+
 Interface tag {
     id: number;
     name: string;
@@ -65,6 +72,7 @@ Interface tag {
     tagged_ids: Fqid[];
     meeting_id: meeting;
 }
+
 Interface projector_message {
     id: number;
     message: HTML;
@@ -72,6 +80,7 @@ Interface projector_message {
     projection_id: projection[];
     meeting_id: meeting;
 }
+
 Interface projector_countdown {
     id: number;
     title: string;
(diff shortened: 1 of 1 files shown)


FinnStutzenstein:
Two points:
- We should keep consistent in naming: In the client we call these relations "reverse relations". We should adopt this to the server, so we do not have two  wordings for one principle.
- I do not like to have the relations antisymmetric regarding read/write. It is confusing to have both sides of a relation available, but you can only write one. A programmer must know, whch one to write which might not be intuitive. Especially, when this case can be done programatically: Somewhere in the server you define the relation. E.g. there are users and groups, which are related in a M:N fashion with keys user_ids (for groups) and group_ids (for users). With this information, you can handle writes to both sides, because users and groups are exchangeable. Adding an id to user_ids and modifying group_ids is the same as the other way around. This "relation managing" logic has to be implemented once. The same goes for 1:M and 1:1, so in conclusion you have three types of operations on relations. They are equal for for every relation and equal for "reversed relations" (So the term "reverse relation" is is "incorrect" in this case, becuase both sides are equally handled; it is symmetric).
I mean you might have already written the code for e.g. M:N relations, then just use this code for both sides and even generate the code/handlers for the second side (we should not declare this on both sides to reduce errors by programmers)

normanjaeckel:
The term reverse relation is ok for me.

Ok I can try to create some generic logic so that you can write on both sides of a relation, but it would be great if we can get a little bit more structure into the models spec. I will try this maybe next week.

FinnStutzenstein:
Ok, you mean defining relations like for users `group_ids: user[] (references group/user_<meeting>_ids)` so it is clear, which fields are the associated in a relation?

normanjaeckel:
Yes. In some cases the reference is already given. This should be completed, so it is also easier to pick the correct terms for the model code (options `to` and `related_name` I use when defining relational fields).

normanjaeckel:
Additional question: Do we want to enable that fields are writeable if they are used as (so called specific relation) in relational fields with ids inside the field name?

Example:

    motion/meeting_id: meeting[] (references meeting/motion_ids)
    motion/supporter_ids: user[] (references user/support_motion_<meeting_id>_ids: motion[])

In this case if I change meeting_id of the motion from 42 to 538 I do not only have to adapt the relation in meeting collection (delete motion id in respective array of meeting 42 and add it in respective array of meeting 538) but also have to handle the support_motion_42_ids array (remove motion id here) and support_motion_538_ids (add id here).

If you think this is necessary I can try to do it (maybe needs more time) or we just disable this feature (meeting id of motions can not be changed any more).

FinnStutzenstein:
> meeting id of motions can not be changed any more

This is the wanted behavior.

normanjaeckel:
Ok.

FinnStutzenstein:
@normanjaeckel Ping

normanjaeckel:
I have to create a new branch and pull request.

PR #5233, opened 2020-02-25
Adds required users for assignment poll

jsangmeister:
Adds all users that voted to the required users for single votes table so that participants without permission to see user names can see the single votes

### DIFF
--- a/openslides/assignments/apps.py
+++ b/openslides/assignments/apps.py
@@ -50,7 +50,12 @@ class AssignmentsAppConfig(AppConfig):
 
         # Register required_users
         required_user.add_collection_string(
-            self.get_model("Assignment").get_collection_string(), required_users
+            self.get_model("Assignment").get_collection_string(),
+            required_users_assignments,
+        )
+        required_user.add_collection_string(
+            self.get_model("AssignmentOption").get_collection_string(),
+            required_users_options,
         )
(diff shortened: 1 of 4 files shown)


openslides/assignments/apps.py
```diff
@@ -50,7 +50,12 @@ def ready(self):
 
         # Register required_users
         required_user.add_collection_string(
-            self.get_model("Assignment").get_collection_string(), required_users
+            self.get_model("Assignment").get_collection_string(),
+            required_users_assignments,
+        )
+        required_user.add_collection_string(
+            self.get_model("AssignmentPoll").get_collection_string(),
```
FinnStutzenstein: Why not `"AssignmentOption"` here?

openslides/assignments/apps.py
```diff
@@ -95,3 +100,21 @@ def get_startup_elements(self):
                 if option:
                     candidates.add(option["user_id"])
     return candidates
+
+
+async def required_users_polls(element: Dict[str, Any]) -> Set[int]:
```
FinnStutzenstein: Please use options directly, so you must not access the element cache.

jsangmeister: Well, that made it a lot shorter ^^

PR #5237, opened 2020-02-26
Integrating the client and datastore into the main repo:

FinnStutzenstein:
- Added each service as a submodule for development
- Added conventions for naming environment variables
- Added required targets in makefiles of services


normanjaeckel:
Hm. Why do we need submodules? What should be the naming conventions (as the commit message says)? The docs do not contain any conventions. (See e. g. my proposal [here](https://github.com/OpenSlides/openslides-backend#environment-variables).)

FinnStutzenstein:
Hi, for a production build and production setup, no submodules are needed (this is not quite true, see below). You can clone the repo without initializing the submodules and run a `make run-prod`.

If you want to develop inside the complete setup and not every service at their own, you need to mount the sourcecode into the containers. This is needed to live edit inside the containers, which is crucial for debugging the complete setup. To mount the volumes, the developer needs to have all services cloned to a specific known path. We may enforce, that all repos needs to sit next to eachother, or use symlinks (like plugins in OS<4). But we have to manage which service-versions matches to each other. And for this management, we can use submodules with these advantages:
- Management of current matching service versions
- Simpler to install. No need to clone all repos by hand nor writing a support-script for this
- No (explicit) enforcement of repo placement

The prod-build reads the commit hashes for each submodule (without initializing) and build all services for these commit hashes. Inf act we are using the submodules as a management tool for matched services. When a services has updated, the main repo must update the submodule commit hash, to bump the service version.

For development, submodules makes it easy to manage versions of all services. E.g. with https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/DEVELOPMENT.md#a-useful-command you can check out all services to master. With so many services it is necessary to have a good overview, what services are checkd out to what branches. With submodules these queries are easy. With more services and development, I guess that we will put some useful commands into a helping Makefile (like the command above). If someone wants to develop on a service alone, it is still possible to clone the service by itself and not touching the main repo at all. E.g. I developed the datastore writer in a single repo. All changes due to "standerization" with env-variables, ... in the Dockerfile I've done from the mainrepo. The current git-workflow in a submodule is exactly the same as a single repo.

Note: The makefile may be replaced later with some manage-scripts. But for the current state it is good to ducoment all commands needed

The requirements are here: https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/DEVELOPMENT.md#requirements-for-services

In this file are examples how to integrate services and manage them with submodules.

PR #5241, opened 2020-03-09
Autoupdate service interface

FinnStutzenstein:


ostcar:
Here the current data: https://github.com/OpenSlides/openslides-autoupdate-service/blob/master/internal/keysbuilder/keysbuilder_test.go

docs/interfaces/autoupdate-service-examples.txt
```diff
@@ -251,11 +259,15 @@ ModelRequest: {
...
+            type: "generic-relation-list",
             fields: {
                 a: null,
                 b: null,
                 not_existent: {
-                    key: null,
+                    type: "relation-list",
+                    fields: {
+                        key: null,
+                    }
```
ostcar: This is a syntax error. If the type is "relation-list", then the object needs the field "collection". I used `generic-relation` in the golang tests: https://github.com/OpenSlides/openslides-autoupdate-service/pull/43/files#diff-a338522804dc3046ef5f239f5a697f8aR188-R191

So either use `generic-relation` or `generic-relation-list` or add the attribute `collection`

docs/interfaces/autoupdate-service-examples.txt
```diff
@@ -291,52 +303,86 @@ ModelData: {
...
+            B_$_ids: ["1", "2", "3"],
+        },
+        2: {
+            d: "d2",
+            B_$_ids: ["1", "4"],
+        }
+    }
+}
+
+V: structured fields without references
```
ostcar: In example IV you renamed it to "template fields". Here you use the term "structured fields" was this intentional?

docs/interfaces/autoupdate-service-examples.txt
```diff
@@ -291,52 +303,86 @@ ModelData: {
...
+    - `B_$_ids` is a template field, which should be resolved
+
+ModelRequest: {
+    collection: "D",
+    ids: [1, 2],
+    fields: {
+        d: null,
+        B_$_ids: {
+            type: "template",
+            fields: null,
```
ostcar: A template field does not have the attribute "fields". It will be ignored.

docs/interfaces/autoupdate-service.txt
```diff
@@ -27,56 +27,52 @@ subscribe(request: ModelRequest): stream<ModelData>;
...
- * follow. There are two types of values:
- * - GenericModelDescriptor: The reference is a generic one. This means, that
+ * follow. There are three types of values:
+ * - GenericRelationFieldDescriptor: The reference is a generic one. This means, that
  *   the actual value from the model is a fqid or an array of fqids.
- * - ModelDescriptor: A collection is given, so it can be expected, that the
+ * - RelationFieldDescriptor: A collection is given, so it can be expected, that the
  *   actual model value is an id or an array of ids.
+ * - StructuredFieldDescriptor: The field is a template field and should be
+ *   handled in this way.
```
ostcar: I make a difference between the `relation`-field and the `relation-list`-field. The same for `generic-relation` and `generic-relation-list`. So for the autoupdate-service, there are 5 cases.

docs/interfaces/autoupdate-service.txt
```diff
@@ -85,10 +81,41 @@ Interface GenericModelDecriptor {
...
     collection: Collection;
 }
 
+/**
+ * Structured Fields: A field with `$` as a placeholder can be structured. The
+ * `$` is a template placeholder. The `template field` holds a list of strings
+ * with options to insert into `$`. All these fields are the `structured
+ * fields`. Example:
+ * B_$_ids is the remplate field. Its value may be `["1", "test"]`, so the
+ * structured fields are B_1_ids and B_test_ids.
```
ostcar: Ahh. Now I understand how you use the terms. Is ok for me. There is a type s/remplate/template/

docs/interfaces/autoupdate-service.txt
```diff
@@ -85,10 +81,41 @@ Interface GenericModelDecriptor {
...
+ *
+ * This interface indicates, that there are structured fields related to the
+ * given template field, that should be resolved. If one would give `B_$_ids:
+ * null` without using this interface, the response would be just the template
+ * field `B_$_ids` and no structured fields.
+ *
+ * For just resolving values, specify `values: null`.
```
ostcar: The attribute `values` is not needed at all. But `values: null` also works.

docs/interfaces/autoupdate-service-examples.txt
FinnStutzenstein: Typo. Should be `values`.

ostcar: ok. "values" does not harm, but is not needed.

docs/interfaces/autoupdate-service.txt
FinnStutzenstein: Ah ok, this wasn't clear from your email

PR #5257, opened 2020-03-17
Client server communication

FinnStutzenstein:


docs/client-server-streaming-communication.txt
```diff
@@ -0,0 +1,53 @@
...
+For normal packets, it is forbidden to send content with an error key in the
+outer most JSON-object.
+
+5) Outlook
+There might be an additional reserved keys. The protocol may be extended to:
+ - a watchdog: the server sends a ping in regular time intervals. The client
+   take note of them and notices, if there was a missing ping. E.g. the timeout
+   for the client is twice the send interval time. If a missing ping was
+   detected, the connection is reestablished.
+
```
ostcar: The additional reserved keys should not be valid fqfields. The autoupdate-service sends the requested keys. So if a key "error" could be requested, it would be a conflict.

FinnStutzenstein: "error" is not a valid fqid, so this is fine

PR #5258, opened 2020-03-17
Adds interface for auth-service

GabrielInTheWorld:


docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,45 @@
+Interface Token {
+    payload: {
+        // The lifetime of the Token in milliseconds.
+        expiresIn: number (ms),
```
FinnStutzenstein:
```suggestion
        expiresAt: number, // the date in unix seconds of the expiration.
```
If you use `lifetime`, the refrence is not clear. E.g. the token expires in two minutes. But from where?

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,45 @@
...
+        // The corresponding userId of the requesting client.
+        userId: string,
```
FinnStutzenstein:
```suggestion
        userId: number,
```

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,45 @@
...
+        // The id of the current session.
+        sessionId: string
+    },
+    signature: string
+}
```
FinnStutzenstein: Please add the `Cookie` interface

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+* The client can login with its credentials for authentication.
+* If they are correct, the service answers with a signed Token and sets a cookie, containing the sessionId of the client. 
+*/
+(Token, Cookie) Login (username: string, password: string);
```
FinnStutzenstein:
```suggestion
Login(username: string, password: string): (Token, Cookie);
```
Please use this formatting

Also `login` may raise an error, if the login wasn't successful.

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+* An example for any protected route. If the client requests protected resources, it has to 
+* send the signed Token and the cookie, it receives from the service at login, to the server.
+*/
+Any Api* (token: Token, cookie: Cookie);
```
FinnStutzenstein: Suggestion: Name this `authenticate` and define exatly, what is returned, e.g. `{userId: number; sessionId: string;}`. Please make clear, that this is a library function.

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+* Function to kill one specific session by its id.
+*/
+void ClearSessionById (sessionId: string);
+
+/**
+* Function to kill all current opened sessions except the one, which is requesting.
+*/ 
+void ClearAllSessionsExceptThemselves ();
```
FinnStutzenstein: Shouldn't be a userId and sessionId (or a token) passed as parameters?

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+* The service deletes the session depending on the given Token.
+*/
+void Logout (token: Token);
```
FinnStutzenstein: the parameters must be token and cookie. Internally you want to call `authenticate` first, to check, if the request is valid. For this, ou need the cookie.

docs/interfaces/auth-service.txt
FinnStutzenstein: @GabrielInTheWorld I think this is not resolved. Please give me a call if something is not clear.

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,99 @@
...
+        userId: number,
+        // The id of the current session.
+        sessionId: string
+    },
+    signature: string
+}
+
+Interface Cookie {
+    // The id for the session corresponding to the client, who has signed in.
+    sessionId: string;
```
FinnStutzenstein: the cookie has a signature, hasn't it?

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,99 @@
...
+* Sends back a new Token.
+*
+* Throws an exception, if the cookie is empty or the transmitted sessionId is wrong.
+*
+* @throws InvalidCredentials
+*/
+WhoAmI(cookie: Cookie): Token;
+
+/**
+* Function to kill one specific session by its id.
```
FinnStutzenstein: Just a thought: You must check the users permissions, if the given sessionId does not belong to the user in the token. So a user can always end his own sesions, but osers cannot end other users sessions. But there is an exception: If the user (given in the token) is a superadmin, he can clear sessions of other users.

This means, that the auth service will need the permission service at this point (which is not tragic, but just for considerations).

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+* The service deletes the session depending on the given Token.
+*
+* @throws InvalidCredentials
+*/
+Logout (token: Token, cookie: Cookie): void publishes LogoutSessionEvent;
+
+/**
+* Returns all currently active sessions.
```
FinnStutzenstein: Here the token and cookie is also missing. It is the same as for ClearSesionById: One user can just list his own sessions unless he is a superadmin.

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+* Function to kill one specific session by its id.
+* 
+* An exception is thrown, if the client has not the necessary permissions to make this action.
+* Also, if there is no session with the given id, an exception is thrown.
+*
+* @throws NoPermissions
+*/
+ClearSessionById (sessionId: string, cookie: Cookie, token: Token): void;
```
FinnStutzenstein: I think this should also publish a logout event.

docs/interfaces/auth-service.txt
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+* Function to kill all current opened sessions except the one, which is requesting.
+*
+* An exception is thrown, if the client has not the necessary permissions to make this action.
+* 
+* @throws NoPermissions
+*/ 
+ClearAllSessionsExceptThemselves (token: Token, cookie: Cookie): void;
```
FinnStutzenstein: I think this should also publish a logout event.

PR #5261, opened 2020-03-18
Refined models

FinnStutzenstein:


FinnStutzenstein:
Note: I renamed motion/identifier to motion/number to avoid further misunderstandings between id and identifier. The name was too ambiguous.

normanjaeckel:
Can you put the models into some kind of ordering (maybe according to old "apps") and also add an empty line between every model. It would be easier to read the file then ...

I propose the following (pseudo-hierarchical) order:

organisation
user
committee
meeting
group
topic
motion
assignment
agenda
projector

and all other models in between.

No problem if you have no time for this. I can do it (but only if I have the time ... :-))

FinnStutzenstein:
OK, will be done later in this week (currently I'm creating example data in this order ;) )

FinnStutzenstein:
Faster than expected: Done :)

normanjaeckel:
Ok. This is very helpful. I looked a bit around and it looked okay for now. I did not check every single line but I think it would be no problem if we want to rename two or three fields later.

Am I right that we do not want to speak about models here but about collections?

FinnStutzenstein:
Your additions are correct! I'll merge this now.

PR #5263, opened 2020-03-19
Adjustments for datastore spec

jsangmeister:


jsangmeister:
Adding to our discussed points, I renamed "model" to "fqid" for the get methods since it's clearer. Also, renamed the filter properties since `not`, `and` and `or` are keywords in python and can therefore not be used as keys

docs/interfaces/datastore-service.txt
```diff
@@ -161,8 +162,8 @@ getManyWithDeleted(collection: Collection, ids: Id[], position?: Position, mappe
 
 /**
  * Returns all (non-deleted) modells of one collection. It is not possible to specify
- * an id, so this method can not be used, if the user browses the history. It should
- * be noted, that it is highly disencouraged to use this method, becuase it might
+ * an id, so this method cannot be used if the user browses the history. It should
```
FinnStutzenstein:
```suggestion
 * an position, so this method cannot be used if the user browses the history. It should
```

docs/interfaces/datastore-service.txt
```diff
@@ -161,8 +162,8 @@ getManyWithDeleted(collection: Collection, ids: Id[], position?: Position, mappe
 
 /**
  * Returns all (non-deleted) modells of one collection. It is not possible to specify
```
FinnStutzenstein:
```suggestion
 * Returns all (non-deleted) models of one collection. It is not possible to specify
```

docs/interfaces/datastore-service.txt
```diff
@@ -119,70 +120,69 @@ Event ModifiedFieldsEvent on topic ModifiedFields {
...
 
-/**
- * Returns a model. Deleted models are not returned (and handled as a ModelDoesNotExit)
- *
- * @throws ModelDoesNotExist
- * @throws InvalidFormat
- */
-get(model: Fqid, position?: Position, mapped_fields?: fields[]): Partial<Model>;
+Enum DeletedModelsBehaviour {
+    NO_DELETED,
```
FinnStutzenstein: Please add values here, so a user know, hwat to send in a request (numbes or strings, which one is not important).

docs/interfaces/datastore-service.txt
```diff
@@ -119,70 +120,69 @@ Event ModifiedFieldsEvent on topic ModifiedFields {
...
+ * that map a collection to the needed ids and fields. If both the lower and the higher
+ * level mapped_fields are given, the higher level one is merged into all lower level
+ * ones.
+ * If an id is not found, it is not included in the response instead of throwing a
+ * ModelDoesNotExist.
  *
  * @throws InvalidFormat
  */
-getManyWithDeleted(collection: Collection, ids: Id[], position?: Position, mapped_fields?: Field[]): Partial<Model>[];
+getMany(requests: Fqid[] | GetManyRequest[], mapped_fields?: Field[], position?: Position, get_deleted_models?: DeletedModelsBehaviour): Partial<Model>[];
```
FinnStutzenstein: I do not like the extra possibility to give fqids. Currently this request is possible:
```
getMany([{collection:"a", ids: [1, 2], mapped_fields: ["f"]}], mapped_fields: ["g"]);
```
I would prefet the following suggestion:
```suggestion
getMany(requests: GetManyRequest[], position?: Position, get_deleted_models?: DeletedModelsBehaviour): Partial<Model>[];
```

jsangmeister: You are mixing 2 things here: the possibility to give fqids instead of request objects and the possible two mapped_fields.
1) The idea was that if there was a situation where you want to get models of different collections, but a common field of all of them, you could just pass in the fqids and the mapped fields. Of course that's not the typical use case. If you want we can remove the possibility of fqids.
2) The 2 mapped_fields are more useful in my opinion and can exist without the possibility to pass in fqids, if one wants to get a common field of multiple models.

FinnStutzenstein: OK

PR #5314, opened 2020-04-15
Adds more permissive html validation for topics and welcome page

jsangmeister:
Currently allows usage of `<video>` tags
an exception for the welcome page greeting has been implemented, this could also be extended to all markup config variables if needed

### DIFF
--- a/openslides/utils/validate.py
+++ b/openslides/utils/validate.py
@@ -1,11 +1,11 @@
-from typing import Any
+from typing import Any, List
 
 import bleach
 
 from .rest_api import ValidationError
 
 
-allowed_tags = [
+allowed_tags_strict = [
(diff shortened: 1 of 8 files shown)


openslides/utils/validate.py
```diff
@@ -37,13 +37,20 @@
...
+    "iframe",
 ]
 allowed_attributes = {
     "*": ["class", "style"],
     "img": ["alt", "src", "title"],
     "a": ["href", "title"],
     "th": ["scope"],
     "ol": ["start"],
+    "video": ["src", "poster", "controls"],
+    "iframe": ["src"],
```
emanuelschuetze, line 53: please remove iframe (security reasons)

PR #5343, opened 2020-04-29
Defined ActionResult as custom type. Added Error-interface.

ThomasJunk:
In order to reflect the result of a bulk operation I decided to define `ActionResult` as a custom type.
`ActionResult` itself is either an actual result in form of an `object` or it is an `Error` such that you could have the following as a result

```
[{
	msg: "user 2 created"
        ...
}, {
	type: 1,
	msg: "failed to create user4 "
}, {
	msg: "user 4 created"
        ...
}]
```

### DIFF
--- a/docs/interfaces/actions-service.txt
+++ b/docs/interfaces/actions-service.txt
@@ -19,7 +19,9 @@ interface Action {
     data: object[];  # An array of action specific data. See JSON schema defintions.
 }
 
-interface ActionResult {
-    success: boolean;
-    message: string;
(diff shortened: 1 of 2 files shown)


FinnStutzenstein:
@normanjaeckel @jsangmeister Is this ok and implemented in the action service?

normanjaeckel:
I think this is not implemented at the moment.

FinnStutzenstein:
Can it be closed then?

FinnStutzenstein:
Outdated.

PR #5344, opened 2020-04-30
Removed vote weight from votes_cast

FinnStutzenstein:
@emanuelschuetze I added a server side error for a wrong vote weight in the user import. The code is not flexible enough to just pass server errors, so there is a default string to display. Another thing: I do get a red entry on an error, but I cannot see the error message in the UI. Maybe @tsiegleauq have a look, where the message should be shown.

I'm not sure if the change, that votes_cast have the amount of users voted, must reflect in the client in some way.

The csv import works with 6 decimal places. With more, an error is now returned per user.


emanuelschuetze:
As discussed with @tsiegleauq we will fix the error handling of csv import later in #5131.

PR #5348, opened 2020-05-02
Toggle to show all amendments in diff view

CatoTH:
This adds a "temporarily show all in text" button to the amendment list when being in Diff Mode. If activated, then also amendments that have not yet been accepted are shown in diff mode.
Additionally, there is a context menu for amendments in Diff Mode to change the state and go to the amendment detail page.

### DIFF
--- a/client/src/app/site/motions/models/view-motion-amended-paragraph.ts
+++ b/client/src/app/site/motions/models/view-motion-amended-paragraph.ts
@@ -15,7 +15,7 @@ export class ViewMotionAmendedParagraph implements ViewUnifiedChange {
     }
 
     public constructor(
-        private amendment: ViewMotion,
+        public amendment: ViewMotion,
         private paragraphNo: number,
         private newText: string,
         private lineRange: LineRange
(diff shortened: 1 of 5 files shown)


CatoTH:
> Some feedback:
> 
>     * show identifier of amendments before the state label (e.g. "A001-Ä01 – eingereicht:")
> 
>     * why the accept/reject menu entries of a change recommendation is disabled if there is a collision in the same line? (We discussed this. Can you always enable these entries?)
> 
>     * Remove the prefix "Set: " in the set state menu
> 
>     * Use the icon 'visibility_on' for "show amendment" entry.
> 
>     * Please remove the last "–" in the summary line, e.g. "Line 1 (Change recommendation) – Replacement –"

Ok, should be fixed.

CatoTH:
> Last remark: Please show _always_ the amendment identifier if the checkbox "Temporarily show all in text" is enabled - also for amendments which are accepted.

Ok, should be fixed by https://github.com/OpenSlides/OpenSlides/pull/5348/files#diff-ff330ee46dcb9ad45b81843b1f4fdc68R130 (the state name is now always shown in this mode, too)

PR #5350, opened 2020-05-05
Added routes to datastore spec

jsangmeister:


docs/interfaces/datastore-service.txt
```diff
@@ -147,6 +147,7 @@ Enum DeletedModelsBehaviour {
 
 /**
  * Returns a model by fqid.
+ * Url: POST to /datastore/reader/get
```
FinnStutzenstein, line 150: `/internal` is missing as a prefix. Same for all others.

PR #5358, opened 2020-05-07
Add reader to docker compose files

jsangmeister:


docker-compose.yml
```diff
@@ -1,6 +1,17 @@
 version: '3'
 services:
     # DATASTORE SECTION
+    datastore-reader:
+        depends_on:
+            - datastore-postgresql
+        environment:
+            - DATASTORE_DATABASE_HOST=datastore-postgresql
+            - DATASTORE_DATABASE_USER=openslides
```
FinnStutzenstein, line 9: Hi, gernot advised me to use useful defaults. Can you check two things?

1) Do we have defaults for these variables?
2) If so, are tey ok for production usage?

I would use this oppertunity to make this clean and only to give env variables, that are changed for a good reason (e.g. ports)

FinnStutzenstein: This will help understanding this file better and would remove some clutter.

jsangmeister: Done in https://github.com/OpenSlides/openslides-datastore-service/pull/38

PR #5375, opened 2020-05-15
Autoupdate performance

FinnStutzenstein:
This PR is in favor of #5360 (I'll leave it open, because it might be merged, to add tracing to important parts of the code).

Things changed:

- [x] Mutex for the client autoupdate code -> no parallel (concurrent) work on AUs
- [x] Parameter for the wait-timeout for read only redis synchronization
- [x] Rewrite projector code to not fetch all data
- [x] Added caching for data access during the projector data creation
- [x] Order autoupdates in consumers on server side
- [x] Delay autoupdates on server side
- [x] Be aware of skipped autoupdates and holes in the change-id-series
- [x] Add delay-parameter
- [x] Test this.
- [x] fix tests

### DIFF
--- a/SETTINGS.rst
+++ b/SETTINGS.rst
@@ -143,3 +143,7 @@ not affect the client.
 operator is in one of these groups, the client disconnected and reconnects again.
 All requests urls (including websockets) are now prefixed with `/prioritize`, so
 these requests from "prioritized clients" can be routed to different servers.
+
+`AUTOUPDATE_DELAY`: The delay to send autoupdates. This feature can be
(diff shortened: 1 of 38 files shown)


FinnStutzenstein:
In production, some other failures happen, but this PR brings the performance up to suooprt votings with up to 2000 users, if the rest behaves fine (I had some redis replication errors, missing autoupdates, ...)

There is one point left: The poll progess bar is not very responsive and do not update on autoupdates. This issue has to be addressed in another PR.

@emanuelschuetze Ready to merge.

PR #5377, opened 2020-05-19
Adds migration to remove deprecated slides

jsangmeister:
@FinnStutzenstein is the `elements_history` always double nested? If so, why? If not, the migration has to be corrected. Please check that I missed nothing important.

### DIFF
--- a/openslides/assignments/migrations/0014_remove_deprecated_slides.py
+++ b/openslides/assignments/migrations/0014_remove_deprecated_slides.py
@@ -1,0 +1,27 @@
+from django.db import migrations
+
+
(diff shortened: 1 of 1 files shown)


FinnStutzenstein:
Yes the double-nesting is correct:
Each entry in the top-level list is the state of a projector. The state can contain multiple slides, e.g. if you have a topic and an non-stable overlay projected. When you project something else, both slides (topic and overlay) will be removed and put in one entry in the `elements_history`and when you click the "back" button, both elements are restored.

But I'm not sure, if we make usage out of this at the moment :D

PR #5389, opened 2020-05-28
Adds reverse relations for motions and blocks

jsangmeister:
...so the projector code doesn't have to iterate over all motions.
Based on #5375

### DIFF
--- a/openslides/motions/projector.py
+++ b/openslides/motions/projector.py
@@ -69,25 +69,24 @@ async def get_amendment_merge_into_motion_final(all_data_provider, amendment):
 
 async def get_amendments_for_motion(motion, all_data_provider):
     amendment_data = []
-    all_motions = await all_data_provider.get_collection("motions/motion")
-    for amendment_id, amendment in all_motions.items():
-        if amendment["parent_id"] == motion["id"]:
-            merge_amendment_into_final = await get_amendment_merge_into_motion_final(
-                all_data_provider, amendment
-            )
-            merge_amendment_into_diff = await get_amendment_merge_into_motion_diff(
-                all_data_provider, amendment
-            )
-            amendment_data.append(
-                {
-                    "id": amendment["id"],
-                    "identifier": amendment["identifier"],
-                    "title": amendment["title"],
-                    "amendment_paragraphs": amendment["amendment_paragraphs"],
-                    "merge_amendment_into_diff": merge_amendment_into_diff,
-                    "merge_amendment_into_final": merge_amendment_into_final,
-                }
-            )
+    for amendment_id in motion["amendments_id"]:
+        amendment = await all_data_provider.get("motions/motion", amendment_id)
+        merge_amendment_into_final = await get_amendment_merge_into_motion_final(
+            all_data_provider, amendment
+        )
+        merge_amendment_into_diff = await get_amendment_merge_into_motion_diff(
+            all_data_provider, amendment
+        )
+        amendment_data.append(
+            {
+                "id": amendment["id"],
+                "identifier": amendment["identifier"],
+                "title": amendment["title"],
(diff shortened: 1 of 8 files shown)


openslides/motions/projector.py
```diff
@@ -69,9 +69,10 @@ async def get_amendment_merge_into_motion_final(all_data_provider, amendment):
 
 async def get_amendments_for_motion(motion, all_data_provider):
     amendment_data = []
-    all_motions = await all_data_provider.get_collection("motions/motion")
-    for amendment_id, amendment in all_motions.items():
-        if amendment["parent_id"] == motion["id"]:
+    for amendment_id in motion["amendments_id"]:
+        amendment = await all_data_provider.get("motions/motion", amendment_id)
+        # seems like this gets executed before the updated data arrives, so check if amendment was found
```
FinnStutzenstein, line 74: Can you give a minimal way to reproduce this. The amendment must be there..

openslides/motions/projector.py
```diff
@@ -334,10 +335,10 @@ async def motion_block_slide(
...
     referenced_motions: Dict[int, Dict[str, str]] = {}
 
-    # Search motions.
-    all_motions = await all_data_provider.get_collection("motions/motion")
-    for motion in all_motions.values():
-        if motion["motion_block_id"] == motion_block["id"]:
+    # iterate motions.
+    for motion_id in motion_block["motions_id"]:
+        motion = await all_data_provider.get("motions/motion", motion_id)
+        if motion:
```
FinnStutzenstein, line 341: Same here. THis if-clause shoudl not be necessary. If there isn't the motion, something is not updated correctly in the cache.

openslides/motions/serializers.py
```diff
@@ -523,6 +544,14 @@ def update(self, motion, validated_data):
 
         inform_changed_data(motion)
 
+        if new_block != old_block:
+            if new_block:
+                new_block.refresh_from_db()
```
FinnStutzenstein, line 549: do you really need the refresh here? the code in `inform_changed_data` does not access the model directly. The model is resolved in the `AutoupdateBundle.done()` method (see `utils/autoupdate.py`)

openslides/motions/serializers.py
```diff
@@ -488,6 +499,14 @@ def create(self, validated_data):
         motion.supporters.add(*validated_data.get("supporters", []))
         motion.attachments.add(*validated_data.get("attachments", []))
         motion.tags.add(*validated_data.get("tags", []))
+
```
FinnStutzenstein, line 502: When does `validate` get called? If it gets called on every creation of the serializer (=creating, updating, ...) unnecessary autoupdates are send here. I guess the code in `create()` and `update()` should be sufficient.

FinnStutzenstein:
Please rebase on #5375

openslides/motions/serializers.py
jsangmeister: This is the create method ;)

openslides/motions/serializers.py
jsangmeister: Well, I'm _really_ sure that it was needed when I wrote it last week, but it seems to be working fine without it now... ¯\_(ツ)_/¯

openslides/motions/serializers.py
FinnStutzenstein: fuu git diff view ;) Sorry..

openslides/motions/views.py
```diff
@@ -693,6 +699,9 @@ def manage_multiple_motion_block(self, request):
                 motion, information=information, user_id=request.user.pk
             )
 
+            # inform motion block of change
+            inform_changed_data(motion_block)
```
FinnStutzenstein, line 703: `motion_block` might be None and the old motion block must be informed, too.

PR #5391, opened 2020-05-28
Fixes motion poll slide icon size & rounding

jsangmeister:


client/src/app/shared/pipes/parse-poll-number.pipe.ts
```diff
@@ -17,7 +17,7 @@ export class ParsePollNumberPipe implements PipeTransform {
             case VOTE_UNDOCUMENTED:
                 return this.translate.instant('undocumented');
             default:
-                return value;
+                return Math.round(value);
```
emanuelschuetze, line 20: This is wrong if you activate vote weight! - In user settings set "Activate vote weight" and add e.g. "1.75" for a user. Then start a named voting. The Yes value is set to 2 instead 1.75.
Exptected for deactivate vote weight: 1 (not 1.0000).
Exptected for activated vote weight: 1.75 (not 1.750000).

jsangmeister: Sorry, my bad. Is fixed now

emanuelschuetze:
Set  maximumFractionDigits to 6 (instead 2)

PR #5424, opened 2020-06-16
Fix some delegate view bugs

### LINKED ISSUE #5415: Prettify topic menu and list of speakers button
los

- [ ] Delegates see a item seperator, that is not needed
- [ ] the list of speaker button amount-indicator if not on the baseline of the list entry.

jsangmeister:
fixes #5415 

Removes dividers and empty menus in several views.

The problem with the shifted `mat-chip` as mentioned in #5415 is the following: For some reason, the `TopicDetailComponent` doesn't load `client/node_modules/@angular/material/bundles/material-chips.umd.min.js` (and, most importantly, the there-defined styles). `AssignmentDetailComponent` (and all other components that have a list of speakers button, as far as I've verified) loads the styles and the chip is displayed correctly. I have absolutely no idea why that's the case, so I'm just copying the important styles to the speaker button. If anyone wants to look into why that's happening and find the cleaner solution, feel free.

### DIFF
--- a/client/src/app/site/motions/modules/motion-list/components/motion-list/motion-list.component.html
+++ b/client/src/app/site/motions/modules/motion-list/components/motion-list/motion-list.component.html
@@ -262,7 +262,8 @@
         </div>
 
         <mat-divider
-            *ngIf="categories.length || motionBlocks.length || hasAmendments() || perms.isAllowed('manage')"
+            *ngIf="(categories.length || motionBlocks.length || hasAmendments() || perms.isAllowed('manage'))
+            && (perms.isAllowed('manage') || selectedView === 'list' || operator.hasPerms('core.can_manage_config'))"
         ></mat-divider>
 
         <div *ngIf="perms.isAllowed('manage')">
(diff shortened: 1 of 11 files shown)


jsangmeister:
Fixed

tsiegleauq:
motion list as delegate:

image

tsiegleauq:
fixed.

client/src/app/site/motions/modules/motion-list/components/motion-list/motion-list.component.html
```diff
@@ -262,7 +262,8 @@ <h2>{{ 'Motions' | translate }}</h2>
         </div>
 
         <mat-divider
-            *ngIf="categories.length || motionBlocks.length || hasAmendments() || perms.isAllowed('manage')"
+            *ngIf="(categories.length || motionBlocks.length || hasAmendments() || perms.isAllowed('manage'))
+            && (perms.isAllowed('manage') || selectedView === 'list' || operator.hasPerms('core.can_manage_config'))"
```
jsangmeister, line 266: Btw, this is really ugly and if more menu points are added, this gets increasingly more complex. I'd love an own component for the menu which only adds dividers if an element is displayed before and after it. But that's probably a bit overkill :D maybe something for the backlog (since we don't have enough things to do ^_^)

tsiegleauq: ~~could probably be solved using a structural directive. Might be complex~~
yes, backlog.

PR #5436, opened 2020-06-22
Fixed client startup

FinnStutzenstein:
@jsangmeister Please check, if you can now start the client

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,6 @@ tests
 .venv/
 .vscode/
 package-lock.json
+server/
+docker/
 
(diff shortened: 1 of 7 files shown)


jsangmeister:
I still only get a white screen when running `make run-dev` and then accessing `localhost:8000`. Do I have to update the submodules manually? I just checked out this branch and then tried to start it.

FinnStutzenstein:
Ah sorry. You need to have https://github.com/OpenSlides/openslides-datastore-service/pull/68 checked out for the DS.

FinnStutzenstein:
-> I merged the PR and updated this Branch. Should work now.

PR #5460, opened 2020-07-09
Add new settings to meeting model

jsangmeister:


docs/example-data.json
```diff
@@ -194,6 +194,8 @@
 
         "conference_show": false,
         "conference_auto_connect": false,
+        "conference_los_restriction": false,
+        "stream_url": "",
```
FinnStutzenstein, line 198: prefix `conference_`?

docs/models.txt
```diff
@@ -139,22 +139,27 @@ Interface committee {
...
     description: string;
     location: string;
     start_time: datetime;
-    end_time: datetime;
-    welcome_title: string;
-    welcome_text: HTML;
+    end_time: datetime; // TODO: is this needed?
     custom_translations: JSON;
 
+    // System
```
FinnStutzenstein, line 153: I'm currently not sure, if those settings should be editable in the settings page, or managed in some kind of committee-detail/meeting-list -> THis is mainly a UI decision, but for now please put them into the settings

docs/example-data.json
jsangmeister: Strictly, this isn't a setting for the conference, that's why in OS3 the prefix isn't present either. But I can add it if you want

docs/models.txt
jsangmeister: Will do

docs/example-data.json
jsangmeister: Done

PR #5476, opened 2020-07-17
Https

FinnStutzenstein:
Use HTTPS. See the updated install instructions in the DEVELOPMENT.md

Note: Copying the cert into the image is enough for dev but will change for production usage.

@ostcar Plase check, if the autoupdateservice works with this setup. Plase check, if `verify no ssl` is required in your case (`haproxy/src/haproxy.cfg`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@
 .env
 *.code-workspace
 
+# certs
+*.pem
+
 # Old OS3 files and folders
 .coverage
 .mypy_cache
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,10 +1,30 @@
 # Development of OpenSlides 4
 
+## Requirements
+
+You need git, bash, docker, docker-compose, make and go installed.
+
+Go is needed to install https://github.com/FiloSottile/mkcert. The development setup uses HTTPS per default. OpenSlides does not work with HTTP anymore since features are required (like http2) that only works in a secure environment.
+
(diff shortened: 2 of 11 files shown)


ostcar:
> @ostcar Plase check, if the autoupdateservice works with this setup. Plase check, if `verify no ssl` is required in your case (`haproxy/src/haproxy.cfg`).

I succeeded to test the full-system with different branches of the autoupdate service. I used `rm -rf openslides-autoupdate-service && ln -s ../openslides-autoupdate-service .` I hope, there is a better way.

It works with the current master (of the autoupdate-service), but the master does currently only support http1.1

I tested different solutions with this pr: https://github.com/OpenSlides/openslides-autoupdate-service/pull/81

With the commit `Allow http2 without tls` it does not work. With this commit, the autoupdate-services only acceppts http2 requests and rejects http1.1 requests. It seems, that the haproxy does not use http2 without tls. 

With the commit `Always use https but create self signed cert` the autoupdate service only supports https but with a self signed cert (if no other cert is given). With this commit, it only works with `ssl verify none`.

FinnStutzenstein:
> With the commit Always use https but create self signed cert the autoupdate service only supports https but with a self signed cert (if no other cert is given). With this commit, it only works with ssl verify none.

Ok, this is the way to go

ostcar:
The autoupdate-service-master now only supports https: https://github.com/OpenSlides/openslides-autoupdate-service/pull/81

PR #5477, opened 2020-07-17
OS4 Schema: agenda items have tags

FinnStutzenstein:


normanjaeckel:
I do not understand why agenda items have tags additionally to the tags that motions, topics and assignments already have.

FinnStutzenstein:
To tag agenda items independently from their content objects.

PR #5483, opened 2020-07-24
Fix docs

FinnStutzenstein:


FinnStutzenstein:
@jsangmeister Do you have a proposal for `inherited_access_groups`?

docs/interfaces/autoupdate-service.txt
```diff
@@ -119,19 +119,22 @@ Interface RelationFieldDescriptor extends Fields {
...
     type: 'template',
-    values?: {
-        [field: Field]: GenericRelationFieldDescriptor
-            | RelationFieldDescriptor
-            | StructuredFieldDescriptor
-            | null;
+    values?: GenericRelationFieldDescriptor
+        | RelationFieldDescriptor
+        | StructuredFieldDescriptor;
     }
```
jsangmeister, line 138: remove

docs/interfaces/autoupdate-service.txt
```diff
@@ -119,19 +119,22 @@ Interface RelationFieldDescriptor extends Fields {
...
  * `GenericRelationFieldDescriptor`.
  */
 Interface StructuredFieldDecriptor {
     type: 'template',
-    values?: {
-        [field: Field]: GenericRelationFieldDescriptor
-            | RelationFieldDescriptor
-            | StructuredFieldDescriptor
-            | null;
+    values?: GenericRelationFieldDescriptor
```
jsangmeister, line 135: I would suggest renaming it to value if it's only holding a single descriptor now

jsangmeister:
@FinnStutzenstein pushed my suggestion. A flag seemed cleaner than setting `inherited_access_group_ids` to `null`.

docs/interfaces/autoupdate-service.txt
FinnStutzenstein: The meaning is "How to handle possible values?"; this is currently fixed in the autoupdate serivce and client. I'm not sure if this is worth the renaming. @ostcar ?

docs/models.txt
```diff
@@ -685,10 +686,11 @@ Interface mediafile {
     mimetype: string;
     pdf_information: JSON;
     create_timestamp: datetime;
-    path: string; // Note: calcuated
-    inherited_access_group_ids: boolean | number[]; // Note: calculated and no
-    // reverse-relation for the group ids.
+    path: string; // Note: calculated
 
+    is_public: boolean;
```
FinnStutzenstein, line 691: I see why it's needed, but this field is calculated, right? If so, please add a note about it (I can smell confusion on both side about not allowing to write to this field ;) )

docs/interfaces/autoupdate-service.txt
FinnStutzenstein: Yes this is correct I think. Double-nested StructuredFieldDesriptors do not make any sense.

jsangmeister: I meant the left-over bracket :)

docs/models.txt
jsangmeister: Yes, right. Also renamed it after looking again at the OS3 code

FinnStutzenstein:
Something I noticed: Currently a structured field can be interpreted as a relation, generic relation and structured relation. Examples for the first ones:

```
x_$_ids = ["a"]
x_a_ids = [2]
```
(normal many-relation)

and

(generic relation)
```
x_$_ids = ["a"]
x_a_ids = "collection/3"
```

But a structured relation doesn't make any sense here, does it? If not, it should be removed...

docs/interfaces/autoupdate-service.txt
FinnStutzenstein: Yes, I noticed, but I think `StructuredFieldDescriptor` is still wrong here. See comment in the main thread...

docs/models.txt
```diff
@@ -685,10 +686,10 @@ Interface mediafile {
     mimetype: string;
     pdf_information: JSON;
     create_timestamp: datetime;
-    path: string; // Note: calcuated
-    inherited_access_group_ids: boolean | number[]; // Note: calculated and no
-    // reverse-relation for the group ids.
+    path: string; // Note: calculated
 
+    has_inherited_access_groups: boolean; // Note: calculated
```
FinnStutzenstein, line 691: Ok. false=public, true=Lookup in `inherited_access_group_ids`; if `inherited_access_group_ids` is `[]`, no perms, correct?

docs/interfaces/autoupdate-service.txt
jsangmeister: Ok, always happy to help :D

docs/models.txt
jsangmeister: Yes (except that `access_group_ids` also has to be checked, that's why I've named it so explicit and not more general)

FinnStutzenstein: > except that access_group_ids also has to be checked

does it? It should be enough to check `*inherited_access_groups` since `access_group_ids` should be evaluated in the inherited ones, too.

FinnStutzenstein:
@jsangmeister I'll merge the PR but let's further discuss the above points.

docs/models.txt
jsangmeister: Ah okay, I thought `inherited_access_group_ids` only included those from above and not `access_group_ids` themselves. But well, the naming is still fine I think

PR #5493, opened 2020-07-31
Fix syntax of models.txt

ostcar:
I am trying to parse the models.txt. Therefore, the syntax errors should be fixed.

I added a TODO in Line 733. Can someone tell me what field it should be?


FinnStutzenstein:
See line 757

ostcar:
> See line 757
Done. Can this be merged?

PR #5496, opened 2020-08-10
Update auth-service's interface

GabrielInTheWorld:
Updates the interface of the auth-service corresponding to PR `openslides-auth-service:#19`.


docs/interfaces/auth-service.txt
```diff
@@ -1,20 +1,45 @@
...
-    // A signature created from the server.
-    signature: string
+}
+
+/**
+* Describes an http-response, which is sent back to any requesting service.
+*/
+Interface Response <V, T> {
+    // Optional headers, which are set in an http-response
+    [ Headers?: V ]
```
FinnStutzenstein, line 22: Do you mean `headers?: {[name: string]: string}?` or `[header: string]: V` -> Are ehaders in an extra object or just extra keys in the response?

docs/interfaces/auth-service.txt
```diff
@@ -35,7 +63,7 @@ Exception InvalidCredentials
 *
 * @throws InvalidCredentials
 */
-Login (username: string, password: string): (Token, Cookie);
+"/login" (username: string, password: string): Response<V = {'authentication': Token, 'set-cookie': refreshId=Cookie)}>;
```
FinnStutzenstein, line 66: Please remove `"/"` -> the name is ust the last bit of the url.

docs/interfaces/auth-service.txt
```diff
@@ -1,20 +1,45 @@
...
+    // The id for the session corresponding to the user's client.
     sessionId: string,
-    // A signature created from the server.
-    signature: string
+}
+
+/**
+* Describes an http-response, which is sent back to any requesting service.
+*/
+Interface Response <V, T> {
```
FinnStutzenstein, line 20: You have two generic types, but below you mostly give one of them. What are the defaults?

docs/interfaces/auth-service.txt
```diff
@@ -49,7 +77,7 @@ Login (username: string, password: string): (Token, Cookie);
 *
 * @throws InvalidCredentials
 */
-Authenticate (token: Token, cookie: Cookie): {userId: number; sessionId: string;};
+"/authenticate" (Headers: {'authentication': Token, cookie: refreshId=Cookie}): Response<T = LoginInformation>;
```
FinnStutzenstein, line 80: Just keep it simple and add `authentication` and `set-cookie` in the response object as optional keys. Instead of using too much generics, maybe create a response type with and without headers.

docs/interfaces/auth-service.txt
```diff
@@ -1,104 +1,134 @@
...
-* and receive knowledge about authentication without request.
+* Sends back a new Token.
 *
-* Throws an exception, if the token is not valid. E.g. if the signature is not valid.
+* Throws an exception, if the cookie is empty or the transmitted sessionId is wrong.
 *
 * @throws InvalidCredentials
 */
-Authenticate (token: Token, cookie: Cookie): {userId: number; sessionId: string;};
+"who-am-i" (Ticket): Response<void>;
```
FinnStutzenstein, line 94: Please remove the `"`

docs/interfaces/auth-service.txt
```diff
@@ -1,104 +1,134 @@
...
+/**
+* Describes an http-response, which is sent back to any requesting service.
+*/
+Interface Response <T> {
+    // Optional headers, which are set in an http-response
+    HttpHeaders: {
+        // Authentication is passed, if a new access-token is returned.
+        authentication?: string,
+        // A cookie containing 'refreshId=(Cookie as string)' is passed, if a user signs in.
+        set-cookie?: refreshId=string
```
FinnStutzenstein, line 36: Suggestion:

```
cookies: {
    [name: string]: string
}
```
This has the advantage, that you can control the cookie name, or even support multiple cookies. This has to be adopted everywhere.

Please add to the docs:
- the lifetime of all cookies:
- other flags, like HTTPOnly

docs/interfaces/auth-service.txt
```diff
@@ -1,104 +1,134 @@
...
-*
-* @throws NoPermissions: Only users themselves can clear their own session and (super-) admins
-* can do this, too.
+* An exception is thrown, if there is no session with the given id, an exception is thrown.
 */
-ClearSessionById (sessionId: string, cookie: Cookie, token: Token): void publishes LogoutSessionEvent;
+"api/clear-session-by-id" (userId: string, Ticket): Response<void> publishes LogoutSessionEvent;
 
 /**
 * Function to kill all current opened sessions except the one, which is requesting.
```
FinnStutzenstein, line 118: Docs are not right: This will terminate all session *of the request user*, except for the one session. Whats about the anonymous?

FinnStutzenstein (changes requested):
See comments.

docs/interfaces/auth-service.txt
```diff
@@ -1,104 +1,134 @@
...
+    // The lifetime of a cookie. Date of expiration in unix seconds.
+    exp: number,
+    // Date of creation of a token in unix seconds.
+    iat: number
+}
+
+// The properties of this interface have to be passed as HTTP-headers in a request. 
+Interface Ticket {
+    authentication: string,
+    cookies: 'refreshId=string'
```
FinnStutzenstein, line 24: Maybe change this equally to line 38?

docs/interfaces/auth-service.txt
```diff
@@ -13,8 +13,42 @@ Interface Token {
...
+    exp: number,
+    // Date of creation of a token in unix seconds.
+    iat: number
+}
+
+// The properties of this interface have to be passed as HTTP-headers in a request. 
+Interface Ticket {
+    authentication: string,
+    cookies: {
+        'refreshId=string',
```
FinnStutzenstein, line 26:
```suggestion
        refreshId: string,
```

docs/interfaces/auth-service.txt
```diff
@@ -13,8 +13,42 @@ Interface Token {
...
+        [name: string]: string
+    }
+}
+
+/**
+* Describes an http-response, which is sent back to any requesting service.
+*/
+Interface Response <T> {
+    // Optional headers, which are set in an http-response
+    HttpHeaders: {
```
FinnStutzenstein, line 36:
```suggestion
    httpHeaders: {
```

docs/interfaces/auth-service.txt
```diff
@@ -49,7 +83,7 @@ Login (username: string, password: string): (Token, Cookie);
 *
 * @throws InvalidCredentials
 */
-Authenticate (token: Token, cookie: Cookie): {userId: number; sessionId: string;};
+who-am-i (Ticket): Response<void>;
```
FinnStutzenstein, line 86:
```suggestion
who-am-i (ticket: Ticket): Response<void>;
```

Same in other places

docs/interfaces/auth-service.txt
```diff
@@ -61,44 +95,44 @@ Authenticate (token: Token, cookie: Cookie): {userId: number; sessionId: string;
...
-ListSessions (token: Token, cookie: Cookie): string[];
\ No newline at end of file
+api/list-sessions (Ticket): Response<string[]>;
+
+/**
+* Hashes a given value with an optional salt. A random salt (64bit) is generated, if a salt is not given.
+*
+* @returns the hashed value. The hashed value is structured as follows: [salt + hash].
+*/
+hash (toHash: string, salt?: string): Response<{hash: string}>;
```
FinnStutzenstein, line 130: Use `oldHash` or so instead of salt. The user of you service does not know how you hash is composed

FinnStutzenstein: In combination with `is-quals`, you do not need a salt here at all. This would than always give a random salt. For new passwords, this is ok.

docs/interfaces/auth-service.txt
```diff
@@ -61,44 +95,44 @@ Authenticate (token: Token, cookie: Cookie): {userId: number; sessionId: string;
...
+
+/**
+* Compare a given value with an given hash.
+* An optional salt can be passed, too.
+*
+* @returns a boolean, if the hashed value of the given value is equals to the passed hash.
+*/
+is-equals (toHash: string, toCompare: string, salt?: string): Response<{isEquals: boolean}>;
```
FinnStutzenstein, line 138: Leave `salt` out here and get it from `toCompare`.

docs/interfaces/auth-service.txt
GabrielInTheWorld: Why should I need an `oldHash` here? Another service can easily get a hashed value. There is `is-equals` to check, if the hash of a given value  is equals to a stored hash.

FinnStutzenstein: Sorry, the reference is outdated. When I wrote the comment, the interface was different.

FinnStutzenstein (approved):
Ok, this looks good now. I still have one request:

why is there sometimes an `api/`. Please add a comment for this. Also add for every route the exact url, e.g. for `login` add to the docstring `POST to /system/auth/login` or so.

PR #5499, opened 2020-08-11
Add calculated fields

FinnStutzenstein:


docs/models.txt
```diff
@@ -306,6 +306,7 @@ Interface meeting {
     present_user_ids: (user/is_present_in_meeting_ids)[];
     temporary_user_ids: (user/meeting_id)[];
     guest_ids: (user/guest_meeting_ids)[];
+    user_ids: (Id)[]; // Calculated: All ids from temporary_user_ids, guest_ids and all users assigned to groups.
```
FinnStutzenstein, line 309: @normanjaeckel Should this get a reverse relation? In general: Should calculated fields get a reverse relation (if possible)?

docs/models.txt
```diff
@@ -425,6 +426,7 @@ Interface motion {
     sort_child_ids: (motion/parent_id)[];
     origin_id: motion/derived_motion_ids; // Note: The related motions may not be in the same meeting
     derived_motion_ids: (motion/origin_id)[]; // Note: The related motions may not be in the same meeting
+    forwarding_tree_motion_ids: (Id)[]; // Calculated: All children (derived_motion_ids), grand children, ... and all parents (origin_id).
```
FinnStutzenstein, line 429: Same here.

docs/models.txt
normanjaeckel: I hope not ...

FinnStutzenstein: I currently also do not see the advantage, so for me it is ok.

PR #5523, opened 2020-08-23
Update models.txt

ostcar:


docs/models.txt
```diff
@@ -430,7 +430,7 @@ Interface motion {
     forwarding_tree_motion_ids: (Id)[]; // Calculated: All children (derived_motion_ids), grand children, ... and all parents (origin_id).
     state_id: motion_state/motion_ids;
     recommendation_id: motion_state/motion_recommendation_ids;
-    recommendation_extension_reference_ids: ({motion}/referenced_in_motion_recommendation_extension_ids)[];
```
FinnStutzenstein, line 433: This is on purpose: It is a generic relation, but currently only populated by motions. We leave the possiility open to add more models here.

ostcar: Fixed

FinnStutzenstein (approved):

PR #5541, opened 2020-09-03
Add vote delegation

jsangmeister:
New fields on model `user`:
- `vote_delegated_to`/`vote_delegated_to_id`
- `vote_delegated_from_users`/`vote_delegated_from_users_id`
Both are read- and writeable for updates.

Adjusted the `vote` view(s), the new payload is:

    {
        "data": <old-payload>
        "user_id": int  // optional
    }

New calculated field on `BasePoll`: `user_has_voted_for_delegations`, which contains all user ids which have delegated their votes to the current user and have already voted.

Adjusted tests and wrote new tests for vote delegation.


server/openslides/users/serializers.py
```diff
@@ -119,6 +134,9 @@ def create(self, validated_data):
         inform_changed_data(user)
         return user
 
+    def get_vote_delegations_id(self, user):
```
FinnStutzenstein, line 137: Maybe be consistent with the server naming and change it to `get_vote_delegated_from_id`.

server/openslides/users/views.py
```diff
@@ -150,6 +150,10 @@ def update(self, request, *args, **kwargs):
             if str(request.user.pk) != self.kwargs["pk"]:
                 self.permission_denied(request)
 
+            # Only admins can set vote delegation
```
FinnStutzenstein, line 153: This should not be necessary. See some lines below: All keys, that are not `("username", "about_me", "email")` will be deleted.

server/openslides/poll/views.py
```diff
@@ -198,25 +199,40 @@ def vote(self, request, pk):
         if isinstance(request.user, AnonymousUser):
             self.permission_denied(request)
 
-        # check permissions based on poll type and handle requests
-        self.assert_can_vote(poll, request)
-
+        # data format is:
+        # { data: <vote_data>, [user_id: int] }
+        # if user_id is not given, the operator is assumed
```
FinnStutzenstein, line 204: Please add, that if the is is given, the request user votes for the given user.

server/openslides/poll/views.py
```diff
@@ -246,18 +262,21 @@ def assert_can_vote(self, poll, request):
...
                 self.permission_denied(request)
 
             try:
-                self.add_user_to_voted_array(request.user, poll)
+                self.add_user_to_voted_array(vote_user, poll)
                 inform_changed_data(poll)
             except IntegrityError:
                 raise ValidationError({"detail": "You have already voted"})
 
+        if request.user != vote_user and request.user != vote_user.vote_delegated_to:
```
FinnStutzenstein, line 277: can this check be moved up before the checking, if the user is present and in the right groups? Because than there must not be done any rollback by the DB since nothing was created.

server/openslides/users/serializers.py
```diff
@@ -97,6 +103,15 @@ def validate(self, data):
         if "about_me" in data:
             data["about_me"] = validate_html_strict(data["about_me"])
 
+        if "vote_delegated_to" in data:
```
FinnStutzenstein, line 106: Here we are missing two autoupdates. If the id changed from 2 to 3, we need autoupdates for users 2 and 3, since the reverse relation needs to be updated. See https://github.com/OpenSlides/OpenSlides/blob/master/server/openslides/motions/serializers.py#L511 for reference (there we are doing it for motion block). You should be able to test this by either capturing the autoupdate in tests or accessing the element cache and get the data for each user and validate the ids.

FinnStutzenstein (changes requested):
Some tests are failing. For the
```
assert count_queries(User.get_elements)() == 3
E       AssertionError: assert 14 == 3
```

error, look at the `UserManager`. The import errors seem strange, but you will figure it out I guess ;)

server/openslides/users/serializers.py
jsangmeister: That was my first idea, too, but it bothered me that the plural is lost there and from the name you just assume its a single id.

server/openslides/poll/views.py
jsangmeister: Yes, good catch

server/openslides/users/serializers.py
FinnStutzenstein: This is a pattern all in OS3. It would be clear if instead of using `_id`, use `users_id`.

jsangmeister: So `vote_delegated_from_users_id`?

server/openslides/users/views.py
jsangmeister: I thought that it was clearer to raise a 403 if a user tries to update something which he is not allowed to, but I dind't realise that that is currently not the case for the other fields, either. But if I leave it out, the following request would return a 200 without actually updating something:

    client.patch(
        reverse("user-detail", args=[user.pk]),
        {"vote_delegated_to_id": admin_pk},
    )

Is that intended? It seems very misleading to me.

server/openslides/users/serializers.py
FinnStutzenstein: ok

server/openslides/users/views.py
FinnStutzenstein: Hm. BTW: This is the same argument as norman always says for OS4 ;)

Yes, it is not nice I guess. Maybe - if you are working on it anyways - always raise an error, if fields are given, which are not expected.

FinnStutzenstein: Note: This has to be checked in the client! (Possibly make explicit request data with only allowed keys)

jsangmeister: Mmh, I fear that I break something when doing this xD if it's consistent I'd say we leave it as is. It works and we don't want to put too much work into os3 ;)

FinnStutzenstein: Ok, fair enough.

jsangmeister:
Now sets up on #5555. The `Vote` model now has a field `delegated_user` which either holds the same user as `user` (when no vote delegation was used) or the user who actually gave the vote in the case of vote delegation.

tsiegleauq:
currently, all requests to the servrs user model are causing an error:

```
[2020-09-14 15:53:59 +0200] [47064] [WARNING] openslides.requests.errors [gQqm] /rest/users/user/25/ 1 unknown exception: User matching query does not exist.
[2020-09-14 15:53:59 +0200] [47064] [ERROR] django.request Internal Server Error: /rest/users/user/25/
openslides.users.models.User.DoesNotExist: User matching query does not exist.
[2020-09-14 15:53:59 +0200] [47064] [INFO] openslides.utils.timing [gQqm] request: 1: 0.03818 2: 0.00001 sum: 0.03818
[2020-09-14 15:53:59 +0200] [47064] [ERROR] django.channels.server HTTP PUT /rest/users/user/25/ 500 [0.05, 127.0.0.1:35152]
```

jsangmeister:
Fixed.

tsiegleauq:
depends on #5555

tsiegleauq:
rebased on current master, added client side.
can be tested @emanuelschuetze

FinnStutzenstein:
> depends on #5555



> rebased on current master, added client side.
> can be tested @emanuelschuetze

Merged #5555. @tsiegleauq You can renew the branch I guess.

tsiegleauq:
@jsangmeister @FinnStutzenstein 
There appears to be something a little more severe left, or I am simply using it wrong.
To determine if a delegated user has already voted on a given poll, I check if the users (the delegation) id exists in the `voted_id` array.
( `rest/motions/motion-poll/` `"voted_id": []` - same for assignments )
during voting, the voted_id list is only readable by the admin. It appears not even to be a permission problem, only administrators get to see that list.
Users need to access this, so we can prevent that our users try to vote twice for their delegation

emanuelschuetze:
Example: A delegates their vote to B. B can vote for A and B.

1. A is NOT allowed to vote! -> Hide voting buttons!
1. Set A to not present. -> B cannot vote for A. That's wrong! It's always valid to vote for A because A cannot be present. (message "You have to be present to vote on a poll." is NOT required)
1. Login as B and set B to not present. Start Voting. Set B to present.-> B gets blue voting prompt but see now voting buttons! Reload is required. This is critical because users need to set present while voting is already started. Then voting buttons have to appear without reload (as status quo in master).

tsiegleauq:
2. is invalid, it appears you are not on the current commit.

jsangmeister:
> during voting, the voted_id list is only readable by the admin.

This is intentional since else everyone could see everyone else who voted. The relevant field before was `has_user_voted`, I will look into how to handle the vote delegation here.

jsangmeister:
@FinnStutzenstein New test is failing because a user update is missing in the cache... Do I have to manually flush the cache in the tests or something?

server/openslides/poll/access_permissions.py
```diff
@@ -71,8 +72,21 @@ async def get_restricted_data(
         """
 
         # add has_voted for all users to check whether op has voted
+        # also fill user_has_voted_for_delegations with all users for which he has
+        # already voted
+        user_data = await element_cache.get_element_data(
```
FinnStutzenstein, line 78: This does currently not work (sometimes, with the right (wrong?) conditions):

First, do not raise errors here. The restricters should not have a reason to have invalid data and there is no other reason to throw errors. So you have to expect the user_data to be available. Proposal: Do an error log here and set `vote_delegated_from_ids` to `[]` and proceed.

The seconds, more subtile but sevre part is the cache building. If the cache is empty, it will be filled (cache.py:125). Since last year all models were independent, so the order is not important. The code can build motions first and then the categories or vice versa. With the vote-weight this changed: https://github.com/OpenSlides/OpenSlides/blob/master/server/openslides/poll/models.py#L227

This line caused a lot of trouble because now we have a dependency from */*_poll to core/config. This resulted in this commit: https://github.com/OpenSlides/OpenSlides/commit/b78372f8a3eb51c5c4e1c5753f500a46416900a0

I manually build the config first, so it is available (the `config[key]` accesses the `element_cache`). With the marked line, we now have a dependency from */*_poll to user. In theory this is not a problem as we have a tree as the dependency graph, but this is not implemented in OS3 yet, since we just didn't had dependencies on serialization, except the one for config. But the config was always a bit special...

Note: Normally We surrounded this problem by using serializermethodfields, but here I see no possiblity to inject `vote_delegated_from_users_id` from all users into all polls only to remove the dependency.

-> I think that your approach is good and works, but we have to manage the dependencies. Proposal: Hard-code the dependencies in the cache.py. The nicer approach would be something like done in the migrations, that we write into a model class that it needs another model class before objects can be serialized. Then the build cache first does the config, than users, than the rest.

Note that this not the only place where models are serialized: A wrong order can potentially cause problems e.g. here: https://github.com/OpenSlides/OpenSlides/blob/master/server/openslides/utils/autoupdate.py#L107 (I wonder myself why I didn't care about this case back then; I just overlooked I guess).

Please think about it and maybe we should speak about possible solutions.

FinnStutzenstein: Another after-thought: When updating a user (or just the `vote_delegated_from_users_id` field), you have to re-serialize all polls. The same is also true vor the vote weight config, but I guess this was missed, too. In fact that we need this information I wonder, if we can put it into the user model instead of polls. This solves the user-update problem, but then there must be autoupdates for the voted user, when a vote is received (more autoupdates here).

tsiegleauq:
Fixed the progress bar and rebased to two commits.
whats left:
 - Having the server side tests fixed

server/openslides/poll/access_permissions.py
FinnStutzenstein: Ok, it is not as severe as I though: Since it is in the restricters this code is actually not called when the cache is built. These fields even doesn't exist there. The config-thing is serialization-related, but this access to the cache is restricter-related. But this needs to be watched carefully...

FinnStutzenstein (approved):

jsangmeister:
@FinnStutzenstein did you change anything or does something still need to be changed? Tests seem to work now...

tsiegleauq:
I just assume that this one is done. Can be tested / merged

jwinzer:
There is a spelling error here: Should be delegations:
https://github.com/jsangmeister/OpenSlides/blob/vote-delegation/client/src/app/site/users/components/user-detail/user-detail.component.html#L324

FinnStutzenstein:
> did you change anything or does something still need to be changed? Tests seem to work now...

It works now. The test failed since this line was missing https://github.com/jsangmeister/OpenSlides/blob/vote-delegation/server/tests/integration/motions/test_polls.py#L905

With the large comment: We were lucky, that we accessed it in the resturctes. I thought they would be used when building up the cache, but no. This saves us a lot of work. (It is like two-staged: first, the cache is built, than the data is restricted on a build cache)

PR #5555, opened 2020-09-10
Fix non-nominal voting on assignment polls with type votes

jsangmeister:


jsangmeister:
Should I write a migrations that gets rid of all user references in these kind of votes?

server/openslides/assignments/views.py
```diff
@@ -493,36 +493,28 @@ def create_votes_type_votes(self, data, poll, user):
...
+            weight = vote_weight if config["users_activate_vote_weight"] else Decimal(1)
             vote = AssignmentVote.objects.create(
-                option=option, user=user, weight=weight, value=data
+                option=option, user=vote_user, weight=weight, value=data
             )
             inform_changed_data(vote, no_delete_on_restriction=True)
             inform_changed_data(option)
             inform_changed_data(poll)
 
-        poll.voted.add(user)
```
FinnStutzenstein, line 513: Where is this `add` now?

jsangmeister: This is done in the `assert_can_vote` method to have it handled the same way for all `vote` requests, see https://github.com/OpenSlides/OpenSlides/blob/04477d9ebd0488f35ecfbb02b8a548306a596251/server/openslides/poll/views.py#L237 (and ff.)

FinnStutzenstein (approved):

FinnStutzenstein:
> Should I write a migrations that gets rid of all user references in these kind of votes?

No, I think it is not necessary

PR #5571, opened 2020-09-18
Add models.simple.yml

ostcar:


ostcar:
@normanjaeckel Please have a look.

The models.simple.yml has the same content then the models.txt, but in yml format.

The models.yml is the file where every change should be done. I started manually with the role- and user-model. Since you go to all models to add your fields, it should not be to hard to also add the type fields, where they are needed.

The comments at the top should also be changed.

When everything is finished, we can remove the models.txt and the models.simple.yml

ostcar:
Norman creates a new pr

PR #5587, opened 2020-09-29
Fix branch.

normanjaeckel:


normanjaeckel:
@ostcar Can you please fix the validator and enable the `HTMLVideo` type?  Thanks.

FinnStutzenstein:
What does `through` do?

normanjaeckel:
If you have a structured field you have to define how to find the content of this field. E. g. in an poll option the placeholder represents a meeting id which can be find by following the foreign key poll_id from option to poll where you then have the field meeting_id.

PR #5593, opened 2020-10-01
Adjust models.yml

### LINKED ISSUE #5590: Rename some fields in the model.yml
- `HTML` and `HTMLVideo` should be named `HTMLStrict` and `HTMLPermissive` as @normanjaeckel already suggested, since the point will most certainly come when we add tags to the permissive type and then the name doesn't fit
- `datetime` should be called `timestamp` since we save it as a timestamp

Feel free to add other suggestions here.

jsangmeister:
fixes #5590 

Adds `on_delete` fields to the model description. Still missing on many fields, but enough for first testing purposes. Rest can be added later.

### DIFF
--- a/docs/models.yml
+++ b/docs/models.yml
@@ -1,12 +1,12 @@
 ---
 # Types:
 #     - Nativ datatypes: string, number, boolean, JSON
-#     - HTML: A string with HTML content.
-#     - HTMLVideo: A string with HTML content (with video tags).
+#     - HTMLStrict: A string with HTML content.
+#     - HTMLPermissive: A string with HTML content (with video tags).
 #     - float: Numbers that are expected to be non-integer. Formatted as in rfc7159.
 #     - decimal(X): Decimal values represented as a string with X decimal places.
 #       At the moment we support only X == 6.
-#     - datetime: Datetime as a unix timestamp. Why a number? This enables queries
+#     - timestamp: Datetime as a unix timestamp. Why a number? This enables queries
 #       in the DB. And we do not need more precision than 1 second.
 #     - <T>[]: This indicates and arbitrary array of the given type. At the moment
 #       we support only some types. You can add JSON Schema properties for items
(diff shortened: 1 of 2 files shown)


docs/models.yml
```diff
@@ -258,7 +264,7 @@ committee:
 meeting:
   id: number
   welcome_title: string
-  welcome_text: HTMLVideo
+  welcome_text: HTMLStrictPermissive
```
normanjaeckel, line 267: This is wrong.

docs/models.yml
```diff
@@ -840,7 +846,7 @@ topic:
   title:
     type: string
     required: true
-  text: HTMLVideo
+  text: HTMLStrictPermissive
```
normanjaeckel, line 849: This is wrong.

FinnStutzenstein (approved):

docs/models.yml
jsangmeister: ops, dont know how that happened. my bad

jsangmeister:
I think @ostcar (or someone else?) has to fix the validator so this can be merged

ostcar:
The validator failed, because you added a lot of new types.

Can you tell me what types I should add? If I see correctly it is `timestamp`, `HtmlPermissive` and `HTMLStrict`. Is this correct? Anything else?

jsangmeister:
> Can you tell me what types I should add? If I see correctly it is `timestamp`, `HtmlPermissive` and `HTMLStrict`. Is this correct? Anything else?

That's correct. I guess you can remove the old types (`datetime`, `HTML`, `HTMLVideo`) since we don't need them anymore. Also it's all upper case: `HTMLStrict`/`HTMLPermissive`

ostcar:
Done. Please use the version v0.1.1 by changing this line: https://github.com/OpenSlides/OpenSlides/blob/a7248c2399942db9864c538acd3467ccdacdabb2/.github/workflows/models.yml#L14

PR #5620, opened 2020-10-08
Add on_delete and meeting_id for all models

jsangmeister:
Please check if I missed something. Also, feel free to suggest other places that should have delete cascading/protection, so that I can add it here.

### DIFF
--- a/docs/models.yml
+++ b/docs/models.yml
@@ -158,6 +158,7 @@ user:
     fields:
       type: relation-list
       to: personal_note/user_id
+    on_delete: CASCADE
   supported_motion_$_ids:
     type: template
     replacement: meeting_id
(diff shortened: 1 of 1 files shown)


FinnStutzenstein:
The default is something like SET_NULL, right?

Protected:

- deleting a comment section if it has comments
- deleting a projector if it has projection defaults
- protect state, if it is used by a motion (since workflow->state cascade: A workflow deletion is also prevented by this; this should be tested in your implementation, that it works).

Cascade:

- deleting motion comments, if a motion is deleted
- commitee -> meetings
- meeting -> all meeting models
- user -> personal note
- motion -> amendments (I think, not sure)
- motion -> submitter
- motion -> change recommendation
- motion -> polls
- assignment -> polls
- assignment -> candidates
- projector -> projection
- workflow -> state
- *polls -> *options -> *votes

jsangmeister:
Changed. Also added the `meeting_id` field to all models.

> * motion -> amendments (I think, not sure)

leaving it out for now.

What about motion -> personal_note, should this also cascade?

jsangmeister:
Also added `equal_fields` (see https://github.com/OpenSlides/openslides-backend/issues/191), which was needed in more places than I thought. Please double check.

@ostcar I was about to ask you to add `equal_fields` to the validator, but it passes, so instead we should probably restrict the possible fields for each node, I think.

FinnStutzenstein:
> What about motion -> personal_note, should this also cascade?

Right. Note that it is a generic relation, so it's more like *->personal_note

jsangmeister:
Merging this now, fixes can be done in a later PR. @normanjaeckel please check `user/vote_delegations_$_from_ids/user/vote_delegated_$_to_id`, I do not know if this is the right syntax for the template fields, but it works for now. It did not work with the previous version from @FinnStutzenstein.

PR #5632, opened 2020-10-13
Change weights and add defaults

jsangmeister:
Anything missing?


FinnStutzenstein:
I think it is ok for now. Maybe we find something else, but we can make another PR if it occurs.

Issue #5638, opened 2020-10-18
Install via docker-compose fails 


simonla82:
**Describe the bug**
When following the [Installation guide](https://github.com/OpenSlides/OpenSlides) the build of openslides/openslides-server:latest aborts at `Step 17/29 : RUN pip install -r requirements/production.txt -r requirements/big_mode.txt     -r requirements/saml.txt &&     rm -rf /root/.cache/pip` with error:

<details>

```
Step 17/29 : RUN pip install -r requirements/production.txt -r requirements/big_mode.txt     -r requirements/saml.txt &&     rm -rf /root/.cache/pip
[...]
Collecting xmlsec>=0.6.0
  Downloading xmlsec-1.3.8.tar.gz (59 kB)
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Installing backend dependencies: started
  Installing backend dependencies: still running...
  Installing backend dependencies: still running...
  Installing backend dependencies: finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /usr/local/bin/python /usr/local/lib/python3.7/site-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-8nyxbpvh/normal --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- pkgconfig 'lxml>=3.8'
       cwd: None
  Complete output (190 lines):
  Collecting pkgconfig
    Using cached pkgconfig-1.5.1-py2.py3-none-any.whl (12 kB)
  Collecting lxml>=3.8
    Downloading lxml-4.6.1.tar.gz (3.2 MB)
  Building wheels for collected packages: lxml
    Building wheel for lxml (setup.py): started
    Building wheel for lxml (setup.py): still running...
    Building wheel for lxml (setup.py): finished with status 'error'
    ERROR: Command errored out with exit status 1:
[...]

PR #5651, opened 2020-10-21
Update Dockerfile to allow build via docker-compose on arm architectures

simonla82:
As stated in issue #5638 the installation via docker-compose fails on arm architectures since two librariers are not correctly installed. This is fixed by installing those manually in the /server/docker/Dockerfile

### DIFF
--- a/server/docker/Dockerfile
+++ b/server/docker/Dockerfile
@@ -29,7 +29,9 @@ RUN apt-get -y update && apt-get -y upgrade && \
     rsync \
     wait-for-it \
     wget \
-    xz-utils
+    xz-utils \
+    zlib1g-dev \
+    libffi-dev # install of zlib1g-dev and libffi-dev is required on armhf architectures
 
 # Install additional tools
 RUN apt-get install --no-install-recommends -y \


server/docker/Dockerfile
```diff
@@ -29,7 +29,9 @@ RUN apt-get -y update && apt-get -y upgrade && \
     rsync \
     wait-for-it \
     wget \
-    xz-utils
+    xz-utils \
+    zlib1g-dev \
```
FinnStutzenstein, line 33: Please add a comment, that both packages are needed for the armfh architecture.

FinnStutzenstein (approved):

FinnStutzenstein:
@simonla82 ping

server/docker/Dockerfile
simonla82: Ok, after some struggle with git I added the required comment...

FinnStutzenstein:
HI, I fixed the git history - can you verify, that the changes are the ones you intended?

simonla82:
Thank you! Looks good to me.

PR #5652, opened 2020-10-22
Split motion comments migration to separate schema and data migrations

FinnStutzenstein:
Similar to #4266


FinnStutzenstein:
Is not needed for current instances.

PR #5666, opened 2020-11-01
Add EMAIL_USE_TLS and EMAIL_USE_SSL to settings

m-ober:
Add EMAIL_USE_TLS und EMAIL_USE_SSL to settings.py.

This allows to use mailservers which require a TLS/SSL connection.


FinnStutzenstein:
Obsolete by #5658

PR #5696, opened 2020-11-10
Enhance the action interface

FinnStutzenstein:
@jsangmeister @normanjaeckel I make this proposal regarding some changes.

- https://github.com/OpenSlides/openslides-backend/issues/261 We have three cases, what can be returned by calling `handle_request`:
    1) No error, everything is fine: An object with `success=true` is returned with a result array for each action containing an array for each payload of each action.  Examples are given in the interface
    2) One action had an error: An object is returned with `success=false`, the `error_index` of the error-causing action and a message.
    3) A general error occured (->`ActionException`): An object with `success=false` but no `error_index` is returned.

The thing is, that everytime an object is returned and it is easy decideable, which of these cases happened (not the current case).

- https://github.com/OpenSlides/openslides-backend/issues/259 The interface is extended (or recreated - the original version hat the result, but someone removed it...), so every action and "partial"-action can provide a result with the flexibility to not give a bunch of null-arrays by omitting them.

Please discuss here - If we decided to stick to an (hopefully final) interface, we all know what to expect from the service, so it can universally be used without any quirks and special return structures.

### DIFF
--- a/docs/interfaces/action-service.txt
+++ b/docs/interfaces/action-service.txt
@@ -3,11 +3,11 @@
 /**
  * Executes multiple actions in the context of the user given by the user_id.
  * There are two modes of execution:
- * single_validation=false (default):
+ * atomic=false (default):
  *   All actions are validated in common, so if one action or one payload of
  *   one action fails, the request is aborted with an ActionException indicating
  *   the problematic action with both indices.
- * single_validation=true:
+ * atomic=true:
  *   Each action is validated by it's own. If there is an error, the error must
  *   be reported via an ActionError in the ActionsResponse. The actions result
  *   is not written into the Datastore. It might raise an ActionException if the
@@ -22,7 +22,7 @@ handle_request(payload: Action[], user_id: Id): ActionsResponse
 interface Action {
     action: string;
     data: object[];
-    single_validation?: boolean;
+    atomic?: boolean;
 }
 
 interace ActionsResponse {
@@ -59,7 +59,7 @@ interface ActionError {
  * use the action_error_index and action_payload_error_index to indicate the errored
  * action and payload. If there were general errors, both indices must be omitted.
  *
- * If the single_validation Flag was true in the request, it is not allowed to send
+ * If the atomic flag was true in the request, it is not allowed to send
  * action-specific errors with this exception. It must be responded through ActionsResponse.
  */
 Exception ActionException {
--- a/docs/interfaces/action-service.txt
+++ b/docs/interfaces/action-service.txt
@@ -1,33 +1,28 @@
-# Actions Service Interface
-
-/**
- * JSON resonse with a status code of !=200:
- * {
- *     success: false;
- *     message: string;
- * }
- *
- */
-Exception ActionException(message: string);
+// Actions Service Interface
 
 /**
  * Executes multiple actions in the context of the user given by the user_id.
- * All actions are depdend: If one fails, nothing is written to the DataStore.
- * It can be seen as one transaction.
- *
- * For each action an ActionResult is returned. If an action has an error,
- * ActionError is used in combination wuth error_index to indicate, which action
- * resulted in an error.
+ * There are two modes of execution:
+ * single_validation=false (default):
+ *   All actions are validated in common, so if one action or one payload of
+ *   one action fails, the request is aborted with an ActionException indicating
+ *   the problematic action with both indices.
+ * single_validation=true:
+ *   Each action is validated by it's own. If there is an error, the error must
+ *   be reported via an ActionError in the ActionsResponse. The actions result
+ *   is not written into the Datastore. It might raise an ActionException if the
+ *   single write request to the Datastore fails (e.g. because of locking there)
  *
  * For general, non specific action-related, errors an ActionException is used.
  *
  * @throws ActionException
  */
-handle_request (payload: Action[], user_id: Id): ActionsResponse | ActionError
+handle_request(payload: Action[], user_id: Id): ActionsResponse
 
 interface Action {
     action: string;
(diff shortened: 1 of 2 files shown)


normanjaeckel:
I will try to implement it and then we can see. No problem if you propose changes inbetween. Then I may change the code again (on my risk).

One thing: Can we please rename `ActionResult` to `ActionResponse`? So I can use this reponse for the object that is send to the client where es my WriteRequestElements are send to the datastore and I can handle both inside a container object called `ActionResult`.

normanjaeckel:
Maybe the error case is not that easy to provide. At the moment we process all actions and bundle all write requests to one big write request that is send to the datastore, containing severel create and update events. What if the datastore answers with an error like model is locked. How can we know which action failed?

normanjaeckel:
See https://github.com/OpenSlides/openslides-backend/pull/286

FinnStutzenstein:
> What if the datastore answers with an error like model is locked. How can we know which action failed?

This would be an `ActionException`. If you do not really know which action failed, you can use it.

FinnStutzenstein:
> Can we please rename ActionResult to ActionResponse

Done

jsangmeister (approved):

FinnStutzenstein:
@normanjaeckel if you approve, we can merge it.

normanjaeckel:
I am not sure: It might be better to merge ActionException and ActionError together: Also in ActionError case I would like to respond with HTTP 4xx but also provide the requested index your the client (if possible). Internal I would like to use a raised exception to do this.

FinnStutzenstein:
Ok, 4xx in ActionError case is ok.

FinnStutzenstein:
Another thing: `error_index` is not enough. We need two indices: `action_error_index` and `action_payload_error_index` -> For multiselect/bulk actions we send (normally) one action with many payloads, e.g. motion.set_state -> We need to know which action (ok, her we have just one, easy) and which payload created the issue (e.g. there is one motion in the wrong workflow...). From this info, we can print the user a detailed error message, which motion (in this case) is the bad one.

FinnStutzenstein:
@jsangmeister I added the `single_validation` flag. If you have a better name, please suggest it. I also adjusted the usage of `ActionException` vs. `ActionError`. Please read the docs and look at the structure. This is what we discussed; I think I did not missed anything.

FinnStutzenstein:
@GabrielInTheWorld This should fit the needs of importing things on the client. Please have a look and we should talk about it.

jsangmeister:
> If you have a better name, please suggest it.

How about `atomic`? This would be analogous to the use in database systems.

FinnStutzenstein:
> How about atomic?

Even single_validation=true is atomic in the sense that the things written to the Datastore (if it succeeds) are atomic, but not everything is included.

jsangmeister:
It would be
- `atomic=true`: All payloads either succeed or fail together
- `atomic=false`: The payloads are processed one by one
I think it's ok this way. The atomic would describe the action as a whole, not the write requests to the datastore, maybe that's what confusing to you.

FinnStutzenstein:
Ok for me

docs/interfaces/action-service.txt
```diff
@@ -0,0 +1,70 @@
...
+ *   Each action is validated by it's own. If there is an error, the error must
+ *   be reported via an ActionError in the ActionsResponse. The actions result
+ *   is not written into the Datastore. It might raise an ActionException if the
+ *   single write request to the Datastore fails (e.g. because of locking there)
+ *
+ * For general, non specific action-related, errors an ActionException is used.
+ *
+ * @throws ActionException
+ */
+handle_request(payload: Action[], user_id: Id): ActionsResponse
```
normanjaeckel, line 20: The atomic flag should be here and not inside the Action interface.

FinnStutzenstein: @jsangmeister I think norman is right. ok?

jsangmeister: Yes

FinnStutzenstein:
For me t is fine. If @normanjaeckel and @jsangmeister give their OK, we can merge.

jsangmeister (approved):

docs/interfaces/action-service.txt
```diff
@@ -0,0 +1,69 @@
...
+    message: string;
+}
+
+/**
+ * JSON resonse with a status code of !=200. If a specific action raised the error,
+ * use the action_error_index and action_payload_error_index to indicate the errored
+ * action and payload. If there were general errors, both indices must be omitted.
+ *
+ * If the atomic flag was true in the request, it is not allowed to send
+ * action-specific errors with this exception. It must be responded through ActionsResponse.
```
normanjaeckel, line 62: Typo `ActionError` and not `ActionResponse`

docs/interfaces/action-service.txt
```diff
@@ -0,0 +1,69 @@
...
+ */
+Exception ActionException {
+    success: false;
+    message: string;
+    action_error_index: number;
```
normanjaeckel, line 67: Please allow me to send `null` instead of omitting the property so I can save some time when parsing / coding this.

normanjaeckel (changes requested):

docs/interfaces/action-service.txt
FinnStutzenstein: Yes and no. **through** `ActionResponse` is correct but I really meant that there must be an `ActionError` in the `ActionResponse`. I made this description a bit more clear.

FinnStutzenstein:
@normanjaeckel done.

FinnStutzenstein:
Since there was no reaction, I'll merge it. Please adjust the backend accordingly

PR #5698, opened 2020-11-11
Restructure polls

FinnStutzenstein:
@jsangmeister This is the proposal for the poll restructure. Essentially each `motion_*` and `assignment_*` is merged together (or it is just the `assignment_*` version, since they had more features). Things to note:

- polls have now a `content_object`, which can be null to allow for standalone polls (Imagine a poll-list with just polls...)
- options have `text` -> This allowes for free-hand polls (analog to BBB)
- options have a `content_object` what to vote about. Either the content object or text must be given. E.g. the content object is a user (analog to assignment_poll), or a motion -> Vote about motions, or what ever to vote about. This results in a problem, since this is a generic relation to meeting-specific models and users. Please see the comment in the models.yml. I change the example-data.json, so you can see how it looks, if the proposal is accepted.
- I explicitly left the `assignment_poll_*` and `motion_poll_*` settings in the meeting, so one can customize the use cases differently - this is just a client thing, so it should be ok. I added also the generic form (`poll_*`) of that.
- I fixed some things in the example data, that were changed in the models.yml but the changes did not reflect into the example data. Don't be confused about it.

### DIFF
--- a/docs/example-data.json
+++ b/docs/example-data.json
@@ -53,18 +53,16 @@
         "supported_motion_$_ids": [],
         "submitted_motion_$_ids": ["1"],
         "submitted_motion_$1_ids": [1, 2, 3, 4],
-        "motion_poll_voted_$_ids": [],
-        "motion_vote_$_ids": [],
         "assignment_candidate_$_ids": ["1"],
         "assignment_candidate_$1_ids": [1],
-        "assignment_poll_voted_$_ids": ["1"],
-        "assignment_poll_voted_$1_ids": [3],
-        "assignment_option_$_ids": ["1"],
-        "assignment_option_$1_ids": [1, 2],
-        "assignment_vote_$_ids": ["1"],
-        "assignment_vote_$1_ids": [4],
-        "assignment_delegated_vote_$_ids": ["1"],
-        "assignment_delegated_vote_$1_ids": [4]
+        "poll_voted_$_ids": ["1"],
+        "poll_voted_$1_ids": [5],
+        "option_$_ids": ["1"],
+        "option_$1_ids": [3, 4],
+        "vote_$_ids": ["1"],
+        "vote_$1_ids": [7],
+        "vote_delegated_vote_$_ids": ["1"],
+        "vote_delegated_vote_$1_ids": [7]
     },
     {
         "id": 2,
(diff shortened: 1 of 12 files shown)


docs/models.yml
```diff
@@ -1505,56 +1500,85 @@ motion_poll:
...
+    #  field:
+    #    name: option_$_ids
+    #    type: structured-relation
+    #    replacement: meeting_id
+    type: generic-relation
+    to:
+      collection:  # Now, we have multiple models to vote about.
+        - motion
+        - assignment
+        - user  # This is problematic: Normally, all relations to user are structured...
```
jsangmeister, line 1539: I'm confused. Why can you vote about an assignment? Is this a new feature? And, the more important question, why did we save the user in the option? We explicitly create the `assignment_candidate` models which we could use to solve this problem here, or not?

FinnStutzenstein: > We explicitly create the assignment_candidate models which we could use to solve this problem here, or not?

No, assignment_candidates are used before an election - they are copied into the poll. You can have two polls in an assignment with different options <- This is why the users are in the options.

With a generic field, we have the possiblity to vote about other thinks if we want to.

FinnStutzenstein:
@jsangmeister Please check again.

docs/modelsvalidator/README.md
```diff
@@ -0,0 +1,29 @@
...
+Modelsvalidate is a tool to validate the models.yml file, that is used in the
+development process of OpenSlides 4.
+
+
+## Install
+
+The easiest way is to install it with go:
+
+```
+go get github.com/OpenSlides/openslides-modelsvalidate/cmd/modelsvalidate
```
jsangmeister, line 12: This is no longer valid, right?

docs/example-data.json
```diff
@@ -53,18 +53,16 @@
         "supported_motion_$_ids": [],
```
jsangmeister, line 53: Just a thought and does not belong to this PR, but should we build a validator for the example data? It would prevent errors such as https://github.com/OpenSlides/openslides-backend/issues/324 and shouldn't take to long on the basis of the `models.yml`. It would be especcially easy with the existing backend code (I think) but it should belong to this repo, like the models validator. What do you think, how long would this take in go, maybe an the basis of the models validator? I think it would be a helpful addition.

docs/models.yml
```diff
@@ -1505,56 +1516,76 @@ motion_poll:
...
+        - collection: motion
+          field: option_ids
+        - collection: topic
+          field: option_ids
+        - collection: user
+          field:
+            name: option_$_ids
+            type: structured-relation
+            replacement: meeting_id
+    equal_fields: meeting_id
```
jsangmeister, line 1554: Ok, this has to be implemented in the backend first before we can merge this...

(this is a new format, right? Or were structured generic relations possible before in the models code? Because you did not change the docs above)

docs/models.yml
```diff
@@ -23,8 +23,8 @@
 #       `motion/category_id`. The type indicates that there are many
```
jsangmeister, line 23: Did not check everything again, only the lines where I put a remark. If you changed something else, please ping me again ;)

jsangmeister (commented):
Did not check the go code.

docs/models.yml
FinnStutzenstein: Yes this is new. Before this, you have one field decribing the field for all collections - now you can give a field descriptor for each collection.

jsangmeister: Then please add this to the docs

FinnStutzenstein:
@jsangmeister Please merge, if you have thebackend ready to take the changed. Or @normanjaeckel May do it.

PR #5704, opened 2020-11-13
Update datastore specs

jsangmeister:
Current use case for partial list updates: `meeting/user_ids`. If this list contains 100.000 or more entries, it should not be needed to fetch the whole list from the datastore just to add one entry. Specs are adjusted specifically for this case. We may use it for the relation handling as well, if we want; we'd have to restructure some of the backend code (@normanjaeckel).

### DIFF
--- a/docs/interfaces/datastore-service.txt
+++ b/docs/interfaces/datastore-service.txt
@@ -84,13 +84,27 @@ Interface CreateEvent {
     }
 }
 
-// Note: For deleting keys, they must be set to `None`. These keys
-// will be removed from the model.
+// Note: For deleting keys, they must be set to `None`. These keys will be removed from
+// the model.
+// list_fields can be used to partially update list fields: the values in `add` will be
+// appended to the given field, the values in `remove` will be removed from the field.
+// The case of double adding or removing something invalid will be silently ignored.
+// All given list_fields must be flat arrays of strings or ints, else an error is
+// thrown.
+// Either fields or list_fields must be given or an error will be thrown.
 Interface UpdateEvent {
     type: 'update';
     fqid: Fqid;
     fields: {
         <field>: Value;
+    }
+    list_fields: {
+        add: {
+            <field>: Value[];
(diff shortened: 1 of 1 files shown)


normanjaeckel:
As I understand the using of list_fields is optional so we do not break anything at the moment. Correct?

Nevertheless please open an issue in backend repo after this work is done in datastore.

docs/interfaces/datastore-service.txt
```diff
@@ -84,14 +84,28 @@ Interface CreateEvent {
     }
 }
 
-// Note: For deleting keys, they must be set to `None`. These keys
-// will be removed from the model.
+// Note: For deleting keys, they must be set to `None`. These keys will be removed from
+// the model.
+// list_fields can be used to partially update list fields: the values in `add` will be
+// appended to the given field, the values in `remove` will be removed from the field.
+// The case of double adding or removing something invalid will be silently ignored.
```
FinnStutzenstein, line 91: "ignored" meaning it is done and not logged/noticed, or just not done? This needs to be more precise

docs/interfaces/datastore-service.txt
```diff
@@ -84,14 +84,28 @@ Interface CreateEvent {
...
+// All given list_fields must be flat arrays of strings or ints, else an error is
+// thrown.
+// Either fields or list_fields must be given or an error will be thrown.
```
FinnStutzenstein, line 94: I miss information about different behaviours:

- fields are given, that are not lists
- fields that are lists, but during comparisons the content are not strings or integers
- adding to a non-existing
- Removing from an empty list.

jsangmeister:
@normanjaeckel: Yes, correct, we can leave everthing as-is for now. It just could be more performant when dealing with long relation list fields. Will create an issue.

jsangmeister:
@FinnStutzenstein updated the description

FinnStutzenstein:
Ok!

PR #5708, opened 2020-11-18
Set relation Committee -> Meeting protected

r-peschke:
Deleting a committee should be protected, if there exist any meeting related to this committee


FinnStutzenstein:
@jsangmeister If this is the right place, please go ahead and merge. We thought that deleting a committee can be really damaging; so we will protect it.

PR #5713, opened 2020-11-19
Multi-select perms

jwinzer:
Multi-select and export should not require can manage.


emanuelschuetze:
I created a new PR - see #5732 which fixed the typo in assignments.can_manage permission. The multi select menu is for managers only as in other apps like motions. A normal user can filter and export (without multi select).

PR #5723, opened 2020-11-25
Change the category prefix to optional

FinnStutzenstein:
The required prefix was a mistake when I created the schema. Changing it to an optional field must be adopted in several places:

- @GabrielInTheWorld Can you do the client? I guess the create/update dialogs may be changed, so an empty prefix is allowed
- @reiterl Please change `motion_category.create` to make the prefix optional
- @reiterl Please change `motion_category.update` to accept `None` or `""` as the prefix to clear an existing prefix.
- @reiterl You can merge this PR if you are ready to update the generated models.py in the backend
- I'll adjust `motion_category.number_motions`.

### DIFF
--- a/docs/models.yml
+++ b/docs/models.yml
@@ -1246,7 +1246,6 @@ motion_category:
     required: true
   prefix:
     type: string
-    required: true
   weight:
     type: number
     default: 10000


GabrielInTheWorld:
I noticed it and will work on it next week.

jsangmeister:
Question: Does the `required` field have any meaning for other services? It is not used in the backend...

FinnStutzenstein:
This is exactly what I tried to ask @normanjaeckel about: https://github.com/OpenSlides/OpenSlides/issues/5743

If not, every unused field can be removed since it is just noise...

normanjaeckel:
The required field is used in the backend for generic actions to describe that this field can not be null or an empty string.

jsangmeister:
Ahh you're right, I overlooked that. To be more specific, it's used for the json schema generation (for instance, the type for a required integer field must be `integer`, but for an optionial field is has to be `["integer", "null"]`).

PR #5728, opened 2020-11-26
Add number_value to motion model.

reiterl:


jsangmeister (dismissed):
Please add a description.

FinnStutzenstein:
This should also be flagged as read only, since the client cannot write it, right?

jsangmeister:
Yes, I thought about that but wasn't sure what exactly the meaning of `read_only` was. If it is meant only as an indicator for the client, then yes, it has to be set to `read_only`. In the backend this flag is currently not checked.

FinnStutzenstein:
@reiterl please add `read_only: true`, so it can be merged.

reiterl:
Added read_only and description to motion/number_value.

PR #5769, opened 2020-12-10
Removed redundant information from models.yml

jsangmeister:


docs/models.yml
```diff
@@ -1629,10 +1569,7 @@ option:
         - collection: topic
```
FinnStutzenstein, line 1569: can this be
```
to:
  - motion/option_ids
  - ...
```

This would simplyfy the schema a lot, when only accepting `to: <string>` and `to: <string>[]`.

jsangmeister: Sure, for the backend this is irrelevant. Just for clarification: For fields with multiple collections which have the same name, do you still want to keep the schema:
```
to:
  collection:
    - a
   ...
  field: ...
```
or do you want to repeat the field name in each definition here?

FinnStutzenstein: > or do you want to repeat the field name in each definition here?

Would be simpler imo

FinnStutzenstein:
When this is merged please create an issue, that the extra schema can be removed from the validator.

docs/models.yml
jsangmeister: I kept it in to avoid repetition. Does not work as soon as one field falls out of line of course, e.g. for the projector. Anyway, should be ready now

jsangmeister:
@FinnStutzenstein the validator doesn't like the changes. Guess it'll have to wait until next year then :D

docs/models.yml
```diff
@@ -223,23 +217,13 @@ user:
     replacement: meeting_id
     fields:
       type: relation
-      to:
-        collection: user
-        field:
-          name: vote_delegations_$_from_ids
-          type: structured-relation
-          replacement: meeting_id
+      to: user/vote_delegations_$_from_ids
```
normanjaeckel, line 220: Replacement is missing. The replacement in line 217 refers to line 215. Do you want so see an implicit replacement here?

docs/models.yml
```diff
@@ -1623,16 +1565,9 @@ option:
...
-        - collection: topic
-          field: option_ids
-        - collection: user
-          field:
-            name: option_$_ids
-            type: structured-relation
-            replacement: meeting_id
+      - motion/option_ids
+      - topic/option_ids
+      - user/option_$_ids
```
normanjaeckel, line 1570: Replacement is missing. Here it is impossible to guess that meeting_id is meant.

normanjaeckel (changes requested):

docs/models.yml
normanjaeckel: The same for many other places ...

jsangmeister: The replacement is defined in the field `user/option_$_ids` as `meeting_id`

PR #5783, opened 2020-12-17
Add config defaults to models.yml settings.

reiterl:
Mostly add the config defaults to the models.yml settings.

Exclude special cases:
conference_auto_connect_next_speakers
user_pdf_url
user_email_body

This is the most setting defaults being set, but some are missing.

### DIFF
--- a/docs/models.yml
+++ b/docs/models.yml
@@ -299,16 +299,22 @@ committee:
 
 meeting:
   id: number
-  welcome_title: string
-  welcome_text: HTMLPermissive
+  welcome_title:
+    type: string
+    default: Welcome to OpenSlides
+  welcome_text:
+    type: HTMLPermissive
+    default: Space for your welcome text.
 
   # General
   name:
     type: string
     maxLength: 100
+    default: OpenSlides
   description:
     type: string
     maxLength: 100
+    default: Presentation and assembly system
   location: string
   start_time: timestamp
   end_time: timestamp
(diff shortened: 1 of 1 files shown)


docs/models.yml
```diff
@@ -458,35 +520,62 @@ meeting:
...
+    type: string
+    default: Motions
   motions_export_preamble: string
-  motions_export_submitter_recommendation: boolean
-  motions_export_follow_recommendation: boolean
+  motions_export_submitter_recommendation:
+    type: boolean
+    default: False
+  motions_export_follow_recommendation:
+    default: boolean
```
FinnStutzenstein, line 578: this is not right. The default is false, the type is boolean

FinnStutzenstein (commented):
Some more defaults:
- projector_default_countdown_time: 60
- projector_countdown_warning_time: 0
- conference_auto_connect_next_speakers: This is actually an integer (changed recently) with default 0
- list_of_speakers_amount_next_on_projector: should be an integer with default -1
- list_of_speakers_show_amount_of_speakers_on_slide: True
- motions_number_type: per_category
- motion_poll_default_type, assignment_poll_default_type, poll_default_type: analog
- *default_100_percent_base: YNA
- users_allow_self_set_present: False

FinnStutzenstein:
@reiterl Ping

reiterl:
I have added the missing setting-default, fix the "default: boolean" bug and rebased it.

FinnStutzenstein (approved):

PR #5790, opened 2021-01-02
models.yml: More clear enums

ostcar:


docs/models.yml
```diff
@@ -212,18 +212,23 @@ user:
     fields:
       type: relation-list
       to: projector/current_element_ids
-  vote_delegated_$_to_id:
```
FinnStutzenstein, line 215: Please undo this change! The template form is required for requesting specific data from the autoupdate service. The format cannot deal with m2m tables spanning over meetings.

docs/models.yml
```diff
@@ -715,9 +723,9 @@ meeting:
     type: relation
     to: group/default_group_for_meeting_id
     required: true
-  superadmin_group_id:
+  admin_group_id:
```
FinnStutzenstein, line 726: Why did you change this? it is already implemented in the backend and client...

FinnStutzenstein (changes requested):

docs/models.yml
FinnStutzenstein: But for me it is ok.

docs/models.yml
normanjaeckel: @ostcar Please discuss this during code sprint. Only to be kept in mind: The problem has something to do with the structure of permission service (only yes or no) and restriction service (restrict relation id fields only if the user can not see any field of the related object). @FinnStutzenstein may give an example of an autoupdate request that requires this template form tomorrow.

ostcar: I removed the commit

FinnStutzenstein (approved):

PR #5794, opened 2021-01-06
Docs in model.yml

normanjaeckel:


FinnStutzenstein:
Are all fields used in the backend?

Issue #5737, opened 2020-11-30
Multiple OpenSlides Instances on one docker host


MichaelZiegler:
Hi,
i am trying to install multiple OpenSlides instances on one docker host, that should be running at the same time.
As reverse proxy I am using that one: https://hub.docker.com/r/jwilder/nginx-proxy.
I cloned the OpenSlides docker directory and changed the URLs in the docker compose file. Also I tried to use different docker networks for the different instances. When I execute "docker-compose up" the first instance is starting and working without any problems. When I now try to start the next instance, "docker-compose up" is also working, but I cannot login the second instance. If I shut down the first one via "docker-compose down" and then start the second one, this instance is working. But I didn't succeeded in running two instances at the same time.
Has anyone any solution for this problem? Thanks!

FinnStutzenstein:
I suspect that docker compose uses equal container names and gets confused by this. The container names are put together by the foldername the `docker-compose.yml` is in and the name in the `docker-compose.yml`. Try to name the folders where the `docker-compose.yml` lies in differently. Another thing: The setup in the docker-compose repo is obsolete. You can use the current master *of this repo* since it has a nicer docker setup and does not involve a secondary repository.

MichaelZiegler:
> 
> 
> I suspect that docker compose uses equal container names and gets confused by this. The container names are put together by the foldername the `docker-compose.yml` is in and the name in the `docker-compose.yml`. Try to name the folders where the `docker-compose.yml` lies in differently. Another thing: The setup in the docker-compose repo is obsolete. You can use the current master _of this repo_ since it has a nicer docker setup and does not involve a secondary repository.

MichaelZiegler:
I also had the idea to change the foldernames. So I tried already folders like "OpenSlidesEvent1/docker_event1" and "OpenSlidesEvent2/docker_event2", but that wasn't the solution. The containernames are so different, but the problem consists.

ManniLoe:
I'm also guessing that there is a problem with duplicate container names.

I have two instances of slides running on the same Docker host without any problems.
However, I set the docker-compose option "project-name" for each of the instances.
Try it at instance 1: `docker-compose -p os1 up --build -d`
and for instance 2: `docker-compose -p os2 up --build -d`
(-p => Specify an alternate project name (default: directory name))

MichaelZiegler:
> 
> 
> I'm also guessing that there is a problem with duplicate container names.
> 
[...]

PR #5805, opened 2021-01-10
Fix compatibility issues with macOS

loelkes:
For development and testing purposes I want to run this on my local computer, a macOS system. 

* 608fcb3 by default, macOS comes with bash version 3 and does not support declarative arrays. This commit adds a check and shows a message if the check fails.
* 77f7802 `getopt` does not support long options under macOS (Darwin/BSD). It seems that getopt is not necessary, so I removed it and added a check and documention for the services option.
* 899dd3e  Some Docker systems (such as Docker Desktop on macOS) have buildkit enabled by default. 
* a5cf33e This could be a possible fix for #5737


loelkes:
Side note: while working on this and building the client I ran into he issue of npm throwing the following error:
```
An unhandled exception occured: Call retries were exceeded
```
After some research I found out that Docker Desktop on macOS has a default memory limit of 2GB. Increasing this to 8GB (didn't tested other values) solves this issue.

docker/build.sh
```diff
@@ -1,4 +1,15 @@
-#!/bin/bash
+
+# Check if associative arrays arrays are supported
+unset assoc
+if ! declare -A assoc > /dev/null 2>&1 ; then
+    echo "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
```
FinnStutzenstein, line 5: Please reduce this error to a one-liner. This is too much.

docker/build.sh
```diff
@@ -24,11 +35,18 @@ usage() {
...
-  --no-cache         Pass --no-cache to docker-build
+  -t                 Tag the Docker image (default: $DOCKER_TAG)
+  -P                 Offer to push newly built images to registry
+  -C                 Pass --no-cache to docker-build
+
+Notes:
+  <service> defaults to 'server client'.
+
+  You can build all services with 'all'. Currently supported are:
+    client server media-service pgbouncer postfix repmgr
```
FinnStutzenstein, line 48: Please dynamically build this line by using `TARGETS`, so one does not forget to update this place if some targets are changed.

docker/build.sh
loelkes: I prefer if errors are clearly visible but thats a personal preference. I don't agree that it's too much but will change it.

docker/build.sh
```diff
@@ -1,4 +1,10 @@
-#!/bin/bash
+
+# Check if associative arrays arrays are supported
+unset assoc
+if ! declare -A assoc > /dev/null 2>&1 ; then
```
FinnStutzenstein, line 4: A comment from @gsiv got lost I think:

How about we actually check the Bash version?  Example:

```bash
[[ "${BASH_VERSINFO[0]:-0}" -ge 4 ]] || {
  echo "Ancient bash version :("
  exit 4
}
```

Finally, please don't remove the shebang.

FinnStutzenstein: https://github.com/OpenSlides/OpenSlides/commit/608fcb38e4ff9680fc3c57f2e12dde5f1aede849

loelkes: Yes, I had considered it but decided to go the other way. Also I think my suggestion clearly states what the problem is. I would not change it.

gsiv: > Yes, I had considered it but decided to go the other way.

Would you like to share your considerations?  I'm curious.

If we were to opt for the proposed method, I would like to point out that it does not make sense to a) `unset` a variable that does not exist and b) create an array that we won't need.  Instead of declaring `assoc`, I would suggest to simply catch the error when we declare `TARGETS`.

loelkes: I didn't overthink it. I was just looking for something that resolves the issue, not the perfect solution. For me it seemed better to check `assoc` instead of the bash version. And I took a decision. I don't care if we do it the other way and your suggestion sounds good, so I implemented it. Let's merge this and move forward.

loelkes:
@FinnStutzenstein Can we move forward here?

FinnStutzenstein:
Sorry, @gsiv does not have the time for this right now. We talked and came to the conclusion, that the bash version check is a higher step to take than discarding longopts. Since you already have to manually install a newer bash version on BSD, you can also install gnu getopt. As far as I know docker does not run natively on BSD (since it uses linux kernel features), so sadly we are not going to support this and stick to the well working linux environment. Regarding longopts: Since getopt will exit with a non-zero exit code, this can be captured and transformed in a nice error message, that the getopt version is not supported. All in all thanks for your patience, but doing so much special things just to work around incompatile systems is not the thing we want to do (good errors are welcome, but no cut-content).

PR #5817, opened 2021-01-20
Performance improvements for OS3+

FinnStutzenstein:
- Cleans up log messages in the client
- Refactored the autoupdate bundle code into an own file
- Added bulk creates for History in Postgresql. This is the only database system
  that supports returning all ids whan multiple elements are inserted. We can
  make usage out of it.
- Added a `disable_history`, that is request-wide
- Disabled history on poll vote requests
- Removed unnecessary user ordering
- Reduced the queries for creating motion vote objects by one


FinnStutzenstein:
@emanuelschuetze If the tests are successful, it can be merged.

emanuelschuetze (approved):

PR #5822, opened 2021-01-24
Improve client-side password generation

mdickopp:
The currently implemented client-side password generation algorithm does not choose all allowed characters with equal probability, but chooses some more often than others.

The reason is that generating an 8-bit random number and reducing it modulo 56 (`characters.length`, the number of allowed characters) does not choose all numbers 0 to 55 with equal probability, but chooses 0 to 31 with higher probability than 32 to 55.

This change improves the password generation algorithms by choosing all characters with equal probability.

### DIFF
--- a/client/src/app/core/repositories/users/user-repository.service.ts
+++ b/client/src/app/core/repositories/users/user-repository.service.ts
@@ -181,11 +181,21 @@ export class UserRepositoryService extends BaseRepository<ViewUser, User, UserTi
      */
     public getRandomPassword(length: number = 10): string {
         let pw = '';
-        const array = new Uint8Array(length);
-        window.crypto.getRandomValues(array);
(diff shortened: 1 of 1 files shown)


FinnStutzenstein:
Thanks fpr the PR. It would be nice, if the solution would have a deteministic runtime. Currently the runtime behavior depends on the random numbers generated.

mdickopp:
Unfortunately, it is impossible to generate a random integer between 0 and 55, such that each choice has equal probability, in deterministic runtime.

Although my proposal does not have deterministic runtime, it is a common solution to this problem. The OpenSlides server uses the same algorithm I'm proposing when it generates passwords: It relies on the `User.objects.make_random_password()` Django function, which ultimately uses the Python function seen at https://github.com/python/cpython/blob/f7b5bacd7a0b2084ce699eda6f6f4b1adfa16590/Lib/random.py#L240 to generate a single character.

FinnStutzenstein:
Ok for me

tsiegleauq (approved):

PR #5828, opened 2021-01-27
Fixes offline bar on successful reconnections (closes #5810).

### LINKED ISSUE #5810: Reset retry counter on successfull message
OS3+: When failing three times, the client goes into offline mode. If one retry was successfull, the retry counter should reset, so the offline mode is not triggered early.

FinnStutzenstein:
Waiting for https://github.com/OpenSlides/openslides3-autoupdate-service/issues/85 and https://github.com/OpenSlides/openslides3-autoupdate-service/issues/96 so this feature works

### DIFF
--- a/client/src/app/core/core-services/autoupdate-throttle.service.ts
+++ b/client/src/app/core/core-services/autoupdate-throttle.service.ts
@@ -68,6 +68,12 @@ export class AutoupdateThrottleService {
                 this.disabledUntil = null;
                 console.log('Throttling autoupdates again');
             }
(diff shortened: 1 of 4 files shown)


haproxy/src/haproxy.common.cfg
```diff
@@ -27,5 +27,5 @@ backend backend_server
 
 backend backend_autoupdate
     mode http
-    timeout server 1h
+    timeout server 1m
```
FinnStutzenstein, line 30: This is for debugging and will be undone!

PR #5837, opened 2021-01-31
Fixed Path error

sanojkl:
The readme path is different, fixed

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -8,7 +8,7 @@ from openslides import __version__ as openslides_version
 from openslides import __license__ as openslides_license
 from openslides import __url__ as openslides_url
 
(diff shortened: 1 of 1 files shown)


sanojkl:
pip build didnt work, please wait i post update

FinnStutzenstein:
We drop pip support in the newest version since the python part of OpenSlides cannot run stand-alone anymore. You need to have additional services spin up to get it working. I think that I forget to delete the setup.py

sanojkl:
@FinnStutzenstein ok, then you should laso update the wiki make release

emanuelschuetze:
Closed because pip support was removed, see #5852

PR #5850, opened 2021-02-04
Own session redis

FinnStutzenstein:


ostcar:
I checked out your branch but the key_salt that django uses is `django.contrib.sessionsSessionStore`. Are you sure that the new session backend is really used?

ostcar:
The docker-compoe setup has to be changed so the secred_key is set as environment variable

FinnStutzenstein:
Hi, please delete your settings.py or remove the `SESSION_ENGINE` key. It is not valid anymore. (I think we should find another solution for the settings.py - since we do have so much config under control now we might force-regenerate the settings.py, if we find invalid keys...)

For the secret key you can use something like this in your entrypoint:
```
# Set DJANGO_SECRET_KEY variable
source /run/secrets/django || true
[[ -n "$DJANGO_SECRET_KEY" ]] || {
  echo "ERROR: Django secret key undefined!  Cannot continue."
  sleep 5
  exit 2
}
export SECRET_KEY="$DJANGO_SECRET_KEY"
```

ostcar:
If I understand your entry point script correctly, then it is not possible to override the variable with a specific value. The value in the file will always be used.

It seems be simpler, that the autoupdate service reads the variable from the file directly.

FinnStutzenstein:
> It seems be simpler, that the autoupdate service reads the variable from the file directly.

Ok for me. But please fail hard, if the file does not exist and do not fall back to something else.

FinnStutzenstein:
I'll add some auto-replacement of the settings.py here as well

FinnStutzenstein:
- Added auto-replacement of the settings.py
- rebased on the current master

ostcar:
I have a working version (when I hardcode the secred_key).

The file /run/secrets/django does not exist in the autoupdate container. Could you change the docker-compose.dev.yml file to mount this file?

FinnStutzenstein:
I would like to have a similar solution to OS4: In development, lets set the secret key to `development`.

Can you add something like this to your development target in your Dockerfile:

```
echo "DJANGO_SECRET_KEY='development'" > /run/secrets/django
```

I've adapted this branch, so that the secret key of the server is also `development`.

FinnStutzenstein:
Done. waits for https://github.com/OpenSlides/openslides3-autoupdate-service/pull/108 to be merged first.

emanuelschuetze:
@FinnStutzenstein Can we merge this? (PR in AU3 is ready for merge.) Do we need to adapt somethig in settings.py after update production instances?

FinnStutzenstein:
> Do we need to adapt somethig in settings.py after update production instances?

Already done: https://github.com/OpenSlides/OpenSlides/pull/5850/files#diff-79615e07e2472517c314271a0c086f45d1fdefa7208d5f3bbd2b66fb0f7f6a87L110

PR #5852, opened 2021-02-04
Removed pip support

emanuelschuetze:
Reason: pip supported is dropped for next OpenSlides release (3.4)
because the python server cannot run stand-alone anymore. You will cannot
install OpenSlides 3.4 via 'pip install openslides'.


emanuelschuetze:
[Wiki page](https://github.com/OpenSlides/OpenSlides/wiki/Make-Release) updated.

FinnStutzenstein (approved):

PR #5860, opened 2021-02-05
Fix Development

ostcar:


ostcar:
I could not test the production build. When I run `build.sh all` it fails with:

```
Building openslides/openslides-repmgr:latest...
[+] Building 0.0s (1/1) FINISHED                                                                                                                                                              
 => ERROR [internal] load git source https://github.com/OpenSlides/openslides-docker-compose.git#:repmgr                                                                                 0.0s
------
 > [internal] load git source https://github.com/OpenSlides/openslides-docker-compose.git#:repmgr:
------
failed to solve with frontend dockerfile.v0: failed to read dockerfile: failed to load cache key: subdir not supported yet
```

emanuelschuetze (approved):
Dev setup works for me now - by using http://localhost:8000/.
(https does not work. Same for production build.)

PR #5863, opened 2021-02-06
Force http2 on requests to the autoupdate-service.

ostcar:
This should only be done in situations, where the browser directly connects to caddy. If another proxy is in between, then this is not necessary.

See: https://caddy.community/t/match-on-http-version/11405

I am not 100% sure if we should merge this. All modern browsers use http2 on https when it is supported by the server. The situation, that a browser connects with a http1 connection should not happen. But it is more configuration, that can be confusing. On the other hand, it is a good example, how to force http2 with caddy.

### DIFF
--- a/caddy/Caddyfile.dev
+++ b/caddy/Caddyfile.dev
@@ -5,9 +5,22 @@
 
 localhost:8000
 
-reverse_proxy /system/* autoupdate:8002 {
+@autoupdate {
+    path /system/*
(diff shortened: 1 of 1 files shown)


emanuelschuetze:
I'm against to force it. - Currently we can test it with older (HTTPv1) oder newer (HTTPv2+) versions. Maybe helpful. So I'll close this PR.

PR #5884, opened 2021-02-11
Add performance scripts

FinnStutzenstein:
@emanuelschuetze This are the basic things I have. There are some TODOs for the go part (autoupdate). For the server there is currently just a simple stress test for setting the presence. (the voting.py does not work currently and requires many users - this is not what you need I guess. I just checked it in so the work is not lost).


performance/server-clients/README.md
```diff
@@ -0,0 +1,29 @@
+## Do requests to stress the server
+
+Setup:
+
+    python3 -m venv .venv
+    source .venv/bin/activate
+    pip install -u -r requirements.txt
```
emanuelschuetze, line 7: use `-U`

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+	"time"
+)
+
+func main() {
+	N := getNFromCommandLine()
+	host, user, pw := getParams()
+	jar := getCookieJar(user, pw, host)
+
+	sumBytes := 0
+	oldSumKB := 0
```
ostcar, line 26: Style: I would write
```go
var sumBytes, oldSumKB int
```

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+	changeIds := make(map[int]int)
+	changeIdsMutex := &sync.Mutex{}
```
ostcar, line 28: style: I would write:
```go
changeIDs := make(map[int]int)
var changeIDsMutex sync.Mutex
```
The mutex does not have to be a pointer.

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+	gotUpdate := false
```
ostcar, line 29: Style: I would write `var gotUpdate bool`

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+
+	interrupt := make(chan os.Signal, 1)
+	signal.Notify(interrupt, os.Interrupt)
```
ostcar, line 32: I saw this TODO
>Pressing Ctrl+C during connecting should stop the process immediatly

If you move this line after the connection, you would solve this TODO

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+	retry := true
+	initialIteration := true
+	for retry {
+		if !initialIteration {
+			time.Sleep(1 * time.Second)
+			log.Println(i, "retry after error")
+		}
+		initialIteration = false
+
+		var client *http.Client
```
ostcar, line 117: This line is not necessary

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+				if m != nil {
+					recv <- *m
+				}
+			}
+		}(resp, recv, done)
+
+	L:
+		for {
+			select {
+			case <-quit:
```
ostcar, line 151: Style: Use context for cancelation

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+
+				m := parseMessage(message, i)
+				if m != nil {
+					recv <- *m
+				}
+			}
+		}(resp, recv, done)
+
+	L:
+		for {
```
ostcar, line 149: It seems, the loop is not necessary. The only cases inside the select-statement end with `break` or `return`. So the loop does not run a second time.

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+				log.Println("recv goroutine exited!")
+				break L
+			}
+		}
+	}
+}
+
+func parseMessage(data []byte, i int) *Message {
+	var result map[string]interface{}
+	json.Unmarshal(data, &result)
```
ostcar, line 169: You should check the error.

If you are sure, that this can never fail, then you should write:
```go
_ = json.Unmarshal(...)
```
This shows, that you did not forget to check the error.

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+				// The recv goroutine errored. break the for and retry it
+				log.Println("recv goroutine exited!")
+				break L
+			}
+		}
+	}
+}
+
+func parseMessage(data []byte, i int) *Message {
+	var result map[string]interface{}
```
ostcar, line 168: You should use a concret struct like:
```go
var result struct{
  ToChangeID float64 `json:"to_change_id"`
  ...
}
```

In other case, you could get a panic in line 171.

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+	if _, ok := result["to_change_id"]; ok {
+		return &Message{Length: len(data), I: i, ChangeId: int(result["to_change_id"].(float64))}
+	}
+
+	if _, ok := result["connected"]; ok {
+		return nil // skip connect messages
+	}
+
+	log.Printf("Invalid message: %v\n", result)
+	return nil
```
ostcar, line 179: You should check for errors early. The last line should be the success-return. On error, you should return it.

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	go func(quit chan int) {
+		for {
+			select {
+			case <-quit:
+				return
+			case <-time.After(time.Second):
+				if !gotUpdate {
```
ostcar, line 47: This is a race condition. The variable is written in line 98 in a separat goroutine.

You should test/run you program with the race detector: https://golang.org/doc/articles/race_detector.html

Use either a lock or a channel to communicate between the goroutines

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+				gotUpdate = false
+
+				// count the amount of change ids per change id
+				amountChangeIds := make(map[int]int)
+				changeIdsMutex.Lock()
+				for _, changeId := range changeIds {
+					if _, ok := amountChangeIds[changeId]; ok {
+						amountChangeIds[changeId] += 1
+					} else {
+						amountChangeIds[changeId] = 1
```
ostcar, line 59: This block is not necessary. Each key in a map has the 0-value as default. If you write
```go
amountChangeIDs[changeId] += 1
```
It does the same. Each map in go is like a defaultdict in python

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+					}
+				}
+				changeIdsMutex.Unlock()
+
+				// get size of update
+				sumKB := int(sumBytes / 1024)
```
ostcar, line 65: This is also a race condition

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+				diffKB := sumKB - oldSumKB
+				oldSumKB = sumKB
+				log.Printf("New update: %d KB (diff %d KB)\n", sumKB, diffKB)
+
+				sortedChangeIds := make([]int, len(amountChangeIds))
+				i := 0
```
ostcar, line 71: It seems, you do not use this variable

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+	recv := make(chan Message)
+	for i := 0; i < N; i++ {
+		go client(quit, recv, i, jar, host)
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	go func(quit chan int) {
+		for {
+			select {
+			case <-quit:
```
ostcar, line 44: Style: Use context for cancelation. See also https://tip.golang.org/doc/go1.16#os/signal

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+				log.Printf("\n")
+			}
+		}
+	}(quit)
+
+	for {
+		select {
+		case <-interrupt:
+			for i := 0; i < (N + 1); i++ {
+				quit <- 0
```
ostcar, line 94: You can close the quit channel to get the same result

performance/autoupdate-clients/main.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+			return
```
ostcar, line 96: This will stop the program immediately. If you do not wait for the goroutines to quit, then you don't need the quit channel at all.

emanuelschuetze (approved):

FinnStutzenstein:
@ostcar Thanks for the review. These lines of code you saw are actually my first lines of go ;) Since I am not going to maintain this script (@emanuelschuetze will take over control), and it works (well, it has bugs as you stated...) @emanuelschuetze will merge it soon I think.

ostcar:
Could you maybe add the possibility to test the projector and notify URL as well?

ostcar:
@FinnStutzenstein I changed the commit id of the autoupdate service by accident. How can I revoke it?

Can you check, if your tools test the same thing then os-performance? If so, we could talk, which one to keep.

FinnStutzenstein:
> How can I revoke it?

Go into the autoupdate service, checkout the commit you want, `cd ..`, commit.

> Can you check, if your tools test the same thing then os-performance?

Not now, If it is urgend, @emanuelschuetze should do it.

FinnStutzenstein:
Closed in favor of #5916

PR #5894, opened 2021-02-15
add optional https for caddy

FinnStutzenstein:
@gsiv Can you check, if this works in production? The only real change is `:8000` -> `http://:8000`, if no certs are in the `certs` folder (which shouldn't in our setup, the folder is new).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,3 +12,5 @@ stop-dev:
 get-server-shell:
 	docker-compose -f docker/docker-compose.dev.yml run server bash
 
+reload-proxy:
+	docker-compose -f docker/docker-compose.dev.yml exec -w /etc/caddy proxy caddy reload
--- a/README.rst
+++ b/README.rst
@@ -36,10 +36,11 @@ first and initialize all submodules::
 Setup Docker images
 -------------------
 
-You need to build the Docker images for the client and server and have to setup some
-configuration.
+You need to build the Docker images and have to setup some configuration. First,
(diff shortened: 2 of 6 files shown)


ostcar:
I think this is even a bigger problem.

If the browses do not support http2 without Https. If someone does use the setup without Https, he can only use one tab in his browser.

See also my comment to Https in the autoupdate readme.

I would say, Https is mandatory. You either have to use Https with caddy or have another Https terminator like haproxy or nginx on front.

FinnStutzenstein:
> he can only use one tab in his browser.

This is not correct (at least for chrome and FF).

We are not responsible how others manage their setups. We can tell them to use https and about the reason behind it. Enforcing it is useless, since there are so many proxy setups (Look into our issues...). They can even use a https proxy, but if it does not support http2, https is useless in this sense. We should add a clear recommendation, but we cannot do more I think.

ostcar:
> > he can only use one tab in his browser.
> 
> This is not correct (at least for chrome and FF).
> 
You are correct, he can use two tabs. Or even three, if is it not a tab with the projector.

You can try this by using the prod build and open many tabs. This does not work with any browser.

FinnStutzenstein:
I changed the wording and explicitly use "required" instead of "highly recommended". People who know what they are doing can always use another approach (like we are doing). Other ones that just want a running OpenSlides instance are warned.

@ostcar Ok?

ostcar (approved):

PR #5897, opened 2021-02-15
Fixes an infinite running spinner

GabrielInTheWorld:
The ApplicationRef propagates never a stable state, when in cinema mode. This is, because in the `cinema.component.ts` asynchronous requests are made, before the app was getting stable.
Also, interval, timeout and similar actions have the same result.

### DIFF
--- a/client/src/app/app.component.spec.ts
+++ b/client/src/app/app.component.spec.ts
@@ -1,35 +1,33 @@
-import { async, fakeAsync, TestBed, tick } from '@angular/core/testing';
+// import { async, fakeAsync, TestBed, tick } from '@angular/core/testing';
 
-import { TranslateService } from '@ngx-translate/core';
+// import { TranslateService } from '@ngx-translate/core';
 
-import { AppComponent } from './app.component';
-import { E2EImportsModule } from './../e2e-imports.module';
-import { ServertimeService } from './core/core-services/servertime.service';
+// import { AppComponent } from './app.component';
+// import { E2EImportsModule } from './../e2e-imports.module';
+// import { ServertimeService } from './core/core-services/servertime.service';
 
 describe('AppComponent', () => {
-    let servertimeService, translate;
-
-    beforeEach(async(() => {
-        TestBed.configureTestingModule({
-            imports: [E2EImportsModule]
-        }).compileComponents();
-
-        servertimeService = TestBed.inject(ServertimeService);
-        translate = TestBed.inject(TranslateService);
-        spyOn(servertimeService, 'startScheduler').and.stub();
-        spyOn(translate, 'addLangs').and.stub();
-        spyOn(translate, 'setDefaultLang').and.stub();
-        spyOn(translate, 'getBrowserLang').and.stub();
-        spyOn(translate, 'getLangs').and.returnValue([]);
-        spyOn(translate, 'use').and.stub();
-    }));
-    it('should create the app', fakeAsync(() => {
-        const fixture = TestBed.createComponent(AppComponent);
-        const app = fixture.debugElement.componentInstance;
-        expect(app).toBeTruthy();
-        tick(1000);
-        fixture.whenStable().then(() => {
-            expect(servertimeService.startScheduler).toHaveBeenCalled();
-        });
-    }));
+    // let servertimeService, translate;
+    // beforeEach(async(() => {
+    //     TestBed.configureTestingModule({
+    //         imports: [E2EImportsModule]
+    //     }).compileComponents();
+    //     servertimeService = TestBed.inject(ServertimeService);
+    //     translate = TestBed.inject(TranslateService);
+    //     spyOn(servertimeService, 'startScheduler').and.stub();
+    //     spyOn(translate, 'addLangs').and.stub();
+    //     spyOn(translate, 'setDefaultLang').and.stub();
+    //     spyOn(translate, 'getBrowserLang').and.stub();
+    //     spyOn(translate, 'getLangs').and.returnValue([]);
+    //     spyOn(translate, 'use').and.stub();
+    // }));
+    // it('should create the app', fakeAsync(() => {
+    //     const fixture = TestBed.createComponent(AppComponent);
+    //     const app = fixture.debugElement.componentInstance;
+    //     expect(app).toBeTruthy();
+    //     tick(1000);
+    //     fixture.whenStable().then(() => {
+    //         expect(servertimeService.startScheduler).toHaveBeenCalled();
+    //     });
+    // }));
 });
--- a/client/src/app/app.component.ts
+++ b/client/src/app/app.component.ts
@@ -4,7 +4,7 @@ import { DomSanitizer } from '@angular/platform-browser';
 import { Router } from '@angular/router';
 
 import { TranslateService } from '@ngx-translate/core';
-import { filter, take } from 'rxjs/operators';
+import { first, tap } from 'rxjs/operators';
 
 import { ChatNotificationService } from './site/chat/services/chat-notification.service';
 import { ConfigService } from './core/ui-services/config.service';
@@ -19,6 +19,7 @@ import { OperatorService } from './core/core-services/operator.service';
 import { OverlayService } from './core/ui-services/overlay.service';
 import { RoutingStateService } from './core/ui-services/routing-state.service';
 import { ServertimeService } from './core/core-services/servertime.service';
+import { StableService } from './core/core-services/stable.service';
 import { ThemeService } from './core/ui-services/theme.service';
 import { VotingBannerService } from './core/ui-services/voting-banner.service';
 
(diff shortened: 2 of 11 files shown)


client/src/app/app.component.ts
```diff
@@ -107,8 +107,7 @@ export class AppComponent {
         appRef.isStable
             .pipe(
                 // take only the stable state
-                filter(s => s),
-                take(1)
+                first(stable => stable)
```
FinnStutzenstein, line 110: Why did you remove the `take(1)`. This event should only be emitted once.

client/src/app/core/core-services/openslides.service.ts
```diff
@@ -35,6 +35,10 @@ export class OpenSlidesService {
         return this.booted.value;
     }
 
+    public get isStable(): Promise<void> {
```
FinnStutzenstein, line 38: `isStable` suggest, that it is a boolean. Either:
- make `private stable = new Deferred();` to `public readonly stable = new Deferred();`
- rename it to `waitForStability`

Sidenote: Both suggestions (as well as your) does not prevent a user from doing `this.openSlidesService.isStable.resolve()`, which should be prevented (Like a "read-only" Promise). Since I do not know if this easily doable, it can be skipped ;)

client/src/app/site/cinema/components/cinema/cinema.component.ts
```diff
@@ -98,6 +100,15 @@ export class CinemaComponent extends BaseViewComponentDirective implements OnIni
...
         super.setTitle('Autopilot');
+        this.loadSubscriptions();
+    }
+
+    public async toggleListOfSpeakersOpen(): Promise<void> {
+        await this.listOfSpeakersRepo.setListOpenness(this.listOfSpeakers, this.isLosClosed).catch(this.raiseError);
+    }
+
+    private async loadSubscriptions(): Promise<void> {
+        await this.openslidesService.isStable;
```
FinnStutzenstein, line 111: I'm curious why this await helps here. On the first glance no request is send here, isn't it?

FinnStutzenstein (changes requested):

client/src/app/app.component.ts
GabrielInTheWorld: As of the description from rxjs:

> first
> Emits only the first value. Or emits only the first value that passes some test.

I think, `first` will unify filter and take. ;)

client/src/app/core/core-services/openslides.service.ts
GabrielInTheWorld: Yeah, you're right. I will clean up this, later. :)

client/src/app/site/cinema/components/cinema/cinema.component.ts
GabrielInTheWorld: You're right. Timeouts, intervals and similar functions have the same result.

emanuelschuetze (approved):

client/src/app/core/core-services/openslides.service.ts
GabrielInTheWorld: > Sidenote: Both suggestions (as well as your) does not prevent a user from doing `this.openSlidesService.isStable.resolve()`, which should be prevented (Like a "read-only" Promise). Since I do not know if this easily doable, it can be skipped ;)

This is possible, if a user knows, it is a `Deferred`. But a plain `Promise` can't be resolved out of its scope. :)

emanuelschuetze:
@GabrielInTheWorld Thanks.

client/src/app/core/core-services/http-interceptors/noop-interceptor.service.spec.ts
```diff
@@ -0,0 +1,14 @@
+// import { TestBed } from '@angular/core/testing';
```
FinnStutzenstein, line 1: Please just delete this file. Same for the app component test

client/src/app/core/core-services/openslides.service.ts
FinnStutzenstein: Ah, right

FinnStutzenstein: Still a todo

client/src/app/site/projector/components/projector-detail/projector-detail.component.ts
```diff
@@ -353,4 +350,16 @@ export class ProjectorDetailComponent extends BaseViewComponentDirective impleme
             this.messageRepo.create(sendData).then(() => {}, this.raiseError);
         }
     }
+
+    private async loadSubscriptions(): Promise<void> {
```
FinnStutzenstein, line 354: Do not wait until an stable app to subscribe. These subscriptions itself do not keep the spinner up, but some other effects. This is too broad...

client/src/app/site/projector/components/projector-detail/projector-detail.component.ts
```diff
@@ -353,4 +350,16 @@ export class ProjectorDetailComponent extends BaseViewComponentDirective impleme
...
+        await this.stableService.isStable;
+        this.countdownRepo.getViewModelListObservable().subscribe(countdowns => (this.countdowns = countdowns));
+        this.messageRepo.getViewModelListObservable().subscribe(messages => (this.messages = messages));
+        this.repo.getViewModelListObservable().subscribe(projectors => (this.projectorCount = projectors.length));
+    }
+
+    private async installUpdater(): Promise<void> {
+        await this.stableService.isStable;
+        this.subscriptions.push(timer(0, 500).subscribe(() => this.cd.detectChanges()));
```
FinnStutzenstein, line 363: This seems hacky. Why do you need a detect changes 500ms after the app stable event? Please put the `detectChanges` to the place, where it is actually needed.

As a sidenote: You can use https://github.com/OpenSlides/OpenSlides/blob/master/client/src/app/core/promises/sleep.ts: `await SleepPromise(500)` if you want to do that kind of waiting.

client/src/app/site/projector/components/projector-list/projector-list.component.ts
```diff
@@ -138,4 +133,16 @@ export class ProjectorListComponent extends BaseViewComponentDirective implement
         super.ngOnDestroy();
         this.cd.detach();
     }
+
+    private async installUpdater(): Promise<void> {
+        await this.openslidesService.isStable;
+        /**
+         * Angulars change detection goes nuts, since countdown and motios with long texts are pushing too much data
+         */
+        this.subscriptions.push(
```
FinnStutzenstein, line 142: Please use the SleepPromise.

client/src/app/core/core-services/stable.service.ts
```diff
@@ -0,0 +1,29 @@
+import { Injectable } from '@angular/core';
+
+import { BehaviorSubject, Observable } from 'rxjs';
+
+import { Deferred } from '../promises/deferred';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class StableService {
```
FinnStutzenstein, line 10: Integrate this into the OpenSlidesStatusService. Use this service only, so remove the `isStable` in the OpenSlides Service

client/src/app/core/core-services/stable.service.ts
```diff
@@ -0,0 +1,29 @@
...
+    public get isStable(): Promise<void> {
+        return this.stable;
+    }
+
+    public get booted(): Observable<boolean> {
```
FinnStutzenstein, line 15: This is only needed in the interceptor - move the BehaviorSubject there

client/src/app/core/core-services/http-interceptors/noop-interceptor.service.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { Observable, Subject } from 'rxjs';
+import { first, mergeMap } from 'rxjs/operators';
+
+import { StableService } from '../stable.service';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class NoopInterceptorService implements HttpInterceptor {
```
FinnStutzenstein, line 12: Is it a Noop interceptor?

client/src/app/core/core-services/stable.service.ts
FinnStutzenstein: Also: booted is another terminology that indicates, that OpenSlides has started up - do not use this wording twice.

FinnStutzenstein (commented):
@emanuelschuetze Please wait for reviews!!!

FinnStutzenstein:
@GabrielInTheWorld This needs some fixes.

client/src/app/app.component.spec.ts
```diff
@@ -1,35 +1,33 @@
-import { async, fakeAsync, TestBed, tick } from '@angular/core/testing';
+// import { async, fakeAsync, TestBed, tick } from '@angular/core/testing';
```
FinnStutzenstein, line 1: Please just remove this file.

PR #5898, opened 2021-02-16
Some cleanup in favor of spinner-fixes

GabrielInTheWorld:
Cleans up #5897 

- Rename: noop-interceptor -> stable-interceptor
- Put logic from stable.service to openslides-status.service

I've tried to get another fix for high cpu usages of projector-list and projector-detail. But I haven't found anything.


client/src/app/core/core-services/openslides-status.service.ts
```diff
@@ -26,13 +29,29 @@ export class OpenSlidesStatusService {
         return !!this.history;
     }
 
+    public get stable(): Promise<void> {
+        return this._stable;
+    }
+
+    public get booted(): Observable<boolean> {
```
FinnStutzenstein, line 36: Not used, can be removed.

FinnStutzenstein (approved):
One change needed, looks good.

PR #5900, opened 2021-02-16
Remodel the projector

FinnStutzenstein:


FinnStutzenstein:
Waiting for https://github.com/OpenSlides/openslides-backend/issues/462

@ostcar You did not do anything with the projector yet, right? So we must not wait for changes from you?

ostcar:
> @ostcar You did not do anything with the projector yet, right? So we must not wait for changes from you?

I did not start yet. You don't have to wait

docs/models.yml
```diff
@@ -1044,11 +1057,7 @@ agenda_item:
     equal_fields: meeting_id
   projection_ids:
     type: relation-list
-    to: projection/element_id
-    equal_fields: meeting_id
-  current_projector_ids:
-    type: relation-list
-    to: projector/current_element_ids
+    to: projection/content_object_id
```
jsangmeister, line 1060: Why did you remove the `equal_fields`? If this was by mistake, please add back (same for all other places below).

FinnStutzenstein: It is in the line below your comment.

FinnStutzenstein:
@jsangmeister If there are missing changes, please add them later.

docs/models.yml
jsangmeister: Oops, my bad

PR #5904, opened 2021-02-18
Proposal for permissions in initial-data.json

r-peschke:
1. Assuming that the permissions in the models.yml for group.permissions are correct, I fixed some bugs in the initial-data.json.
 2. Then I deleted all permissions that are dependent from given permissions following the structure in permissions.yml
 3. Finally I compared the remaining permissions with those from OS3 (transferred to =S4-permissions following the models.yml/group.permissions)
There are some (json-forbidden) comment lines in the proposal of initial-data.json.
There I mentioned some missing permissions compared with the OS3-version. For the group delegates there is a real difference of what is allowed to do with motions (initial: support and manage, OS3: support, but only create and create_amendments) that has to be decided.


docker/initial-data.json
```diff
@@ -258,14 +258,17 @@
...
             "assignment.can_see",
             "meeting.can_see_frontpage",
             "projector.can_see",
             "mediafile.can_see",
             "motion.can_see",
-            "user.can_see"
+            "user.can_see",
+
+            # Missing compared with os3 and existing permissions in models.yml
+            "list_of_speakers.can_see",
```
FinnStutzenstein, line 270: Both right.

docker/initial-data.json
```diff
@@ -304,25 +307,20 @@
...
             "mediafile.can_manage",
-            "mediafile.can_see",
-            "motion.can_create",
             "motion.can_manage",
-            "motion.can_see",
             "user.can_manage",
-            "user.can_see_extra_data",
-            "user.can_see"
+
+            # Missing compared with os3 and existing permissions in models.yml
```
FinnStutzenstein, line 320: all ok

docker/initial-data.json
```diff
@@ -342,16 +340,19 @@
...
             "mediafile.can_see",
             "motion.can_create",
-            "motion.can_see",
             "motion.can_support",
-            "users.can_see"
+            "user.can_see"
+            # Missing permissions compared with os3 and existing permissions in models.yml
+            "list_of_speakers.can_see",
+            "motion.can_create_amendments",
+            "user.can_see_extra_data",
```
FinnStutzenstein, line 354: this is not included for committees

docker/initial-data.json
```diff
@@ -342,16 +340,19 @@
...
             "assignment.can_see",
             "meeting.can_see_frontpage",
             "projector.can_see",
             "mediafile.can_see",
             "motion.can_create",
-            "motion.can_see",
             "motion.can_support",
-            "users.can_see"
+            "user.can_see"
+            # Missing permissions compared with os3 and existing permissions in models.yml
```
FinnStutzenstein, line 351: All except `user.can_see_extra_data` are correct

docker/initial-data.json
```diff
@@ -371,20 +372,25 @@
...
+            "agenda_item.can_see_internal",
+            "list_of_speakers.can_be_speaker",
             "assignment.can_nominate_other",
             "assignment.can_nominate_self",
-            "assignment.can_see",
             "meeting.can_see_frontpage",
             "projector.can_see",
             "mediafile.can_see",
-            "motion.can_create",
             "motion.can_manage",
```
FinnStutzenstein, line 382: this one is wrong - a delegate does not have manage perms. It should be these perms: `motion.can_create`, `motion.can_create_amendments`, `motion.can_support`

FinnStutzenstein: This resolves the discussion below

docker/initial-data.json
```diff
@@ -371,20 +372,25 @@
...
-            "motion.can_see",
             "motion.can_support",
             "user.can_see"
+            # Missing compared with os3 and existing permissions in models.yml
```
FinnStutzenstein, line 385: both correct

FinnStutzenstein (approved):
most things are ok, please order the permissions by collections.

If this is done, please copy these permissions into `docs/example-data.json`

docker/initial-data.json
r-peschke: done

docker/initial-data.json
r-peschke: done

docker/initial-data.json
r-peschke: Removed  "user.can_change_own_password"! Are you sure or is it just a remark for "user.can_see_extra_data"?

docker/initial-data.json
r-peschke: done

docker/initial-data.json
r-peschke: done

docker/initial-data.json
FinnStutzenstein: Committees have `user.can_see` and `user.can_change_own_password`.

r-peschke: Done

PR #5910, opened 2021-02-21
Make configurable whether a newly created list of speakers is initial…

mdickopp:
Add a new configuration setting that determines whether newly created lists of speakers are initially open or closed. When the new configuration setting is activated, lists of speakers created from that point on are initially in a closed state instead of an open state.

Motivation: For organizations that follow parliamentary conventions that speaking can only be requested once an agenda item has been called, starting with closed lists of speakers is the preferred default. Especially in the case where motions can be added shortly before or even during an assembly, this reduces the need for the organizer to close lists of speakers before or during the assembly.

The default behavior is not changed, i.e., lists of speakers are only created in a closed state if the user has activated the new configuration setting.

(I also added myself to the `AUTHORS` file.)


server/openslides/agenda/models.py
```diff
@@ -383,7 +390,7 @@ class ListOfSpeakers(RESTModelMixin, models.Model):
     Field for generic relation to a related object. General field to the related object.
     """
 
-    closed = models.BooleanField(default=False)
+    closed = models.BooleanField(default=list_of_speakers_initially_closed)
```
emanuelschuetze, line 393: This changes required a new migration file (python manage.py makemigrations). Or maybe we found an alternative to set this list-of-speakers toggle (e.g. on client side)?

emanuelschuetze (changes requested):
Thanke for your PR. See my remark for missing migration file.
And, please, don't add translations to PRs. We will do translations in a separate PR later. So please remove your changes for template.pot, de.json|po. Thx.

server/openslides/agenda/models.py
FinnStutzenstein: AN alternative would be to add it here: https://github.com/OpenSlides/OpenSlides/blob/master/server/openslides/agenda/signals.py#L88 This is the single place lists of speakers are created and do not add migrations.

mdickopp:
Thank you for the feedback. I'll rework the PR.

server/openslides/agenda/models.py
mdickopp: After looking into these options, setting the field when the `ListOfSpeakers` object is created still seemed the cleanest solution to me, so I added a migration file.

emanuelschuetze (approved):
Works as expected. Thanks!

emanuelschuetze:
@FinnStutzenstein Please review and merge.

FinnStutzenstein (approved):

PR #5915, opened 2021-02-24
Docker: Improve Prometheus integration

gsiv:


prometheus/prometheus.yml
```diff
@@ -0,0 +1,10 @@
+global:
+  scrape_interval: 5s
+  scrape_timeout: 5s
```
ostcar, line 3: At least for production, 1 minute should be enough.

caddy/Caddyfile
```diff
@@ -13,5 +13,7 @@ import endpoint
 
     reverse_proxy /media/* media:8000
 
+    reverse_proxy /prometheus/* prometheus:9090
```
ostcar, line 16: Is there some sort of restriction? I don't think, this URL should be public.

prometheus/prometheus.yml
gsiv: Thanks, I see.  So we should include two different configuration files.  I'll add one.

You presumably mean that `scrape_interval` should be set to 1 minute, correct?  Or should it be set for `scrape_timeout` as well?

caddy/Caddyfile
gsiv: As part of this setup, no.  My idea is to manage access using HTTP basic auth with the reverse proxy.

prometheus/prometheus.yml
ostcar: I would suggest to remove the values, so it uses the defaults.

scrape_interval 1m
scrape_timeout 10s

caddy/Caddyfile
ostcar: Ok

gsiv:
I pushed changes with which I believe this patch is ready to be merged:

  - squashed commits
  - added warning to `.env`
  - removed the scrape settings for the production config

ostcar (approved):
Did not test it

emanuelschuetze (approved):

PR #5916, opened 2021-02-25
Performance

ostcar:
This is the same as the performance in the pr of @FinnStutzenstein  #5884 .

I remaned the tool to `performance`. The help text is not in the readme anymore. You have to call the command to see it. This helps me to write the text only once. If you would like examples etc. in the help text, I can add them.

The vote command changed. I added a login-retry flag. The default is 3.

If you still get the error, I could add a workpool, so not all login requests are send at the same time.


FinnStutzenstein:
So #5884 can be closed?

emanuelschuetze (approved):

PR #5918, opened 2021-02-26
Remove voted_ids

FinnStutzenstein:
Wait for the autoupdate service.


ostcar:
Should be done in https://github.com/OpenSlides/openslides3-autoupdate-service/pull/125

PR #5921, opened 2021-03-01
Docker: Allow overriding image names from .env

gsiv:
For a subset of the images required by a Compose or Swarm Setup,
non-default names can be configured in .env.  Originally, the names were
treated as the images' complete names, i.e., including an optional
registry domain.  Using this setup, it was possible to pull the
irregularly updated auxiliary images from a default registry while, at
the same time, obtaining certain images from a different registry.

Commit e225a57f97 changed this behavior.  Since then, the names in .env
can only be used to change part of the image name, excluding the
registry.  If a default registry is configured it is always prepended to
the given image name, breaking the above use case and compatibility with
older .env files.

This patch reintroduces (almost) full control over image names from
.env: if the given image name contains a slash, no additional registry
path is prepended.  At the same time it also continues to support the
new logic – debatable as it may be.


FinnStutzenstein:
Should the .env be adopted as well (e.g. adding one example)?

gsiv:
> Should the .env be adopted as well (e.g. adding one example)?

I'm not sure but I'm leaning against it.  The only reason, in my mind, to ever
change image names is to override their repository.  There is no use case that
I can think of – and correct me if I'm wrong – that would include renaming,
e.g., `openslides-server` to something else.  Unless it involves the registry,
versions can and should be managed using tags.  For this reason, I think `.env`
would work as expected (again) and I don't think the behavior warrants much
documentation at this point.

Come to think of it, however, another way to both fix and simplify the
situation would be to remove the option to rename images entirely and instead
introduce image-specific registry overrides.  In other words, we could replace
`DOCKER_OPENSLIDES_BACKEND_NAME` with `DOCKER_OPENSLIDES_BACKEND_REPOSITORY`
etc.

I think I may prefer this solution because these new options would state their
purpose more clearly and would be simpler to handle in the template.  On the
flip side, this reintroduces the question of backwards compatibility but I'm
going to push a proposal anyway.

FinnStutzenstein:
Ok for me!

gsiv:
> Ok for me!

I have squashed my commits.  After review, this should be ready to be merged.

Do these changes need to be added to a NEWS file or something similar?

FinnStutzenstein (approved):
Looks good, not tested. No need for a news/changelog entry or so. I'll port this to OS4 after merge, so go ahead and merge it, if the deployment is ready.

PR #5924, opened 2021-03-01
Fixed showSubtitle config. Reordered agenda config.

emanuelschuetze:


FinnStutzenstein:
You have changed the default from `False` to `True`, so [this test](https://github.com/OpenSlides/OpenSlides/blob/master/server/tests/integration/core/test_viewset.py#L364) must be adopted (switch false/true).

FinnStutzenstein:
To get it working redis needs to be flushed and the autoupdate service needs to be restarted (mostly a note for @gsiv).

PR #5925, opened 2021-03-01
Adjusted models

FinnStutzenstein:
- Added new permissions poll.can_manage
- Reorder permissions.yml alphabetically
- provide some more default values

@normanjaeckel Maybe we should also add `poll.can_see` for non assignment/motion polls.


FinnStutzenstein:
@ostcar Small ping, too.

FinnStutzenstein:
I'll take this as an OK

ostcar:
Sorry. It's ok for me.

PR #5931, opened 2021-03-01
Disabled caddy debug mode in dev setup.

emanuelschuetze:


caddy/Caddyfile.dev
```diff
@@ -1,5 +1,5 @@
 {
-    debug
+    # debug
```
FinnStutzenstein, line 2: Please completly remove this section.

emanuelschuetze: ok

PR #5937, opened 2021-03-04
Delete unused code

FinnStutzenstein:
@oscar: Only the user restricter is still present in the code since it is needed for whoami


FinnStutzenstein:
@emanuelschuetze I would like to merge this PR; whats about the alternatives for the rest api?

PR #5940, opened 2021-03-06
Don't show diff view on slides if no change recos exist

CatoTH:
Moves the logic to decide which crMode to use to the class shared between desktop view and slides.
This should fix the slides showing an empty view when projecting the diff mode of an amendment but no change recommendation exists. The original version will be shown instead.


emanuelschuetze (approved):
Thanks. Works as expected.

PR #5945, opened 2021-03-08
Add default_vote_weight default

FinnStutzenstein:
@GabrielInTheWorld This should fix https://github.com/OpenSlides/openslides-backend/issues/523

### DIFF
--- a/docs/models.yml
+++ b/docs/models.yml
@@ -102,7 +102,9 @@ user:
   email: string
   default_number: string
   default_structure_level: string
-  default_vote_weight: decimal(6)
+  default_vote_weight:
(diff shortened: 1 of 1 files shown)


jsangmeister:
I think this has to be `1.000000`, right? So we stay consistent with our decimal schema.

FinnStutzenstein:
Can the backend handle a "1" or does it need to have 6 decimals?

jsangmeister:
It should be able to, but I still prefer to have the default values match the schema of the field.

FinnStutzenstein:
Ok, done in #5962

PR #5951, opened 2021-03-11
Use ENV for redis in db-setup and fix defaults

promasu:
* Add default value for `DATABASE_HOST`
* Use ENV for `wait-for-it` redis and redis-slave

Signed-off-by: Adrian Nöthlich <git@promasu.tech>

### DIFF
--- a/server/docker/entrypoint-db-setup
+++ b/server/docker/entrypoint-db-setup
@@ -24,7 +24,7 @@ source /run/secrets/django || true
 }
 export SECRET_KEY="$DJANGO_SECRET_KEY"
 
-until pg_isready -h "${DATABASE_HOST:-postgres}" -p "${DATABASE_PORT:-5432}"; do
+until pg_isready -h "${DATABASE_HOST:-pgbouncer}" -p "${DATABASE_PORT:-5432}"; do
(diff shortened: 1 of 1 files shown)


server/docker/entrypoint-db-setup
```diff
@@ -24,14 +24,14 @@ source /run/secrets/django || true
 }
 export SECRET_KEY="$DJANGO_SECRET_KEY"
 
-until pg_isready -h "${DATABASE_HOST}" -p "${DATABASE_PORT:-5432}"; do
+until pg_isready -h "${DATABASE_HOST:-postgres}" -p "${DATABASE_PORT:-5432}"; do
```
gsiv, line 27: This is probably more of a nit-pick than anything else.  `DATABASE_HOST` defaults to `pgbouncer` in the default configuration templates, not `postgres`.  Ideally, that default would be matched here.

However, I am aware that `docker-compose.dev.yml` diverges and uses `postgres`.  Since `postgres` is also an alias for `pgbouncer` in the other configuration templates, it should be alright.

gsiv (approved):

server/docker/entrypoint-db-setup
promasu: Done

PR #5956, opened 2021-03-13
Add manage service

ostcar:
This PR introduces a new service for management command.

To be used, it opens the port 8001. In production, this port has to be protected by htaccess or some other methods.

To use the management service, there is a helper-tool called `manage` You can find/build it in the openslides-manage-service with:

```
go build ./cmd/manage
./manage
```


FinnStutzenstein (approved):
The exact prod usage will be done later

PR #5969, opened 2021-03-19
Use fresh client dependencies in CI

FinnStutzenstein:


FinnStutzenstein:
Did not get it to work..

PR #5970, opened 2021-03-19
Client dependencies

FinnStutzenstein:
- Update ng2-pdf-viewer
- Lock down tsparticles and ng-particles to the version in the
package-lock.json. In fresh installs newer versions do not work
currently


matteobruni:
> * Update ng2-pdf-viewer
> * Lock down tsparticles and ng-particles to the version in the
>   package-lock.json. In fresh installs newer versions do not work
>   currently

Hello,

Can you tell me more about what is not working? If I broke something I'll fix it ASAP.

FinnStutzenstein:
Hi,
It is not an error with your code:
```
src/app/shared/components/applause-particle-display/applause-particle-display.component.ts:9:26 - error TS2307: Cannot find module 'tsparticles/dist/Plugins/Emitters/Emitters' or its corresponding type declarations.

import { Emitters } from 'tsparticles/dist/Plugins/Emitters/Emitters';
```
I think there were some API changes that needs to be reflected for us. This was the quick solution for other developers who got the errors to be able to develop again ;)

Thanks for the comment - how did you find this PR? :D

matteobruni:
Oh sorry, I make refactoring of undocumented files.

I found the PR using the GitHub search, I look for issues around sometimes, if someone needs help I’d like to help

FinnStutzenstein:
Cool, that quite some effort.

The import is a direct import:
```
import { Emitters } from 'tsparticles/dist/Plugins/Emitters/Emitters'
```
It seems, that it is not exported directly by tsparticles. I looked into the `node_modules` and the files `Emitter.js` and `Emitter.d.ts` are present - I am not sure why it does not want to import it.

matteobruni:
> Cool, that quite some effort.
> 
> The import is a direct import:
> 
> ```
> import { Emitters } from 'tsparticles/dist/Plugins/Emitters/Emitters'
> ```
> 
> It seems, that it is not exported directly by tsparticles. I looked into the `node_modules` and the files `Emitter.js` and `Emitter.d.ts` are present - I am not sure why it does not want to import it.

That file/class is for configuring the emitters plugin, if you want to add dynamically some emitters (outside options), you can use the `addEmitter` method on container. For now it needs a cast to `any` since it's not in the `Container` class, but it's added by that plugin and I'm figuring out which is the best method to let TypeScript know about that.

```Typescript
tsParticles.load({ /* options */ }).then(c => {
    (c as any).addEmitter({ /* emitter options, like the emitters section of the main options */, position /* optional, if not specified a random one will be used */); // this one is not documented yet, because of the any trick, I prefer to document it in a better way
});
```

If you just need emitters you can use the standard options like the samples you can find on CodePen (like [this](https://codepen.io/matteobruni/pen/qBqGPgQ))

FinnStutzenstein:
I'll pass this forward to @tsiegleauq . He has much more experience with the library :)

PR #5976, opened 2021-03-22
Fix LOS initially closed migration issue

FinnStutzenstein:


FinnStutzenstein:
@emanuelschuetze @gsiv This should fix `KeyError: 'agenda_list_of_speakers_initially_closed'` in the current tracebacks.

FinnStutzenstein:
@emanuelschuetze Another solution would be to delete the new migration (Currently we have two ones, the old and the one I created to revert it), but this is only possible if we have **no** instance updated and used the new list of speaker initial closed migration.

emanuelschuetze (approved):

PR #5977, opened 2021-03-22
Enable pytest in docker

FinnStutzenstein:
The DJANGO_SETTINGS_MODULE is set in the develop container, so the
tests.settings from the setup.cfg was ignored. Executing pytest with

    DJANGO_SETTINGS_MODULE=tests.settings pytest

works fine. The line added in the setup.cfg takes over precedence: It
auto-adds the --ds parameter which has a higher precedence than the
environment variable.

Also removed an unnecessary setting.

### DIFF
--- a/server/setup.cfg
+++ b/server/setup.cfg
@@ -34,6 +34,7 @@ disallow_untyped_defs = true
 
 [tool:pytest]
 DJANGO_SETTINGS_MODULE = tests.settings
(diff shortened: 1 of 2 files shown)


FinnStutzenstein:
https://pytest-django.readthedocs.io/en/latest/configuring_django.html#order-of-choosing-settings

jsangmeister (approved):

PR #5978, opened 2021-03-22
Implement voting changes

jsangmeister:
- add `user_token` to each vote object, add migrations for correct associations
- add `is_pseudoanonymized` field to polls for easier differentiation
- add single votes table for pseudoanonymous polls
- remove obsolet code for empty votes
- fix error where the deletion of a user would change the amount of valid votes for a poll
- fix error in Makefile where UID/GID was not set correctly
- add list of all users which are entitled to vote at stop of voting process
  - filterable by name & voting status (`voted:true` and `voted:false`)
  - not visible for users which have not `user.can_see_names`
- fix error where deleting candidates would break the single votes table in assignment polls
- add new 100% basis (amount of entitled users)


server/openslides/motions/migrations/0038_motionvote_user_token.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("motions", "0037_motionvote_delegated_user"),
+    ]
+
+    operations = [
+        migrations.AddField(
```
FinnStutzenstein, line 17: You need to split the three fields into three migration files. Postgresql cannot mix data and schema changes. Same for assignments

jsangmeister:
`npm run cleanup` has internal conflicts: prettier's uses its option `printWidth` more as a "preferred length" (https://prettier.io/docs/en/options.html#print-width), while the linter interprets it as a hard limit. This leads to the formatting seen in the first commit of this PR: prettier formats them into one line since it somehow "prefers" that and the linter complains.

I've found no way to set a hard line limit for prettier, any ideas @tsiegleauq @FinnStutzenstein ?

FinnStutzenstein:
> Specify the line length that the printer will wrap on.

Does not sound like a soft limit.

I actually never had this problem before. If this is just one single line that behaves strange, you can try to write it manually different...

jsangmeister:
> > Specify the line length that the printer will wrap on.
> 
> Does not sound like a soft limit.
> 
> I actually never had this problem before. If this is just one single line that behaves strange, you can try to write it manually different...

From the linked section: "Prettier’s printWidth option does not work the same way. It is not the hard upper allowed line length limit. It is a way to say to Prettier roughly how long you’d like lines to be. Prettier will make both shorter and longer lines, but generally strive to meet the specified printWidth."

It is two places (`current-list-of-speakers-overlay-slide.component.ts` and `motion-detail-original-change-recommendations.component.ts`) and yes, I can change that manually, but that has to be done after each time `npm run cleanup` is executed, since prettier reformats the line (I think) and it doesn't solve the underlying problem...

jsangmeister:
> but that has to be done after each time `npm run cleanup` is executed, since prettier reformats the line (I think)

Just tested it and yes, it reformats it every time.

jsangmeister:
Done from my side. @emanuelschuetze please test. We probably need some way to make the entitled users table more readable, like column headers, tooltips or just more verbose content.

The change detection does not work correctly for the entitled users table and I couldn't figure out why, e.g. when someone changes his name, the change is not reflected in the table until after a reload. But since the single votes table works the same way, I guess this is not so much of a problem.

jsangmeister:
(but still waiting for https://github.com/OpenSlides/openslides3-autoupdate-service/pull/148)

emanuelschuetze (changes requested):
Some feedback:
* entitled users table is missing for assignment polls! (motion polls works)
* TypeError if you project a motion poll with 100% base = all entitled users
* result table of motion/assignment polls should show the number of entitled users if 100% base = all entitled users

server/openslides/assignments/migrations/0019_assignmentvote_add_user_token.py
```diff
@@ -0,0 +1,18 @@
+# Generated by jsangmeister on 2021-03-18 16:27
+
```
FinnStutzenstein, line 2: Can you name them `user_token_1` to `user_token_3`, so it is clear, that they have to be seen as one unit.

server/openslides/assignments/migrations/0022_assignmentpoll_add_and_alter_fields.py
```diff
@@ -0,0 +1,63 @@
+# Generated by jsangmeister on 2021-03-22 12:44
```
FinnStutzenstein, line 1: Same here.

server/openslides/motions/migrations/0038_motionvote_add_user_token.py
```diff
@@ -0,0 +1,18 @@
+# Generated by jsangmeister on 2021-03-18 16:27
```
FinnStutzenstein, line 1: And the same for all motion migrations.

FinnStutzenstein (approved):
Looks good!

jsangmeister:
Pushed raw autoupdate changes since these are needed to solve the projection bug and restrict the data properly. I hope you can just start this, will open a seperate PR in the autoupdate repo.

emanuelschuetze (commented):
ok, tested

PR #5982, opened 2021-03-27
Update Autoupdate-service

ostcar:
I refactored same parts of the autoupdate. Most are tests and fixes.

One new "feature" is, that the cache gets cleared from time to time. For the moment, it is configured to clear the cache every 10 second. Later, I will change this to a hour or a day. I hope, the short time help to find bugs.


FinnStutzenstein:
Please set labels next time

FinnStutzenstein:
outdated

PR #6006, opened 2021-04-12
Lock poll to prevent race conditions

jsangmeister:
To test, run dev setup with gunicorn, multiple workers and postgres backend and add a `time.sleep(5)` in `server/openslides/poll/views.py:304`. Then you can vote in a poll, which will hang for 5 seconds, during which you can stop the poll. This request will in turn wait until the vote is finished and then stop the poll and count the votes correctly.


jsangmeister:
Added a migration to fix wrong vote values. Only executing them for motion polls since assignment polls might have differing user tokens for votes from the same user, which would lead to new wrong values.

The only other possible scenario where this migrations leads to a wrong result would be if vote weight was enabled during a poll, then later disabled, all users had weight 1 except those which happened to vote right as the poll was stopped and whose votes therefore did not count into `votesvalid`/`votescast`. This, however, is unlikely enough and moreover, the new result would still be nearer to the correct value than the old one.

jsangmeister:
Added a migration for named assignment polls. Pseudoanonymous assignment polls are excluded because the correct amount for the votes fields is calculated based on the user tokens. And pseudoanonymous assignment polls which were created before we introduced the user tokens do not have the same token for votes by the same user, so it is impossible to determine the correct values there.

emanuelschuetze (approved):

FinnStutzenstein (approved):

emanuelschuetze:
Rebased

PR #6016, opened 2021-04-19
poll backend

FinnStutzenstein:


jsangmeister (approved):

ostcar (approved):

normanjaeckel:
The opposite of long is short, not fast. Or the opposite of fast is slow, not long.

ostcar:
The goal of the Postgresql backend is not to be slow and the goal of the redis backend not to be short.

Both backend have a different, not related, not opposite, purpose. The name tells, what the advantage is. One is fast and the other can be used for long going polls.

PR #6018, opened 2021-04-19
Change locked fields to a list

jsangmeister:


jsangmeister:
This is needed since with the filter locking options, a field can now be potentially locked multiple times - e.g. if `filter` is called multiple times on the same field with differing filters and the results should all be locked.

jsangmeister:
Or, other suggestion: Since this is only needed for collection field locks, we could also change only the collection field key too accept an array of locks instead of only one value. This would probably the better way to do it, since we then have no problems with validating that e.g. fqids are not given multiple times.

jsangmeister:
Changed interface so the change is non-breaking. Also added possibility to leave filter empty, so it can be locked with filters and without filters together.

FinnStutzenstein (approved):

jsangmeister:
waiting for https://github.com/OpenSlides/openslides-datastore-service/pull/125

PR #6024, opened 2021-04-22
Fix vote delegation update error

jsangmeister:
It was wrongly checked that the user must not have delegated his own vote in order to receive vote delegations even when the list of received delegations was empty.

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -19,6 +19,8 @@ services:
     depends_on:
       - redis
       - postgres
+    stdin_open: true
+    tty: true
 
   postgres:
     image: postgres:11
--- a/server/openslides/users/views.py
+++ b/server/openslides/users/views.py
@@ -167,6 +167,11 @@ class UserViewSet(ModelViewSet):
         ):
             request.data["username"] = user.username
 
+        # handle delegated_from field seperately since its a SerializerMethodField
+        new_delegation_ids = request.data.get("vote_delegated_from_users_id")
+        if "vote_delegated_from_users_id" in request.data:
+            del request.data["vote_delegated_from_users_id"]
+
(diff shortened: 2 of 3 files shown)


docker/docker-compose.dev.yml
```diff
@@ -19,6 +19,8 @@ services:
     depends_on:
       - redis
       - postgres
+    stdin_open: true
+    tty: true
```
jsangmeister, line 23: I think this is generally useful for debugging.

tsiegleauq (changes requested):
Removing the "Vote right delegated to (Proxy)" does not update the recipient of the vote.
Not sure what side effects that might cause.
image

- userA "vote right delegate to" userB
  - so userB can vote for userA
- remove userA "vote right delegate to" userB
- userB still has the delegated vote, although userA does not delegate their vote anymore

jsangmeister:
Good catch. There was a missing autoupdate, it should be fixed now (the server-side data was correct).

tsiegleauq (changes requested):
That might be minor, since no direct error would occur in most cases:

full re-alignment of voting rights for one user does not work in one request.

before:
image

change to:
image

results in: `Error: You cannot delegate a vote to a user who has already delegated his vote.`

It works just fine in two requests. If this cannot be fixed (trivially) we might just ignore it, yet it's counter intuitive in some cases.

- a received voting rights from d
- d does wants to withdraw the voting proxy from a *and* wants to get the voting rights from e

docker/docker-compose.dev.yml
FinnStutzenstein: ok

jsangmeister:
Fix that. Also fixed the reverse case: resetting the `from` side while simultaniously setting the `to` side.

PR #6027, opened 2021-04-24
Manage setup command

normanjaeckel:


normanjaeckel:
Please test this pull request. Attention: It will not work as described in new README because the manage tool references to openslides4-dev branch and not to this pull request. It will work after this pull request is merged.

To test this you have to run the following steps:

- Get the manage tool: Later we provide prebuild binaries for most systems but at the moment you have to build the manage tool yourself. Checkout this pull request, run submodule update, go to openslides-manage-service and run `go build ./cmd/manage`
- Copy the `manage` binary to a nice place in your filesystem.
- Then you can follow the steps in the new README. After the first setup command you have to edit the docker-compose.yml: Go to the entry for the manage service and remove the commit id from the build context URL, so it points to main branch and not to the wrong commit of openslides4-dev.

Please let me know if this works for you and is a useful way to run OpenSlides 4 with Docker Compose for the community.

Already planned enhancements:
- Using XDG_DATA_PATH instead of local directory.
- Building images and creating initial-data in one step. For this I want to extend the setup command.

The company version setup is another step that we can discuss later.

Attention: The new setup command also provides a way to setup a production environment from a cloned repository. Do we want to explain this in README or do we want to keep this as hidden feature?

normanjaeckel:
@ostcar Please also check this!

normanjaeckel:
Update: Maybe this should already work because someone updated the branch in another pull request.

normanjaeckel:
I updated this pull request. You can test it in two ways:
1. Go to a empty directory in your filesystem and follow the new README. This way you should get a running OpenSlides 4.0.0-dev with Docker Compose.
2. Use your local codebase:
    - Check out this branch.
    - Get the mange tool either from GitHub (as stated in README) or go inside `openslides-manage-service` and run `go build ./cmd/manage`.
    - Run `./manage setup --cwd --local`. This way you get a docker-compose.yml that can be used with your local code base.
    - Run e. g. `docker-compose up --build` and go on as usual.

After this pull request is merged I will update the manage tool so that it fetches the images from OpenSlides organisation account and not my personal one.

The setup with Docker Swarm etc. is an extra part. Please lets talk about the requirements.

.github/workflows/models.yml
```diff
@@ -1,7 +1,9 @@
-name: Validate models.yml
+---
+name: Validate models.yml and initial and example data
 on: [push, pull_request]
 env:
   PYTHON_VERSION: 3.8.5
+  GO_VERSION: 1.15
```
ostcar, line 6: Why not using go 1.16?

.github/workflows/models.yml
```diff
@@ -10,7 +12,7 @@ jobs:
     - name: Set up Go 1.15
```
ostcar, line 12: You should rename this to "set up Go"

README.md
```diff
@@ -8,48 +8,59 @@ https://openslides.com for more information.
...
 
-For a non-development setup, clone this repo and run it via docker compose. The make command is a handy shortcut for this:
+### Setup OpenSlides
 
-    $ git clone git@github.com:OpenSlides/OpenSlides.git
-    $ cd OpenSlides
-    $ git checkout openslides4-dev  # needed, until OS4 is released
-    $ make run-prod
+Go to a nice place in your filesystem and get the [OpenSlides manage
+tool](https://github.com/OpenSlides/openslides-manage-service/releases/tag/latest).
```
ostcar, line 29: Nobody wants to go to another place. I would suggest to add a wget command: like `wget https://github.com/OpenSlides/openslides-manage-service/releases/download/latest/manage`. If you don't want to add wget as a "requirement", then you could add this line as an example:

```
Get the manage tool from .... For example with wget ...
```

ostcar: Since the user needs docker anyway, we could release the tool as docker-container. Then you would call it like this:
```
docker run manage setup --cwd
docker  run manage initial-data
...
```
Then you could add a github action to create the container. This would also work on other platforms like Windows and Mac.

README.md
```diff
@@ -8,48 +8,59 @@ https://openslides.com for more information.
...
+have to run this as root:
+
+    $ docker-compose up --detach
+
+Setup initial-data:
 
+    $ ./manage initial-data
 
-## Productive environment
+Now you can open http://localhost:8000 and use OpenSlides.
```
ostcar, line 44: The setup has to start openslides on https. In other case the browser will use http1.1 that does not perform good with the autoupdate service. You can try this by opening more then four tabs. They will not start the client (since they can not create any http connections. The max connections of your browser to localhost is reached).

ostcar:
I can confirm that it works

README.md
ostcar: For this to work, the manage-container has to write files into the host container: https://stackoverflow.com/questions/31448821/how-to-write-data-to-host-file-system-from-docker-container

(Or it can write the files into its own container. Then all configuration is bundled in this container and we do not need the xdg path)

README.md
normanjaeckel: https://github.com/OpenSlides/OpenSlides/issues/6044

normanjaeckel:
Can I merge this? The next step would be to use a smarter way to start OpenSlides with one single command. This is already work in progress ...

ostcar (approved):

PR #6029, opened 2021-04-26
Remodel users

FinnStutzenstein:


docs/models.yml
```diff
@@ -126,18 +126,17 @@ user:
...
-    to: committee/manager_ids
+    to: committee/user_ids
+
+  # committee specific permissions
+  committee_$_management_level:
+    type: template
+    replacement: committee_id
+    description: Hierarchical permission level for the each committee organisation.
+    enum:
+    - manager
```
jsangmeister, line 139: I would suggest to keep the naming scheme of our other permissions and name this `can_manage`

jsangmeister: But regardless of that, it says `manage` instead of `manager` in the `initial_data`. Shouldn't this be caught by the CI? Can you look into this?

docs/models.yml
```diff
@@ -126,18 +126,17 @@ user:
...
-  committee_as_manager_ids:
+  committee_ids:
     type: relation-list
-    to: committee/manager_ids
+    to: committee/user_ids
+
+  # committee specific permissions
+  committee_$_management_level:
+    type: template
+    replacement: committee_id
```
jsangmeister, line 136: This _probably_ has to say committee_id**s** since it must refer to an existing field on the model. I am not quite sure if the backend is able to handle relation list fields here, but we will see, I guess

jsangmeister: Same goes for all structured meeting fields which have to be changed to meeting_id**s**

jsangmeister (changes requested):

docs/models.yml
r-peschke, line 139: "can_manage_committees"?

docs/models.yml
```diff
@@ -868,12 +869,6 @@ meeting:
...
     to: user/is_present_in_meeting_ids
-  temporary_user_ids:
-    type: relation-list
-    to: user/meeting_id
-  guest_ids:
-    type: relation-list
-    to: user/guest_meeting_ids
   user_ids:
     type: number[]
     decription: Calculated. All ids from temporary_user_ids, guest_ids and all users assigned to groups.
```
r-peschke, line 874: The calculation base may be slightly different;-)

docs/example-data.json
```diff
@@ -188,7 +187,9 @@
         "current_projector_$_ids": [],
         "vote_delegated_vote_$_ids": [],
         "vote_delegated_$_to_id": [],
-        "vote_delegations_$_from_ids": []
+        "vote_delegations_$_from_ids": [],
+
+        "meeting_ids": [1]
```
r-peschke, line 192: Isn't it a calculated field based on group-membership? Then I would expect to be a group set and this one calculated

docs/models.yml
FinnStutzenstein: structured fields are not implemented by the datavalidator yet.

docs/models.yml
FinnStutzenstein: So it is already wrong everywhere? If so, it better be fixed in another PR to isolate this fix.

jsangmeister: No, it became wrong because you removed the field `meeting_id` (or renamed it to `meeting_ids`, whatever) to which all structured fields previously referred.

docs/models.yml
jsangmeister: Ah ok. But checking just the value (ignoring every aspect of them being structured) should be easy to implement, right?

docs/example-data.json
FinnStutzenstein: type, fixed. Thanks!

docs/models.yml
FinnStutzenstein: I've chosen `can_manage`

FinnStutzenstein:
Note: user 3 is not assigned to the committee in the example data to emulate a guest.

docs/models.yml
```diff
@@ -300,12 +304,9 @@ committee:
...
     type: relation
     to: meeting/default_meeting_for_committee_id
-  member_ids:
-    type: relation-list
-    to: user/committee_as_member_ids
-  manager_ids:
+  user_ids:
     type: relation-list
-    to: user/committee_as_manager_ids
+    to: user/committee_ids
```
GabrielInTheWorld, line 309: Does this mean, every user in a committee can also manage the committee?

jsangmeister: No, that is what the new `committee_management_level` is for.

docs/models.yml
r-peschke: I'm not quite sure about the meeting_id. Formerly it has been the signal for the temporary user. The meeting_ids is/should IMO a calculated field of meetings from group-assignments.
The meeting related fields are template fields (=structured fields?), which have their own meeting_id.

docs/models.yml
GabrielInTheWorld: Okay, that sounds good.

GabrielInTheWorld (approved):

docs/models.yml
jsangmeister: Oh, you are right! I didn't even think about that. That is highly unfortunate. This probably means we have to restructure the way template relations are defined and work in the backend, since `user` does not have a direct relation field to `meeting` any more. My first idea would be two enhance the `replacement` field in this schema to take two possible values:
- A field of the same model, just like before. It must be ensured that the field exists on the model. We could theoretically (while we're at it) also ensure in the relation handling that all replacements given for the template field [1] must be present in the given relation. Should we do this? Does anyone know a field where this would not be the case?
- A CollectionField, like `meeting/id`. This also suffices to check if the referred model exists. If we want to implement the idea from above, we could add another field to the model definition for this case like `calculated_field` to indicate which field on this model corresponds to the given collection field. Then, we could implement the same check also in this case.

Let me know what you think about that or if you have other ideas.

[1] Reminder: _Template fields_ are the "main" fields with just a `$` which hold all _replacements_. When inserting the replacement in the template field's name, you get a _structured field_

FinnStutzenstein:
Please review again

jsangmeister (changes requested):
See my other comment. Maybe we should talk about it later

docs/models.yml
FinnStutzenstein: structured fields checking in general should not be a huge problem. Maybe I find some spare time...

(Currently, every field containing a `$` is just skipped for the sake of simplicity)

docs/models.yml
r-peschke: Why do you think the replacement field must exist in the model? If we want we can check in the backend, whether the user belongs ( = is member of a group of that meeting) if the user wants to store for instance an about_me_$ for meeting 222.

FinnStutzenstein: Resolved: Use a new `replacement_collection` to point to the foreign collection the replacement ids belonging to.

FinnStutzenstein:
Done, please review again.

docs/models.yml
```diff
@@ -46,8 +46,8 @@
 #     - There are template fields (see autoupdate service interface) with a `$` as
 #       the placeholder.
 #     - The type `template` describes a structured field for the given model. If the
-#       property `replacement` is given, it describes which field of the same model
-#       is used as the replacement (=> structured relation). If it not given, the field
+#       property `replacement_collection` is given, it describes for which model the
```
jsangmeister, line 49:
```suggestion
#       property `replacement_collection` is given, it describes which model the
```

docs/models.yml
```diff
@@ -46,8 +46,8 @@
...
+#       replacement ids are belonging to (=> structured relation). If it not given, the field
```
jsangmeister, line 50:
```suggestion
#       replacement ids are belonging to (=> structured relation). If it is not given, the field
```

docs/models.yml
```diff
@@ -868,15 +868,9 @@ meeting:
...
-  temporary_user_ids:
-    type: relation-list
-    to: user/meeting_id
-  guest_ids:
-    type: relation-list
-    to: user/guest_meeting_ids
   user_ids:
     type: number[]
-    decription: Calculated. All ids from temporary_user_ids, guest_ids and all users assigned to groups.
+    decription: Calculated. All user ids from all users assigned to groups.
```
jsangmeister, line 873:
```suggestion
    decription: Calculated. All user ids from all users assigned to groups of this meeting.
```

jsangmeister:
Minor improvements possible. Looks good structurally

PR #6038, opened 2021-04-29
Fix LOS (closes #6035)

### LINKED ISSUE #6035: pro/contra/mark speakers buttons / read permission for own point of order note
splitted our from PR #6023:

1. If the config "agenda_list_of_speakers_speaker_note_for_everyone" is disabled the note is visible for LOS-managers only - the submitter can NOT see his/her own note!
2. The mark-speakers-with-star-button should work as radio button. The server should disable pro/contra option if client sends mark option. (Otherwise the client would have to send two requests...)

FinnStutzenstein:


FinnStutzenstein:
Wait, this may have a bug - I'll see

emanuelschuetze (approved):

PR #6041, opened 2021-04-29
Pro/contra/marked/note for OS4

FinnStutzenstein:
@GabrielInTheWorld I splitted the `pro_speech` (OS3) to `pro_speech` and `contra_speech` since the backend does not support nullable booleans (tristate booleans).

### DIFF
--- a/docs/example-data.json
+++ b/docs/example-data.json
@@ -276,6 +276,9 @@
         "list_of_speakers_present_users_only": false,
         "list_of_speakers_show_first_contribution": false,
         "list_of_speakers_enable_point_of_order_speakers": true,
+        "list_of_speakers_enable_pro_contra_speech": true,
+        "list_of_speakers_can_set_contribution_self": true,
+        "list_of_speakers_speaker_note_for_everyone": true,
         "list_of_speakers_initially_closed": true,
 
         "motions_default_workflow_id": 1,
@@ -1065,7 +1068,8 @@
         "begin_time": 1584512636,
         "end_time": 1584512638,
         "weight": null,
-        "marked": false,
+        "speech_state": null,
+        "note": null,
         "point_of_order": false,
 
         "list_of_speakers_id": 1,
@@ -1077,7 +1081,8 @@
         "begin_time": null,
         "end_time": null,
         "weight": 2,
-        "marked": false,
+        "speech_state": null,
+        "note": null,
         "point_of_order": false,
 
         "list_of_speakers_id": 1,
@@ -1089,7 +1094,8 @@
         "begin_time": null,
         "end_time": null,
         "weight": 3,
-        "marked": false,
+        "speech_state": null,
+        "note": null,
         "point_of_order": false,
 
         "list_of_speakers_id": 1,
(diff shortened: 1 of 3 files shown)


FinnStutzenstein:
When this is merged, these actions has to be taken:
- @GabrielInTheWorld Adapt https://github.com/OpenSlides/openslides-client/pull/202
- https://github.com/OpenSlides/openslides-backend/issues/695 can be done
- https://github.com/OpenSlides/openslides-backend/issues/694 can be done

docs/models.yml
```diff
@@ -515,7 +525,7 @@ meeting:
     default: outside
   motions_line_length:
     type: number
-    minimium: 40
+    minimum: 40
```
jsangmeister, line 528: Shouldn't this have been caught by the validator?

docs/models.yml
```diff
@@ -1163,7 +1173,18 @@ speaker:
...
-  marked: boolean
+  marked:
+    type: boolean
+    default: False
+  pro_speech:
+    type: boolean
+    default: False
+  contra_speech:
+    type: boolean
+    default: False
```
jsangmeister, line 1184: I don't like the split into two fields, it just makes room for undefined datastore content and situations. Why not something like
```
speech_status:
  type: string
  enum:
    - pro
    - contra
```
and if it's null, no status is set?

jsangmeister (changes requested):

docs/models.yml
FinnStutzenstein: I thought so.....

I'll have a look later and also add template fields

docs/models.yml
FinnStutzenstein: In combination with the second point in https://github.com/OpenSlides/OpenSlides/issues/6035 I would propose to do this:
```
speech_status:
  type: string
  enum:
    - none
    - pro
    - contra
    - contribution
```

I think we need the `none` case, since null is also not allowed here if I'm correct

jsangmeister: If the field is not required, `None` should be allowed. The other problem with your suggestion is that `none` would be interpreted as a string which means we would save the literal string `"none"` in the datastore and I hope you are not seriously suggesting this :D

Besides that point, that seems fine. I did not look closely into #6035 but if that warrants its own enum value, that is exactly the way to do it.

FinnStutzenstein: > If the field is not required, None should be allowed

ok it is fine than

docs/example-data.json
```diff
@@ -1065,7 +1068,8 @@
         "begin_time": 1584512636,
         "end_time": 1584512638,
         "weight": null,
-        "marked": false,
+        "speech_state": null,
+        "note": "",
```
jsangmeister, line 1072: Shouldn't note be either `null` or not existent if `point_of_order` is false?

FinnStutzenstein: `null` is better; otherwise the datavalidator complains about a missing field (We do want to have every field explicitly here even though `undefined===null`)

jsangmeister (approved):

FinnStutzenstein:
@GabrielInTheWorld Note the renaming from `list_of_speakers_can_set_mark_self` to `list_of_speakers_can_set_contribution_self` since we do not have the "mark" anymore

Also note the new `speech_state` which is `None`/`null` by default.

PR #6054, opened 2021-05-06
Speed up stopping a poll

FinnStutzenstein:
Reduced the amount of queries in BasePoll.stop from 4+`<amount entitled users>` to 3.

### DIFF
--- a/server/openslides/poll/models.py
+++ b/server/openslides/poll/models.py
@@ -264,18 +264,25 @@ class BasePoll(models.Model):
             self.is_pseudoanonymized = False
         self.save()
 
-    def calculate_votes(self):
-        if self.type != BasePoll.TYPE_ANALOG:
-            self.votescast = self.voted.count()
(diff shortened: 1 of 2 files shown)


jsangmeister (approved):

jsangmeister:
@FinnStutzenstein did you run your new test? :D The count of 3 was too low since there are additional queries being run from other code places. I changed the test to not execute a full client request, but only the stop call, and changed the count to 13.

emanuelschuetze (approved):
Works great in my tests (now 4sec instead 51sec for stopping voting with 200 of 2000 voting users!)

FinnStutzenstein:
> did you run your new test?

There were 36 additional queries when doing the full request. But with a previos-current-comparison, this was the amount of queries of the one function inside the complete request. It might be more queries now, but mainly due to the fact that when executing the complete query, some DB-queries are done before the stop method and already fetched.

PR #6056, opened 2021-05-08
Performance assignment poll

ostcar:


emanuelschuetze (changes requested):
I tested your assigment voting script:
- why an admin login is required for assignment voting? (e.g. performance votes -a 100 --interreupt -i 42 -d example.com -u admin -p admin) -> works without admin login for motion poll.
- YNA election does not work, only Y election works
- I prefere to use "--motion" when we introduce "--assignment"
- Can you add short flags for --answer (e.g. -c for 'choice') --assignment (-a), --motion (-m) and --amount (-n)

ostcar:
>     * why an admin login is required for assignment voting? (e.g. performance votes -a 100 --interreupt -i 42 -d example.com -u admin -p admin) -> works without admin login for motion poll.
On assignment-votes the client needs the option-ids. Therefore a read request before the votes is required. I changed the code to use "dummy1" for this read request.

>     * YNA election does not work, only Y election works
Currently only "Y" elections are supported. Do you think it is necessary for the performance tool to test different votes? Do you think this makes a difference in performance?

>     * I prefere to use "--motion" when we introduce "--assignment"
I changed it as arguments. Now you have to run:

```
performance votes motion ...
performance votees assignment ...
```

The arguments `m` and `a` do also work.

>     * Can you add short flags for --answer (e.g. -c for 'choice') --assignment (-a), --motion (-m) and --amount (-n)
done

emanuelschuetze:
I retested your performance script. Works good! Two minor documentation issues:
* The votes  flags `motion` and `assignment` are not listed in the help (`performance votes --help`). Maybe an example command would be helpful.
* The help text sentence with "old_url" is outdated and can removed, right?

emanuelschuetze:
@ostcar ping for the last changes ;-)

ostcar:
Pong

ostcar:
I fixed the help text. I don't know how to automatically add the arguments to the help text by the cli-lib that I am using. I hope, the examples in the help text is enough.

emanuelschuetze (approved):

PR #6116, opened 2021-06-14
Move custom_translation from organization to meeting

r-peschke:


jsangmeister (changes requested):
example/initial data has to be adjusted

jsangmeister (approved):

docker/initial-data.json
```diff
@@ -104,6 +103,7 @@
         "url_name": "openslides_1",
         "template_for_committee_id": null,
         "enable_anonymous": false,
+        "custom_translations": "",
```
FinnStutzenstein, line 106: `""` seems to be a mistake, can you change it to `[]` please?

jsangmeister: Ah, I wondered about it, but I thought maybe it should be saved in a stringified version ;) But if that's not the case, shouldn't it be `{}` instead of `[]`? Or what is the schema of the translations?

r-peschke: I think it is {}, a dictionary in Python notation. The only test with non empty data I found, is a projection.options field with a python dict.

PR #6124, opened 2021-06-21
Enhance datavalidator for template fields

FinnStutzenstein:


docs/datavalidator/check_json.py
```diff
@@ -125,9 +220,25 @@ def check_collections(self) -> None:
             raise CheckException(err)
 
     def check_model(self, collection: str, model: Dict[str, Any]) -> None:
-        model_fields = set(x for x in model.keys() if "$" not in x)
+        errors = self.check_normal_fields(model, collection)
+
+        if not errors:
```
jsangmeister, line 225: Isn't it better to check everything, regardless if there were errors?

jsangmeister (approved):
Looks good on first sight, did not test it further. Will https://github.com/OpenSlides/OpenSlides/pull/6062 be also fixed with this?

jsangmeister:
Also https://github.com/OpenSlides/OpenSlides/pull/6052: please check if the two errors in the linked issue can no longer occur.

docs/datavalidator/check_json.py
FinnStutzenstein: The later checks rely on these checks to succeed. E.g. when checking relations, I can be sure, that all fields are there, I just have to check values.

FinnStutzenstein:
No, these issues belong to the models validator which I didn't change

jsangmeister:
Ahh ok, my bad

PR #6131, opened 2021-06-27
Add ICC Service

ostcar:


docker/docker-compose.dev.yml
```diff
@@ -27,14 +28,14 @@ services:
       - COMMAND=create_initial_data
       - DATASTORE_INITIAL_DATA_FILE=https://raw.githubusercontent.com/OpenSlides/OpenSlides/openslides4-dev/docs/example-data.json
       - OPENSLIDES_DEVELOPMENT=1
-    ports:
```
FinnStutzenstein, line 30: This breaks https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/dev-commands/set-ds.sh

Please add it back or implement a different solution to `set-ds.sh`

tsiegleauq (approved):
Tested, developed client-icc using this branch.
can be merged after resolving the previous request

PR #6152, opened 2021-07-09
Moved production setup to manage service.

normanjaeckel:


normanjaeckel:
See our current README. It contains a instruction how to setup the production system with Docker Compose at the moment. HTTPS support is still missing. Also the solution for Docker Swarm is under development.

docs/datavalidator/check_json.py
```diff
@@ -10,14 +10,13 @@
 MODELS_YML_PATH = "../../docs/models.yml"
```
FinnStutzenstein, line 10: You need to adjust this in the backend (https://github.com/OpenSlides/openslides-backend/pull/826) This file ist just left-over now. Also using two styles is not good, since this formet is the current way to import/export OpenSlides data.

FinnStutzenstein: Also I'm not a fan of moving this data out since updates like this (https://github.com/OpenSlides/OpenSlides/pull/6153) are not captured.

jsangmeister: Me neither, we just talked about it this morning and the consent was to leave both data files in the main repo and validate it with the data validator from the backend.

normanjaeckel: So how can I compile this file into my Go code without too much effort? The manage tool should contain the default initial data and it is part of the binary.

normanjaeckel: Ok, I readded the new initial-data.json file in array style format. Changing the format to object-style is an extra issue (see https://github.com/OpenSlides/openslides-backend/issues/846). Whether the check_json.py script is obsolete here or not is no urgent question for me.

An extra point is that I strongly recommend moving initial data (and maybe also example data) to the manage service so that updates are mentioned there (and lead to a new commit and a new binary in main branch). Anything else would be extra work every time ...

FinnStutzenstein: Are the changes in `check_json.py` still needed?

PR #6192, opened 2021-07-31
Add field group info in organization

ostcar:


ostcar:
@FinnStutzenstein @tsiegleauq  Please merge this as fast as possible. I don't want to do this again if there are any merge conflicts

docs/datavalidator/check_json.py
```diff
@@ -202,6 +202,9 @@ def is_structured_field(self, field: str) -> bool:
 
     def is_normal_field(self, field: str) -> bool:
         return "$" not in field
+    
+    def is_calculated_field(self, field: str) -> bool:
+        return field == "content"
```
ostcar, line 207: See #6195

docs/models.yml
```diff
@@ -1190,382 +1531,506 @@ list_of_speakers:
...
     type: relation
     to: motion/derived_motion_ids  # Note: The related motions may not be in the same meeting
+    restriction_mode: B
   derived_motion_ids:
     type: relation-list
     to: motion/origin_id  # Note: The related motions may not be in the same meeting
-  all_origin_ids: number[]  # Calculated: All parents (origin_id), grandparents, etc. of this motion.
-  all_derived_motion_ids: number[]  # Calculated: All children (derived_motion_ids), grandchildren, etc. of this motion.
+    restriction_mode: B
+  all_origin_ids:
```
FinnStutzenstein, line 1739: `restriction_mode` is missing here and below

docs/models.yml
```diff
@@ -1769,256 +2292,351 @@ poll:
...
       field: poll_ids
     equal_fields: meeting_id
+    restriction_mode: A
   option_ids:
     type: relation-list
     to: option/poll_id
     on_delete: CASCADE
     equal_fields: meeting_id
+    restriction_mode: A
+
```
FinnStutzenstein, line 2358: please remove

FinnStutzenstein (changes requested):
Two small remarks. Otherwise ok. I think we does not have to wait for the backend since the models-generation is a manual step.

FinnStutzenstein:
I'll enhance the models-validator the next days. Maybe I'll find some issues then ;)

docs/models.yml
```diff
@@ -2028,26 +2646,32 @@ mediafile:
...
     type: template
     fields:
       type: relation
       to: meeting/font_$_id
+    restriction_mode: B
 
 projector:
-  id: number
-  name: string
+  id:
```
ostcar, line 2671: I forgot the projector

docs/models.yml
ostcar: The two fields are not documented in the wiki. Which group should I use?

https://github.com/OpenSlides/OpenSlides/wiki/motion-restrictions

docs/models.yml
ostcar: done

docs/models.yml
FinnStutzenstein: Oh, the field `forwarding_tree_motion_ids` was removed and replaced with both of them; needs a fix in the wiki, too.

docs/models.yml
ostcar: done

docs/models.yml
ostcar: So maybe it would be a good idea if the fields where only defined in the models.yml 😉

done

FinnStutzenstein (approved):

PR #6194, opened 2021-08-02
Add missing defaults & fix existing ones

jsangmeister:
Some defaults that were present in the client where missing in the models file - good thing that we are unifying them.


FinnStutzenstein (approved):

ostcar (changes requested):
Please first merge  #6192

ostcar (approved):

PR #6203, opened 2021-08-09
Updated manage service.

normanjaeckel:


normanjaeckel:
Please not the new manage service has some changes in how to use it. The binary is now called `openslides` and the setup command is used a little bit different. Just look at the help text and the README (here and in manage repository).

I kindly ask all of you to help me with the following questions:
- I prefer moving the initial-data.json to the manage service repository. If not, please help me to get a consistant version. How can we ensure that further changes are reflected in the manage service (where the initial-data become part of the compiled binary).
- Someone may refactor the development version. The productive version contains several important changes that are not backported to the dev version yet. Who has some time for this?
- Which dev tools/scripts/snippets can be removed here? I think we do not need everything, that can be found here at the moment.

Thanks in advance.

normanjaeckel:
Please merge https://github.com/OpenSlides/openslides-backend/pull/902 first. Then we can update the backend here and then the validation check should work.

FinnStutzenstein:
> I prefer moving the initial-data.json to the manage service repository. If not, please help me to get a consistant version. How can we ensure that further changes are reflected in the manage service

Using migrations. We have two ways to go:
- leave the migration index of the initial-data.json at -1. This requires to update this file on every schema change.
- Fix "Day X" (We internally call it "Day X" since we do not know when it is yet): This is the point in time where we fix the `models.yml` to have MI 1. From this point in time we always have to write migrations, but this allows to fix the MI of the initial data to 1. So the manage service can inject "old" data (with MI 1) and since the migrations must always run before starting the backend, the initial data will be migrated.

See also https://github.com/OpenSlides/openslides-manage-service/issues/58#issuecomment-897381544 since it is highly coupled to this question. But we are free to decide: Both ways work, are fine and both have pro's:

- Leaving at MI=-1: One (Especially Emanuel) is flexible to provide data that may differ from things a future migration would do
- Setting MI=1: You do not have to update this file too much.

I guess you and emanuel (and others who might be affected) have to decide the way to go.

> Which dev tools/scripts/snippets can be removed here?

IMO removing is a bad option, better replace them. E.g. the scripts in dev-commands/ are very useful to get access to the datastore to check, what actual content is in it and you can export the content and some other dev can import it to e.g. trace down a bug.

But I do not know if any other than me is using it. These are nice options, but if no one uses/wants them they may be removed. So I will forward the question to others: @jsangmeister @GabrielInTheWorld @tsiegleauq

FinnStutzenstein:
> We have two ways to go

With https://github.com/OpenSlides/OpenSlides/pull/6205 we decided to fix the `initial-data.json`. If someone wants to modify migrated data, one has to provide a new `initial-data.json` with a higher MI.

I'll add a parameter to the `write` route of the datastore to provide an initial migration index. This will be only available, if the datastore is empty, so for the first write request.

PR #6205, opened 2021-08-10
Issue6204: Define fields for archiving of meetings

r-peschke:
@FinnStutzenstein : Please make this job required for merging pull requests

### DIFF
--- a/.github/workflows/md5_check_initial-data.yml
+++ b/.github/workflows/md5_check_initial-data.yml
@@ -1,0 +1,30 @@
+---
+name: Prevent manual changes of initial-data.json without md5 hash
+# The file initial-data.json has a migration-index (MI) set to an active migration,
+# In consequence setting up a new system triggers migrations beginning at this MI.
+# That's the reason this file is protected against unwanted/unintended edits.
+#
+# If one is sure to change the initial-data.json file it is allowed!
+# 1. Change the file
+# 2. Get the md5sum for changed initial-dta.json
+# 3. Change the md5-check-literal in this workflow-file
+on:
+  push:
+    paths:
+      - "docker/initial-data.json"
+  pull_request:
+    paths:
+      - "docker/initial-data.json"
+jobs:
+  validate-data:
+    name: Validate example-data.json and initial-data.json
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Show hash for initial-data.json
+        run: md5sum docker/initial-data.json
+
+      - name: Validate and protect initial-data.json
+        run: echo "79e29bd433253fac11899059d59cebe3 docker/initial-data.json" | md5sum -c -
--- a/docker/initial-data.json
+++ b/docker/initial-data.json
@@ -1,5 +1,5 @@
 {
-    "_migration_index": -1,
+    "_migration_index": 1,
     "organization": [
         {
             "id": 1,
@@ -11,9 +11,7 @@
             "theme": "openslides-default-light-theme",
             "reset_password_verbose_errors": false,
             "enable_electronic_voting": false,
-            "committee_ids": [
-                1
-            ],
+            "committee_ids": [1],
             "resource_ids": [],
             "organization_tag_ids": []
         }
--- a/docs/models.yml
+++ b/docs/models.yml
@@ -92,16 +92,25 @@ organization:
     type: boolean
     restriction_mode: B
 
-
   # Configuration (only for the server owner)
   enable_electronic_voting:
     type: boolean
     restriction_mode: B
+  limit_of_meetings:
+    type: number
+    description: Maximum of active meetings for the whole organization. 0 means no limitation at all
+    restriction_mode: B
+    default: 0
+    minimum: 0
 
   committee_ids:
     type: relation-list
     restriction_mode: B
     to: committee/organization_id
+  active_meeting_ids:
+    type: relation-list
+    restriction_mode: B
+    to: meeting/is_active_in_organization_id
   resource_ids:
     type: relation-list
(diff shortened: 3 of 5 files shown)


docs/models.yml
```diff
@@ -92,16 +92,23 @@ organization:
     type: boolean
     restriction_mode: B
 
-
   # Configuration (only for the server owner)
   enable_electronic_voting:
     type: boolean
     restriction_mode: B
+  limit_of_meetings:
```
FinnStutzenstein, line 99: One question here: Is there a way to disable the limiting (Would be good to have!)? Whats about the values 0, -1, ... here. I think you should set a minimum.

docs/models.yml
```diff
@@ -450,6 +457,12 @@ meeting:
     maxLength: 100
     default: OpenSlides
     restriction_mode: A
+  is_active:
+    type: relation
+    to: organization/active_meetings
+    restriction_mode: A
+    default: 1
```
FinnStutzenstein, line 464: invalid attribute

docker/initial-data.json
```diff
@@ -14,6 +14,8 @@
             "committee_ids": [
                 1
             ],
+            "active_meetings": [1],
+            "limit_of_meetings": 3,
```
FinnStutzenstein, line 18: Please set "no limit" for the initial data.

FinnStutzenstein (changes requested):

docs/models.yml
r-peschke: > One question here: Is there a way to disable the limiting (Would be good to have!)? Whats about the values 0, -1, ... here. I think you should set a minimum.

Values below 1 disable the limit, see last point in [impact on backend service](https://github.com/OpenSlides/OpenSlides/wiki/MeetingArchived#impact-on-backend-service)

docs/models.yml
r-peschke: removed the default value, must be set in software to organization id

docker/initial-data.json
r-peschke: Done, set to 0 (=no_limit

FinnStutzenstein:
One thing:

we have the convention that all id fields must end with `_id` or `_ids`. This is expected, e.g. in the client. The client *can* adopt other naming schemes, but having this naming scheme allows at a quick glance to identify relations vs. non-relations.

I think about adding it to the modelsvalidator now...

docs/models.yml
FinnStutzenstein: Please set a minimum of -1

FinnStutzenstein (changes requested):
Please rebase, so #6206 is included

I also wrote a comment about the naming of the relations.

docker/initial-data.json
```diff
@@ -1,667 +1,628 @@
 {
```
FinnStutzenstein, line 1: Please undo all formatting changes.

FinnStutzenstein: Also in the example-data.json

docs/models.yml
```diff
@@ -92,16 +92,24 @@ organization:
...
-
   # Configuration (only for the server owner)
   enable_electronic_voting:
     type: boolean
     restriction_mode: B
+  limit_of_meetings:
+    type: number
+    description: Maximum of active meetings for the whole organization. 0 means no limitation at all
+    restriction_mode: B
+    default: 0
```
FinnStutzenstein, line 103: please add `minimum: 0`

FinnStutzenstein (approved):
One small remark. Before we merge this, we have to decide if we want an initial migration or not (https://github.com/OpenSlides/openslides-backend/pull/901)

docs/models.yml
r-peschke: Done

FinnStutzenstein:
Fith fixint the initial data to MI 1, you should remove the check from the CI and add another one:
- get the md5sum of the current initial data and assert in the CI, that the checksum does not change -> detect (accidental) modifications to this file

.github/workflows/md5_check_initial-data.yml
```diff
@@ -0,0 +1,24 @@
+---
```
FinnStutzenstein, line 1: Please add a comment that this is only for protecting against unwanted/unintended edits. But if one is sure to change this file it is allowed!

FinnStutzenstein (approved):
CI looks good, one comment

.github/workflows/md5_check_initial-data.yml
```diff
@@ -0,0 +1,24 @@
...
+name: Prevent manual changes of initial-data.json without md5 hash
+on:
+  push:
+    paths:
+      - "docker/initial-data.json"
+  pull_request:
+    paths:
+      - "docker/initial-data.json"
+env:
+  PYTHON_VERSION: 3.9.6
```
jsangmeister, line 11: Can be removed

jsangmeister (commented):
Rest looks good. Is the update of the backend intended?

.github/workflows/md5_check_initial-data.yml
r-peschke: done

.github/workflows/md5_check_initial-data.yml
r-peschke: done

jsangmeister (approved):

FinnStutzenstein:
The linked backend version has a broken `cleanup-standalone.sh`. Please use the current commit (6218992fe4512b8d00242131de4ce952669ab11a)

r-peschke:
> Rest looks good. Is the update of the backend intended?

yes:)

FinnStutzenstein (approved):

PR #6218, opened 2021-09-01
Integration tests using cypress

tsiegleauq:
Test client integration
Test auth integration
Test backend integration
Test autoupdate integration

Add manual cypress tests
Add dockered cypress tests

Add Readme
Add test execution to makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,11 @@
-run-system-tests:
-	echo "TODO: write complete system tests"
+run-integration-tests:
+	@echo "Start OpenSlides Dev"
+	make run-dev ARGS="-d"
+	@echo "Start integration tests"
+	make cypress-docker
+	docker-compose -f integration/docker-compose.yml up
+	@echo "Stop OpenSlides Dev"
+	make stop-dev
 
 run-service-tests:
 	git submodule foreach 'make run-tests'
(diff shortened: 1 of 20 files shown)


Makefile
```diff
@@ -1,5 +1,12 @@
 run-system-tests:
-	echo "TODO: write complete system tests"
+	@echo "Start OpenSlides Dev"
+	docker-compose -f docker/docker-compose.dev.yml up -d
+	@echo "Wait  until the system is actually up"
+	sleep 120
```
tsiegleauq, line 5: @gsiv maybe you have an idea for that.
What I would need to run the system tests is to "ensure" here that `docker-compose -f docker/docker-compose.dev.yml up -d` was started in a way the the client and the proxy is actually "up and running".

In this case, up and running would imply that the system is actually reachable. This is not necessarily the case when `docker-compose -f docker/docker-compose.dev.yml up -d` finished the first time. Docker still needs some time untill the proxy (caddy?) Actully delivers the client.
Doing this immediately after docker compose finished would return "bad gateway" upon requesting asking `localhost:8000`

integration/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+HOST="https://host.docker.internal:8000"
+
+echo "wait until OpenSlides is up"
+until [[ $(curl -k -s -o /dev/null -w %{http_code} $HOST) -eq 200 ]];
+do
+    sleep 5
+done
+echo ready
+
+exec npx cypress run
```
GabrielInTheWorld, line 22: This can be put into the `Dockerfile`-file, can't it? But it is okay for me, if it remains here!

GabrielInTheWorld (approved):
Looks good! I am happy, that we have integration tests for our project!
I wrote one comment, but it is beside the point.

integration/entrypoint.sh
tsiegleauq: from the dockerfile:
```
COPY entrypoint.sh /usr/local/bin/entrypoint
```

PR #6225, opened 2021-09-15
Remove permission service

ostcar:
The permission-service is (hopefully) not used anymore. At least the repo was archived some time ago: https://github.com/openslides/openslides-permission-service


tsiegleauq (approved):
also changes the remote from autoupdate from master to main, which fixes `make services-to-master`

PR #6236, opened 2021-09-28
Fixed typo and changed some format.

normanjaeckel:


normanjaeckel:
The format changes are really optional but this is the output of some auto formating tools. ;-)

PR #6242, opened 2021-10-04
Add theme collection to models.

reiterl:
Add theme collection with relations to organization.
Update initial-data.json and example-data.json


docker/initial-data.json
```diff
@@ -19,7 +18,9 @@
                 1
             ],
             "resource_ids": [],
-            "organization_tag_ids": []
+            "organization_tag_ids": [],
+            "default_theme_id": null,
+	    "theme_ids": []
```
jsangmeister, line 23: fix indentation

docker/initial-data.json
```diff
@@ -68,6 +69,7 @@
     },
     "resource": {},
     "organization_tag": {},
+    "theme": {},
```
jsangmeister, line 72: Please create a default theme here and in the example data so `default_theme_id` can be required - it is not valid to have no theme

docs/example-data.json
```diff
@@ -331,6 +332,7 @@
             "organization_id": 1
         }
     },
+    "theme": {},
```
jsangmeister, line 335: see initial data comment

docs/models.yml
```diff
@@ -119,6 +116,14 @@ organization:
     type: relation-list
     restriction_mode: B
     to: organization_tag/organization_id
+  default_theme_id:
```
jsangmeister, line 119: should be required

jsangmeister: and if only one theme is ever in use, `default` is maybe the wrong prefix, since this suggests that others can be selected, too. Maybe `used_theme_id` or something like that (but I'm not really happy with that, either). Do you (or @GabrielInTheWorld) have a suggestion?

docs/models.yml
```diff
@@ -393,6 +397,153 @@ organization_tag:
...
+    restriction_mode: A
+    type: color
+  warn_a700:
+    restriction_mode: A
+    type: color
+  default_theme_for_organization_id:
+    restriction_mode: A
+    to: organization/default_theme_id
+    type: relation
+  organization_id:
```
jsangmeister, line 542: should also be required

.github/workflows/md5_check_initial-data.yml
```diff
@@ -21,4 +21,4 @@ jobs:
         run: md5sum docker/initial-data.json
 
       - name: Validate and protect initial-data.json
-        run: echo "f6274a3f6f638f55a4c66f963d3abbca  docker/initial-data.json" | md5sum -c -
+        run: echo "99fc335aa19a73085432c0e45d417856  docker/initial-data.json" | md5sum -c -
```
jsangmeister, line 24: Not for this PR, but: If we change the initial data with every change to the models, this job does not make sense. It was intended to prevent accidental changes to the file, but if the file gets changed too often, this is just useless work.

jsangmeister (changes requested):

docker/initial-data.json
reiterl: Done.

docker/initial-data.json
reiterl: Add a theme to the inital data and the example data.

docs/models.yml
reiterl: Rename it into `theme_id` and the reverse relation into `theme_for_organization_id`

PR #6244, opened 2021-10-06
Add motion state weight

jsangmeister:


reiterl (approved):
I'm not sure of withdrawed vs withdrawn and referred vs refered.
I think it is okay, else.

jsangmeister:
https://dict.leo.org/german-english/withdrawn
https://dict.leo.org/german-english/referred

PR #6246, opened 2021-10-07
Set required attribute on relation and generic-relation fields

ostcar:
This PR sets the `required` flag on all relation and generic-relation fields.

Some of them are `required=false`, this is the default and therefore not necessary. But I still would like to explicit add the attribute, so it is clearer that the relation is not requied and the "null" case has to be handled.

There are a lot of field where I added `required=true`. Please check if my assumption was correct.

I need this PR for the autoupdate-service. There are bugs, where I did not check the `null`-case (https://github.com/OpenSlides/openslides-autoupdate-service/issues/327 https://github.com/OpenSlides/openslides-autoupdate-service/issues/323). With this change, I want to use the `required` field in my auto generated functions to enforce, that the `null` case is checked or a nicer error message is returned.

### DIFF
--- a/docs/models.yml
+++ b/docs/models.yml
@@ -57,7 +57,8 @@
 #     - The property `read_only` describes a field that can not be changed by an action.
 #     - The property `default` describes the default value that is used for new objects.
 #     - The property `required` describes that this field can not be null or an empty
-#       string. If this field is given it must have some content.
+#       string. If this field is given it must have some content. On relation and generic-relation
+#       fields the value as to be an id of an existing object.
 #     - The property `equal_fields` describes fields that must have the same value in
 #       the instance and the related instance.
 # Restriction Mode:
@@ -328,6 +329,7 @@ user:
     fields:
       type: relation
       to: user/vote_delegations_$_from_ids
+      required: false
     restriction_mode: C
   vote_delegations_$_from_ids:
     type: template
@@ -366,6 +368,7 @@ resource:
     type: relation
     to: organization/resource_ids
     restriction_mode: A
+    required: true
 
 organization_tag:
   id:
@@ -392,6 +395,7 @@ organization_tag:
     type: relation
     to: organization/organization_tag_ids
     restriction_mode: A
+    required: true
 
 committee:
   id:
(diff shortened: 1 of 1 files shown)


docs/models.yml
```diff
@@ -1349,11 +1363,13 @@ group:
     to: meeting/default_group_id
     on_delete: PROTECT
     restriction_mode: A
+    required: false
   admin_group_for_meeting_id:
     type: relation
     to: meeting/admin_group_id
     on_delete: PROTECT
     restriction_mode: A
+    required: false
```
GabrielInTheWorld, line 1372: Does this mean, a group can optionally be the default or admin group of one meeting or does this also mean, that a meeting can optionally have a default or an admin group?
I think, only the first one is right.

docs/models.yml
```diff
@@ -2041,6 +2074,7 @@ motion_block:
     on_delete: CASCADE
     equal_fields: meeting_id
     restriction_mode: B
+    required: true
```
GabrielInTheWorld, line 2077: What exactly means the `required: true`? Does it mean at this point, that every motion_block will always have an agenda_item assigned? If yes, than this is wrong. The `agenda_item_id` of a motion_block can be `null`.

docs/models.yml
```diff
@@ -2559,6 +2605,7 @@ assignment:
...
```
GabrielInTheWorld, line 2608: Same for `motion_block`

docker/initial-data.json
```diff
@@ -521,7 +521,7 @@
             "motion_ids": [],
             "motion_recommendation_ids": [],
             "workflow_id": 1,
-            "first_state_of_workflow_id": null,
+            "first_state_of_workflow_id": 1,
```
GabrielInTheWorld: I don't know, what your intention is, but a workflow can only have one first state. Therefore, only one state should have the property `first_state_of_workflow_id` set to `1`...

ostcar: You are correct. I already reverted that change.

docs/models.yml
GabrielInTheWorld: Not automatically, for each one an agenda_item can be created. But, if an agenda_item was created, this can be deleted afterwards and the related content_object's agenda_item is `null`.

docker/initial-data.json
GabrielInTheWorld: Nevermind if you have just deleted these changes...

PR #6265, opened 2021-10-22
ClientIssue553: Add Testmailserver to openslides-dev-setup

r-peschke:


tsiegleauq (changes requested):
`make run dev` fails with the following error:
> Error response from daemon: failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:76: mounting "/opt/store/Arbeit/OpenSlides/proxy/Caddyfile.dev" to rootfs at "/etc/caddy/Caddyfile" caused: mount through procfd: not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
make: *** [Makefile:18: run-dev] Error 1

This was both reproducible locally and inside the integration tests of cypress

PR #6274, opened 2021-11-10
Add vote service to OS4 setup

### LINKED ISSUE #6273: Integrate vote service to our docker setup
The vote service needs to be integrated to out docker setup so other services can use it by now.

tsiegleauq:
Integrate vote service, makes according changes to configuration

closes #6273


gsiv:
I haven't tested it but it looks functional overall.  Just a few comments:

- The indentation in `docker/docker-compose.dev.yml` is partially off.
- You may want to add the service to `docker/build.sh`, assuming that it's still required.
- There is a Makefile change that appears to be unrelated.  It doesn't do any harm, for course, but just so you know in case that's by accident.

peb-adr:
Looks good,
just one thought I had: Wasn't there a pretty hard fight to get the manage service to not using the DB directly and go through the backend instead?
And now seeing the vote-service getting passed DB credentials seems irritating to me.
Although this should probably be an issue in the vote service and not discussed here if not easily disproven

ostcar:
> Looks good, just one thought I had: Wasn't there a pretty hard fight to get the manage service to not using the DB directly and go through the backend instead? And now seeing the vote-service getting passed DB credentials seems irritating to me. Although this should probably be an issue in the vote service and not discussed here if not easily disproven

This is different. The problem with the manage service was, that it was accessing the raw data of the datastore. But the vote-service uses its own database. It has nothing to do with the datastore. It is not a problem to use the same postgres server and the same postgres database, but this is not required.

ostcar:
To integrate the vote service, the autoupdate service needs new environment varialbes: https://github.com/OpenSlides/openslides-autoupdate-service/pull/363/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R166-R168

docker/docker-compose.dev.yml
```diff
@@ -151,3 +151,27 @@ services:
...
+            - MESSAGE_BUS_HOST=message-bus
+            - VOTE_REDIS_HOST=cache
+            - VOTE_DATABASE_HOST=postgres
+            - VOTE_DATABASE_USER=openslides
+            - VOTE_DATABASE_PASSWORD=openslides
+            - VOTE_DATABASE_NAME=openslides
+            - AUTH=ticket
+            - AUTH_HOST=auth
+    ports:
+      - "9013:9013"
```
ostcar, line 177: Why is it necessary to publish the port of the vote-service?

ostcar:
I also tested this branch. It Starts the vote-service. But when I see it correctly, it does not use it. Are the backend and the client already able to work with the new vote service? If not, are there issues to track there progress?

tsiegleauq:
> I also tested this branch. It Starts the vote-service. But when I see it correctly, it does not use it. Are the backend and the client already able to work with the new vote service? If not, are there issues to track there progress?

That is correct. This merely integrates the vote service. I "think" the current backend main branch already made the according changes, the client side is WIP.
https://github.com/OpenSlides/openslides-client/pull/640

docker/docker-compose.dev.yml
tsiegleauq: I don't know if it really is. I wanted to see if I can `curl` requests directly to the vote service and took the other services in the compose file as template.

tsiegleauq:
> I haven't tested it but it looks functional overall. Just a few comments:
> 
> * The indentation in `docker/docker-compose.dev.yml` is partially off.

corrected

> * You may want to add the service to `docker/build.sh`, assuming that it's still required.

added

> * There is a Makefile change that appears to be unrelated.  It doesn't do any harm, for course, but just so you know in case that's by accident.

Thats intended. That helps to set all origin-remotes to upstream, so the `make services-to-master` actually works. If I knew how to make propper conditions in makefiles that would be more powerfull already

docker/docker-compose.dev.yml
```diff
@@ -151,3 +151,27 @@ services:
...
+            - DATASTORE_READER_HOST=datastore-reader
+            - MESSAGING=redis
+            - MESSAGE_BUS_HOST=message-bus
+            - VOTE_REDIS_HOST=cache
+            - VOTE_DATABASE_HOST=postgres
+            - VOTE_DATABASE_USER=openslides
+            - VOTE_DATABASE_PASSWORD=openslides
+            - VOTE_DATABASE_NAME=openslides
+            - AUTH=ticket
+            - AUTH_HOST=auth
```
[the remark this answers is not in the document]
tsiegleauq: I suppose other changes in the vote-service are necessary here. For example, the env. var `AUTH=ticket` was only used in by the vote service (not currently present in the `services.env`). Should probably renamed to `VOTE_AUTH`.

This cannot be done in this scope

PR #6276, opened 2021-11-13
Makes the motion block detail and motion list more user friendly

GabrielInTheWorld:


emanuelschuetze (approved):
Added some style changes after testing.

PR #6290, opened 2021-11-22
add vote-service to build-images workflow

peb-adr:
An oversight during integrating the vote-service.
It of course needs to be built by github actions as well.


tsiegleauq:
ahh. ok.

PR #6293, opened 2021-11-24
Run cypress workflow without docker

tsiegleauq:
Runs the cypress workflow directly without docker


jsangmeister (approved):
not tested, looks good

PR #6309, opened 2021-12-09
add local production setup

peb-adr:
As @GabrielInTheWorld suggested, this is an easy way of using the manage-tool to build the local repository into a production-like setup.

closes https://github.com/OpenSlides/openslides-manage-service/issues/123

### DIFF
--- a/localprod/README.md
+++ b/localprod/README.md
@@ -1,0 +1,35 @@
+# Local production setup
+
+This folder allows you to use the
+[manage-tool](https://github.com/OpenSlides/openslides-manage-service) to launch
+the local repository state as a production setup.
+The `config.yml` instructs the manage-tool to include build tags for all service
+images.
+
+## Usage
+
+Run
+
+    ./setup.sh
+
+It will download the latest manage-tool (the `openslides`-executable) - if not
+present already - and setup this directory like a production environment using\
+the `config.yml`.
+
+Now run
+
+    docker-compose up --build
+
+After all services are up initial data needs to be set to be able to login.
+In a different terminal run
+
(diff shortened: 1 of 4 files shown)


GabrielInTheWorld (approved):
Looks good!

localprod/setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+[[ -f ./openslides ]] ||
+  wget https://github.com/OpenSlides/openslides-manage-service/releases/download/latest/openslides
+
+[[ -x ./openslides ]] ||
+  chmod +x ./openslides
+
+./openslides setup .
```
normanjaeckel, line 9: This can be a oneliner here `./openslides setup --config config.yml .`

localprod/README.md
```diff
@@ -0,0 +1,27 @@
...
+## Usage
+
+Run
+
+    ./setup.sh
+
+It will download the latest manage-tool (the `openslides`-executable) - if not
+present already - and setup this directory like a production environment using\
+the `config.yml`.
+
```
normanjaeckel, line 18: Either in setup.sh or as an extra step the user mus run `./openslides initial-data`

normanjaeckel (changes requested):
Nice work. Thanks.

localprod/setup.sh
peb-adr: For first time setup yes.
But I want to be able to re-run `setup.sh` after e.g. changing the config file. So for that to work I think this should stay like this

normanjaeckel: Ok for me

PR #6310, opened 2021-12-09
Improves the 'services-to-master' function

GabrielInTheWorld:


services-to-master.sh
```diff
@@ -19,13 +19,12 @@ function get_upstream_branch {
...
 function get_upstream_name {
-   { 
-      # try
-      git ls-remote --exit-code upstream 2>/dev/null && echo "upstream"
-   } || {
-      # catch
+   git ls-remote --exit-code upstream &>/dev/null
+   if test $? = 0; then
+      echo "upstream"
+   else
```
peb-adr, line 25:
```suggestion
function get_upstream_name {
   git ls-remote --exit-code upstream &>/dev/null || {
      echo "origin"
      return
   }
   echo "upstream"
}
```

PR #6315, opened 2021-12-22
proxy: add ALLOWED_HOSTS env var

peb-adr:
will setup caddy to check if the Host-header matches one of the space
delimited fields in ALLOWED_HOSTS. Or do no such thing if ALLOWED_HOSTS
is not set.

### DIFF
--- a/proxy/caddy_base.json
+++ b/proxy/caddy_base.json
@@ -27,6 +27,27 @@
             {
               "handle": [
                 {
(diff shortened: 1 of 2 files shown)


gsiv (approved):
Looks very good to me.

peb-adr:
I'll mege this, since there will be an update soon and this should be included.

PR #6322, opened 2022-01-07
Opentelemetry

ostcar:

PR #6354, opened 2022-02-09
Opentelemetry

peb-adr:
This is the cleaner and up-to-date version of #6322 

Also I adjusted the compose-files and added suitable make targets to be able to view traces for the dev-stack

I also made PRs in the backend (https://github.com/OpenSlides/openslides-backend/pull/1201) and datastore (https://github.com/OpenSlides/openslides-datastore-service/pull/179).
More services can gradually be instrumented for sending spans to the collector.
They should however all only generate telemetry data if `OPENTELEMETRY_ENABLED=1`. In the other PRs I did this by encapsulation all otel-related code (such as initialization and span-creation) into a python module.
This setup has also been tested for compatibility with our production setup.


docker/otel-collector-config.yml
```diff
@@ -0,0 +1,21 @@
...
+
+processors:
+  batch:
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [ jaeger]
```
jsangmeister, line 21:
```suggestion
      exporters: [jaeger]
```

docker/dc.otel.dev.yml
```diff
@@ -0,0 +1,13 @@
+version: "3"
+services:
+
```
jsangmeister, line 3: Analogous to the backend, I would think that `OPENTELEMETRY_ENABLED` needs to be set to true for all services here

jsangmeister (changes requested):

ostcar:
OpenTelemetry uses some [default environment variables](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sampling). I did not find a default variable to "disable" or "enable" OpenTelemetry. So I am not sure, if there is a replacement for `OPENTELEMETRY_ENABLED`. Maybe it is possible to use `OTEL_TRACES_SAMPLER` instead and disable any sampling with `always_off`? In this case, this would also work with other services like caddy.

I am not sure if we should use `jaeger`. To store the spans on disk, it uses elastic search or casantra. An alternative seems to be [grafana tempo](https://grafana.com/oss/tempo/). I have not tested it, but it could be worth a look. This does not have to be done in this PR. We have to instrument the services in any case.

ostcar (approved):

peb-adr:
merged via e343bb7e6be7055573e74bd0617eb2cfb776fa77
@r-peschke FYI

PR #6356, opened 2022-02-10
without this the installation won't function as described (missing m4 file etc)

bug0001:
as requested, without this the installation won't function as described (missing m4 file etc)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@ instructions <DEVELOPMENT.rst>`_.
 Get OpenSlides
 --------------
 
-First, you have to clone this repository, then move to the latest stable branch::
+First, you have to clone this repository, then checkout to the latest stable branch::
 
     git clone https://github.com/OpenSlides/OpenSlides.git --recurse-submodules
     cd OpenSlides/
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -21,10 +21,11 @@ instructions <DEVELOPMENT.rst>`_.
 Get OpenSlides
 --------------
 
-First, you have to clone this repository::
+First, you have to clone this repository, then move to the latest stable branch::
```
emanuelschuetze, line 24: better: "... then _checkout_ to the latest stable branch:"

emanuelschuetze:
Please use a better and shorter commit message, e.g. "Updated install instructions in Readme"

README.rst
```diff
@@ -21,10 +21,11 @@ instructions <DEVELOPMENT.rst>`_.
 Get OpenSlides
 --------------
 
-First, you have to clone this repository::
+First, you have to clone this repository, then checkout to the latest stable branch::
 
     git clone https://github.com/OpenSlides/OpenSlides.git --recurse-submodules
     cd OpenSlides/
+    git checkout remotes/origin/stable/3.4.x
```
jsangmeister, line 28: Nitpick, but I think
```suggestion
    git checkout stable/3.4.x
```
should be enough, shouldn't it?

bug0001: That's one thing, but the bigger issue at the moment is that some tests fail (python installation, dependencies) even with minor modifications.

jsangmeister:
Closing in favor of https://github.com/OpenSlides/OpenSlides/pull/6359

PR #6359, opened 2022-02-15
Dependency resolve

bug0001:
solving dependency dev. req.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@ instructions <DEVELOPMENT.rst>`_.
 Get OpenSlides
 --------------
 
-First, you have to clone this repository, then move to the latest stable branch::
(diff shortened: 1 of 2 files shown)


bug0001:
tests sill not working after restricting black req. to less than 22

jsangmeister:
Annoyingly, the version mismatch only occured on python 3.8, not on python 3.10 because of the specific requirements of black. Something to keep in mind in case something similar occurs again.

Should however be fixed now.

jsangmeister:
Afternote: `black`'s first stable version 22.1.0 requires a higher version of `typing-extensions` than 3.8, which is the max in our current requirements. Since the former is only a dev dependency and the latter a productive one, we chose to just fix `black` to an older (beta) version instead of updating `typing-extensions`.

PR #6372, opened 2022-02-23
simplify services-to-main.sh

peb-adr:
- remove -B from checkout, didn't work as expected
- all submodules are now using main, fallback to master no longer needed

### DIFF
--- a/services-to-main.sh
+++ b/services-to-main.sh
@@ -2,42 +2,64 @@
 
 set -e
 
-function get_upstream_branch {
-   local BRANCH_NAME=main
-   local exists=`git show-ref refs/heads/$BRANCH_NAME`
-   if [[ -z $exists ]]; then
-      BRANCH_NAME=main
-   fi;
-   echo "$BRANCH_NAME"
+ME=$(basename "$0")
+
+BRANCH_NAME=main
+REMOTE_NAME=
+OPT_PULL=
+
+usage() {
+  echo "USAGE:"
+  echo "  $ME [ --pull | -p ]"
+  echo
+  echo "By default $ME will fetch the latest upstream changes for every"
+  echo "service/submodule and directly checkout the upstream's $BRANCH_NAME branch."
+  echo "This will leave them in detached HEAD state."
+  echo "Use --pull to instead forward the local $BRANCH_NAME branch."
 }
 
-function get_upstream_name {
-   git ls-remote --exit-code upstream &>/dev/null || {
-      echo "origin"
-      return
-   }
-   echo "upstream"
+set_remote() {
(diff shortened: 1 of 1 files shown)


services-to-main.sh
```diff
@@ -2,42 +2,34 @@
...
-   git checkout -B $BRANCH_NAME;
-   git pull $REMOTE_NAME $BRANCH_NAME;
+  BRANCH_NAME=main
 }
 
-export -f pull_latest_commit
-export -f get_upstream_branch
-export -f get_upstream_name
+pull_latest_commit() {
+  echo "git checkout $BRANCH_NAME && git pull $REMOTE_NAME $BRANCH_NAME ..."
```
emanuelschuetze, line 17: We should go back to `git fetch $REMOTE_NAME` and `git checkout $REMOTE_NAME/$BRANCH_NAME`.
Otherwise you get an unclean main branch which is updated by pull but contains something custom commits.

emanuelschuetze (changes requested):

services-to-main.sh
```diff
@@ -2,42 +2,34 @@
...
-
-function get_upstream_name {
-   git ls-remote --exit-code upstream &>/dev/null || {
-      echo "origin"
-      return
-   }
-   echo "upstream"
-}
+REMOTE_NAME=
+BRANCH_NAME=
```
jsangmeister, line 6: I would set the branch name directly here and remove the line in the function

jsangmeister (changes requested):

services-to-main.sh
GabrielInTheWorld: Which custom commits can this be?
I like the current state more, because then I can use it, too, to get the most recent state in every sub-repo.

GabrielInTheWorld (approved):
I agree to @jsangmeister's comment, but in general I like the changes.

services-to-main.sh
GabrielInTheWorld: Maybe we can make use of command line flags to switch between `fetch & checkout` and `checkout & pull`.

peb-adr:
@emanuelschuetze @GabrielInTheWorld @jsangmeister 
What do you think about this?
Will revert to the old `fetch && checkout` behavior by default.
Adds the `--pull | -p` flag to instead do `checkout && pull`

GabrielInTheWorld:
> @emanuelschuetze @GabrielInTheWorld @jsangmeister What do you think about this? Will revert to the old `fetch && checkout` behavior by default. Adds the `--pull | -p` flag to instead do `checkout && pull`

That will be nice!

emanuelschuetze (approved):

PR #6376, opened 2022-02-28
Multiple votes accumulated on one person

MJJojo97:
For several years now, we are using OpenSlides for two conferences per year within our organization. Within these conferences we have apart from lots of "normal" elections some more special elections. In these special ones everyone has a certain number of votes which can also be accumulated onto one person with the restriction that there musn't be more than three votes combined onto one person.
In the last years, we made this special election somehow outside of OpenSlides, which introduced much more work, more overhead in the evaluation and so on. For this reason, I now implemented it:

When creating a new election-poll with the option "Yes" you have a third option "maximum votes per person". If this option equals 1 (default), nothing of the original voting behavior changes. If it is >1, the "thumb-up" button in the election is replaced by an input field constrained to the corresponding value limits.

I would be really happy, if this feature is added to the official OpenSlides in order to allow these types of elections in the future. Of course, feedback is also very welcome.

[Note: If it is merged, it would be necessary to remove the "replace"-statements in the go.mod files of vote-service and the icc-service and adapt the corresponding versions.]


jsangmeister:
(Note: for the repos where you just patched the commit, e.g. icc-service, you of course do not have to create a PR)

MJJojo97:
Hi! I rebased everything and added pull requests for all relevant repos, i.e.,
- autoupdate-service,
- backend,
- client, and
- vote-service,
and look forward to hearing from you!

ostcar:
Thank you very mutch for your work. I think you idea is good and we should merge it.

OpenSlides is documented in the wiki. Before we change the code, the wiki has to be updated. Do you have the required permissions to do this? https://github.com/OpenSlides/OpenSlides/wiki/poll.vote

The next step is to update the backend with the models.yml. https://github.com/OpenSlides/openslides-backend/pull/1251/files#diff-f507c083dbbc3decaabe4ccefeb830ddb9419f66dd0729f1e15abdf4ef9f53d7R2686-R2689

I would prefere another name for the field. This would be the first time, that we use the term "person". I think `max_votes_per_option` sounds better.

The go code (autoupdate and vote service) looks good. It would be nice, if you could some more tests to the vote-service. For example a test, if someone sets to many votes on one option.

MJJojo97:
Apart from the documentation, where I still don't have permission for writing, I changed/added the code according to your issues. Do the tests correspond to your expectations?

ostcar:
From my perspective, everything looks good. @emanuelschuetze wants to have a look, how you solved the ui in the client. He will probably try to contact you directly about it.

MJJojo97:
Yes, we talked yesterday, and afterwards, I added/changed several aspects in the client. They just need to be reviewed again.

ostcar:
I changed the wiki. I hope I found all relevant places.

I would propose that we merge all the server side code. It is not a problem if the work on the client is not done yet. The new server code is compatible with the old client.

The next step is to update the models.yml in the backend. Is it possible to merge https://github.com/OpenSlides/openslides-backend/pull/1251 when the vote service is merged a bit later? If not, we need a PR that only adds the new field to the models.yml but without the python-code, that requires the changes to the vote service.

emanuelschuetze:
All services are merged. We will update all services in this main repo in the next days. This PR can be closed.

PR #6401, opened 2022-04-12
Update Readme & fix dependencies

jsangmeister:
Shorten the time of cloning by directly cloning the correct branch and not fetching the other branches. Also gets rid of the warnings regarding non-empty directories when switching the branch and of the need to execute `git submodule update --init` again afterwards.


jsangmeister:
Set all dependencies to a fixed version since we do not develop on OS3 actively and if only some versions are fixed, simple updates like this will fail because of some conflict in the requirements.

jsangmeister:
Also, as observed in #6403, breaking changes in dependencies which do not have a max version will break our code and make it impossible to set up new instances.

emanuelschuetze (approved):

PR #6409, opened 2022-04-24
Make the style of the project titles customizable

m-schieder:
* Fix candidate display at the polls

### DIFF
--- a/client/src/assets/styles/projector.scss
+++ b/client/src/assets/styles/projector.scss
@@ -11,13 +11,31 @@
         line-height: 1.1em;
     }
 
+    .slidetitle > h1,
+    .slidetitle > h1 *,
+    h1.projector_h1,
+    h1.projector_h1 * {
+        font-size: 1.8em;
+        font-weight: 500;
+        line-height: 1.1em;
+        font-family: $font-projectorH1;
+    }
+
     h2 {
         line-height: 40px;
         font-weight: normal;
     }
 
+    .slidetitle > h2,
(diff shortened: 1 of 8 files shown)


emanuelschuetze (approved):

client/src/assets/styles/projector.scss
```diff
@@ -11,13 +11,26 @@
         line-height: 1.1em;
     }
 
+    .slidetitle > h1,
+    .slidetitle > h1 *,
+    h1.projector_h1,
+    h1.projector_h1 * {
+        font-family: $font-projectorH1;
+    }
```
GabrielInTheWorld, line 19:
```suggestion
    .slidetitle > h1,
    h1.projector_h1 {
        &, * {
            font-family: $font-projectorH1;
        }
    }
```

I think, this is equivalent. Maybe just a detail, but in my opinion it is more readable.

client/src/assets/styles/projector.scss
```diff
@@ -11,13 +11,26 @@
...
         line-height: 40px;
         font-weight: normal;
     }
 
+    .slidetitle > h2,
+    .slidetitle > h2 *,
+    h2.projector_h2,
+    h2.projector_h2 *  {
+        font-family: $font-projectorH2;
+    }
```
GabrielInTheWorld, line 31: Same as above.

GabrielInTheWorld (approved):
Looks good!
Some hints, which can be considered but do not have to.

PR #6413, opened 2022-04-28
Remove topic/option_ids from export

jsangmeister:


r-peschke (approved):
Okay, but could also remove tag_ids and projection_ids: empty and the fields are not required

PR #6450, opened 2022-06-22
updated DS

peb-adr:
for fixing tests.
Failed due to incompatible version of protobuf py-package.


emanuelschuetze:
already updated in #6451

PR #6457, opened 2022-07-07
Fix OS4 export when no motion, but assignment options are present

### LINKED ISSUE #6456: OS3: export to OS4 - Error: Tried to overwrite option/1
steps to reproduce:
1. create 2 assignments ("elections")
2. create a voting in both (i did non-nominal)
3. do exportToOs4 in settings
4. backend answers {"detail":"Tried to overwrite option/1"}

@jsangmeister this is probably the easiest for you to fix.

jsangmeister:
fixes #6456

Interestingly, this error only occurs when there are no motion polls yet, but some assignment polls, which might explain why it did not come to our attention until now.

### DIFF
--- a/server/openslides/core/export.py
+++ b/server/openslides/core/export.py
@@ -290,8 +290,8 @@ class OS4Exporter:
 
         self.option_id_counter = (
             max_or_zero([x["id"] for x in self.get_collection("motions/motion-option")])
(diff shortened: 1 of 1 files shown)


peb-adr (approved):
Problem solved, thanks ^^
can be merged.

PR #6465, opened 2022-07-21
add script to import sql into dev setup

peb-adr:


jsangmeister:
A similar script is also present in the datastore: https://github.com/OpenSlides/openslides-datastore-service/tree/main/cli/import-events.sh

Maybe you can use that or merge them together? I would like the script itself to stay in the datastore, since then it can also be used if the full dev setup is not started, but e.g. only the backend and datastore.

peb-adr:
@jsangmeister Thanks for the input. I think this should do what you asked for.
However I also needed to make some changes in the datastore (https://github.com/OpenSlides/openslides-datastore-service/pull/208).
Could you please review both these PRs and verify that the script in the DS continues to behave as expected.

This should then be merged together with said PR in the DS.

dev-commands/reset-db-import-sql.sh
```diff
@@ -0,0 +1,18 @@
...
+DC="$(dirname $0)/../dc-dev.sh"
+FILE="$1"
+
+[ -n "$FILE" -a -r "$FILE" ] || {
+  echo "ERROR: cannot read file $FILE"
+  exit 1
+}
+
+$DC rm -svf postgres
+$DC up -d postgres
```
jsangmeister, line 14: you can also call the `truncate_db` route of the writer, which is only enabled in dev mode. See https://github.com/OpenSlides/openslides-datastore-service/blob/main/docs/development.md#development-exclusive-tools

jsangmeister (changes requested):

dev-commands/reset-db-import-sql.sh
peb-adr: I tried this and it generally seems to work.
I think I still prefer my way of completely resetting postgres here.
This script is meant to import complete DB-dumps and only using the truncate_db route could result in many errors.
Though there were many "relation already exists" messages from CREATE statements,
I also saw some constraint violations, which don't happen when postgres is completely fresh. Also it's only a time loss of a few seconds.

jsangmeister: ok

Issue #6473, opened 2022-07-19
OS3 Meeting Import: amendment import is broken


MSoeb:
OS3 amendments are imported incorrectly into OS4. The effect is that changes are taken over incorrectly or amendments are deleted completely.

Reproduction: 

Examples in https://testintern.openslides.com/77/motions
Import file could be found in the meeting under FILES -> OS3 Import - ÄA Fehlerhaft.json

1. Download the OS3 export file
2. Go back ot dashboard, navigate to committees and choose one.
3. Use the import function in the three-dot-menu and import the meeting file via "Import meeting"
4. open the imported meeting and compare data in motions with the ones in OS3 demo. (Example motion F002)

What should happen: All motions and amendments with all associated data should be imported correctly . 

Example of broken amendment
Left side OS3 - original content
Right side OS4 - Import
![Screenshot_20220719_085657](https://user-images.githubusercontent.com/105491030/179686104-e6890f82-74d4-4499-b8b2-5d2a64719604.png)

jsangmeister:
done with https://github.com/OpenSlides/OpenSlides/pull/6474 (why did github not close this automatically?)

PR #6474, opened 2022-08-08
Fix amendment paragraph numbering

jsangmeister:
fixes https://github.com/OpenSlides/OpenSlides/issues/6473

@MSoeb the export seems to just have been wrong. To test, you have to export the instance again with this fix and then import it in OS4. It should work now.


MSoeb:
@m-schieder Need test instance.

emanuelschuetze (approved):
Tested.

PR #6483, opened 2022-09-07
Updated services

emanuelschuetze:


emanuelschuetze:
@peb-adr Note: `UPDATE positions SET information=NULL`  in postgres required in new version

PR #6486, opened 2022-09-15
Remove integration tests

jsangmeister:
Since they were transferred to the [client](https://github.com/OpenSlides/openslides-client/tree/main/client/tests/integration), these files are deprecated and no longer kept up-to-date, so they should be removed. It does not help us at all to keep the integration tests in two places.


luisa-beerboom (commented):
I generally have no objections to this, but it seems like some of these tests do not exist in the client.
Of course I don't know if these tests are even functional, but if they are I don't think it would be bad if they were moved to the client.

tsiegleauq:
Besides committee create, they used to be functional alright, and probably still are. Maybe the necessity has lowered since (e)manu(e/a)l testing has grown.

Synergy is better in the client, for you are able to alter html attributes by cypress's needs. As long as you remember to start the whole stack it matters little were the tests are located.

jsangmeister:
missing tests transferred to the client in https://github.com/OpenSlides/openslides-client/pull/1621. Workflow adjusted to use the tests in the client repo to still have the "full" tests available in this repo.

PR #6509, opened 2022-10-28
Update Readme to fix localhost httpS note

CuzImBisonratte:
Fix localhost httpS note


emanuelschuetze (changes requested):
We use https only because it is required by HTTP/2. So we should remove the part "(or https, if configured)". Can you do it?
Or is there an other reason why you want to use http?

PR #6518, opened 2022-12-02
Set meeting specific number instead of default number.

rrenkert:


emanuelschuetze:
closed in favour of #6519

PR #6519, opened 2022-12-02
Export vote weight instead of default vote weight

rrenkert:


server/openslides/core/export.py
```diff
@@ -1039,9 +1039,9 @@ def migrate_users(self):
 
             new["is_physical_person"] = not old["is_committee"]
             new["password"] = ""
-            new["default_number"] = old["number"]
+            new["default_number"] = ""
             new["default_structure_level"] = ""
-            new["default_vote_weight"] = old["vote_weight"]
+            new["default_vote_weight"] = ""
```
r-peschke, line 1044: Default value in models.yml/py: `new["default_vote_weight"] = "1.000000"`

r-peschke (changes requested):

r-peschke (approved):

emanuelschuetze (approved):
Tested

PR #6535, opened 2023-01-14
Add script to automatically set labels and milestones

jsangmeister:


.github/meta/update_labels_and_milestones.py
```diff
@@ -0,0 +1,196 @@
...
+        },
+        {
+            "name": "waiting",
+            "color": "cccccc",
+            "description": "Waiting for some other PR/feature; more details in comments"
+        },
+    ],
+    "OpenSlides": [
+        {
+            "name": "help wanted",
```
normanjaeckel, line 90: Proposal: Move `help wanted` to general.

.github/meta/update_labels_and_milestones.py
```diff
@@ -0,0 +1,196 @@
...
+        {
+            "name": "migration",
+            "color": "a046d0",
+        },
+        {
+            "name": "performance",
+            "color": "006b75",
+        },
+        {
+            "name": "tests",
```
normanjaeckel, line 120: Proposal: Move `tests` to general.

.github/meta/update_labels_and_milestones.py
jsangmeister: If someone needs help on an issue, he or she should communicate this with the team using internal channels and a solution will be found. I don't see the value of such a label except in the main repo, where users will ask for help without anyone specific to ask and can label their issue accordingly.

.github/meta/update_labels_and_milestones.py
jsangmeister: Yes, I thought about this, too, but this would clash with the client-specific label `tests` which has another meaning. @m-schieder @MSoeb would it be okay for you to remove the `tests` label in the client? Or rename it?

.github/meta/update_labels_and_milestones.py
normanjaeckel: Ok

.github/meta/update_labels_and_milestones.py
jsangmeister: Done.

jsangmeister:
Updated file and re-executed it.

normanjaeckel (approved):

PR #6543, opened 2023-01-20
Update INSTALL.md

Magolv:
Add refined instructions for usage of openslides manage tool when SSL encryption is disabled.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -169,6 +169,15 @@ setup configuration YAML file.
 
     enableLocalHTTPS: false
 
+Note, that some commands of the manage tool require the `--no-ssl` flag when SSL encryption is disbaled, e.g:
+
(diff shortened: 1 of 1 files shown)


jsangmeister:
@normanjaeckel maybe the manage tool could automatically evaluate that setting when executing commands like these and remove the need for the user to explicitely provide the flag? This would remove a (seemingly relatively common) source of error.

jsangmeister:
@Magolv thanks for your contribution!

normanjaeckel:
I am not sure how the manage tool could evaluate this. The respective setup configuration YML is not used by the respective commands like initial-data. This setting is intended to be used only for the setup process (create container configuration YML).

normanjaeckel (approved):

PR #6550, opened 2023-02-15
Update jitsi API and config

m-schieder:


emanuelschuetze (approved):
tested with meet.jit.si and webconf-staging

PR #6582, opened 2023-04-24
Exclude `/system/action/internal` from backend matcher in proxy

jsangmeister:
The backend provides the route `/system/action/internal/handle_request` for stack internal access to actions, protected by a password. For additional security, the proxy shouldn't even match this route, since it should only be called from inside the stack.

I did not test this since I wasn't sure how and didn't want to invest too much time. @peb-adr it would be nice if you could confirm that this works as intended.

### DIFF
--- a/proxy/caddy_base.json
+++ b/proxy/caddy_base.json
@@ -72,7 +72,10 @@
               ],
               "match": [
                 {
-                  "path": ["/system/action*"]
+                  "path": ["/system/action*"],
+                  "not": [
+                      { "path": ["/system/action/internal*"] }
+                  ]
                 }
               ]
             },


peb-adr:
I was under the impression that the route used for those internal actions is `/internal`.
I have just tested that running `./openslides create-user '{"first_name":"Albert","last_name":"Einstein"}'`
and the other manage subcommands
sends a POST request to http://backendManage:9002/internal/handle_request.

From where do you get the `/system/action/internal` path? Is that a route actually being served by the backend? If so I'd argue it's obsolete and shouldn't be served.

To address to leading question `/internal/*` is not routed by caddy and if we can agree, that `/system/action/internal` shouldn't be used (or maybe doesn't actually exist in the backend?) I don't think this change is necessary.
But I may also be missing something..

jsangmeister:
Right, I misread the code! Sorry for that.

r-peschke:
Removed myself from reviewer, because don't know nothing about the caddy configuration. If it do, what I think it do, it's okay for me!

jsangmeister:
Already closed, since the route is `/internal/handle_request` instead of `/system/action/internal/handle_request`, like I assumed.

PR #6586, opened 2023-05-08
Add script and workflow for creating staging and stable updates separately

peb-adr:


peb-adr:
I think this finally reached a state where it's working as intended.

TODOs left (ar least):
- revert submodule links peb-adr -> OpenSlides
- Write wiki documentation
- come to consensus about naming (main <> staging vs. stable <> main)
- Prepare github repos
  - create new branches
  - protect branch for stable updates (direct push)

Issue #6593, opened 2023-05-22
Use redis:alpine?


peb-adr:
I would like to use the `redis:alpine` image rather than `redis:latest` (debian based) for production.
In order to keep dev and prod setup similar, would it be a problem for anyone to change the `docker-compose.dev.yml` to use `redis:alpine`?

jsangmeister:
I don't see any problem with it, functionality-wise it should be the same

ostcar:
@normanjaeckel Can you also make the change in the manage-tool?

peb-adr:
https://github.com/OpenSlides/OpenSlides/pull/6602
https://github.com/OpenSlides/openslides-vote-service/pull/204
https://github.com/OpenSlides/openslides-manage-service/pull/223

have been merged. Closing.

PR #6602, opened 2023-06-06
Use redis:alpine in dev setup

peb-adr:
As discussed in #6593 we want to use the `redis:alpine` image instead of the the normal debian based one.

This change has to be done in several places so these PRs are relevant for the same issue:
- https://github.com/OpenSlides/openslides-vote-service/pull/204
- https://github.com/OpenSlides/openslides-manage-service/pull/223


peb-adr:
The related PRs are merged.
Also the `redis:alpine` image was tested for a week now on our staging system with no problems arising.
I'll merge.

PR #6612, opened 2023-06-19
Pin debian based python docker images to bullseye

peb-adr:
debian docker images are currently being migrated to version 12 (bookworm). This is currently leading to errors when building as the bookworm-based images are not stable yet. In order for the stable/3.4.x branch to stay stable (and buildable) we can keep using the debian bullseye images.


jsangmeister (approved):
Sure, looks good

PR #6638, opened 2023-08-25
Manage versioning

peb-adr:
Call a build action in the manage tool when creating stable updates.

See also: https://github.com/OpenSlides/openslides-manage-service/pull/231/files


jsangmeister (approved):
Looks good, ofc not tested

PR #6639, opened 2023-08-30
Outsource proxy to its own submodule

peb-adr:
Tested:
- `make run-dev`
- `./dev/docker/build.sh` (for staging and stable builds)
- `./dev/localprod`


rrenkert (approved):
Looks good so far. Will test tomorrow if you think its necessary.

PR #6649, opened 2023-09-06
Staging update 20230906

peb-adr:


peb-adr:
This is an empty commit to rebuild the staging images from yesterday.
Strangely it seems the translations didn't make it into the images even though they are included in the client commit.
I hope this will rectify that.

peb-adr:
conclusion: it did!

PR #6651, opened 2023-09-08
Update integration test action

bastianjoel:
For more details see https://github.com/OpenSlides/openslides-client/pull/2773


bastianjoel:
Should currently fail because client changes are missing to make this work.

PR #6653, opened 2023-09-14
Enable OpenTelemetry in services via env var

jsangmeister:
`datastore-reader` and `datastore-writer` now log correctly with this change, the backend does not. This should be investigated further.


jsangmeister:
see https://github.com/OpenSlides/openslides-backend/issues/1907

r-peschke (approved):

peb-adr (approved):

PR #6698, opened 2023-11-28
Add co-branch script

peb-adr:
This script has been around and used by developers for quite some months.
It makes sense to include it in the repo.

### DIFF
--- a/dev/scripts/co-branch.sh
+++ b/dev/scripts/co-branch.sh
@@ -1,0 +1,116 @@
+#!/bin/bash
+
+set -e
+
+ME=$(basename "$0")
+
+REMOTE_NAME=${1:-origin}
+BRANCH_NAME=${2:-main}
+OPT_PULL=${3:-0}
+
+usage() {
+  echo "USAGE:"
+  echo "  $ME [REMOTE:origin] [BRANCH:main] [pull]"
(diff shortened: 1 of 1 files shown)


dev/scripts/co-branch.sh
```diff
@@ -0,0 +1,116 @@
...
+    if [[ `git status --porcelain --ignore-submodules --untracked-files=no` ]]; then
+        echo "The repository $mod has changes. Stash them? [y/n]"
+        read decision
+        if [ $decision == "y" ]; then
+            git stash
+        else
+            exit 0
+        fi
+    fi
+    for mod in $(git submodule status | awk '{print $2}'); do
```
jsangmeister, line 58: Small note, but couldn't you use `git submodule foreach` instead? Seems simpler

peb-adr: For singular commands yes.
But when doing loops with multiple command per submodule and - as done here - getting user input, doing this is often easier to type and read IMO.
Also dealing with escaping and variables is more straightforward.

jsangmeister (approved):

peb-adr:
As this script was written by @rrenkert, he won't review here.
So I'll merge now.

PR #6702, opened 2023-12-05
Update translations and script

jsangmeister:
`--sort-output` has been deprecated, so I removed it as well


emanuelschuetze (approved):
I pushed an updated template-en.pot.

PR #6728, opened 2024-02-08
add `make-update hotfix` workflow

peb-adr:


jsangmeister:
The CI fails because the search service is not updated yet (as https://github.com/OpenSlides/OpenSlides/pull/6729 proves). When we do the update with all new services, this can be rebased and merged afterwards.

rrenkert (approved):

PR #6733, opened 2024-02-13
Add compatibility and checks for meta repo

peb-adr:


jsangmeister (approved):
From the bash commands that I understand this looks sensible :D

PR #6757, opened 2024-03-14
Fix typo in de.po

topelrapha:
There is a german typo mistake.

Should be "bearbeiten" not "bearbeitet"


jsangmeister:
Thanks for the hint! We manage our translations in transifex, the `.po` files in this repo are merely the mirrors. I updated the string there and the changed translation will be included in the next update. Therefore, I'm closing this PR.

PR #6770, opened 2024-04-04
Update docs and scripts to new docker compose syntax

jsangmeister:
closes https://github.com/OpenSlides/OpenSlides/issues/6767

PR #6774, opened 2024-04-16
FIX Adjust Makefiles To New Docker Compose Syntax; ADD Danny Reichelt To Authors

reicda:


jsangmeister:
The changes double the ones just merged in https://github.com/OpenSlides/OpenSlides/pull/6770/, don't they?

bastianjoel:
Yes but the changes do not seem to appear in changed files. So it should be fine after squashing.

PR #6810, opened 2024-09-27
OIDC/Auth2 integration

boehlke:
This PR contains a POC for a OIDC/oauth2 integration into OpenSlides. The dev setup is based on keycloak.


.gitmodules
peb-adr, line 1: Before merging these submod pointers must be reverted to default.

peb-adr (changes requested):

dev/docker/services.env
```diff
@@ -22,8 +22,7 @@ SEARCH_HOST=search
 SEARCH_PORT=9050
 RESTRICTER_URL=http://autoupdate:9012/internal/autoupdate
 
-AUTH_HOST=auth
-AUTH_PORT=9004
+KEYCLOAK_URL=http://keycloak:8080
```
peb-adr, line 25: Is there a specific reason this has to be `_URL`?
Intuitively I would prefer it being consistent with the other services defining `KEYCLOAK_HOST` and `KEYCLOAK_PORT` variables (if needed also `KEYCLOAK_PROTOCOL` which can be concatenated to `KEYCLOAK_PROTOCOL://KEYCLOAK_HOST:KEYCLOAK_PORT` within the service. As is done everywhere else.

https://github.com/OpenSlides/OpenSlides/blob/main/DEVELOPMENT.md#environment-variables

dev/docker/docker-compose.dev.yml
```diff
@@ -1,5 +1,32 @@
 version: "3"
 services:
+  keycloak:
+    image: quay.io/keycloak/keycloak:23.0.5
```
peb-adr, line 4: I would maybe suggest to create our own keycloak-derived docker image for this container as we did for the `openslides-proxy`. Name proposals:
  - `

Currently when thinking about running this in production, I (for example) do not like passing admin credentials plain text via env vars.\
The password should go into a secret. Then we can - in line with the other openslides services - set a var like `KEYCLOAK_ADMIN_PASSWORD_FILE=/run/secrets/.../keycloak_admin_password`.\
I do not see a mechanism like this in the official keycloak image on quay.io. We could implement this in our wrapper

Another benefit is we would have a good way of controlling the keycloak version being used without having to maintain many different compose files or k8s deployment definitions for it.

dev/docker/services.env
peb-adr: This seems to already be expected here:
https://github.com/OpenSlides/openslides-proxy/pull/18/files#diff-590bd89681945f557a56099d2c2b293d216d845ce1e2709dd652bbed1e199177R199 and
https://github.com/OpenSlides/openslides-proxy/pull/18/files#diff-590bd89681945f557a56099d2c2b293d216d845ce1e2709dd652bbed1e199177R119

boehlke: The endpoint could have a path, so another variable KEYCLOAK_PATH would be needed:
KEYCLOAK_PROTOCOL://KEYCLOAK_HOST:KEYCLOAK_PORT$KEYCLOAK_PATH
... I like KEYCLOAK_URL much better.

dev/docker/docker-compose.dev.yml
boehlke: I think this setup is great for development. No images build is required when changing the configuration script and it's easy to test with other keycloak versions.

Is the production environment directly derived from this? Perhaps we can build something different for production...

dev/docker/docker-compose.dev.yml
```diff
@@ -167,6 +172,28 @@ services:
...
+      - MEDIA_HOST=media
+      - MEDIA_PORT=9006
+      - MANAGE_HOST=manage
+      - MANAGE_PORT=9008
+      - CLIENT_HOST=client
+      - CLIENT_PORT=9001
+      - VOTE_HOST=vote
+      - VOTE_PORT=9013
+      - KEYCLOAK_HOST=keycloak
+      - KEYCLOAK_PORT=8080
```
peb-adr, line 196: Here probably the `env_file: services.env` directive should be used as done for the other services instead of redefining the whole environment, doubling the code.

dev/docker/services.env
peb-adr, line 1: This file will need `KEYCLOAK_HOST` and `KEYCLOAK_PORT`, right?

Also it seems `CLIENT_HOST` and `CLIENT_PORT` are missing, which would be a oversight from years ago (lol).

PR #6845, opened 2024-12-10
Add action to create a new release for every patch

peb-adr:


peb-adr:
I decided to keep the release title simple and only use the tag string (which is default).

peb-adr:
https://github.com/peb-adr/OpenSlides/releases

For example releases generated by this action.

rrenkert (approved):
Looks good!

PR #6846, opened 2024-12-11
Improve actions for status badges

peb-adr:


peb-adr:
Re-opened, because I thought the required check was orphaned i some way.

Then I understood:
The old integration test is still set as required in the repo settings, which doesn't exist on this branch anymore. @rrenkert talk about intuitive ;)

I removed the requirement temporarily to merge this.
Setting the new check as required after this PR is merged

PR #6857, opened 2024-12-16
Use backend directory structure

hjanott:
Since the backend directory structure will change as per OpenSlides/openslides-backend#2604 we need some adaptions in the main repository. 
This means that the docker global volume will be split and paths to the data and meta directory change.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -14,7 +14,7 @@ jobs:
           submodules: recursive
 
       - name: Use example data instead of initial data
-        working-directory: "./openslides-backend/global/data/"
+        working-directory: "./openslides-backend/data/"
         run: cp example-data.json initial-data.json
 
       - name: Start setup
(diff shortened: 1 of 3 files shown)


peb-adr (changes requested):
I thought about the `organization.initial_import` action.
Looks like this get's the default initial data using functionality from

    openslides_backend/shared/util.py

So that file must be adjusted too.

A `grep -r initial-data` also reveals the following files:

    tests/system/action/mediafile/test_upload.py: get_initial_data_file("global/data/initial-data.json")
    Makefile: check-initial-data-json:
                python cli/check_json.py global/data/initial-data.json
    docs/actions/meeting.create.md: [Links used in documentation]

peb-adr (approved):
For some reason I thought this PR was in the backend. That's how my above comment came to be.
Now I see that all files I mentioned are being adjusted in OpenSlides/openslides-backend/pull/2769.
For the main repo I couldn't find or think of more files to be adjusted.

I approve.

PR #6871, opened 2025-02-12
Add go workspace

bastianjoel:
**This is a proposal** for how to integrate the `openslides-go` repo into our development setup. 

This requires changes to all go subrepo Dockerfiles which leads to the `development` being definitely not executable without additional mounts. So we might need to check if the `development` target is used somewhere else. 

Also we still need a solution for production builds. 

* [x] `make build-dev` not working

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .DS_Store
 .idea
 *.code-workspace
+go.work.sum
 
 # docs
 docs/modelsvalidator/modelsvalidator
--- a/go.work
+++ b/go.work
@@ -1,0 +1,9 @@
+go 1.23.6
+
+use (
+	./lib/openslides-go
+	./openslides-autoupdate-service
+	./openslides-icc-service
+	./openslides-search-service
+	./openslides-vote-service
+)
--- a/go.work
+++ b/go.work
@@ -1,4 +1,4 @@
-go 1.23.6
+go 1.24.0
 
 use (
 	./lib/openslides-go
--- a/.gitmodules
+++ b/.gitmodules
@@ -40,3 +40,6 @@
 [submodule "openslides-search-service"]
 	path = openslides-search-service
 	url = https://github.com/OpenSlides/openslides-search-service.git
+[submodule "openslides-go"]
+	path = lib/openslides-go
+	url = https://github.com/OpenSlides/openslides-go.git
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 DC_PATH=dev/docker
 SCRIPT_PATH=dev/scripts
 DC=docker compose -f $(DC_PATH)/docker-compose.dev.yml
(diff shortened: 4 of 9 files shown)


ostcar:
I like this idea

.gitignore
```diff
@@ -6,6 +6,7 @@
 .DS_Store
 .idea
 *.code-workspace
+go.work.sum
```
ostcar, line 9: I think the .sum file should be checked in. There is a reason to ignore `go.work` in some cases (not our case) but if `go.work` is checked in, `go.work.sum` should also be checked in. Its the same as `go.mod` and `go.sum`

go.work
```diff
@@ -0,0 +1,9 @@
+go 1.23.6
```
ostcar, line 1: Could be updated to `1.24.0`. See https://github.com/OpenSlides/openslides-autoupdate-service/pull/1147

ostcar (approved):
Only suggestions.

I have not tested it. If its work, this seems like the correct solution.

.gitignore
bastianjoel: I think in this case it actually makes sense to not include it as this is just for the local setup. If I understand correctly the data is collected from the modules mod files anyway. Therefore IMHO this would be just an unnecessary source for merge conflicts.

go.work
```diff
@@ -0,0 +1,9 @@
...
+
+use (
+	./lib/openslides-go
+	./openslides-autoupdate-service
```
rrenkert, line 5: Can we change this `use` to reduce the dependencies between the go serviceres? I don't like the mount of all go-services into every container in the `docker-compose.dev.yml` above that is a result of this `use` block.

rrenkert (changes requested):

go.work
rrenkert: Does the versioning here mean that all repos in the `use` block have to use the same version and vice versa?

go.work
bastianjoel: IMHO for the `go.work` file this is okay. We could change the `go.work` files that are added to the containers to just contain the openslides-go module.

rrenkert: If we get rid of the unnecessary "dependencies", thats the way.

peb-adr:
> Also we still need a solution for production builds.

@bastianjoel Not sure what you meant here. Tested building prod images, they just work. Which is what I expected since the go module is just loaded as any other and then the binary gets compiled - just like before.

Discussed and tested this with @rrenkert. With latest changes we're both happy.
Merging this now to introduce openslides-go as submodule and then be able to create first staging update with it \o/ .

bastianjoel:
> Not sure what you meant here. 

I meant local prod which is used for the integration tests here and in the client. But not sure if this is a problem. I guess it is sufficient to react on this if it get relevant.

PR #6873, opened 2025-02-16
Improve documentation

nevyen:
Add info for using comands after changing port


peb-adr:
@nevyen thanks for that input!
That was definitely a detail missing in the documentation.

PR #6926, opened 2025-05-22
Dockerfile Rework

Janmtbehrens:
Reworked Dockerfile & Makefiles, so that there is one singular Dockerfile that supports development, testing and production images.
Also adding features like healthchecks and use of restricted app users, as well as fixing some bugs


rrenkert (changes requested):

m-schieder:
merged

PR #6960, opened 2025-07-28
Makefile Rework

Janmtbehrens:


bastianjoel (commented):
Local prod does not seem to work (`dev/localprod`). Postgres does not start.

bastianjoel (approved):

PR #6993, opened 2025-10-01
Make `Makefile` work for macOSX devices

xela1601:
There is a difference in behavior when using sed with the `-i` flag on GNU `sed` (available on linux) vs. BSD `sed` (available on macOS).

This caused the following error:

```bash
$ make run-dev
sed -i "1s/.*/go 1.25.0/" dev/docker/workspaces/*.work
sed: 1: "dev/docker/workspaces/a ...": extra characters at the end of d command
make: *** [build-dev] Error 1
```

For more information see [here](https://www.baeldung.com/linux/gnu-bsd-stream-editor).

This is resolved by detecting the os and building the full command dynamically based on that.


Makefile
```diff
@@ -19,7 +27,7 @@ run-dev-otel: | build-dev
 
 # Build the docker dev images for all services in parallel
 build-dev:
-	sed -i "1s/.*/$(GO_VERSION)/" $(DOCKER_PATH)/workspaces/*.work
+	sed $(SED_INPLACE) "1s/.*/$(GO_VERSION)/" $(DOCKER_PATH)/workspaces/*.work
```
m-schieder, line 30: Why not just use the `-e` flag as described [here](https://www.baeldung.com/linux/gnu-bsd-stream-editor#bd-2-the--i-option)?

That looks much simpler.

PR #7085, opened 2026-03-06
Replace services-to-main with checkout

Janmtbehrens:


dev/scripts/makefile/checkout.sh
```diff
@@ -150,6 +152,11 @@ checkout() {
         # If branch couldn't be found, user has the option to either checkout main branch instead or skip checkout for this service
         if [ -n "$BRANCH_NOT_FOUND" ]
         then
+            if [ "$AUTO_MAIN_FALLBACK" == 1 ]
+            then
+                info "Automatically skipping checkout for $SUBMODULE, because no branch found named $SOURCE/$BRANCH and automatic main fallback is active"
+                eixt 0
```
peb-adr: `exit`

dev/scripts/makefile/localprod-with-local-manage.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Moving openslides to localprod directory
+info  "Moving openslides executable"
+mv ./openslides ../dev/localprod/openslides
+cd ../dev/localprod || exit 1
+
+# Setup and generate localprod docker compose
+info  "Executing openslides setup"
+./openslides setup .
+./openslides config --config config.yml .
```
peb-adr: Again I'm not too happy to see these `cd`s, especially with the multilevel relative paths going back and forth.
IMO it's confusing and error prone.

`make` can be called with `-C` to provide a working directory without needing to change there.
`openslides setup` can also be called with another directory then `.` .

Maybe a solution to this could be to export some variable like `OPENSLIDES_REPO` from which paths can be build without need for `../[...]/../` constructs.

These concerns probably don't need to be addressed in this PR but I still have them.. @Janmtbehrens let's talk about it some time.

dev/scripts/make-update.sh
```diff
@@ -241,8 +225,7 @@ add_changes() {
 }
 
 choose_changes() {
-  ask y "Fetch all submodules $BRANCH_NAME changes now?" &&
-    OPT_PULL=1 fetch_all_changes
+  ask y "Fetch all submodules $BRANCH_NAME changes now?" && bash ./dev/scripts/makefile/checkout.sh "" "" "" "true" ""
```
peb-adr: I'd prefer this on two lines. More readable IMO.

peb-adr (changes requested):
Only one real issue.

dev/scripts/makefile/localprod-with-local-manage.sh
Janmtbehrens: I get the ugliness of `cd`s and I've removed them entirely with the new commit. The nested `../..` structures are still there however and I don't see a way to resolve this issue in another way. I exported the relevant paths into their own variables, so it should at least be more readable

peb-adr (approved):

PR #7106, opened 2026-04-24
Expanding Stash Options on Checkout

Janmtbehrens:
At the moment users are only able to stash changes or ignore a submodule entirely on checkout. This should be expanded, so that local changes can be reverted

### DIFF
--- a/dev/scripts/makefile/checkout.sh
+++ b/dev/scripts/makefile/checkout.sh
@@ -116,12 +116,14 @@ checkout() {
             info "The repository has changes"
             info "$GIT_CHANGES"
 
-            read -rp $'\n'"Stash them (Y) delete them (d) or skip this submodule (s): " </dev/tty
-            local STASH_OUTPUT=2
+            read -rp $'\n'"Stash them (Y) soft reset them (d) hard reset them (D) or skip this submodule (s): " </dev/tty
+            local STASH_OUTPUT=0
             case "$REPLY" in
             Y|y|Yes|yes|YES) STASH_OUTPUT=0;;
-            D|d) STASH_OUTPUT=1 ;;
-            *) STASH_OUTPUT=2 ;;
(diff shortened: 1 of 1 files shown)


peb-adr (approved):
Since I'm not the target audience for this features I'm not sure how relevant my judgement really is for this.

My gut feeling says automation on this level approaches the border of over engineering. One way of handling these situations would be to just error out whenever there are unstaged changes, leaving the dev to handle the situation as they prefer.
I personally only very rarely use `git stash` or `git reset --soft`.
And it would mean the tooling can make more simple assumptions about the state of the repo, which probably would improve robustness.

However this feature is not for me.
If the dev like to see this change, I'm all for it. The diff looks just fine.

Janmtbehrens:
> Since I'm not the target audience for this features I'm not sure how relevant my judgement really is for this.
> 
> My gut feeling says automation on this level approaches the border of over engineering. One way of handling these situations would be to just error out whenever there are unstaged changes, leaving the dev to handle the situation as they prefer. I personally only very rarely use `git stash` or `git reset --soft`. And it would mean the tooling can make more simple assumptions about the state of the repo, which probably would improve robustness.
> 
> However this feature is not for me. If the dev like to see this change, I'm all for it. The diff looks just fine.

Yes, I mainly wanted/needed your feedback on the technical implementation of this feature. I agree that it's somewhat unwieldy at this point. At the same time, both the devs who requested this feature and myself would like this option to exist and manually switching to each submodule to handle each case individually feels like a huge time sink (and nerve sink)
Maybe this could be solved with flags instead? Then again flags are harder to realize since the checkout.sh script is called via maketargets

peb-adr:
If there is demand for it right now, I'd say just merge it.

If in the future this seems to expand further "out of control" we might revisit on changing this behavior more fundamentally.

PR #7111, opened 2026-05-09
fix: use BSD-compatible sed -i syntax for macOS portability

xela1601:
Replace GNU-style `sed -i` with `sed -i ''` in all 5 occurrences so the Makefile works on both macOS (BSD sed) and Linux (GNU sed).

Related to https://github.com/OpenSlides/OpenSlides/pull/6993

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,11 +27,11 @@ $(.FLAGS):
 .PHONY: dev
 
 devstop:
-	@sed -i "1s/.*/$(GO_VERSION)/" $(DOCKER_PATH)/workspaces/*.work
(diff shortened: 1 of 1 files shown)


peb-adr:
Hi @xela1601,

Sorry that the `Makefile` again introduced incompatibility for you. We will try to raise awareness for it and keep it compatible in the future.

However as in the related issue you already mentioned we would prefer the `sed -i -e` way.
Calling `sed -i ''` breaks GNU sed. An alternative could be `sed -i''` - the optional suffix must be passed without a space. However I would tend towards the solution that was agreed on previously (`sed -i -e`).

If you adjust, we are happy to merge ;) .

PR #7128, opened 2026-06-04
Checkout fork fallback feature expansion

Janmtbehrens:
When checking out a branch on a fork which does not exist, the user is prompted to either checkout main instead of skip. When main checkout is chosen, the main branch of the fork is checked out. This is not desirable, since it requires the forked main branch to be up-to-date.
Instead the upstream/origin main should be pulled


peb-adr (approved):
LGTM and does what I need it to do.

PR #7147, opened 2026-07-15
Buildsh fix

LinKaKling:
simply supress unwanted grep error messages in build.sh


m-schieder (changes requested):
This will suppress all error messages. A nicer solution would be to test whether './go.mod' exists beforehand.
A nice touch would also be to use '--max-count=1' in grep since we only want to know whether there is at least one hit.

PR #7149, opened 2026-07-28
Feat: Make Dev GitHub Workflow

Janmtbehrens:
This workflow ensures that changes to future code will not break `make dev` functionality and by extension check if the development setup is working or not.

### DIFF
--- a/.github/workflows/make-dev-ci.yml
+++ b/.github/workflows/make-dev-ci.yml
@@ -1,5 +1,6 @@
 name: Runs and tests various make dev targets
-on: [pull_request]
+on:
+  pull_request:
 
 jobs:
   run-tests:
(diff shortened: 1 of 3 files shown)


peb-adr (approved):
I approve the implementation.
But I'm not yet convinced this is really needed or might be clutter.
@rrenkert can you maybe give an opinion?

bastianjoel:
> But I'm not yet convinced this is really needed or might be clutter.

I second that

We have the prod workflow and just testing the log commands feels unnecessary. Also you are not checking any of those outputs. I don't really see a point here.

Janmtbehrens:
> We have the prod workflow and just testing the log commands feels unnecessary. Also you are not checking any of those outputs. I don't really see a point here.

The prod workflow only tests if the docker compose file (and services) are starting correctly. This workflow tests if the make targets and make dev script is running as expected. There are no tests for that yet and there have been updates to main that break the script. 
Testing the log outputs themselves is not the goal per se, rather checking if it will execute at all without errors. We can change or expand the specific set of commands used here if that's the main issue

